PR #8, opened 2013-02-27
ComponentUnderTest rule

Kornel:
Hi, 
   There is a feature I use a lot with my Spring projects where dependencies are injected based on annotations. The idea is to automate the creation of mocks and get rid of @Mock and @InjectMocks annotations. It could be used anywhere where dependencies are injected based on annotations.

As described in https://groups.google.com/forum/?fromgroups=#!topic/mockito/aBv81zqU-p8 (rather unnoticed ;))

 Please take a look at ComponentUnderTestRuleIntegrationTest for
documentation / sample.

What do you think?

Kornel

### DIFF
--- a/src/org/mockito/rules/CGLibObjectCreationEngineImpl.java
+++ b/src/org/mockito/rules/CGLibObjectCreationEngineImpl.java
@@ -1,0 +1,32 @@
+package org.mockito.rules;
+
+import org.mockito.cglib.proxy.Enhancer;
+import org.mockito.cglib.proxy.NoOp;
+
+public final class CGLibObjectCreationEngineImpl implements ObjectCreationEngine {
+
+    private static class Holder {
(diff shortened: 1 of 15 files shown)


TimvdLippe:
This PR has been stale for quite some time. Mockito 2.0 is now using Bytebuddy instead. @Kornel are you willing to work with us updating this PR to be compatible with the new engine?

Kornel:
Hi,

 I guess the real question is - do we want to include this pull request? :)
Since is stale since a few years, maybe that's not the proper way to mock
things? I think the current trend is to have dependencies in the
constructor for reasons like explicit setting in tests etc

Kornel

pt., 17.06.2016 o 22:39 użytkownik Tim van der Lippe <
notifications@github.com> napisał:

TimvdLippe:
@Kornel Hm yes after reconsideration closing might be a better option. If I remember correctly there were some discussion regarding Spring integration, so let's wait till that discussion has been resolved. Thank you anyways for your contribution and sorry that we were not able to integrate it :(

PR #20, opened 2013-11-08
Implemented support for InOrder evaluation of Timeout VerificationMode

sirchia:
Fixed for [issue 292](https://code.google.com/p/mockito/issues/detail?id=292)

Note that while the provided wrapping mechanism is generic enough to support both 'timeout' and 'after' usage, the current implementation of VerificationWithTimeoutTest.java prevents correct usage with 'after'. This is because it repeatedly verifies until the timeout has elapsed, thus marking all invocations that may have matched in the meantime as verified. This will of course 'exhaust' correct invocations and more often than not fail on the verification before the 'ager time' has elapsed.

### DIFF
--- a/src/org/mockito/Mockito.java
+++ b/src/org/mockito/Mockito.java
@@ -779,8 +779,6 @@ import org.mockito.verification.VerificationWithTimeout;
  * <p>
  * It feels this feature should be used rarely - figure out a better way of testing your multi-threaded system.
  * <p>
- * Not yet implemented to work with InOrder verification.
- * <p>
  * Examples:
  * <p>
  * <pre class="code"><code class="java">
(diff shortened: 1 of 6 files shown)


Jazzepi:
I could also use this feature! :+1:

TimvdLippe:
@sirchia could you please rebase this PR on current master, then I can take a look and review. Sorry for the extremely delayed response, hopefully you are still interested in making this contribution up-to-date.

If this PR is not updated within 2 weeks, I will close it as stale.

Jazzepi:
@TimvdLippe I had actually done that for my own private usage a while back. I've provided a https://github.com/mockito/mockito/pull/424 that rebases the original PR by @sirchia

PR #39, opened 2014-03-05
Allow calling real implementation of jdk8 extension methods

alberskib:
See also the google code ticket: https://code.google.com/p/mockito/issues/detail?id=456

Enable calling real implementation on extensions method from jdk8
Additionally change add validation checking whether callRealMethod is invoked on abstract methods.
In fact resolving issue could be achieved by removing this validation because when someone do next thing:
1: when(someInteface.abstractMethod()).thanCallRealMethod();
2: someInteface.abstractMethod()
Exception will be throwned on second line informing that there were attempt to invoke abstract method

### DIFF
--- a/src/org/mockito/exceptions/Reporter.java
+++ b/src/org/mockito/exceptions/Reporter.java
@@ -580,12 +580,12 @@ public class Reporter {
         ));
     }
 
-    public void cannotCallRealMethodOnInterface() {
+    public void cannotCallAbstractRealMethod() {
         throw new MockitoException(join(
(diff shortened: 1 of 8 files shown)


fbastian:
Hi, 

what is the status of this feature? The release note 1.10.0 (2014-09-25 22:25 UTC) says "Allow calling real implementation of jdk8 extension methods", but actually, it is not possible, see for instance: http://stackoverflow.com/questions/27663252/can-you-make-mockito-1-10-17-work-with-default-methods-in-interfaces

From my own tests, it is indeed not possible to call real implementation of default methods.

PR #40, opened 2014-03-11
issue #345 - custom answer class can be specified in annotation

kretes:
[issue 345](https://code.google.com/p/mockito/issues/detail?id=345)

### DIFF
--- a/src/org/mockito/internal/configuration/MockAnnotationProcessor.java
+++ b/src/org/mockito/internal/configuration/MockAnnotationProcessor.java
@@ -4,11 +4,16 @@
  */
 package org.mockito.internal.configuration;
 
+import org.mockito.Answers;
 import org.mockito.Mock;
 import org.mockito.MockSettings;
 import org.mockito.Mockito;
+import org.mockito.exceptions.Reporter;
+import org.mockito.stubbing.Answer;
 
+import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
 
 /**
  * Instantiates a mock on a field annotated by {@link Mock}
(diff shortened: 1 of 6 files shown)


scubbo:
What's the status of this change? Was another PR ever created? I was thinking about adding this feature myself - if kretes@ is no longer interested in driving this, I could submit a new PR?

src/org/mockito/internal/configuration/MockAnnotationProcessor.java
```diff
@@ -24,12 +29,50 @@ public Object process(Mock annotation, Field field) {
...
+        if (hasCustomAnswerDefined(customAnswer)) {
+            reportCustomAnswerMisuse(defaultAnswer, field);
+
+            return instantiateCustomAnswer(customAnswer);
+        }
+        return defaultAnswer.get();
+    }
+
+    private Answer<?> instantiateCustomAnswer(Class<? extends Answer> customAnswer) {
+        if(customAnswer.getConstructors().length == 0) {
```
scubbo: Going through your comments and attending to them in my own (planned) Pull Request - not sure what you mean here? Doesn't getConstructor() (with no arguments) check for precisely that public parameter-less constructor? If no such constructor exists, a NoSuchMethodException will be thrown (and then caught by the catch Exception block that you suggested)

test/org/mockitousage/annotation/MockWithCustomAnswerAnnotationsTest.java
```diff
@@ -0,0 +1,140 @@
...
+    public void should_choose_default_constructor() {
+        assertSame(MyCustomAnswer.returnedObject, mockWithCustomAnswerHavingMultipleConstructors.objectReturningMethodNoArgs());
+    }
+
+    @Test
+    public void should_throw_exception_when_custom_answer_cannot_be_instantiated() throws NoSuchFieldException {
+        Field aField = MyCustomAnswer.class.getDeclaredFields()[0];
+
+        try {
+            new MockAnnotationProcessor().process(aMockAnnotation().withCustomAnswer(NotInstantiableClass.class).build(), aField);
```
scubbo: Not sure how that's possible without splitting all the test cases out into their own test classes - consider the following:

````
public class MockWithCustomAnswerAnnotationsTest {
  @Mock(answerClass = NotInstantiableClass.class)
  Object unMockableObject;

  @Mock(answerClass = AnswerThrowingConstructor.class)
  Object objectWithAnswerClassThrowingException;
  ...

  @Test
  public void testNonInstantiableObject() {
    try {
      initMocks(this);
      fail("RuntimeException expected ");
    } catch (MockitoException ex) {
      assertEquals("Custom answer cannot be instantiated",ex.getMessage());
    }
  }

  @Test
  public void testAnswerClassThrowsException() {
    ...
  }
````

Since `initMocks(this);` will initialise all the mocks in the Test Class at once (and, presumably, in unpredictable order), at-most one of the tests will see expected behaviour - the others will see exceptions from instantiating other "failing answer classes". We can't define the @Mock inside the test case, because it's not applicable to a local variable.

Was splitting out test cases to multiple test classes, each with one @Mock, what you intended?

kretes:
Hi, I can see you submitted https://github.com/mockito/mockito/pull/302 - good luck!

scubbo:
Thanks!

PR #54, opened 2014-05-05
An initial pass at throwing an exception for unstubbed invocations.

RohanTalip:
This is an initial solution for https://code.google.com/p/mockito/issues/detail?id=346

It doesn't solve all the use cases on that issue but may be a good enough solution for now.

FYI, I moved ThrowsSmartNullPointer out of ReturnsSmartNulls into its own package private class so it could be reused by the new ThrowsSmartNulls class.

Please let me know if the name of the new ThrowsSmartNulls class could be improved, however I felt that keeping "SmartNulls" as part of the name made sense, since it is
a) still returning smart null objects for unstubbed methods that return non-primitive and non-final types (the smart null objects throw smart null exceptions on access via method calls, except for toString() calls), and
b) now throwing smart null exceptions immediately for methods that return primitive or final types.


TimvdLippe:
@RohanTalip Friendly ping to process the feedback. If this PR is not updated within 2 weeks, I will close it as stale.

PR #55, opened 2014-05-06
Allowed for the configuration to be set via the MockitoAnnotations and Mockito classes

RohanTalip:
Allowed for the configuration to be specified through a new `MockitoAnnotations.initMocks(IMockitoConfiguration, Object)` method and also through `Mockito.setConfiguration(IMockitoConfiguration)`.

This allows the specification of a default answer programmatically rather than through a class on the classpath.  This is useful for not needing to remember to set the classpath in Eclipse, especially when there are many team members working on a project.

### DIFF
--- a/src/org/mockito/Mockito.java
+++ b/src/org/mockito/Mockito.java
@@ -4,7 +4,9 @@
  */
 package org.mockito;
 
+import org.mockito.configuration.IMockitoConfiguration;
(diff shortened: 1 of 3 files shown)


TimvdLippe:
Closing as this pull request is stale. Thank you for your effort @RohanTalip !

PR #60, opened 2014-07-01
Adding "getRealClass()" method to MockingDetails and DefaultMockingDetails

davidmichaelkarr:
In some situations, it's useful to get from a mock object the original class object that was mocked to produce the mocked class. The logical place to provide this information is in the MockingDetails interface, and in the DefaultMockingDetails implementation. The method that provides this information should return a reasonable result if the object is either a mock, a spy, or neither.

This change adds the method to the interface and implementation, and also adds a unit test for DefaultMockingDetails which includes tests for this new method and some tests for some of the existing methods in the class.

### DIFF
--- a/src/org/mockito/MockingDetails.java
+++ b/src/org/mockito/MockingDetails.java
@@ -41,4 +41,13 @@ public interface MockingDetails {
      * @since 1.9.x
      */
     Collection<Invocation> getInvocations();
+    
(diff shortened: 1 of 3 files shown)


bric3:
fixed in misspelled commit  a258ae81a164f794ca58fe197da33b8101823135

PR #81, opened 2014-09-19
Update Mockito CGLIB and ASM

guw:
This is an update to Mockito's embedded ASM and CGLIB libraries.


bric3:
@martin-g With ASM 5.x some API and opcodes should be used.

bric3:
Closing due to byte buddy being merge #171, also see my last comment in #83 

Thank you so much for your effort, it me reused if we ever publish a mockito-cglib-mockmaker.

PR #82, opened 2014-09-19
New sub-project for simulating OSGi fragment class-loading issues

guw:
This commits adds a sub-project with a small Ant script for creating three jars - Mockito, sample source and tests. It also adds a new tests for verifying package protected mocking.

The three jars by themselves are enough to trigger the same issue which has been seen in OSGi environment with mocking package visible members. It fails if tests and sources under tests are in different jar files even though they are loaded by the same class loader.

The same tests runs fine as part of the regular Gradle build.

### DIFF
--- a/subprojects/osgiTest/build.xml
+++ b/subprojects/osgiTest/build.xml
@@ -1,0 +1,121 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project basedir="." default="all" name="mockito-osgi-tests">
+
+  <target name="init">
+    <property name="compile.target" value="1.5" />
+    <property name="compile.source" value="1.5" />
+
+    <property name="build.target" value="build" />
+
+    <property name="key.store" value="${build.target}/.keystore" />
+    <property name="key.storepass" value="mockito" />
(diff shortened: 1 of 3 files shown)


Tremelune:
Note that since some Android tests must be (or are at least recommended to be in the Google docs) in a different project than the test code, this bug affects them, though I suspect it's only an issue with Gradle projects.

TimvdLippe:
@raphw could you take a look at this PR? I think it can be closed as we fixed the OSGI issues in the other PRs.

bric3:
Actually the idea of this PR was to build an OSGI project woth OSGI artefacts and test mockito with those artifacts. The trouble was the fact that gradle wasn't supporting well OSGI at that time. That explains why this PR still uses the ant build script.

We may want to create an OSGI project the gradle way.

raphw:
Since this ticket is really old and we use Gradle, should we close this PR but rather open an issue for it?

TimvdLippe:
Yes that seems like a good resolution.

bric3:
That seems fine to me as well.

PR #83, opened 2014-09-19
Enhance Gradle build script with repackaging process

guw:
This change adds new tasks to the Gradle build script for repackaging
asm and cglib sources downloaded from Maven central.


bric3:
Hi @guw, as you ay have noticed byte-buddy now replaces CGLIB in mockito.
At the moment ByteBuddy is not shadowed, but it may be later the case, anyway we'd like to avoid further dependencies on `ant`. Old CGLIB mockmaker code is still there but may be moved in the subproject, and eventually released (which means CGLIB could be updated).

PR #141, opened 2014-12-16
Fix anyX() methods to not match nulls

### LINKED ISSUE #134: Argument matcher anyXxx() (i.e. anyString(), anyList()) should not match nulls
This is a bug I'm seeing in 1.10.8 version (older version has the same issue - tested with 1.9.0).

Given:
```java
Function<Object, Integer> function = Mockito.mock(Function.class);
when(function.apply(Mockito.anyString())).thenReturn(1);
Integer result = function.apply(2);
```
Expected behavior:
```java
result == null;
```

Actual behavior:
```java
result == 1;
```

Note that the function is called with an integer (not a string), and still the mocked function return the value which it should return only when a string is passed. The same works when using anyBoolean() or any other methof from any* family.

pimterry:
This fixes #134. To be specific:

`anyX()` - matches all instances of X (Integer, String, List), but not nulls.
`anyList(X.class)`, `anySet(X.class)`, etc - matches all list/set/map instances, ignoring the X parameter entirely (just used to fix generic type hassle). Doesn't match null.
`any()`, `anyObject()` - matches anything, including null.
`any(X.class)` - matches anything, including null, but with a more convenient return type.

We might want to take this further and deprecate some of these, or make the suggested API slightly clearer, but this is a good start.

Since this is definitely not backward compatible, I've now bumped to 2.0.0-beta!

### DIFF
--- a/src/org/mockito/Matchers.java
+++ b/src/org/mockito/Matchers.java
@@ -50,13 +50,6 @@ import java.util.Set;
  * This implementation is due static type safety imposed by java compiler.
  * The consequence is that you cannot use <code>anyObject()</code>, <code>eq()</code> methods outside of verified/stubbed method.
  *
- * <p>
- * <b>Warning 2:</b>
- * <p>
- * The any family methods <b>*don't do any type checks*</b>, those are only here to avoid casting
- * in your code. If you want to perform type checks use the {@link #isA(Class)} method.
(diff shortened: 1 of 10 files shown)


szczepiq:
Included in 2.0.0-beta. Thanks for contribution!!!

PR #142, opened 2014-12-16
Fix #128 : Deep stubbing with generic responses in the call chain is not working

### LINKED ISSUE #128: Deep stubbing with generic responses in the call chain is not working
Deep stubbing will throw an Exception if multiple generics occur in the call chain. For instance, consider having a mock `myMock1` that provides a function that returns a generic `T`. If `T` also has a function that returns a generic, an Exception with the message "Raw extraction not supported for : 'null'" will be thrown.

As an example the following test will throw an Exception:

``` Java
public class MockitoGenericsDeepStubTest {

    @Test
    public void discoverDeepMockingOfGenerics() {
        MyClass1 myMock1 = mock(MyClass1.class, RETURNS_DEEP_STUBS);

        when(myMock1.getNested().getNested().returnSomething()).thenReturn("Hello World.");
    }

    public static interface MyClass1 <MC2 extends MyClass2> {
        public MC2 getNested();
    }

    public static interface MyClass2<MC3 extends MyClass3> {
        public MC3 getNested();
    }

    public static interface MyClass3 {
        public String returnSomething();
    }
}
```

You can make this test run if you step into the class `ReturnsDeepStubs` and change the method `withSettingsUsing` to return `MockSettings` with `ReturnsDeepStubs` instead of `ReturnsDeepStubsSerializationFallback` as default answer:

``` Java
private MockSettings withSettingsUsing(GenericMetadataSupport returnTypeGenericMetadata, MockCreationSettings parentMockSettings) {
    MockSettings mockSettings = returnTypeGenericMetadata.hasRawExtraInterfaces() ?
            withSettings().extraInterfaces(returnTypeGenericMetadata.rawExtraInterfaces())
            : withSettings();

    return propagateSerializationSettings(mockSettings, parentMockSettings)
            .defaultAnswer(this);
}
```

However, this breaks other tests and features.

I think, the issue is that further generics are not possible to be mocked by `ReturnsDeepStubsSerializationFallback` since the `GenericMetadataSupport` is "closed" at this point.

Thanks and kind regards
Tobias

pbielicki:


pbielicki:
I was off for few days - I see Szczepan did the job, so there's noting left for me :)

btw. Happy New Year!
cheers

PR #144, opened 2014-12-19
enum Answers implements now Answer<Object>

ChristianSchwarz:
The enum type Answers implements now the Answer interface. This allows us to use the enum values in `Mockito.mock(Class,Answer)`. This is handy when you have to test legacy code with PowerMockito and want to use both `@Mock` and `Mockito.mock(..)`.

```java
@Mock(answers=RETURNS_DEEP_STUBS) 
ClassA a;
ClassB b;

public void setUp(){
  b= mock(ClassB,Mockito.RETURNS_DEEP_STUBS)   //<< static import is not possible here, but with this pull request we can use RETURNS_DEEP_STUBS we used in @Mock.
}
```


bric3:
Thanks for the change

PR #158, opened 2015-01-15
[#157] Source files should not be put in binary JAR

### LINKED ISSUE #157: Source files should not be put in binary JAR
Source files (`*.java`) should not be put into binary `mockito-core.jar`. It stupefies Idea to show decompiled file even when source jar is available.

szpak:
It stupefies IntelliJ Idea to show decompiled file even when source jar is available.

Fix #157.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -71,7 +71,7 @@ def licenseFiles = copySpec {
 
 jar {
     baseName = 'mockito-core'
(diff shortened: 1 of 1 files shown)


bric3:
Duplicate of #28, however we felt that sources could remain in the jar for projects that don't rely on maven like repositories

Issue #109, opened 2014-10-23
Improve automated release notes look


szczepiq:
- Let's ditch the top level "Improvements:" element.
- Let's remove the 'Other' label and put unlabelled elements under 'Enhancements'

Anyone interested to pick this up? :)

TimvdLippe:
I think the current look is correct (https://github.com/mockito/mockito/blob/master/doc/release-notes/official.md) therefore closing this issue.

Issue #108, opened 2014-10-22
Clarify Spy vs Mock CALLS_REAL_METHODS


wernight:
I'd suggest adding a comment or side note in `[@InjectMocks](http://mockito.github.io/mockito/docs/current/org/mockito/InjectMocks.html)` to explain that when you want to inject real instances (and not mocks) you have the choice (unless I'm mistaken) between `[@Spy](http://mockito.github.io/mockito/docs/current/org/mockito/Spy.html)` and `[@Mock(answer = CALLS_REAL_METHODS)](http://mockito.github.io/mockito/docs/current/org/mockito/Mockito.html#CALLS_REAL_METHODS)` but the former is preferable, because...

PR #161, opened 2015-01-25
Excluded missing transitive dependency of the coveralls gradle plugin to fix failing build.

brewerc:


bric3:
Hey thanks for the PR :)

However this issue being worked on by upgrading dependencies. It will be merged soon.

brewerc:
Oh good, that will be a better solution. I just thought I'd share a
quick/easy fix to unblock the build in case it was affecting others.

bric3:
Well, upgrading dependency, just got me under an avalanche of... different kind of issues. I still can't explain all at the moment. So in the mean time I will merge your PR ;)

bric3:
And... no in the end there's the same new issue... with `GitHubTicketFetcherTest`
https://travis-ci.org/mockito/mockito#L147

brewerc:
The new problem seems to be unrelated. The test is pulling a list of closed issues from github @ https://api.github.com/repos/mockito/mockito/issues?access_token=a0a4c0f41c200f7c653323014d6a72a127764e17&state=closed&filter=all

and looking for issues #109, 108, 99999, and 112. 108 and 112 are legitimate closed issues that it is expecting in response to verify against. However, it looks like there is a limit to the number of issues that the API returns. In this case, issue #108 has rolled off onto the next "page" (you can see it on page 2 if you add "&page=2" to the query) because of the pull request that I created (apparently a PR gets assigned an issue #).

bric3:
Yes I know, increasing version numbers got me a lot of _somehow unrelated_ issues, but then I wanted to sleep. Creating this very PR, bumped the test issue number to the next page, that's why I wanted to close yours :)

I am currently fixing the code.

PR #164, opened 2015-01-25
Upgrade coveralls plugin

bric3:
Should have been just changing numbers, but it isn't. This PR contains several fixes in the source code in order to fix the coverage reports.

However with the last version there's still a pretty big issue in the travis logs, cobertura fails with some classes hence the lower coverage result (I suspect it is related to the missing dependency, but I'm not sure...)

* Replacing Cobertura with Jacoco, make Gradle fail with obscure stacktraces

  ```gradle
apply plugin : 'jacoco'
apply plugin: 'com.github.kt3k.coveralls'

  jacocoTestReport {
      reports {
          xml.enabled = true // coveralls plugin depends on xml format report
          html.enabled = true
      }
}
```

  ```
  Failed to notify ProjectEvaluationListener.afterEvaluate(), but primary configuration failure takes precedence.
  java.lang.NullPointerException: Cannot get property 'plus' on null object
  	at org.codehaus.groovy.runtime.NullObject.getProperty(NullObject.java:57)
    ...

* Where:
Script '/Users/brice/work/opensource/mockito/gradle/coverage.gradle' line: 4

* What went wrong:
A problem occurred evaluating script.
> Could not find method jacocoTestReport() for arguments [coverage_bz56tm90of2yklk5dd11r2maq$_run_closure1@aca79de] on root project 'mockito'.
```

  ... yet it is documented like that here : http://gradle.org/docs/current/userguide/jacoco_plugin.html#N138AF

  I don't get what I'm missing there.

* Updated Cobertura / Coveralls plugin, however the coverage phase fails to instrument classes, for a unknown reason at this moment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,4 +17,4 @@ script:
 - ./gradlew ciBuild
 after_success:
 - ./gradlew release
-- ./gradlew cobertura coveralls
+- ./gradlew --debug --stacktrace cobertura coveralls
--- a/test/org/mockitointegration/NoJUnitDependenciesTest.java
+++ b/test/org/mockitointegration/NoJUnitDependenciesTest.java
@@ -4,36 +4,75 @@ import org.hamcrest.Matcher;
 import org.junit.Test;
 import org.mockito.Mockito;
 import org.mockito.cglib.proxy.Enhancer;
+import org.mockito.junit.MockitoJUnit;
 import org.mockitoutil.ClassLoaders;
 import org.objenesis.Objenesis;
 
 import java.util.Set;
 
+import static org.fest.assertions.Assertions.assertThat;
+
 public class NoJUnitDependenciesTest {
+
     @Test
-    public void pure_mockito_should_not_depend_JUnit() throws Exception {
+    public void pure_mockito_should_not_depend_JUnit() throws Throwable {
         ClassLoader classLoader_without_JUnit = ClassLoaders.excludingClassLoader()
                 .withCodeSourceUrlOf(
                         Mockito.class,
                         Matcher.class,
                         Enhancer.class,
-                        Objenesis.class
+                        Objenesis.class,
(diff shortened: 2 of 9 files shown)


bric3:
For reference purpose I added a comment on the gradle cobertura plugin about the decrease in coverage : https://github.com/stevesaliman/gradle-cobertura-plugin/issues/67

bric3:
Also for reference purpose, there's a thread on the gradle forums about the issue to get Jacoco working http://forums.gradle.org/gradle/topics/hans-dockter-gradleware-com?rfm=1

.travis.yml
```diff
@@ -17,4 +17,4 @@ script:
 - ./gradlew ciBuild
 after_success:
 - ./gradlew release
-- ./gradlew cobertura coveralls
+- ./gradlew --debug --stacktrace cobertura coveralls
```
szczepiq: --debug is not very useful (very internal stuff)

test/org/mockitointegration/NoJUnitDependenciesTest.java
```diff
@@ -4,36 +4,75 @@
...
 
 import java.util.Set;
 
+import static org.fest.assertions.Assertions.assertThat;
+
 public class NoJUnitDependenciesTest {
+
     @Test
-    public void pure_mockito_should_not_depend_JUnit() throws Exception {
+    public void pure_mockito_should_not_depend_JUnit() throws Throwable {
```
szczepiq: I like the idea of using isolated class loaders to test classpath related features.

szczepiq:
Thanks for looking into this! This is great stuff!!!

.travis.yml
bric3: Yeah but at that time build/coverage broke without notice just for fun.
I had no clue what was the issue. gradle (and plugin) wasn't much helping on that front. Anyway I removed coveralls and cobertura.

Issue #154, opened 2015-01-04
stop depending on hamcrest internally


szczepiq:
In order to avoid compatibility/upgrade problems we will stop depending on hamcrest internally. Hamcrest will be an optional dependency of Mockito.

tobier:
What's the progress on this? I'd like to use Mockito with my unit tests and I'm running JUnit 4.11 that is bundled with Eclipse 4.4 that I'm using.

bric3:
Is Hamcrest dependency a blocker for using Mockito with Eclipse 4.4 ?

-- Brice

tobier:
@bric3 It seems that way. Eclipse 4.4 comes with JUnit 4.11 and Hamcrest 1.3, which apparently doesn't work with Mockito? Therefore when I try to run my JUnit plug-in tests for an Eclipse 4.4 RCP application it doesn't work.

martin-g:
I use Mockito 1.9.5 and 1.10.16 with JUnit 4.11 and 4.12 combined with Hamcrest 1.3 in several applications and didn't have any issues.

bric3:
Fells more of an Eclipse issue, if an IDE cannot work with the project
classpath...

The team / the code is clearly not near any Mockito 2.0 _final_ release.
Anyway, this task is not even started, and it will target Mockito 2.0 as
API will change.

tobier:
@bric3 @martin-g My bad, I made a plugin for Mockito and forgot to add the dependencies to Hamcrest and Objenesis. Seems to work fine now as far as I can tell..

philipwhiuk:
Would be nice to at least upgrade to hamcrest 1.3 (latest version I think).

bric3:
@philipwhiuk Maybe in version 2.0. Maybe we could eventually drop the dependency (at this time nothing yet decided)

szczepiq:
Working on this. We will use hamcrest 1.3, too.

szczepiq:
This is completed on master.

PR #167, opened 2015-02-07
Fixes Google Code issue 409 - added invocation of describeMismatch where possible so it's...

### LINKED ISSUE #409: Selective capturing matcher
Creating an issue tracker for a feature request.

A small team i've been working with has created an extended use case wrapping Mockito. We'd like to add in some of the functionality that we've found to be particular useful during our own development.

One such item is an ArgumentCaptor that matches on a provided ArgumentMatcher. We've had several use cases within our own library where we'd like capture an argument that meets a certain condition, and leverage that captured argument for additional verification or reporting.

You _could_ capture the value, then match it against the provided matcher in a separate step. This alleviates the extra work.

Usage example:
We have an API that emits a host of Kafka (http://kafka.apache.org/) events. It's hyper-critical that we have high test coverage for the various events that we emit, so we created a separate TestNG reporter (http://testng.org/javadocs/org/testng/IReporter.html) to generate a basic coverage report (events tested / total events). As the user makes verification on the events they've emitted (using custom matchers), we capture and log them to the reporter.

mzoltan:
Reflectively invoking describeMismatch when possible.
[Google code issue 409](https://code.google.com/p/mockito/issues/detail?id=409)

### DIFF
--- a/src/org/mockito/internal/invocation/ArgumentsProcessor.java
+++ b/src/org/mockito/internal/invocation/ArgumentsProcessor.java
@@ -38,14 +38,18 @@ public class ArgumentsProcessor {
         return newArgs;
     }
 
+    public static Matcher argumentToMatcher(Object arg) {
+        if (arg != null && arg.getClass().isArray()) {
+            return new ArrayEquals(arg);
+        } else {
+            return new Equals(arg);
+        }
+    }
+
     public static List<Matcher> argumentsToMatchers(Object[] arguments) {
         List<Matcher> matchers = new ArrayList<Matcher>(arguments.length);
         for (Object arg : arguments) {
-            if (arg != null && arg.getClass().isArray()) {
-                matchers.add(new ArrayEquals(arg));
-            } else {
-                matchers.add(new Equals(arg));
-            }
+            matchers.add(argumentToMatcher(arg));
         }
         return matchers;
     }
--- a/src/org/mockito/internal/invocation/InvocationImpl.java
+++ b/src/org/mockito/internal/invocation/InvocationImpl.java
@@ -94,7 +94,7 @@ public class InvocationImpl implements Invocation, VerificationAwareInvocation {
     }
 
     public String toString() {
-        return new PrintSettings().print(ArgumentsProcessor.argumentsToMatchers(getArguments()), this);
+        return new PrintSettings().printArguments(this);
     }
 
     public Location getLocation() {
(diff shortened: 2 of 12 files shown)


bric3:
Could you replace in the commit message the `#` by `issue`, github inspect each commits and will link to the PR 409 (even if not yet existing), GH may even mark the PR / issue merged.

mzoltan:
You mean "Fixes issue409 - ..."? CONTRIBUTING.md should be updated with
this info I think.

Nope I did not, I only tested with the shipped version. Is it enough if I
test with the latest or did you have other specific versions in mind?

2015-02-08 18:38 GMT+01:00 Brice Dutheil <notifications@github.com>:

bric3:
> You mean "Fixes issue409 - ..."? CONTRIBUTING.md should be updated with this info I think.

Good point, thx :smile: 

> Nope I did not, I only tested with the shipped version. Is it enough if I test with the latest or did you have other specific versions in mind?

Actually I it would good to see if this change is compatible across hamcrest versions 1.1, 1.2, 1.3 as long as mockito depends on Hamcrest internally see #154

mzoltan:
So I created a test application with only the mockito jar, objenesis 1.2
and hamcrest versions 1.1, then 1.2 and 1.3. All work as expected, starting
from version 1.2 the describeMismatch method is of course mandatory and is
being picked up all the same by mockito.

2015-02-08 19:00 GMT+01:00 Brice Dutheil <notifications@github.com>:

bric3:
cool, I'll test your code asap.

One task that may prove useful to automatize this test is to make "integration" sub-projects

mzoltan:
Hey Brice,

I was wondering if there's an estimate as to when/if this change will be
released? As it happens I could really use this in my work but I'm only
supposed to work with disted versions :)

Thanks

2015-02-15 22:42 GMT+01:00 Brice Dutheil <notifications@github.com>:

mzoltan:
Hey Brice,

Will this be merged do you think? It's been a while but I'm happy to rebase if it's considered for 2.0.

szczepiq:
Hey, we cannot merge this as I have changed hamcrest integration rather drastically recently. Mockito no longer uses hamcrest internally.

If you still believe that this feature is useful, can you show us how? E.g. can you show as textual view of a verification failure that demonstrates the usefulness of describing mismatches?

mzoltan:
Sure, I had this Network#send(EncodedMessage msg) method and I mocked Network and used a ContainsEntity matcher to make assertions on the contents of the encoded message. The matcher would decode the msg and check whether the contents contained what I expected. You can see how it was not very useful when the match failed, I got something like:

Expected: "a message containing: MyEntity"
Got: 129dh3jaksjSI&Dh2k8ad... etc. (the toString of EncodedMessage)

It would have been useful to know what was in that message, rather than its toString.

PR #170, opened 2015-02-15
Fixes #169

### LINKED ISSUE #169: @InjectMocks and @Spy on same field should cause MockitoException
Some time ago there was [@InjectMocks and @Spy cannot be used together when object initialized by mockito](https://code.google.com/p/mockito/issues/detail?id=489) issue rejected.
Checking the codebase I found a test case `WrongSetOfAnnotationsTest.shouldNotAllowSpyAndInjectMock()` which, I guess, was supposed to cover Brice's explanation. The test passes because the MockitoException **is** thrown, but it's thrown for different reason: the test tries to instantiate an interface (List). The test was there at list since 2010, so I think the behavior could have been basically forgotten.
So, although I find Brice's explanation convincing, I still think the exception should be thrown in that case, informing of wrong Mockito usage.

makasprzak:
First fixed the test so that it doesn't try to instantiate an interface, then added the Spy annotation to the set of not allowed together with InjectMocks. Assumed the behavior should be same as for InjectMocks + Mock combination.

### DIFF
--- a/src/org/mockito/internal/configuration/injection/scanner/InjectMocksScanner.java
+++ b/src/org/mockito/internal/configuration/injection/scanner/InjectMocksScanner.java
@@ -4,10 +4,7 @@
  */
 package org.mockito.internal.configuration.injection.scanner;
 
-import org.mockito.Captor;
-import org.mockito.InjectMocks;
-import org.mockito.Mock;
-import org.mockito.MockitoAnnotations;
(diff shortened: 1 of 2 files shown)


makasprzak:
Seeing the details of CI failure I guess there is some conflict of expected behaviours, I guess I give up:/

PR #171, opened 2015-02-16
Replace CGLIB by Bytebuddy

bric3:
Preliminary steps to replace CGLIB by ByteBuddy, thanks to @raphw for bootstrapping the code.

What does it bring :

* This PR replaces CGLIB entirely by ByteBuddy as the default mockmaker
  * ByteBuddy offers a more modern take on byte code manipulation than CGLIB
  * ByteBuddy offers more features than CGLIB
  * ByteBuddy is actively developed
  * ByteBuddy may be the trigger to support `final` class support
* The current test suite show it's feature complete
* This fixes the long standing issue [212](https://code.google.com/p/mockito/issues/detail?id=212) where a mock misbehave if the mocked type has a non public parent.
* This PR has been the playground for some test utils that have already been cherry picked in the master like the test tool `ClassLoaders`
* This PR uses different source folders for the mockmaker code. Both CGLIB and ByteBuddy relative code have been moved.

This PR may need review and potential leftovers are :

* Create a `mockito-bytebuddy-mockmaker` maven release, as the `MockMaker` interface din't yet change, it could provide an opportunity to test ByteBuddy before it is integrated in Mockito 2.x
* Create a `mockito-cglib-mockmacker` maven release, while the default mock maker is replaced, the legacy mock maker has been the foundation of mockito for a long time and other project may still depend on it (my first thought is PowerMock) 
* As mentioned in the 1st point I plan to add some API like `isTypeMockable`, as it would prove useful for different mock maker implementations for different reasons, android may have different capabilities, and bytebuddy may finally be the trigger to support `final` class mocking.
* Check that ByteBuddy is OSGI compatible, ie doesn't break like in #82 (OSGI stuff)
* Open question do we want to repackage ByteBuddy & ASM.
  CGLIB / ASM were repackaged to avoid class generation collision in internal CGLIB cache, this implementation does not suffer form that, as it has its own class cache.

Before merging if changes are to be done, please make a PR on [bric3/mockito:bytebuddy-mockmaker](https://github.com/bric3/mockito/tree/bytebuddy-mockmaker)

_A step closer to Mockito 2.0 #123_


raphw:
Great to see this! 

As for the versioning: Byte Buddy is still in a 0.X version, simply because I did not yet have the time to implement all features that I see inside the scope of the project. However, I do not want to bloat the library either and Byte Buddy will always be a code generation library only. Currently, I am missing support for defining generic times at runtime (for dealing with frameworks that read and require this information) which will require some minor API changes in Byte Buddy 0.6. However, these API changes do not effect the general API that Mockito uses. 

After introducing this feature - hopefully sometime this year - Byte Buddy promotes to a 1.X version and will guarantee byte-level compatibility for the non-byte-code level API (which Mockito does not use either). Further guarantees cannot be given as Java byte code - other than Java source code - is not backwards-compatible itself and one might need adapt the API to make new byte code features accessible.

Also, as Byte Buddy uses Mockito intensively for its own testing, I am glad that I can fallback to cglib myself. ;) As further information, I recently added Android support for Byte Buddy. It only requires a change of the class loading strategy which is nothing more than changing a line of code. I am happy to add a Byte-Buddy-Android-Mock-Maker for that some time soon, as well.

bric3:
Yeah we are fairly safe on API wise, and if binary compatibility is a problem change we may even repackage bytebuddy. It's undecided but I would like to avoid that.

Also :
* I followed the improvements on bytebuddy to support dex format, that is great. Is it included in the bytebuddy-dep ?
* And finally about android, I believe that mobile developpers have to perform this trick to make dexmaker + mockito work :

  ```java
  // Work around for a known issue with Mockito and Dexmaker
  System.setProperty("dexmaker.dexcache", getInstrumentation().getTargetContext().getCacheDir().getPath());
  ```

  Does it need such trick ?

Now the most important leftovers are packaging / releasing, as my current state of knowledge my gradle skills are rather poor, I may need help from @szczepiq

raphw:
Understandable that you like to avoid it, it is a pain which is another reason why I want to avoid any further API changes. The DEX support is found in its own dependency https://github.com/raphw/byte-buddy/tree/master/byte-buddy-android

You could add the dependency but only use it in case that you are running on Android, thus avoiding class loader issues. As Mockito is mainly used in test environments, this extra dependency should not cause too much trouble. In the code, one would simply need to use the: https://github.com/raphw/byte-buddy/blob/master/byte-buddy-android/src/main/java/net/bytebuddy/android/AndroidClassLoadingStrategy.java instead of using the default subclass loading strategy which is currently used. No property setting necessary.

bric3:
Interesting, however I believe android developers use interchangeably the JVM and dalvik on a daily basis. With the questions on SO, I came to think that dexmaker should know if it should run or not, i.e if Dalvik / ART, then dex code is generated, if the VM is not Dalvik then it could fall back to the default mockmacker. I do not have the hand on dexmaker but if there bytebuddy mockmaker for android that could be a idea that avoids classloading issues your are mentioning.

I think it would be interesting to collect usage from actual developers to get it right (some people I know @pyricau, @a-thomas).

raphw:
I agree, this would not be implicit with the approach I suggested: When the runtime system property suggests that the VM is Dalvik, the Android class loader would used which then creates a `dex` file. Otherwise, the "normal" subclass loader would be used which creates a normal `class` file. At least, this would be my goal for the implementation.

bric3:
Also an idea once this PR is merged is that it would be possible to take the work from @guw on PR #81

pyricau:
This is very interesting (cc @rjrjr). We currently still run our unit tests on a JVM, and run UI tests on device. We do not use mocks for UI tests.

PR #175, opened 2015-02-22
Fixes #172: Updated hamcrest-core from version 1.1 to 1.3

### LINKED ISSUE #172: Update hamcrest-core to version 1.3
Hi there,

It would be nice if mockito-core would depend on hamcrest-core 1.3 instead of 1.1 so that both mockito and junit [latest versions] would depend on the same 3rd party library.

Thanks in advance!

soniro:
Updated the `hamcrest-core` version to `1.3`.

Therefore I needed to add a new method to the `LocalizedMatcher` and change all usages of `assertThat(<interface>, is(<interfaceImpl>))`. IMHO `instanceOf` is fine in the unit tests.

In the ExtraMatchers class, `is` could also be replaced by `instanceOf`. I decided to change the signature of the method. But maybe it is safer (and more user-friendly) to leave the public signature as it is and just change the logic.

Could you please review my solution and give me feedback? Thank you!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -61,7 +61,7 @@ tasks.withType(JavaCompile) {
 //TODO we should remove all dependencies to checked-in jars
 dependencies {
     provided "junit:junit:4.10"
-    compile "org.hamcrest:hamcrest-core:1.1", "org.objenesis:objenesis:2.1"
+    compile "org.hamcrest:hamcrest-core:1.3", "org.objenesis:objenesis:2.1"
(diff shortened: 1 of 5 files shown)


bric3:
Hi @soniro thanks for the pull request :)
However we will drop hamcrest internally in version 2.0.x so anyone can choose the version of hamcrest they see fit :)

See https://github.com/mockito/mockito/issues/172#issuecomment-108542272

PR #176, opened 2015-02-24
Make PropertyAndSetterInjection field sorting consistent

### LINKED ISSUE #155: Internal Comparator violates its general contract
Log from my test run:

java.lang.IllegalArgumentException: Comparison method violates its general contract!
	at java.util.TimSort.mergeHi(TimSort.java:868)
	at java.util.TimSort.mergeAt(TimSort.java:485)
	at java.util.TimSort.mergeForceCollapse(TimSort.java:426)
	at java.util.TimSort.sort(TimSort.java:223)
	at java.util.TimSort.sort(TimSort.java:173)
	at java.util.Arrays.sort(Arrays.java:659)
	at java.util.Collections.sort(Collections.java:217)
	at org.mockito.internal.configuration.injection.PropertyAndSetterInjection.orderedInstanceFieldsFrom(PropertyAndSetterInjection.java:125)
	at org.mockito.internal.configuration.injection.PropertyAndSetterInjection.injectMockCandidates(PropertyAndSetterInjection.java:100)
	at org.mockito.internal.configuration.injection.PropertyAndSetterInjection.processInjection(PropertyAndSetterInjection.java:77)
	at org.mockito.internal.configuration.injection.MockInjectionStrategy.process(MockInjectionStrategy.java:68)
	at org.mockito.internal.configuration.injection.MockInjectionStrategy.relayProcessToNextStrategy(MockInjectionStrategy.java:89)
	at org.mockito.internal.configuration.injection.MockInjectionStrategy.process(MockInjectionStrategy.java:71)
	at org.mockito.internal.configuration.injection.MockInjectionStrategy.relayProcessToNextStrategy(MockInjectionStrategy.java:89)
	at org.mockito.internal.configuration.injection.MockInjectionStrategy.process(MockInjectionStrategy.java:71)
	at org.mockito.internal.configuration.injection.MockInjection$OngoingMockInjection.apply(MockInjection.java:92)
	at org.mockito.internal.configuration.DefaultInjectionEngine.injectMocksOnFields(DefaultInjectionEngine.java:20)
	at org.mockito.internal.configuration.InjectingAnnotationEngine.injectMocks(InjectingAnnotationEngine.java:100)
	at org.mockito.internal.configuration.InjectingAnnotationEngine.processInjectMocks(InjectingAnnotationEngine.java:62)
	at org.mockito.internal.configuration.InjectingAnnotationEngine.process(InjectingAnnotationEngine.java:56)
	at org.mockito.MockitoAnnotations.initMocks(MockitoAnnotations.java:108)

josephw:
#155 is caused by calling `Collections.sort` with a comparator that isn't always transitive, as in the case where sorting by name and sorting by type hierarchy imply different orders.

For example, with fields of type `[Iterable, Integer, Number]`, it gives `Iterable == Integer` and `Iterable == Number`, but also an inconsistent `Integer < Number`.

I'm not sure it's possible to sort consistently with a stateless `Comparator`, so this change switches to a `sortSuperTypesLast` method that first sorts by name and then shuffles any subclasses to after their superclasses. It's slower, but should be deterministic. The existing tests are preserved by leaving behind a dummy `Comparator` that uses the new sort method. New tests make sure that the sorted ordering doesn't depend on the original order.

Fixes #155.

### DIFF
--- a/src/org/mockito/internal/configuration/injection/PropertyAndSetterInjection.java
+++ b/src/org/mockito/internal/configuration/injection/PropertyAndSetterInjection.java
@@ -14,6 +14,7 @@ import org.mockito.internal.configuration.injection.filter.TypeBasedCandidateFil
 import org.mockito.internal.util.collections.ListUtil;
 import org.mockito.internal.util.reflection.FieldInitializationReport;
 import org.mockito.internal.util.reflection.FieldInitializer;
+import org.mockito.internal.util.reflection.SuperTypesLastSorter;
 
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
@@ -124,45 +125,6 @@ public class PropertyAndSetterInjection extends MockInjectionStrategy {
         List<Field> declaredFields = Arrays.asList(awaitingInjectionClazz.getDeclaredFields());
         declaredFields = ListUtil.filter(declaredFields, notFinalOrStatic);
 
-        sortSuperTypesLast(declaredFields);
-
-        return declaredFields;
+        return new SuperTypesLastSorter().sort(declaredFields);
     }
-
-    /**
-     * Sort first by name, then move any fields after their supertypes.
-     */
-    static void sortSuperTypesLast(List<Field> fields) {
-        Collections.sort(fields, compareFieldsByName);
-
-        int i = 0;
-
-        while (i < fields.size() - 1) {
-            Field f = fields.get(i);
-            Class<?> ft = f.getType();
-            int newPos = i;
-            for (int j = i + 1; j < fields.size(); j++) {
-                Class<?> t = fields.get(j).getType();
-
-                if (ft != t && ft.isAssignableFrom(t)) {
-                    newPos = j;
-                }
-            }
-
-            if (newPos == i) {
-                i++;
-            } else {
-                fields.remove(i);
-                fields.add(newPos, f);
-            }
-        }
-    }
-
-    private static Comparator<Field> compareFieldsByName = new Comparator<Field>()
-    {
-        public int compare(Field o1, Field o2)
-        {
-            return o1.getName().compareTo(o2.getName());
-        }
-    };
 }
--- a/src/org/mockito/internal/configuration/injection/PropertyAndSetterInjection.java
+++ b/src/org/mockito/internal/configuration/injection/PropertyAndSetterInjection.java
@@ -33,7 +33,7 @@ import static org.mockito.internal.util.collections.Sets.newMockSafeHashSet;
  *   <li>for each fields of a class in @InjectMocks type hierarchy
  *     <ul>
  *     <li>make a copy of mock candidates
- *     <li>order fields rom sub-type to super-type, then by field name
+ *     <li>order fields from sub-type to super-type, then by field name
  *     <li>for the list of fields in a class try two passes of :
  *         <ul>
  *             <li>find mock candidate by type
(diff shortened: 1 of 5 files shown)


bric3:
Wow, nice PR, thanks for diving in the problem !

So I agree on the idea, but I've mixed feelings on having a custom sort there.
Also I agree speed is not the real issue here anyway, I'd prefer the algorithm to be deterministic.

src/org/mockito/internal/configuration/injection/PropertyAndSetterInjection.java
```diff
@@ -125,27 +124,45 @@ private boolean injectMockCandidatesOnFields(Set<Object> mocks, Object instance,
...
+                }
             }
-            if(field2Type.isAssignableFrom(field1Type)) {
-                return -1;
+
+            if (newPos == i) {
+                i++;
+            } else {
+                fields.remove(i);
+                fields.add(newPos, f);
```
bric3: I'm not comfortable to have a custom sort algorithm I would prefer the use of comparator, it could be stateful of course.

bric3: Also I didn't thought about that, but changing the type to a `TreeSet` may be interesting.

test/org/mockito/internal/configuration/injection/PropertyAndSetterInjectionTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+        for (Field[] o : initialOrderings) {
+            List<Field> l = new ArrayList<Field>(Arrays.asList(o));
+
+            PropertyAndSetterInjection.sortSuperTypesLast(l);
+            results.add(l);
+        }
+
+        assertThat(results).hasSize(1);
+    }
```
bric3: nice test !

bric3:
Finally I like the comment in the commit messages, though it would be better if the last 2 commit were amended properly :)

bric3:
Overall it's really nice work :sparkles: 

I you think it's possible to rewrite with a comparator and rewrite a clean history I'll merge _asap_ :)

josephw:
Thanks; I've rebased to clean up history.

Using `Comparator` is trickier. I'd really prefer to do that, and I spent some time thinking about what state could be calculated to allow pairwise comparison. However, this is essentially a topological sort, and I can't think of a way to do this short of using a custom sort and then preparing a `Comparator` that checks indexes in that version.

Actually, checking now, that's exactly the fallback suggested in [this Stackoverflow answer](http://stackoverflow.com/questions/28099307/how-to-write-a-transitive-comparator-when-equality-implies-order-doesnt-matt):

> The only way I see that it would be possible to do using a Comparator would be to first do a topological sort, and then using the indexes of the objects in this sorting as keys when implementing the comparator.

I'd be happy to add any more tests, or change the style to increase confidence that it's a safe implementation.

bric3:
Mmmmh OK, it would definitely be better to have the comparator, but that seems a bit complex as you said. So I propose 2 things then :

1. Without comparator it would be better to externalize the sorting in some class instead of a method, could be used like 

    ```java
    sorted = new org.mockito.internal.util.reflection.TopologySorter().sort(...)
    ```

2. Thanks for the amend, I propose that the proper commits starts `Fixes #155, ` like you did as github can track such keyword and issue reference.

Sorry to be nitpicky, but as soon as the PR is merged there will be an automated release :)
Also tests are fine.

Thanks for the feedback, and the interesting links :)

bric3:
:star2: 

Thanks for the modifications @josephw 
Perfect PR :)

bric3:
By the way do you have any remarks on the contributing guide (missing info, style guideline, git stuff, etc..)

josephw:
It's great that you've got a contribution guide. I think a couple of the things I missed are mentioned by it but, I might over-emphasise:

>  At least one commit message in the PR contains "Fixes #id"

and suggest a template commit message.

> Clean commit history

You could make the history rewriting explicit here. Something like 'We prefer a clean commit history, rebased and squashed if necessary, so that each commit clearly changes one things and there are no extraneous fix-ups'?

Thanks for the feedback -- everything else seemed specific enough that a review was the only way to pick it up.

bric3:
@josephw Ok thanks for the feedback too

PR #180, opened 2015-03-15
Comparison Should Use Equals Method Instead of ==

saden1:
I am writing framework that relies on Mockito which requires proxying mocks during concurrent test runs. The comparison operator in MockHandlerImpl does not give me the opportunity to intercept equality check during the verification process. I would like to propose changing the test from == to equals() which would give me the opportunity to intercept calls to equals() and to do a proper comparison with my proxy's underlying mock instance.

### DIFF
--- a/src/org/mockito/internal/handler/MockHandlerImpl.java
+++ b/src/org/mockito/internal/handler/MockHandlerImpl.java
@@ -67,7 +67,7 @@ class MockHandlerImpl<T> implements InternalMockHandler<T> {
         if (verificationMode != null) {
             // We need to check if verification was started on the correct mock
             // - see VerifyingWithAnExtraCallToADifferentMockTest (bug 138)
-            if (((MockAwareVerificationMode) verificationMode).getMock() == invocation.getMock()) {
+            if (((MockAwareVerificationMode) verificationMode).getMock().equals(invocation.getMock())) {
                 VerificationDataImpl data = createVerificationData(invocationContainerImpl, invocationMatcher);
                 verificationMode.verify(data);
                 return null;


bric3:
Equals may work here because it's not mocked. Unfortunately though, mocking `equals`, `hashCode` though mockito is not possible b/c it's used internally. And especially inside the `MockHandler` which is a sensible part of the mock handling, and mocking equals there _feels_ a bit hazardous at best.

I know it's about proxies of mockito mocks (which are proxies as well), but I can't help have miixed feeling on this one.

saden1:
By default using equals() instead of == results in the delegation of equality test to  MethodInterceptorFilter which simply resorts to using [==] (https://github.com/mockito/mockito/blob/6d665caedab2366313c8e14a770077383c206c14/src/org/mockito/internal/creation/cglib/MethodInterceptorFilter.java#L45) so this change will not have any impact on existing codebase.  

The change proposed amounts to doing the following:
```java
ProxyedInterface mock = mock(ProxyedInterface.class);
//handled by MethodInterceptorFilter and is equivalent to performing mock == mock
assert true == mock.equals(mock); 
```

bric3:
> simply [resorts to using ==](https://github.com/mockito/mockito/blob/6d665caedab2366313c8e14a770077383c206c14/src/org/mockito/internal/creation/cglib/MethodInterceptorFilter.java#L45)

I am well aware of that, the point is that the method interceptor is _internal_ to the `MockMaker`. Our own mock maker can certainly work as the build of this PR shows, but `MockMaker` can be provided by other third parties like _dexmaker-mockmaker_, that may or may not respect such non obvious contract.

saden1:
I wasn't aware of the MockMaker extension point. My understanding is that most Type systems including [Java](http://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals(java.lang.Object)) and [Android](http://developer.android.com/reference/java/lang/Object.html#equals(java.lang.Object)) have a well defined equals() contract with reference == test being the most discriminating possible equivalence test employed first. 

Any implementation of MockMaker has to return true for "mock.equals(mock)" and therefore must correctly implement and handle calls to equals(). Looking at the DexMaker code base it appears that it does correctly [handle calls to equals](https://android.googlesource.com/platform/external/dexmaker/+/master/src/mockito/java/com/google/dexmaker/mockito/InvocationHandlerAdapter.java).

Bottom line, this change shouldn't have any impact on existing systems and if it does that implies those systems can not satisfy "mock.equals(mock) == true" and therefore broken!

saden1:
@bric3, will this pull request be accepted? Any time frame on when it will be merged?

bric3:
Sorry I got caught in something else at work.

Sorry it's a no go.
Even if dexmaker is doing it correctly at the moment, code can change and there's no guarantee another mock maker will behave correctly. If something is wrong this could lead to side effects difficult to identify. Maybe later if the equality code can be improved this could change.

By the way I don't if it can help but I remember facing a similar issue, to get around that I used a specific collection that wrapped the object, this wrapped had custom `equals` and `hashcode` methods.

saden1:
I truly don't understand your objects and concerns regarding this pull. In fact, I have discovered that Mockito already using the [equals method in line 61 of InvocationMatcher.matches](https://github.com/mockito/mockito/blob/master/src/org/mockito/internal/invocation/InvocationMatcher.java#L61) method to test for equality and this change does exactly the same thing.

Please reconsider.

saden1:
@szczepiq can you chime in on this please.

bric3:
> Mockito already using the equals method in line 61 of InvocationMatcher.matches method

Just because it is right in our own and in the current dexmaker version doesn't mean other mockmaker are / will be correctly implemented. At the moment a correct equality check in a handler cannot be enforced in a mock maker in the current base code.

I understand it may feel over protective / defensive. Still I'm open to suggestions to make mockito more resistant to unreliable mock makers, that includes raising an error when one is detected.

saden1:
I understand and appreciate being protective but the rejection of this pull goes way beyond that. By rejecting this pull you are saying "mockito can never and will never run in a concurrent environment on the off chance that a third-party developer incorrectly implements the equals contract." To me that's shortsighted and baffling in light of your existing usage of the equals contract elsewhere in the codebase.

I'm trying to enhance mockito here not break things and this request literally has zero impact on existing code base.

saden1:
For clarification here's the existing usage of equals() in InvocationMatcher.matches(..) :
```java 
public boolean matches(Invocation actual) {
        return invocation.getMock().equals(actual.getMock())
                && hasSameMethod(actual)
                && new ArgumentsComparator().argumentsMatch(this, actual);
}
```

Here's the proposed change MockHandlerImpl.handle(...) in the pull that simply wants to do the same thing:
```java
            if (((MockAwareVerificationMode) verificationMode).getMock().equals(invocation.getMock())) {
                VerificationDataImpl data = createVerificationData(invocationContainerImpl, invocationMatcher);
                verificationMode.verify(data);
                return null;
            } 
        }
```

The proposed change is simply trying to do what you already do.

bric3:
Mockito mocks already work in a multithreaded environment (just not the stubbing).
About shortsight, I do not agree, but let's dive in here. 

> zero impact on existing code base

That is true for the mockito code base only, there are others.

saden1:
If there are others that could be impacted then wouldn't that imply InvocationMatcher.matches method is broken?

bric3:
I agree `InvocationMatcher.matches` is subject to the issue I mentioned.

bric3:
What if we replaced **all** mock equality checks in the mockito base code, by a single new class responsible for that, a `MockEqulityAndHashcode` that will be an extension point in [`Plugins`](https://github.com/mockito/mockito/blob/master/src/org/mockito/internal/configuration/plugins/Plugins.java). This extensioin point would be used check equality between two mocks something like `Plugins. getMockEqulityAndHashcode().equals(mockA, mockB)`.

There one could override equality the way he want.

bric3:
Note there's some big changes comming in 2.x regarding the mockmaker.

saden1:
If you want to be explicit and codify equals and hashcode contracts for third party plugin developers that's great. For me, as long as I have the ability intercept equality tests I'll be a happy camper.

With the proposed changes will mockito by default use equals rather than == to check for equality?

saden1:
FYI, running the following test case does not currently work when using surefire's "methods" based parallelism. It won't work for many reasons but one of them an issue with verification equality check. Now imagine if there were TestNG/JUnit hooks that could inspect and inject the class with proxies of the mocks and enable parallelism. This is exactly what I am trying to do.


```xml
<plugin>
    <version>2.18.1</version>
    <configuration>
        <parallel>methods</parallel>
        <threadCount>4</threadCount>
        <includes>
            <include>**/*Test.java</include>
        </includes>
    </configuration>
</plugin>
```
```java
import static org.mockito.BDDMockito.given;
import org.mockito.Mock;
import org.mockito.MockitoAnnotations;
import static org.testng.Assert.assertEquals;
import org.testng.annotations.BeforeClass;
import org.testng.annotations.Test;

public class ConcurrentNGTest {

    @Mock
    Messenger sut; //has to be proxied 

    @BeforeClass 
    public void init() {
       //won't need to do this as it will be handled by TestNG/Junit Listener that does class inspection
       //and injection of proxies
        MockitoAnnotations.initMocks(this);
    }

    @Test
    public void test1() {
        String message = "test1";
        given(sut.speak()).willReturn(message);
        String result = sut.speak();
        verify(sut).speak();
        assertEquals(message, result);
    }

    @Test
    public void test2() {
        String message = "test2";
        given(sut.speak()).willReturn(message);
        String result = sut.speak();
        verify(sut).speak();
        assertEquals(message, result);
    }

    public static class Messenger {

        public String speak() {
            return "hello";
        }
    }
}
```

bric3:
> With the proposed changes will mockito by default use equals rather than == to check for equality?

It will probably use the current equality check i.e. reference equality.

Also about the concurrent test execution, i falls exactly in what I wrote earlier :

> Mockito mocks already work in a multithreaded environment (just not the stubbing).

What fails is most probably the stubbing. And this Surefire configuration is running methods in multiple threads, with a mock shared among threads for his whole life cycle (stub, record, verification), mocks can be shared only after stubbing.

Actually I don't recommend the `parallel` settings anyway, because it's single process with multiple threads, it uses probably the same classloader, and there's not only mockito that can be affected by race conditioin issues. It depends on the code of course, but if going for `parallel` I'd prefer the following mode : `classes`. Otherwise I would definitely go for the **fork** mode (documented [here](https://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html)), e.g.

```xml
  <plugin>
    <version>2.18.1</version>
    <configuration>
        <forkCount>2C</forkCount>
        <reuseForks>true</reuseForks>
    </configuration>
  </plugin>
```

saden1:
> It will probably use the current equality check i.e. reference equality.

Is it fair to say:
a) when MockEqulityAndHashcode is introduced it will be used throughout the code base.
b) the default mockito MockMaker will resort to using reference equality checks.
c) anyone wishing to intercept equality calls will need to implement their own MockMaker 


p.s. Will you create an issue to track progress on MockEqulityAndHashcode?

PR #184, opened 2015-03-21
Fixes typo

hansjoachim:


hansjoachim:
Travis:
>Execution failed for task ':test'.

Hm, I don't see how my change would break the build. (I did run it locally too, just in case, but wasn't able to reproduce this issue.) Is there any way to get some more detailed information on what went wrong?

bric3:
Yes no idea, it clearly is not related  to a mardown file ;)

bric3:
Thx for the typo fix btw :)

PR #192, opened 2015-04-02
Added a default answer for java.util.Optional

### LINKED ISSUE #191: Mock returning java.util.Optional should return Optional.empty() by default (Java 8)
(From the discussion at https://code.google.com/p/mockito/issues/detail?id=497 )

> A stubbed method that returns `Optional<?>` should return `Optional.empty()` if no explicit behavior is defined. This is equivalent to returning null for regular methods.
> 
> Introducing this feature would add no extra dependencies. However, it would require a test whether the code is running in Java 8.

Since the thread ends with the request for a GitHub ticket and I couldn't find one... here it is :)

phjardas:
I've introduced a new default answer for `java.util.Optional` that returns an empty optional.

The code uses reflection, so it should compile and run cleanly in older versions than Java 8.

--------------------------------------
**EDIT by mockito team** : fixes #191

### DIFF
--- a/src/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValues.java
+++ b/src/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValues.java
@@ -5,14 +5,9 @@
 
 package org.mockito.internal.stubbing.defaultanswers;
 
-import org.mockito.internal.util.MockUtil;
-import org.mockito.internal.util.ObjectMethodsGuru;
-import org.mockito.internal.util.Primitives;
-import org.mockito.invocation.InvocationOnMock;
-import org.mockito.mock.MockName;
-import org.mockito.stubbing.Answer;
-
 import java.io.Serializable;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
@@ -27,6 +22,15 @@ import java.util.SortedMap;
 import java.util.SortedSet;
 import java.util.TreeMap;
 import java.util.TreeSet;
+
+import org.mockito.internal.creation.instance.InstantiationException;
+import org.mockito.internal.util.JavaEightUtil;
+import org.mockito.internal.util.MockUtil;
+import org.mockito.internal.util.ObjectMethodsGuru;
+import org.mockito.internal.util.Primitives;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.mock.MockName;
+import org.mockito.stubbing.Answer;
 
 /**
  * Default answer of every Mockito mock.
@@ -45,6 +49,9 @@ import java.util.TreeSet;
  * </li>
  * <li>
  *  Returns zero if references are equals otherwise non-zero for Comparable#compareTo(T other) method (see issue 184)
+ * </li>
+ * <li>
+ *  Returns an {@link java.util.Optional#empty() empty Optional} for Optional (see issue 191).
  * </li>
  * <li>
  *  Returns null for everything else
@@ -116,8 +123,11 @@ public class ReturnsEmptyValues implements Answer<Object>, Serializable {
         }
         // TODO return empty Iterable ; see issue 175
 
+        if ("java.util.Optional".equals(type.getName())) {
+        	return JavaEightUtil.emptyOptional();
+        }
+        
         //Let's not care about the rest of collections.
         return null;
     }
-
 }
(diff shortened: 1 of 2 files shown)


bric3:
For reference this would help on #191 .

Thanks for the PR :sunglasses: 
A nice start to suport Java 8, I had something like that in mind, however I'd like to go a bit further :

- On types because there's other Java 8 types as well
- Introducing something like `Platform.saneDefaults()` that provide empty values, but maybe other _service_ as well

I won't have time until next weeks though to work on it.

phjardas:
What other Java 8 types would you like to support? Maybe `java.util.Stream`? I can't really see default null values for the new `java.time` types.

I personally really dislike the class `JavaEightUtil` that I introduced, but couldn't think of a better way to encapsulate the reflection magic.

And I have no idea why the build is failing. Works on my machine, as always, and the travis output is not really helpful. :-/

I'm going to be on a two-week vacation from now on, but I would like to help after that. :-)

bric3:
Don't worry about the build, it's broken for a unknown reason.

Also I'm quite overwhelmed at the moment, 2 real life projects are in a critical phase. So I won't be able to work on this as soon as possible, and as I would like.

> What other Java 8 types would you like to support? Maybe java.util.Stream? I can't really see default null values for the new java.time types.

Actually I would like to first return default values for `Iterable`, `Iterator`, `Stream`, `BigInteger`, `BigDecimal`.

For time related classes, I have mixed feelings, I'd like to not return `null` but I'm not sure about the default value, `now` seems reasonable though. Thoughts ?

phjardas:
I guess the collection and number types should be quite easy. I can work on that, adding support by reflection depending on the Java version.

Looking at the date and time classes, i don't see a sensible default value. "now" is not the right way to go IMHO. It feels like always returning 1 for an integer... I guess `null` is quite ok, possibly failing tests to force people to actually mock this specific method.

I'll also have a look at how much trouble would it be to upgrade the build process to Java 8 while maintaining backwards compatibility.

bric3:
I wanted to work on such things in few weeks. I wanted to have integration modules, with possible different versions of library JDK. Specifically of the JDK 8 I had this idea of using [JavaVersion](http://gradle.org/docs/current/javadoc/org/gradle/api/JavaVersion.html).

```java
JavaVersion.current().isJava8Compatible()
```

and disable the module if false, I don't know well gradle, maybe something like could work

```java
subprojects {
    afterEvaluate { project ->
        if (project.name.equals("java8-integration-tests")) 
            && !JavaVersion.current().isJava8Compatible()) {
            test.disabled = true
        }
    }
}
```

phjardas:
Back from vacation. :-)

I don't think it is advisable to do the destinction at build time. No need for gradle from my point of view.

How about introducing the following interface:

    public interface DefaultAnswerProvider {
        boolean isApplicable(Class<?> type);
        <T> T provideDefaultAnswer(Class<T> type);
    }

All the different strategies to create empty values in `ReturnsEmptyValues` would be moved to one individual implementation of this interface. For example:

    public class PrimitiveOrWrapperDefaultAnswerProvider implement DefaultAnswerProvider {
        public boolean isAcceptable(Class<?> type) {
            return Primitives.isPrimitiveOrWrapper(type);
        }

        public <T> T provideDefaultAnswer(Class<T> type) {
            return Primitives.defaultValueForPrimitiveOrWrapper(type);
        }
    }

`ReturnsEmptyValues` would contain a list of these providers, select the first one applicable to a given type and return the default answer.

Any provider that requires an external library or specific java version (`OptionalDefaultAnserProvider`, `Stream` etc.) would use reflection to check for acceptability (`"java.util.Stream".equals(type.getName())`) and for the creation of the default answer.

Throw in some caching to speed up provider resolution.

This approach has the following benefits:
* separates different strategies to create default values
* keeps nasty reflection logic isolated
* allows default values for external libraries, e.g. `commons-collections`

I can think of more ways to enhance this:
* allow users to register their own default answer providers
* allow JARs to automatically register default answer providers by use of the SPI. This would lead to the possibility for third-party libraries to extend mockito like `mockito-java8.jar` or `mockito-commons-collections.jar`

I would be happy to contribute, if you are interested.

bric3:
@phjardas Sorry for late reply. Yes for runtime I had this idea ;)

But I was specifically speaking about a java 8 integration testing

phjardas:
Allright, then I'll just wait. If you need any help, let me know! :-)

bric3:
Reopening, so we'll keep track of it, when done

TimvdLippe:
This has landed in master with https://github.com/mockito/mockito/commit/c992dde1cedbb890b6bb56493e2e67926398445d

PR #201, opened 2015-04-16
Fixes #200 : ArgumentCaptor.forClass is more friendly with generic types

### LINKED ISSUE #200: ArgumentCaptor.fromClass's return type should match a parameterized type
`ArgumentCaptor.fromClass`'s return type should match a parameterized type.  I.e. the expression `ArgumentCaptor.fromClass(Class<S>)` should be of type `ArgumentCaptor<U>` where `S` is a subtype of `U`.   

For example:

    ArgumentCaptor<Consumer<String>> captor = ArgumentCaptor.fromClass(Consumer.class)

does not type check (i.e. it is a compile time error). It should type check. 

The reasons that it is desirable for `ArgumentCaptor.fromClass` to allow expressions such as the example above to type check are:

1) `ArgumentCaptor.fromClass` is intended to be a convenience method to allow the user to construct an ArgumentCaptor without casting the returned value.

Currently, the user can devise a workaround such as: 

    ArgumentCaptor<? extends Consumer<String>> captor 
    = ArgumentCaptor.fromClass(Consumer.class)

This workaround is inconvenient, and so contrary to `ArgumentCaptor.fromClass` being a convenience method.

2) It is inconsistent with `@Captor`, which can be applied to a field with a paramterized type.  I.e.

    @Captor ArgumentCaptor<Consumer<String>> captor 

type checks.

bruceeddy:
We modify the signature of `ArgumentCaptor.fromClass(Class<T>)` to better express the type relationship between the argument and the return type.  There is no change to the behavior of the method.  This change is to allow expressions such as:

```java
ArgumentCaptor<Consumer<String>> captor = ArgumentCaptor.fromClass(Consumer.class)
```

to type check, which is desirable as a convenience to users of `ArgumentCaptor`.

We also add two tests to document this extension to the api, both of which fail to compile without this change to `ArgumentCaptor.fromClass`.  These tests are unusual in that they make no assertions; this is justified because the change which they test is a change to the formal, or compile-time, properties of `ArgumentCaptor`, not its behavior.

### DIFF
--- a/src/org/mockito/ArgumentCaptor.java
+++ b/src/org/mockito/ArgumentCaptor.java
@@ -65,7 +65,7 @@ public class ArgumentCaptor<T> {
     HandyReturnValues handyReturnValues = new HandyReturnValues();
 
     private final CapturingMatcher<T> capturingMatcher = new CapturingMatcher<T>();
-    private final Class<T> clazz;
+    private final Class<? extends T> clazz;
 
     /**
      * @deprecated
@@ -87,7 +87,7 @@ public class ArgumentCaptor<T> {
         this.clazz = null;
     }
 
-    ArgumentCaptor(Class<T> clazz) {
+    private ArgumentCaptor(Class<? extends T> clazz) {
         this.clazz = clazz;
     }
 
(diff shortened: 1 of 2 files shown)


bric3:
Almost perfect PR :)
Maybe the contributing guide is missing something here, but the exemple `#id` should have been expended to the real GH issue number. Actually the commit message should be something like :

```
Fixes #200 : Tweak generic signature of ArgumentCaptor.fromClass
```

I'll merge it asap :)

Thanks for the PR by the way !

bric3:
I updated the contributing guide with an exemple :)

If you're unsure how to rename the commit :

```bash
git commit --amend --message="Fixes #200 : Tweak generic signature of ArgumentCaptor.fromClass"
git push --force
```

bric3:
The travis build broken in some way, your code is fine, thanks for the change :)

Issue #187, opened 2015-03-24
java.lang.ClassCastException: java.lang.Class cannot be cast to java.lang.String


mktitov:
Exception throws on verifyZeroInteractions when using mock with default answer.
checked on versions 1.10.5-2.0.5
all ok on 1.9.5

bric3:
Hi,

Do you have a JUnit test so that we could reproduce.
Also please see the [contributing guide](https://github.com/mockito/mockito/blob/master/CONTRIBUTING.md).

mktitov:
public class MockitoTest extends Mockito {
    public interface TestMock {
        public boolean m1();
    };
    
    @Test
    public void test() {
        TestMock test = mock(TestMock.class, new Answer() {
            @Override public Object answer(InvocationOnMock invocation) throws Throwable {
                return false;
            }
        });
        test.m1();
        verifyZeroInteractions(test);
    }
}

mktitov:
```
public class MockitoTest extends Mockito {
    public interface TestMock {
        public boolean m1();
    };
    
    @Test
    public void test() {
        TestMock test = mock(TestMock.class, new Answer() {
            @Override public Object answer(InvocationOnMock invocation) throws Throwable {
                return false;
            }
        });
        test.m1();
        verifyZeroInteractions(test);
    }
}
```

bric3:
Hi, sorry for the late reply.

Thanks for the testcase, indeed we missed something there.

bric3:
Actually the usage report was improved to include the mock name, however in this usage the default answer always returns `false`. So when the String message is generated there's a CCE because the default answer returns a `boolean`.

On your side the workaround is to return valid values for these `Object` inherited methods.

On mockito side there should be two possible thing to do:
 - get the name safely without invoking `toString`
 - validate more aggressively answer's result

puneethpgowda0825:
Hello, I've one doubt, can we do casting for mocking objects?. When I tried to do that I got "java.lang.ClassCastException".

PR #202, opened 2015-04-19
Better protection against incompatible returned value of default answer, and get safely mock name

bric3:
Fixes issue #187

The value returned by the default answer is now validated to make sure there won't be a `ClassCastException` if the default value is incompatible with return type.

Reporter now fetch in safer way the mock name, in case the default answer is incorrectly implemented. This avoid a `ClassCastException` when reporting a verification issue.


bric3:
Build is ok, but fail on travis.

PR #207, opened 2015-05-08
Fixes #197 : Blocks ability to use negative value for timeout() and after() method.

### LINKED ISSUE #197: Mockito.after() method accepts negative timeperiods and subsequent verifications always pass
e.g.

    Runnable runnable = Mockito.mock(Runnable.class);
    Mockito.verify(runnable, Mockito.never()).run(); // passes as expected
    Mockito.verify(runnable, Mockito.after(1000).never()).run(); // passes as expected
    Mockito.verify(runnable, Mockito.after(-1000).atLeastOnce()).run(); // passes incorrectly

lukasz-szewc:
It should fix issue #197. 

If user uses Mockito.after() or Mockito.timeout() with negative values, then friendly reminder exception with details will be thrown.

### DIFF
--- a/src/org/mockito/exceptions/Reporter.java
+++ b/src/org/mockito/exceptions/Reporter.java
@@ -816,6 +816,15 @@ public class Reporter {
         throw new MockitoException("Mocks instantiated with constructor cannot be combined with " + mode + " serialization mode.");
     }
 
+    public void cannotCreateTimerWithNegativeDurationTime(long durationMillis) {
+        throw new FriendlyReminderException(join("",
+                "Don't panic! I'm just a friendly reminder!",
+                "It is impossible for time to go backward, therefore...",
+                "You cannot put negative value of duration: (" +  durationMillis +  ")",
+                "as argument of timer methods (after(), timeout())",
+                ""));
+    }
+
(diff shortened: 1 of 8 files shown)


lukasz-szewc:
First commit fixes the issue, second one is a little bit of refactoring: (mostly reducing visibility and increasing encapsulation).

Fixes #197 : Refactoring: 
- Removed unused field from VerificationOverTimeImpl and also from constructor's signature.
- improved encapsulation by removing 3 getters from VerificationOverTimeImpl.
- reduced visibility of few constructors in 'Timeout' and 'After' classes

bric3:
Nice pull request, thank you !
If comments are addressed I'll merge right away !

lukasz-szewc:
About TimeoutTest... Please look at it. The whole point of this test is to pass timer object and do mocking/verifications on it. If I use public constructor for this test I will not to able to pass timer object.
This is why constructor with this signature Timeout(long pollingPeriodMillis, VerificationMode delegate, Timer timer) was introduced by author in first place... 
Should I really touch it?

bric3:
Oh yeah I see, fair point. Just ignore what I said on that.

PR #208, opened 2015-05-11
Added .gitignore

naimdjon:


bric3:
There's already [`.gitignore`](https://github.com/mockito/mockito/blob/master/.gitignore) in `master` ;)

naimdjon:
Is it possible to cherry-pick it to 1.9.5 since many developers out there would be still using a stable version.

bric3:
Latest stable version is 1.10.19

PR #215, opened 2015-06-02
Improves InjectMocks behavior when injectee has multiple fields of the same type

### LINKED ISSUE #205: InjectMocks injects mock into wrong field
Using `1.10.19`.

When using `@InjectMocks` on some Android `TextView`s, the mock is injected into the wrong field.

We have two fields, `txtGateView` & `txtNextStep` in a class, and our test mocks out `txtNextStep`, then tried to inject. This field is injected wrong, see screenshot.

image

From our quick testing, the name `txtNextView` doesn't matter, that can be changed. But both `txtGateView` and `txtGateLabel` messed things up. If we mock out both fields, it works correctly.

Testproject: https://github.com/SimenB/emptyandroid

I don't know if it's because it's Android, but it was easiest for me to create a minimal test from existing code.

JeremybellEU:
Fix to @InjectMocks which injected incorrectly when multiple fields
of the same type could be injected into, but only supplying 1 mock.
This makes sure that whenever there are multiple type candidates,
it will hold off injecting if it finds a different field which has the correct matching name.


bric3:
Will fix issue described in #205

PR #219, opened 2015-06-04
Add .gitattributes to enforce LF

SimenB:
Noticed the massive https://github.com/mockito/mockito/commit/d26f5db9023b6b7cd331017551ea4353fff0ef91.

This allows you to let git handle it.

It seems some files are still wrong. See here for a way to fix all https://help.github.com/articles/dealing-with-line-endings/

There might be a valid reason for not doing this, if so just close it :smile:

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# Auto detect text files and perform LF normalization
+* text=auto
+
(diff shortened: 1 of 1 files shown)


bric3:
Nice thanks, I wanted to do that at some point.

Issue #206, opened 2015-05-07
NoJUnitDependenciesTest is failing on Windows machine.


lukasz-szewc:
I found NoJUnitDependenciesTest unit test is the only unit test that fails when you run the build on windows machine (I have also Ubuntu, and it is fine for some reason on Unix platform).

Root cause: Caused by: java.lang.NoClassDefFoundError: org\mockito\internal\invocation\InvocationMarker (wrong name: org/mockito/internal/invocation/InvocationMarker)

Full Stacktrace:

java.lang.AssertionError: 'org\mockito\internal\invocation\InvocationMarker' has some dependency to JUnit
	at org.mockitointegration.NoJUnitDependenciesTest.checkDependency(NoJUnitDependenciesTest.java:36)
	at org.mockitointegration.NoJUnitDependenciesTest.pure_mockito_should_not_depend_JUnit(NoJUnitDependenciesTest.java:28)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:601)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:45)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:15)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:42)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:20)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:263)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:68)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:47)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:231)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:60)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:229)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:50)
[...]

PR #223, opened 2015-06-06
[#206] Fix issue related to windows path

alberskib:
Problem was caused by file separator. Replacing hardcoded char '/' by File.seperatorChar solves issue. In the same file there is second occurence of this character but it is used with getResource(path) method -  in such case '/' is only acceptable separator.


alberskib:
Do I need to do something special in order to build with java 8? It looks like problem with default method - I create branch from master.

bric3:
Don't worry about that errors, I'll push the fix at some point today.

Thanks for fixing the windows test issue :)

PR #227, opened 2015-06-07
Fixed over one thousand warnings

JeremybellEU:
Originally the repository contained over 1100 warnings, this has been reduced to 47.

Most of the fixes are simple suppression warning fixes/additions or imports. I also auto-added (<3 Eclipse save-actions) the final modifiers to speed up the compiler. Therefore the test suite should run a bit faster now.

The number of files changes is insanely high and I used (and fixed) the Checkstyle configuration in the repository. I hope this is still the desired configuration.

### DIFF
--- a/.checkstyle
+++ b/.checkstyle
@@ -1,15 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<fileset-config file-format-version="1.2.0" simple-config="false">
-    <local-check-config name="mockito-code" location="conf/checkstyle-code.xml" type="project" description="">
-        <additional-data name="protect-config-file" value="false"/>
-    </local-check-config>
-    <local-check-config name="mockito-test" location="conf/checkstyle-test.xml" type="project" description="">
-        <additional-data name="protect-config-file" value="false"/>
-    </local-check-config>
-    <fileset name="mockito-code" enabled="true" check-config-name="mockito-code" local="true">
-        <file-match-pattern match-pattern="^src[/\\].*java$" include-pattern="true"/>
-    </fileset>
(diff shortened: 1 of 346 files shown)


bric3:
Well, thanks for caring on the code base, but too much `final` changes, 99% of them are not necessary.

And too much changes intertwined, I think it's been a while since we abandoned checkstyle. Maybe we should remove it from the project or updates its rules. And anyway experience taught me to pay attention to PMD and / or findbugs rather than checkstyle.

So I'd rather skip this change.

JeremybellEU:
Yes I completely understand that :)

I just saw all the errors in my IDE and lost track if I produced errors myself.

If you dont look at the errors then I think it is a good idea to remove Checkstyle completely

Aside from that, afaik the compiler does optimize using the final modifier and it does increase overall speed up as it does not need to check for modifications constantly.

Shall I close the PR?

szpak:
> Aside from that, afaik the compiler does optimize using the final modifier and it does increase overall speed up as it does not need to check for modifications constantly.

I have never tested that, but despite I personally like using 'val' in Scala very much in Java it often descreses readability which in many cases can be more important.

JeremybellEU:
Yes it boils down to developer preference then, so I will go with the majority :)

bric3:
> Aside from that, afaik the compiler does optimize using the final modifier and it does increase overall speed up as it does not need to check for modifications constantly.

Substantial gains are only when there's inheritance involved, interesting impacts are inlining, method dispatch.

bric3:
Actually the core team (i.e. me and Szczepan) uses IntelliJ and don't use checkstyle.

Yes you can close the PR.

In another PR I'll probably accept your generic signature fixes.

PR #229, opened 2015-06-11
Fixes #228: fixed a verify() call example in @Captor javadoc

### LINKED ISSUE #228: @Captor javadoc contains a wrong call example
In the javadoc of `@Captor`:
```
verify(mock.doStuff(captor.capture()));
```
which is incorrect.

eugene-ivakhno:


bric3:
Thanks for the fix :)

PR #238, opened 2015-06-21
Moves responsibility of isTypeMockable to MockMaker

bric3:
Mock makers may have different capabilities. Historically CGLIB and Mockito were deeply related, later Mockmaker was created to allow Android developers to use dexmaker to create mocks. But logic that decided if a type could be mocked still reside in the mockito base, while it should be the responsibility of the mockmaker.

Also for example one could create his own mockmaker that forbid to mock type if some annotation is present.


szczepiq:
Perfect. Thanks!

I've added the incompatibility label so that we can generate migration guide automatically.

bric3:
OK, I'm still polishing this one a bit.

PR #242, opened 2015-06-22
Upgraded to Byte Buddy 0.6.11 and took improved features in use.

raphw:
Over the recent versions, several new features were added to Byte Buddy:

1. An interceptor is now automatically detecting if a default interface method can be invoked if no super method exists.
2. `@Origin` annotated methods are now automatically cached.
3. A field value can be read without generating a proxy using the `@FieldValue` annotation in order to avoid the overhead of the accessor.
4. Replaced Mockito's `SearchingClassLoader` with Byte Buddy's `MultipleParentClassLoader` to avoid the overhead of multiple class loader creation as class loaders are rather heavy objects for a VM.
5. Removed duplicated interception methods.
6. Added `@StubValue` to return correct default value for intercepted constructors.
7. Generified several `Class` values in order to make the project compile against Byte Buddy's `Class<?>` values.
8. Updated *out of sync* OSGi manifest.

Also: Why does Mockito explicitly set the Java version for generated classes to version 5? If not specified explicitly, Mockito discovers the JVM version and generates the newest possible version". This way, Byte Buddy can support Java 8 default methods on demand.

### DIFF
--- a/mockmaker/bytebuddy/main/java/org/mockito/internal/creation/bytebuddy/MockBytecodeGenerator.java
+++ b/mockmaker/bytebuddy/main/java/org/mockito/internal/creation/bytebuddy/MockBytecodeGenerator.java
@@ -4,6 +4,7 @@ import net.bytebuddy.ByteBuddy;
 import net.bytebuddy.ClassFileVersion;
 import net.bytebuddy.dynamic.DynamicType;
 import net.bytebuddy.dynamic.loading.ClassLoadingStrategy;
+import net.bytebuddy.dynamic.loading.MultipleParentClassLoader;
 import net.bytebuddy.dynamic.scaffold.subclass.ConstructorStrategy;
 import net.bytebuddy.implementation.FieldAccessor;
 import net.bytebuddy.implementation.MethodDelegation;
@@ -12,7 +13,6 @@ import net.bytebuddy.implementation.attribute.TypeAttributeAppender;
 import org.mockito.internal.creation.bytebuddy.ByteBuddyCrossClassLoaderSerializationSupport.CrossClassLoaderSerializableMock;
 import org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.DispatcherDefaultingToRealMethod;
 import org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.MockAccess;
-import org.mockito.internal.creation.util.SearchingClassLoader;
 
 import java.util.Random;
 
@@ -49,16 +49,14 @@ class MockBytecodeGenerator {
                              .intercept(to(MockMethodInterceptor.ForWriteReplace.class));
         }
         return builder.make()
-                      .load(SearchingClassLoader.combineLoadersOf(allMockedTypes(features)), ClassLoadingStrategy.Default.INJECTION)
+                      .load(new MultipleParentClassLoader.Builder()
+                              .append(features.mockedType)
+                              .append(features.interfaces)
+                              .append(MultipleParentClassLoader.class.getClassLoader())
+                              .append(Thread.currentThread().getContextClassLoader())
+                              .filter(isBootstrapClassLoader())
+                              .build(), ClassLoadingStrategy.Default.INJECTION)
                       .getLoaded();
-    }
-
-    private <T> Class<?>[] allMockedTypes(MockFeatures<T> features) {
-        Class<?>[] allMockedTypes = new Class<?>[features.interfaces.size() + 1];
-        allMockedTypes[0] = features.mockedType;
-        System.arraycopy(features.interfaces.toArray(), 0,
-                         (Object[]) allMockedTypes, 1, features.interfaces.size());
-        return allMockedTypes;
     }
 
     // TODO inspect naming strategy (for OSGI, signed package, java.* (and bootstrap classes), etc...)
--- a/mockmaker/bytebuddy/main/java/org/mockito/internal/creation/bytebuddy/MockBytecodeGenerator.java
+++ b/mockmaker/bytebuddy/main/java/org/mockito/internal/creation/bytebuddy/MockBytecodeGenerator.java
@@ -6,15 +6,11 @@ import net.bytebuddy.dynamic.DynamicType;
 import net.bytebuddy.dynamic.loading.ClassLoadingStrategy;
 import net.bytebuddy.dynamic.scaffold.subclass.ConstructorStrategy;
 import net.bytebuddy.implementation.FieldAccessor;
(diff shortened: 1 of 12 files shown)


osgi.gradle
```diff
@@ -21,7 +21,7 @@ afterEvaluate {
 
             instruction 'Import-Package',
                     '!org.mockito.asm.signature',
-                    'net.bytebuddy.*;version=0.2.1',
+                    'net.bytebuddy.*;version=0.6.11',
```
bric3: Good catch

mockmaker/bytebuddy/main/java/org/mockito/internal/creation/bytebuddy/MockBytecodeGenerator.java
```diff
@@ -50,7 +38,7 @@ public MockBytecodeGenerator() {
                 byteBuddy.subclass(features.mockedType, ConstructorStrategy.Default.IMITATE_SUPER_TYPE)
                          .name(nameFor(features.mockedType))
                          .implement(features.interfaces.toArray(new Class<?>[features.interfaces.size()]))
-                         .method(any()).intercept(toDispatcher)
+                         .method(any()).intercept(MethodDelegation.to(DispatcherDefaultingToRealMethod.class))
```
bric3: Indeed better

mockmaker/bytebuddy/main/java/org/mockito/internal/creation/bytebuddy/MockBytecodeGenerator.java
```diff
@@ -24,24 +20,16 @@
...
+import static net.bytebuddy.matcher.ElementMatchers.*;
 
 class MockBytecodeGenerator {
     private final ByteBuddy byteBuddy;
     private final Random random;
-    private final Implementation toDispatcher;
 
     public MockBytecodeGenerator() {
-        byteBuddy = new ByteBuddy(ClassFileVersion.JAVA_V5)
+        byteBuddy = new ByteBuddy(ClassFileVersion.JAVA_V5) // TODO: Consider to allow for VM version discovery (remove explicit) to support Java 8 features.
```
bric3: Agreed I've seen that the other day, I think bytebuddy should discover VM version, also it may prove to be useful for a better java 8 support

mockmaker/bytebuddy/main/java/org/mockito/internal/creation/bytebuddy/MockBytecodeGenerator.java
```diff
@@ -61,18 +49,16 @@ public MockBytecodeGenerator() {
...
         }
         return builder.make()
-                      .load(SearchingClassLoader.combineLoadersOf(allMockedTypes(features)), ClassLoadingStrategy.Default.INJECTION)
+                      .load(new MultipleParentClassLoader.Builder()
+                              .append(features.mockedType)
+                              .append(features.interfaces)
+                              .append(MultipleParentClassLoader.class.getClassLoader())
+                              .append(Thread.currentThread().getContextClassLoader())
+                              .filter(isBootstrapClassLoader())
+                              .build(), ClassLoadingStrategy.Default.INJECTION)
```
bric3: I'd like to split this one in another commit, same PR is ok though. Also `SearchingClassLoader` had issues with other tools like eclipse, maven surefire, or spring (fixed along the road). Do you have tested them ?

bric3:
Nice update overall :+1:

mockmaker/bytebuddy/main/java/org/mockito/internal/creation/bytebuddy/MockMethodInterceptor.java
```diff
@@ -29,46 +29,6 @@ public MockMethodInterceptor(InternalMockHandler handler, MockCreationSettings m
...
-                                    @Origin(cache = true) Method invokedMethod,
-                                    @AllArguments Object[] arguments) throws Throwable {
-        return doIntercept(
-                mock,
-                invokedMethod,
-                arguments,
-                InterceptedInvocation.SuperMethod.IsIllegal.INSTANCE
-        );
-    }
-
```
bric3: So the idea is that with the dispatcher those are not necessary anymore ?

mockmaker/bytebuddy/main/java/org/mockito/internal/creation/bytebuddy/MockBytecodeGenerator.java
```diff
@@ -24,24 +20,16 @@
 import static net.bytebuddy.description.modifier.Ownership.STATIC;
 import static net.bytebuddy.description.modifier.Visibility.PRIVATE;
 import static net.bytebuddy.implementation.MethodDelegation.to;
-import static net.bytebuddy.matcher.ElementMatchers.any;
-import static net.bytebuddy.matcher.ElementMatchers.isEquals;
-import static net.bytebuddy.matcher.ElementMatchers.isHashCode;
+import static net.bytebuddy.matcher.ElementMatchers.*;
```
bric3: I'd like to avoid star import if possible :P

mockmaker/bytebuddy/main/java/org/mockito/internal/creation/bytebuddy/MockBytecodeGenerator.java
raphw: The class search path of this class loader ia now fully identical to that of the searching one. It just aggregates all parents within a single class loader instead of creating one per parent.

mockmaker/bytebuddy/main/java/org/mockito/internal/creation/bytebuddy/MockMethodInterceptor.java
raphw: Byte Buddy delegates to the dispatcher. The dispatcher can then decide what to do with an invocation based on of the field is initialized.

raphw:
I can break up the commits tomorrow.

bric3:
OK thanks again for the PR :)

mockmaker/bytebuddy/main/java/org/mockito/internal/creation/bytebuddy/MockMethodInterceptor.java
bric3: OK thanks for the confirmation.

bric3:
Awesome thanks @raphw

PascalSchumacher:
Thanks! After updating to the latest beta the Eclipse Infinitytest Plugin can run my JUnit Theories with Mockito. :)

PR #254, opened 2015-07-08
Fixes #253 create unsynchronized interceptors for mocks

### LINKED ISSUE #253: Can't verify on synchronized methods of objects
Because they are holding the object's lock when verify(object).method() is called, so if another thread tries to call the method it can't.

bonifaido:


raphw:
This causes both the loss of meta data as well as inherited, synchronized methods. In my understanding, Mockito is not supposed to be thread-safe. But I see the point. I will adress this in some future release of Byte Buddy.

bonifaido:
Yeah the implementation is kinda sketchy.
According to the [FAQ](https://github.com/mockito/mockito/wiki/FAQ) Mockito should support this kind of verification especially in timeout scenarios (timeout() is a sign of a concurrency tests). This kind of verification used to work in Mockito 1.x.

raphw:
Indeed. I will have a look at it. Ideally, Byte Buddy would be able to register a transformer that allows a modifier to be changed. I am currently down the rabbit hole to support generic types so I wonder when I'll be able to tackle this. @bric3, is this a blocker?

szczepiq:
I'd rather not release 2.0 final if we regress on the thread-safety front.
There's still some time to get this sorted out, though.

On Wed, Jul 8, 2015 at 11:57 PM, Rafael Winterhalter <
notifications@github.com> wrote:

> Indeed. I will have a look at it. Ideally, Byte Buddy would be able to
> register a transformer that allows a modifier to be changed. I am currently
> down the rabbit hole to support generic types so I wonder when I'll be able
> to tackle this. @bric3 <https://github.com/bric3>, is this a blocker?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mockito/mockito/pull/254#issuecomment-119846695>.
>



-- 
Szczepan Faber
Founder mockito.org; Core dev gradle.org
tweets as @szczepiq; blogs at blog.mockito.org

raphw:
I addressed the issue by slightly extending Byte Buddy's API. Pull request: https://github.com/mockito/mockito/pull/257

bric3:
Belatle answer yes that should a major drawback of 2.0. @raphw I'll look at #257 
Also thanks for reporting and crafting a fix with this PR @bonifaido :)

bric3:
Closing as #257 should fix the issue.

PR #255, opened 2015-07-09
Mockito plugin annotation

mat013:


mat013:
I found another issue

The class GlobalConfiguration contains a ThreadLocal which contains the latest global configuration, and when running the test the TestBase invokes ConfigurationAccess.getConfig which casts the GlobalConfiguration to a MockitoConfiguration, which is not the case when using the @MockitoPlugin. 

I guess the GlobalConfiguration is meant to serve as a singleton, however I guess that is not possible when you have a annotation engine for each test instance.

Are there any special reason why global configuration has to use a ThreadLocal ?

src/main/java/org/mockito/internal/configuration/GlobalConfiguration.java
```diff
@@ -25,6 +26,10 @@
     IMockitoConfiguration getIt() {
         return GLOBAL_CONFIGURATION.get();
     }
+    //back door for testing
+    public static void removeIt() {
+        GLOBAL_CONFIGURATION.remove();
+    }
```
bric3: Don't add this method here, for unit tests there's this class `org.mockito.configuration.MockitoConfiguration` in the test sources

src/test/java/org/mockitousage/MockitoAnnotationsTest.java
```diff
@@ -0,0 +1,89 @@
...
+        public SomeMock someMock;
+    }
+    
+    public static interface SomeMock {
+        
+    }
+    
+    @RunWith(MockitoJUnitRunner.class)
+    @MockitoPlugin(annotationEngine=MyAnnotationEngine.class)
+    public static class MockitoPluginAnnotationTestClass {
```
bric3: I think such annotations shouldn't be on the test class, but rather as a class field (or maybe both), so a plugin could be  be instantiated with custom code (like parameters, other dependencies, etc.)

src/test/java/org/mockitousage/MockitoAnnotationsTest.java
```diff
@@ -0,0 +1,89 @@
...
+    public void tearDown() {
+         GlobalConfiguration.removeIt();
+    }
+
+    @Test
+    public void should_process_the_testcase_using_default_annotation_engine() throws Exception {
+        NoMockitoPluginAnnotationTestClass noMockitoPluginAnnotationTestClass = new NoMockitoPluginAnnotationTestClass(); 
+
+        MockitoAnnotations.initMocks(noMockitoPluginAnnotationTestClass);
+        assertThat(isMock(noMockitoPluginAnnotationTestClass.someMock), is(true));
```
bric3: You may want to prefer AssertJ assertions, it's in our dependencies (and it replaces FEST Assert).

bric3:
> Are there any special reason why global configuration has to use a ThreadLocal ?

I don't remember. @szczepiq do you?

Nice feature overall :)

src/main/java/org/mockito/MockitoPlugin.java
```diff
@@ -0,0 +1,73 @@
...
+ * {@link org.mockito.runners.MockitoJUnitRunner}.
+ * </p>
+ *
+ * @see MockitoAnnotations#initMocks(Object)
+ * @see org.mockito.runners.MockitoJUnitRunner
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+@Inherited
+public @interface MockitoPlugin {
```
bric3: Open thinking mode, I wonder it should be named `MockitoAnnotationEngine`, instead of adding multiple fields here for each new mockito plugin

bric3:
By the way your commits should be squashed ;)

src/test/java/org/mockitousage/MockitoAnnotationsTest.java
mat013: I think this can be achieved by adding a custom the annotation engine on class level.
I think if it is needed to have an annotation for class fields then we can build a annotation engine which could support the new annotation for the class field. 

What do you think?

mat013:
> By the way your commits should be squashed ;)

I tried let's see if I did it correctly :p

mat013:
I moved the removeIt from the globalconfiguration to ConfigurationAccess instead, I think there is an issue with the global configuration that it can only be instantiated once. thereby it will not be possible to have different annotation engine for different test cases.

I think it would make sense to do that so I change the new constructor in GlobalConfiguraton which I implemented to _always_ instantiate a new configuration with the provided annotation engine.

bric3:
>> By the way your commits should be squashed ;)
> I tried let's see if I did it correctly :p

Nope, you have imported commits from master , and you somehow did some merge commits from another branch. On _pull request_ branches it is better to keep a clean _flat_ history.

I'm not sure why you needed those merges, but you definitely want to avoid those. _Squashing_ means to put commits together in order to have a clean and meaningful history. This is done via a `git rebase --interactive <the commit references>`. Check our [wiki](https://github.com/mockito/mockito/wiki/Using-git-to-prepare-your-PR-to-have-a-clean-history) on that matter, it contains the bare minimum to understand how to have a clean history. Nevertheless there's better and more complete tutorial out there.

Don't worry I know git is a bit hard when beginning, but once you know your way and why it matters, you won't regret it :)

mat013:
So should I make a new branch and transfer all the changes into the new
branch rather than continue working on this messy branch?

Br

Meang

bric3:
Yes you can.
Can you close this one when you are ready :)

mat013:
Killed it as it the branch was messed up. Have created a new PR and branch for that feature

PR #257, opened 2015-07-10
Upgraded to Byte Buddy 0.6.12. 

raphw:
Fixes https://github.com/mockito/mockito/issues/253

- Adapted minor API changes. 
- Added modifier resolver that removes the synchronization flag from intercepted methods.


bric3:
Looks good to me, thanks again @raphw

PR #263, opened 2015-07-11
Issue262 verification with description test seems to be in the wrong package

mat013:
aligning package name


bric3:
I think this pull request is misnamed and it contains the non wanted test package change.

I'll review that one later

mat013:
No Need :)... The Issue 262 was rejected as this was intentional...
but if you could answer the question posed in issue 262 I would be happy

Thanks in advance

br

Meang

bric3:
So I can close this one ?

PR #265, opened 2015-07-13
Providing functionality for setting the annotation engine for a test …

mat013:
…case through annotation

In order to customize the annotation engine the current implementation requires that one implements a specific class (MockitoConfiguration) which specifies which annotation engine to use for the whole project.

As a consequence it is not possible to have several custom annotation engines within the same project.

This is a proposal how to achieve several annotation engine by using an annotation.

```java
@MockitoPlugin(annotationEngine = MyAnnotationEngine.class)
 public class ArticleManagerTest extends SampleBaseTestCase {
     @Mock private ArticleCalculator calculator;
     @Mock private  articleMonitor;
     private ArticleManager manager;
 }

 public class SampleBaseTestCase {
     @Before public void initMocks() {
         MockitoAnnotations.initMocks(this);
     }
 }
 
 public class MyAnnotationEngine implements AnnotationEngine {
     void process(Class<?> clazz, Object testInstance) {
         ... processing of logic to do the injection ...
     }
```

### DIFF
--- a/src/main/java/org/mockito/MockitoAnnotations.java
+++ b/src/main/java/org/mockito/MockitoAnnotations.java
@@ -91,11 +91,16 @@ public class MockitoAnnotations {
             throw new MockitoException("testClass cannot be null. For info how to use @Mock annotations see examples in javadoc for MockitoAnnotations class");
         }
 
-        AnnotationEngine annotationEngine = new GlobalConfiguration().getAnnotationEngine();
+        MockitoPlugin mockitoPluginAnnotation = getAnnotationForMockitoPlugin(testClass);
+        GlobalConfiguration globalConfiguration = mockitoPluginAnnotation == null ?
+                new GlobalConfiguration() : new GlobalConfiguration(mockitoPluginAnnotation);
+        
+        AnnotationEngine annotationEngine = globalConfiguration.getAnnotationEngine();
         Class<?> clazz = testClass.getClass();
 
         //below can be removed later, when we get read rid of deprecated stuff
-        if (annotationEngine.getClass() != new DefaultMockitoConfiguration().getAnnotationEngine().getClass()) {
+        if (annotationEngine.getClass() != new DefaultMockitoConfiguration().getAnnotationEngine().getClass() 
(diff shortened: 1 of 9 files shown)


mat013:
Hi Brice

Any news on this PR? Does it need further refinement or anything?

br

Meang

bric3:
Hi, sorry for the delay, I'm a bit busy atm, holidays plus catching up at work, with usual production issues plus private life ;)

TimvdLippe:
@mat013 what was your original use-case that required this feature? I am not sure if we want to incorporate this in Mockito official without a clear usecase.

mat013:
Well it is about year ago I submitted this pull request so I have forgotten the details, but as I recall it. 

let's imagine we have a dependence hierarchy where C depends on B and B depends on A.

If we have a test that looks as follows:

public class Test {

@InjectMocks
B b;

@InjectMocks
C c;

@Test
public void someTest() {
} 

}
 
I would have liked that the b instance was instantiated with a mock of type A, and C was instantiated with b. However when the test is executed a null for the B dependency is injected into C because B is not annotated with @Mock.Since B had already been instantiated I wanted to add inject that instance into C, however this is not the behaviour of the InjectingAnnotationEngine. So I made the suggestion to add some flexibility to Mockito where one makes it possible to provide an alternative injection strategy.

You can argue that one can manually inject the right dependencies when setting up the test (Which I ended up doing), however I think since Mockito already provide a shorthand for automatically injecting dependencies it would be nice to also be able to have an alternative strategy when injecting. 

Hope that made sense otherwise let me know 
br

Meang

TimvdLippe:
Ah that makes it a lot clearer. Actually to solve that problem you can annotate `B` with `@Spy`. Now the bummer is that it is currently broken (https://github.com/mockito/mockito/issues/285) but I think that would be the best solution. If you agree, shall we close this pull request? Thanks again for your effort!

mat013:
Hi Tim

Sounds like that is a more simple solution, so it is fine to close my PR down... I'll check once #285 has been fixed... If I have time I may look at it and see if I can fix it... :) 

Have a nice day 

Meang

PR #266, opened 2015-07-17
Ignore Groovy meta methods when instrumenting.

raphw:
Added a matcher to ignore any methods that are defined by the Groovy meta class system from instrumentation. This way, it is possible to create mocks of Groovy classes that behave appropriately in Groovy code.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -54,7 +54,7 @@ test {
 }
 
 dependencies {
-    compile 'net.bytebuddy:byte-buddy:0.6.12'
+    compile 'net.bytebuddy:byte-buddy:0.6.14'
 
     provided "junit:junit:4.10", "org.hamcrest:hamcrest-core:1.3"
     compile "org.objenesis:objenesis:2.1"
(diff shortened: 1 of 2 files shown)


bric3:
The change seems safe, but for some reason it breaks coverage (jacoco)

raphw:
Thats strange. Ill try to run jacoco from command line and see what is wrong.

raphw:
Can you add

sudo: false

to the Travis file? I had some issues with my builds since Travis switched to Docker.

bric3:
It's already set in the `.travis.yml`

bric3:
Agreed I took a quick look the issue is repeatable the same method fails during coverage.

raphw:
Its a bug in Byte Buddy where the "old" ignored method matcher was unapplied. This way, bridge methods were suddenly instrumented instead of ignored what caused the tests to fail. I'll release a patch and fix this up.

raphw:
Et voilà. I tested with Groovy locally as well. I am wondering if you want a test case? Would require a Groovy compiler to be installed for building the project, though.

bric3:
Awesome :sparkles: 
Thnaks @raphw

bric3:
Also I think we're good here without groovy. Maybe another integration project may be good.

PR #274, opened 2015-07-31
Added InvocationsCounter as VerificationMode. Fixes #77

### LINKED ISSUE #77: Getting an absolute count of calls to a mocked method
We can do this:

    verify(mock1, times(3)).thisMethod();
    verify(mock2, times(3)).thatMethod();

But we cannot do something like this, as far as I can see:

    verifyEquals(times(mock1.thisMethod()), times(mock2.thatMethod()));

That is, I would like to verify that two mocked methods are called the same unspecified number of times. The number of calls made to a method appears to be already available, so this would be just querying the value directly, not indirectly via `times()`, `atLeast()`, etc.

I also [posed this question to StackOverflow](http://stackoverflow.com/q/25543250/1270789), and one answer, on using an `Answer`, seems good; perhaps a standard `Answer` implementation `CountsCalls` or similar should be the solution?

JeremybellEU:
In order to count the number of invocations,
InvocationsCounter can keep track of the number of invocations
using an AtomicInteger.
This AtomicInteger can then be used in Mockito.times(AtomicInteger)
or with Assert.assertEquals(int, int) in order to verify behaviour.

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -17,6 +17,8 @@ import org.mockito.stubbing.*;
 import org.mockito.verification.*;
 import org.mockito.junit.*;
 
+import java.util.concurrent.atomic.AtomicInteger;
+
 /**
  * <p align="left"><img src="logo.png" srcset="logo@2x.png 2x" alt="Mockito logo"/></p>
  * Mockito library enables mocks creation, verification and stubbing.
@@ -2228,6 +2230,22 @@ public class Mockito extends Matchers {
     }
 
     /**
+     * Allows verifying exact number of invocations. E.g:
+     * <pre class="code"><code class="java">
+     *   AtomicInteger counter = new AtomicInteger();
+     *   counter.set(2);
+     *   verify(mock, times(counter)).someMethod("some arg");
+     * </code></pre>
+     *
+     * @param wantedNumberOfInvocations wanted number of invocations
+     *
+     * @return verification mode
+     */
+    public static VerificationMode times(AtomicInteger wantedNumberOfInvocations) {
+        return VerificationModeFactory.times(wantedNumberOfInvocations.get());
+    }
+
+    /**
      * Alias to <code>times(0)</code>, see {@link Mockito#times(int)}
      * <p>
      * Verifies that interaction did not happen. E.g:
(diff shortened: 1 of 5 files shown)


JeremybellEU:
As a personal note:
This PR does fix the issue, but semantically is not 100% correct. Simply because CountInvocations does not verify anything by itself. I am not sure if this would be a problem, but that would be up to the Mockito maintainers to decide.

A possible improvement could be to force the user to invoke times() with the specified counter and also create tests that make sure the user does so. (Similar to the tests that assert that verify() calls without a method call resolve in an MockitoException).

src/main/java/org/mockito/Mockito.java
```diff
@@ -2228,6 +2230,22 @@ public static VerificationMode times(int wantedNumberOfInvocations) {
     }
 
     /**
+     * Allows verifying exact number of invocations. E.g:
+     * <pre class="code"><code class="java">
+     *   AtomicInteger counter = new AtomicInteger();
+     *   counter.set(2);
+     *   verify(mock, times(counter)).someMethod("some arg");
+     * </code></pre>
```
bric3: I don't see any benefit here. It clutters the API without bringing value, why not just 

```
verify(mock, times(counter.get())).someMethod("some arg");
```

src/main/java/org/mockito/Mockito.java
```diff
@@ -2333,6 +2351,30 @@ public static VerificationMode only() {
...
+     *   verify(otherMock, times(counter)).someMethodThatShouldAlsoHaveBeenCalled();
+     * </code></pre>
+     *
+     * or by directly asserting the counter:
+     *
+     * <pre class="code"><code class="java">
+     *   AtomicInteger counter = new AtomicInteger();
+     *   verify(mock, countInvocations(counter)).someMethod("some arg");
+     *   assertEquals(2, counter.get());
+     * </code></pre>
```
bric3: I'd rather see something like that

```java
InvocationStats someMethodStats = AdditionalVerifications.invocationStats();
verify(mock, stats).someMethod("some arg");
assertEquals(2, someMethodStats.count());
```

Note I used the word stat to allow possible other metrics (**if relevant**). Also this should be annotated with `@Incubating`

bric3: Also mving such verification mode follows what we have with matchers and answers. This will avoid cluttering more the _BIG_ `Mockito` class

JeremybellEU:
@bric3 As outlined by my original note when submitting the PR, I think we both agree that the current solution is not the way to go? If yes, it is probably better to close this PR and use the comments you placed to develop a more intuitive and clearer API regarding statscollecting.

bric3:
Oh ok I forgot about that, I was kinda pretty busy. Thanks for the reminder. Let's close it then.

PR #277, opened 2015-08-02
Add @Real annotation for @Injectmocks

### LINKED ISSUE #174: Annotation to instantiate and inject non-mock into @InjectMocks
This may seem to stray from the core mission of Mockito, but it seems like the natural and relatively simple extension to the @InjectMocks functionality.  We can already use

@Mock(answer=Answers.CALLS_REAL_METHODS) MyObject myObject;
or
@Spy MyObject myObject;

But in those cases we are truly mocking/spying the collaborator.  But if there is really no mocking or spying required but we just need to inject an instance of a collaborator class with the same ease that we do for mocks and spies, it would be nice to not have to manually wire them.

Some kind of annotation such as @Real would be nice, to instantiate a real, non-proxied instance with Objenesis, assign it to the annotated field, and inject it into the @InjectMocks field.

TimvdLippe:
**EDITED by mockito team**

This PR is supposed to fix #174

---------------------------------------------------------

With these changes the following code should be possible (injecting non mocks) :

```java
public class EventWatcher {
    public EventWatcher(String eventName, Notifier notifier) { /* ... */ }
    // ...
}
```

```java
@Real String eventName = "PLAYER_ADDED_EVENT";
@Mock Notifier notifier;
@InjectMocks EventWatcher watcher;
```

Also sub-dependency graph should be supported :

```java
@InjectMocks Root root = new Root();

@Mock EndLeaf endMockedDependency;
@Real @InjectMocks MiddleBottom middleBottom;
@Real @InjectMocks MiddleTop middleTop;
```


JeremybellEU:
I have tried to get rid of the instanceof code-smell [here for example](https://github.com/mockito/mockito/pull/277/files#diff-0b522b0de53cdff487ca66dc11564c75R78), but with no success. 

The reason for this is that all `Set<Object>` are actually [HashCodeAndEqualsSafeSets](https://github.com/mockito/mockito/blob/master/src/main/java/org/mockito/internal/util/collections/HashCodeAndEqualsSafeSet.java). While this is nice to get of the `NullPointerException`, real objects don't have this problem. Therefore I tried to replace the `Set<Object>` with `Set<HashCodeAndEqualsMockWrapper>` in order to safely use the set, but with the extra implementation of the wrapper.

However, every place where the actual objects are required, the [get method](https://github.com/mockito/mockito/blob/master/src/main/java/org/mockito/internal/util/collections/HashCodeAndEqualsMockWrapper.java#L35) should be used. I have tried to change all of this, just to get rid of that `instanceof`, but was unable to succeed all tests again.

If anyone is able to fix this issue, that would be very nice.

JeremybellEU:
I have some additional tweaks in mind for the injection package, but if I would do them in this PR, it gets a bit bloated. Shall I wait for the review until continueing work in this package?

vbuell:
What blocks this PR to be approved?

JeremybellEU:
The Mockito team is currently very busy, so we have to wait for them when they can find some free time. But I got promised they will look into the PR's as soon as they do have time =]

JWGmeligMeyling:
:+1:

bric3:
What's wrong the feature, _they have been discussed a long time ago already when the **shorthand** injection annotation came in mockito_.

* The first thing to note is that `@Real` doesn't really express that it's a dependency, the name is chosen as an opposition to what a mock is. For reference in [google code issue 290](https://code.google.com/p/mockito/issues/detail?id=290), there was ideas that better fit the intent, like `@Collaborator`, `@Dependency`, etc.
* The current work allows to inject a single value object in the constructor of the same type. Suppose the constructor takes several `Integer`, then it doesn't work anymore :

    ```java
    new ComponentClient(String host,
                        Callback callback,
                        Integer connectionTimeout,
                        Integer socketTimeout,
                        Integer requestTimeout,
                        Integer slowAbortThreshold)
    ```

    Worst, providing support for multiple arguments of the same type will be either difficult, or impossible to solve without adding yet another concept or making the API awkward.
* On the matter of managing sub-dependencies — _aka_ dependency graph — there's several issues :
  * This brings dependency injection noise in a test.
  * This somehow doesn't feel right in a *unit test*. 
  * If that is really, really necessary then wiring things manually in a `@Before` method or simply using a real injection framework makes more sense anyway.
* A side note too, the more mockito support injection, the more there will be shortcomings because mockito is not a real DI framework, we'll have to maintain that code, and keep us out of the mocking feature focus.

That may feel like opinionated points, but I'm not anymore in favor of mockito supporting more injection concerns that stay with actual DI frameworks.

JeremybellEU:
Yes I do agree that this would cause Mockito to fix DI. Maybe we should approach the problem from a different angle: Is there a possibility to ease the integration of Mockito and inserting Mocks with current DI frameworks like Guice?

If you need to `@Inject` a real value, this could greatly clutter you code. I think the perfect example to test the possibilities on is the one mentioned in the docs:

```java
public class EventWatcher {
    
   private String eventName;

   private Notifier notifier;

   public EventWatcher(String eventName, Notifier notifier) {
       this.eventName = eventName;
       this.notifier = notifier;
    }

   public void addMessage(String event, String message) {
       if (this.eventName.equals(event)) {
           notifier.notify(message);
       }
   }
}

```

Is it maybe possible to notify Guice automatically: we have created a Mock, please use this whenever you encounter class T.
This would delegate all DI problems to the actual DI framework and Mockito only offers the DI framework the created Mock.

WDYT?

bric3:
Actually I don't see typing stuff as a problem here, should it be my example

```java
@Mock Callback callbackMock;
ComponentClient client;

@Before public void initialize_client() {
    client = new ComponentClient("http://",
                                 callbackMock,
                                 C_CONN_TIMEOUT,
                                 C_SOCKET_TIMEOUT,
                                 C_REQ_TIMEOUT,
                                 C_SLOW_ABORT_THRESH);
```

or yours 

```java
@Mock Notifier notifier
EventWatcher watcher;

@Before public void set_up__overflow_event__watcher() {
    watcher = new EventWatcher("overflow", notifier);
}
```


And for Spring or Guice injection, one can just rely on their powerful tooling:, I think Spring is a bit more convenient in this case though.

```java
@RunWith(SpringJUnit4ClassRunner.class)
@ContextConfiguration
public class OrderServiceTest {
    @Inject private Dep depMock
    @Inject private Undertest undertest;

    @Test
    public void should_not_fail() {
        // ensure Undertest behaviour is good
    }

    @Configuration
    static class NotAnymoreAJUnitConfig {
        @Bean public Dep dep() {
            return mock(Dep.class);
        }

        @Bean public Undertest undertest() {
            return new Undertest();
        }
    }
}
```

TimvdLippe:
Closing for now per above discussion. Let's revisit this issue if people encounter this more often.

PR #281, opened 2015-08-07
Correcting public website url in Maven POM

smarkwell:
http://www.mockito.org does not resolve correctly, updating to http://mockito.org


bric3:
Nice thanks

PR #287, opened 2015-08-23
Implement VerificationCollector which can collect multiple verifications.

### LINKED ISSUE #124: Concise way to collect multiple verify failures, ideally with JUnitCollector or  derivative
If you are using more than one verify statement as test post-conditions, it would be nice to be able to collect each failure and continue with the remaining verify statements or asserts that follow.  JUnit provides the ErrorCollector @Rule to facilitate this kid of thing, but the current [ErrorCollector API] (http://junit.org/apidocs/org/junit/rules/ErrorCollector.html) requires either

* a Matcher<T> and a value to compare
* or a Callable<Object>

The Mockito verify statements only return void, however, since they depend on throwing exceptions.  I have not thought of an ideal, non-disruptive way to use JUnit's ErrorCollector to aggregate multiple Mockito verify failure

JeremybellEU:
It will report in the end the result of multiple statements. Fixes #124

### DIFF
--- a/src/main/java/org/mockito/junit/VerificationCollectorImpl.java
+++ b/src/main/java/org/mockito/junit/VerificationCollectorImpl.java
@@ -1,0 +1,89 @@
+package org.mockito.junit;
+
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+import org.mockito.exceptions.base.MockitoAssertionError;
+import org.mockito.internal.MockitoCore;
+import org.mockito.internal.verification.api.VerificationData;
+import org.mockito.verification.VerificationMode;
+
+import static org.mockito.Mockito.times;
+
+/**
+ * Mockito implementation of VerificationCollector.
+ */
+class VerificationCollectorImpl implements VerificationCollector {
+
+    private static final MockitoCore MOCKITO_CORE = new MockitoCore();
+
+    private StringBuilder builder;
+    private int numberOfFailures;
+
+    public VerificationCollectorImpl() {
+        this.resetBuilder();
+    }
+
+    public Statement apply(final Statement base, final Description description) {
+        return new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                base.evaluate();
+                VerificationCollectorImpl.this.collectAndReport();
+            }
+        };
+    }
+
+    public <T> T verify(T mock) {
+        return MOCKITO_CORE.verify(mock, new VerificationWrapper(times(1)));
+    }
+
+    public <T> T verify(T mock, VerificationMode mode) {
+        return MOCKITO_CORE.verify(mock, new VerificationWrapper(mode));
+    }
+
+    public void collectAndReport() throws MockitoAssertionError {
+        if (this.numberOfFailures > 0) {
+            String error = this.builder.toString();
+
+            this.resetBuilder();
+
+            throw new MockitoAssertionError(error);
+        }
+    }
+
+    private void resetBuilder() {
+        this.builder = new StringBuilder()
+                .append("There were multiple verification failures:");
+        this.numberOfFailures = 0;
+    }
+
+    private void append(String message) {
+        this.numberOfFailures++;
+        this.builder.append('\n')
+                .append(this.numberOfFailures).append(". ")
+                .append(message.substring(1, message.length()));
+    }
+
+    private class VerificationWrapper implements VerificationMode {
+
+        private final VerificationMode delegate;
+
+        private VerificationWrapper(VerificationMode delegate) {
+            this.delegate = delegate;
+        }
+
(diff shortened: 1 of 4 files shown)


src/main/java/org/mockito/verification/JUnitVerificationCollector.java
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Use this rule in order to collect multiple verification failures and report at once.
+ *
+ * In the example below, the verification failure thrown by {@code byteReturningMethod()} does not block
+ * verifying against the {@code simpleMethod()}. After the test is run, a report is generated stating all
+ * collect verification failures.
+ *
+ * <pre class="code"><code class="java">
+ *   &#064;Rule
+ *   public JUnitVerificationCollector collector = new JUnitVerificationCollector();
```
bric3: I think this rule should be instantiated this way (Also the method should only expose an interface for example `VerificationCollector`) :

```java
@Rule VerificationCollector collector = MockitoJUnit.collector();
```

src/main/java/org/mockito/verification/JUnitVerificationCollector.java
```diff
@@ -0,0 +1,108 @@
+package org.mockito.verification;
```
bric3: Since it is JUnit specific, this whole PR should live within the package `org.mockito.junit`, we'd like to move runners as well under this package too.

src/test/java/org/mockitousage/verification/rules/JUnitVerificationCollectorTest.java
```diff
@@ -0,0 +1,132 @@
...
+            assertThat(error).hasMessageContaining("1. Wanted but not invoked:");
+            assertThat(error.getMessage()).doesNotContain("2.");
+        }
+    }
+
+    /*
+        Dirty fix in order to only execute the inner test class when the test case below is run.
+        Otherwise the inner test class will be executed in the suite and fail as we expect.
+     */
+    private static boolean SHOULD_RUN_RULE = false;
```
bric3: Don't do these hacks ;)
Instead play with core JUnit functionality like `JUnitCore`, there's some example in the test code base (like `org.mockitousage.bugs.MockitoRunnerBreaksWhenNoTestMethodsTest`)

JeremybellEU: I honestly didn't have a clue how to fix this and internet wasn't helpful either. I will check out the test you mentioned!

bric3: No problem ;)

JeremybellEU:
Rebased branch and changed code according to comments of @bric3 .

To make the inner test test working, the Inner test should NOT end with `Test`.

src/main/java/org/mockito/junit/VerificationCollectorImpl.java
```diff
@@ -0,0 +1,89 @@
...
+import org.mockito.internal.MockitoCore;
+import org.mockito.internal.verification.api.VerificationData;
+import org.mockito.verification.VerificationMode;
+
+import static org.mockito.Mockito.times;
+
+/**
+ * Mockito implementation of VerificationCollector.
+ */
+public class VerificationCollectorImpl implements VerificationCollector {
```
bric3: I don't think this class needs to be public anymore ;)

bric3:
nice rework

JeremybellEU:
Made the Impl class package-protected.

bric3:
Looks good to me

JeremybellEU:
:+1: Thanks for the reviews!

szczepiq:
Hey Tim,

Thanks for cool contributions!!!

What are your thoughts on improving the VerificationCollector idea: the rule would automatically validate all 'verify' methods at the end of the test. This way, there would be no need to call 'collector.verify' at all (we could leave it as optional for gnarly cases).

TimvdLippe:
@szczepiq I am not really sure what you mean, could you provide a code snippet with what you would like to achieve?

szczepiq:
Current API:

```java
     @Rule
     public VerificationCollector collector = MockitoJUnit.collector();
  
     @Test
     public void should_fail() {
         IMethods methods = mock(IMethods.class);
  
         collector.verify(methods).byteReturningMethod();
         collector.verify(methods).simpleMethod();
     }
```

Desired API:

```java
     @Rule
     public MockitoRule rule = MockitoJUnit.rule().lazyVerify(); //just an example, no strong opinion here
  
     @Test
     public void should_fail() {
         IMethods methods = mock(IMethods.class);
  
         //below just works (strong opinion :)
         verify(methods).byteReturningMethod();
         verify(methods).simpleMethod();
     }
```

TimvdLippe:
@szczepiq PR is up at #389 :smile:

PR #288, opened 2015-08-24
Add optional answer to support mocked Builders

TimvdLippe:
When googling the mocking of builders, a lot of people said it was not possible with Mockito to support them in a clean way. `RETURNS_DEEPS_STUBS` does allow cascading, but can become quite fast very cumbersome.
The added `RETURNS_SELF` will try to return itself by looking at the method `returnType` and compare this with the class of the mock.

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -1087,10 +1087,10 @@ import org.mockito.junit.*;
  * <p>
  * <pre class="code"><code class="java">
  *
- * // will print a custom message on verification failure 
+ * // will print a custom message on verification failure
  * verify(mock, description("This will print on failure")).someMethod();
- * 
- * // will work with any verification mode 
+ *
+ * // will work with any verification mode
  * verify(mock, times(2).description("someMethod should be called twice")).someMethod();
  * </code></pre>
  *
@@ -1130,7 +1130,7 @@ public class Mockito extends Matchers {
      * <p>
      * Example:
      * <pre class="code"><code class="java">
-     *   Foo mock = (Foo.class, RETURNS_SMART_NULLS);
+     *   Foo mock = mock(Foo.class, RETURNS_SMART_NULLS);
      *
      *   //calling unstubbed method here:
      *   Stuff stuff = mock.getStuff();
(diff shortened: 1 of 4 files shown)


src/main/java/org/mockito/internal/stubbing/defaultanswers/TriesToReturnSelf.java
```diff
@@ -0,0 +1,29 @@
...
+import org.mockito.internal.util.MockUtil;
+import org.mockito.internal.util.reflection.GenericMetadataSupport;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+
+import java.io.Serializable;
+
+public class TriesToReturnSelf implements Answer<Object>, Serializable{
+
+    private static final MockUtil mockUtil = new MockUtil();
```
Ariel-Isaacm: Does mockUtil needs to be static?

JeremybellEU: I don't see any reason not to, but probably @bric3 if it should (not).

bric3: Yes please avoid `static`

src/main/java/org/mockito/Mockito.java
```diff
@@ -1278,6 +1278,50 @@
...
+     *
+     *     public ExtendedBuilder extendThis() { ... }
+     * }
+     *
+     * public class ExtendedBuilderTest {
+     *
+     *     &#064;Test
+     *     public void mockBuilder() {
+     *         ExtendedBuilder builder = mock(ExtendedBuilder.class, RETURN_SELF);
+     *     }
```
bric3: This exemple is a bit convoluted, maybe add something more _concrete_ like `AnimalsBuilder`/`DogsBuilder`, or whatever.

Also both exemples lack stubbing and verification.

bric3: Also the example should show how to use (stub/verify) the terminal method of a builder (`.make()`, etc.)

src/test/java/org/mockitousage/stubbing/BuilderStubbingTest.java
```diff
@@ -0,0 +1,98 @@
+package org.mockitousage.stubbing;
+
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.RETURNS_SELF;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class BuilderStubbingTest {
```
bric3: I suggest to rename to `ReturnsSelfStubbingTest` thoughts ?

bric3: Also as this class is about usage, I'd try to make a whole example of code for a builder, for a builder that behaves like this for example.

```java
Built built = someBuilder.withA(...)
                         .withB(...)
                         .withC(...)
                         .build()
```

JeremybellEU:
Processed feedback, squashed the two commits to one and rebased on master.

src/main/java/org/mockito/Mockito.java
```diff
@@ -1278,6 +1278,78 @@
...
+             public HttpBuilder withHeader(String header) {
+                 this.headers.add(header);
+                 return this;
+             }
+
+             public String request() {
+                return uri + headers.toString();
+             }
+        }
+     * </code></pre>
```
bric3: I wonder if this lack of `*` won't be causing javadoc issue.

TimvdLippe: My IDE showed the Javadoc without any problems. After quick googling, I stumbled upon http://www.triemax.com/products/jalopy/manual/javadoc.html#d0e19685 which mentions the stars are ignored in `<pre>` blocks. I therefore assume it is no problem. It also makes modifying of code examples easier in Javadoc.

bric3: I don't think you can bet on the IDE on this one, as the javadoc is generated by the `javadoc` tool which is quite picky (I've spent hours to have the javadoc to look as it is now).
Also the link you provide is a documentation for **jalopy** not the official `javadoc` tool. Plus it is not it the style of of project.

For these reasons I'd rather take a conservative approach there.

bric3:
Code wise it's good to be merged

TimvdLippe:
Added the `*` to the Javadoc.

raphw:
The build failed initially due to an unrelated issue. I restarted the build and everything went well. Merged.

bric3:
Cool thanks @TimvdLippe and @raphw

PR #290, opened 2015-09-11
RFC: Allow the collection of stacktraces in LocationImpl to be disabled.

rototor:
I use mockito not only for my test drivers, but also in production to stub objects in my "Frontend-Facade-Model", which is used for Freemarker templates. Instead of null, which the people writing the frontend templates dont really understand, they always get a "valid" object. In the case that an object really does not exist i create (and reuse) a stub only mock for it.

This works quite well, but in my server profiling LocationImpl() is very dominating. This is because of the new Throwable(), which gathers the stack trace on *every* method invocation on the mock. Which is of course fine and needed in test divers. But for my production use I don`t need the stack traces, they just slow everything down...

This is a minimal quick&dirty implementation, which allows to disable the stack trace gathering globally. It is not intended to be commit as is. 

What would be the clean way to disable the stack trace gathering? Is the "stubOnly()" setting enough indication that the user does not want / need the stack trace? Or should there be an other setting?

There are (if IDEA is correct) currently 31 new LocationImpl() calls in the mockito source. Making the construction of the LocationImpl  configurable would mean patching many places. I have no idea if it is worth the effort.

### DIFF
--- a/src/main/java/org/mockito/internal/debugging/LocationImpl.java
+++ b/src/main/java/org/mockito/internal/debugging/LocationImpl.java
@@ -4,31 +4,45 @@
  */
 package org.mockito.internal.debugging;
 
-import java.io.Serializable;
 import org.mockito.internal.exceptions.stacktrace.StackTraceFilter;
 import org.mockito.invocation.Location;
 
+import java.io.Serializable;
+
 public class LocationImpl implements Location, Serializable {
 
-    private static final long serialVersionUID = -9054861157390980624L;
-    private final Throwable stackTraceHolder;
-    private final StackTraceFilter stackTraceFilter;
+	private static final long serialVersionUID = -9054861157390980624L;
+	private final Throwable stackTraceHolder;
+	private final StackTraceFilter stackTraceFilter;
 
-    public LocationImpl() {
-        this(new StackTraceFilter());
-    }
+	/**
+	 * Ugly flag to disable the collection of stack traces. When writing tests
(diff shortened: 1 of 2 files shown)


JeremybellEU:
You probably use tabs instead of spaces, which causes the diff to be horrible.

Personally I dont really see usage in this proposal, but then again I havent ever used Mockito in production. Is there really not a cleaner way to fix this on your side instead of altering Mockito with a public variable?

rototor:
Well the clean way on my side would be to skip Mockito and directly create the mocks via ByteBuddy.  The mocks just need to return the predefined "empty" anwsers for boolean, lists etc. like Mockito does. Currently I do an defaultAnswer() on those mocks for some extra special cases (i.e. if a method returns an other facade I return the corresponding mock object for that facade and not null).

The more I think about this, Mockito seems the wrong tool for this use case. It works, yes. But it is slow (not only because of the constructed Throwable on each invocation, which is *really* slow, but also the additional checks and created objects when capturing the method invocations - which are fine for test drivers).

I will try my luck with ByteBuddy directly.

bric3:
Hi, I second @JeremybellEU on this, mockito is not really suited for production use, and on several point.

Anyway just a rapid feedback on your PR, please no ugly static flag ;)
Instead we have some sort of naive plugin mechanism. Using this or something else to have a way to disable location gathering when `stubOnly` is toggled would prove more mockito friendly ;)

Also @rototor thanks for pointing this out :)

PR #296, opened 2015-09-26
Update StackOverflow link to Mockito tag

JeremybellEU:
Tagging StackOverflow questions with the Mockito tag will get them most likely faster resolved.


szczepiq:
Thanks!

PR #301, opened 2015-10-08
Refactor some utilities and TODO done

Ariel-Isaacm:
Hey guys, awesome framework +1 i've been looking at the code and i did some improvements, also i did a TODO but i'm not quite sure how to handle that null check it would be awesome if you could check it out @bric3 @szczepiq 

Answers.java : private is not required 
ClassPathLoader.java: Removed unnecessary imports
ScenarioPrinter.java: Changed the string concat into another append
MockHandlerFactory.java: Return statement changed to be inline
ValuePrinter.java: TODO done, null check added

Edit 
ReleaseNotesExtension : Made authToken a final String

Edit 2: 
Some improvements to the ThreadTests :)

### DIFF
--- a/src/main/java/org/mockito/internal/matchers/text/ValuePrinter.java
+++ b/src/main/java/org/mockito/internal/matchers/text/ValuePrinter.java
@@ -41,7 +41,19 @@ public class ValuePrinter {
      * @return neatly formatted value list
      */
     public static String printValues(String start, String separator, String end, Iterator values) {
-        //TODO SF null check
+        if(start == null){
+            start = "(";
+        }
+        if (separator == null){
+            separator = ",";
+        }
+        if (end == null){
+            end = ")";
+        }
+        if (values == null){
+            values = new ArrayIterator(new String[]{""});
+        }
+
         StringBuilder sb = new StringBuilder(start);
         while(values.hasNext()) {
             sb.append(print(values.next()));
--- a/src/test/java/org/concurrentmockito/ThreadsRunAllTestsHalfManualTest.java
+++ b/src/test/java/org/concurrentmockito/ThreadsRunAllTestsHalfManualTest.java
@@ -170,7 +170,7 @@ public class ThreadsRunAllTestsHalfManualTest extends TestBase {
         boolean failed = false;
         for (AllTestsRunner t : threads) {
             t.join();
-            failed = failed ? true : t.isFailed();
+            failed |= t.isFailed();
         }
 
         return failed;
--- a/buildSrc/src/main/groovy/org/mockito/release/gradle/notes/ReleaseNotesExtension.groovy
+++ b/buildSrc/src/main/groovy/org/mockito/release/gradle/notes/ReleaseNotesExtension.groovy
@@ -12,7 +12,7 @@ class ReleaseNotesExtension {
     final static String EXT_NAME = "notes"
 
     File notesFile;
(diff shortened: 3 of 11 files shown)


src/test/java/org/concurrentmockito/ThreadsRunAllTestsHalfManualTest.java
```diff
@@ -170,7 +170,7 @@ public static boolean runInMultipleThreads(int numberOfThreads) throws Exception
         boolean failed = false;
         for (AllTestsRunner t : threads) {
             t.join();
-            failed = failed ? true : t.isFailed();
+            failed = failed || t.isFailed();
```
JeremybellEU: Does a `|=` also work?

src/main/java/org/mockito/internal/matchers/text/ValuePrinter.java
```diff
@@ -41,7 +41,9 @@ public static String print(Object value) {
      * @return neatly formatted value list
      */
     public static String printValues(String start, String separator, String end, Iterator values) {
-        //TODO SF null check
+        if(start == null || separator == null || end == null || values == null){
+            return null;
```
JeremybellEU: Returning null seems dangerous for future usages. Maybe it is better to return empty string or use an exception to notigy the developer he made an error.

Ariel-Isaacm: You're right :o! i'll put a empty string for now, thanks!

src/test/java/org/concurrentmockito/ThreadsRunAllTestsHalfManualTest.java
Ariel-Isaacm: Yes! you're right i already made the changes thanks for your comments :)

Ariel-Isaacm:
Commits squashed*

pmduaree:
Good job! The refactor is well done and the code is cleaner now. 

Maybe the coverage is gonna be decreased thanks to the null check. Maybe a new test will do.

src/main/java/org/mockito/internal/matchers/text/ValuePrinter.java
```diff
@@ -41,7 +41,10 @@ public static String print(Object value) {
      * @return neatly formatted value list
      */
     public static String printValues(String start, String separator, String end, Iterator values) {
-        //TODO SF null check
+        if(start == null || separator == null || end == null || values == null){
+            return "";
+            //TODO handle this situation
+        }
```
bric3: I think you can just replace `null` by some default values of each fields

bric3:
Looks good to me, I think you can just fixup the todo with my [comment](https://github.com/mockito/mockito/pull/301#discussion_r46943080)

Ariel-Isaacm:
Hi, i did the changes, are they what you expected? and should i squash the two commits?

raphw:
Yes please. Thank you!

Ariel-Isaacm:
Done :)

bric3:
Thanks @Ariel-Isaacm for the contribution

Ariel-Isaacm:
Thanks you for your patience :) @bric3 @raphw

raphw:
No, thank you!

bric3:
Yes exactly it was quite the opposite, thanks for the patience ;)

PR #302, opened 2015-10-09
Add support for answerClass in @Mock

scubbo:
See [this issue](https://code.google.com/p/mockito/issues/detail?id=345).

All credit for this change goes to kretes@, who actually authored it and created the [initial PR](https://github.com/mockito/mockito/pull/40) - I simply attended to the review comments and resubmitted since it appeared to have lain dormant for over a year.

### DIFF
--- a/src/main/java/org/mockito/Mock.java
+++ b/src/main/java/org/mockito/Mock.java
@@ -4,6 +4,8 @@
  */
 package org.mockito;
 
+import org.mockito.stubbing.Answer;
+
 import java.lang.annotation.Documented;
 import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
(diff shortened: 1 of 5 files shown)


bric3:
I wonder if this is really necessary the following proposal 

``java
@Mock(answerClass = `WhateverAnswer.class`) SomeType mocked;
```

can be easily replaced with the more powerful `mock` factory method, and it supports answer injection as well.

```java
SomeType mocked = mock(SomeType.class, new WhateverAnswer())
```

scubbo:
Sure - I agree that this functionality is already present in the factory method `mock` - that was precisely the motivation for this change, to bring `@Mock`'s functionality into line with `mock`'s.

bric3:
Yeah but that makes the feature more limited when used with `@Mock` annotation. I'm kinda wary to just bring the answer **class** in the annotation ; if we find a better way later on then we'll have to live with that API.

TimvdLippe:
I would prefer to keep the annotation as simple as possible. If additional features are required, `Mockito.mock()` should be used. Therefore closing this pull request.

Issue #337, opened 2016-01-03
Introduce functional interfaces to improve Java 8 utilisation of Mockito 2


ashleyfrieze:
There are two places where this would be very useful:

1. Custom matchers - used with argThat
2. Custom answers - for situations where you would normally create a subclass of Answer<T>

Some ideas here - https://solidsoft.wordpress.com/2015/06/30/more-compact-mockito-with-java-8-lambda-expressions-and-mockito-java8-add-ons/ - have proposed Java 8 specific extensions outside of mockito, but a small modification to mockito itself, usable by Java 8 but not dependent on it, would be sufficient to get a lot of benefit.

Example - if I wanted to check that the object passed to something had a child with the right value I should be able to write

    verify(mock, times(1)).someMethod(
        argThat(input -> input.getChild().getName().equals("Jim")));

Or if I wanted to answer a request

    doAnswer(answer(
        string -> string.length()
    )).when(mock).execute(anyString());

I will be submitting a pull request to demonstrate these capabilities

ChristianSchwarz:
What is wrong with [ArgumentMatcher ](http://site.mockito.org/mockito/docs/current/org/mockito/ArgumentMatcher.html)and [Answer](http://site.mockito.org/mockito/docs/current/org/mockito/stubbing/Answer.html) in mockito 2 ? They are functional interfaces.
 
```
when(callable.apply(anyString())).thenAnswer(invocation ->  invocation.getArgumentAt(0, String.class).length());

verify(callable).apply( argThat( text ->  ((String)text).length()>2 ));
```

ashleyfrieze:
Thanks for the comment. I was feeling a bit ignored :)

The answer is that these are harder to use than the change I'm proposing.

In the case of the boiler plate

    invocation.getArgumentAt(0, String.class).length()

My proposal provides a wrapper that can infer all the types and avoid you having to manipulate the invocation object directly.

In the case of the argument matcher, I took the code which I think was probably made that way to comply with something from Hamcrest, which uses Object as the input parameter to the ArgumentMatcher<T> rather than T and just made it more strongly typed. This again reduces the need for boilerplate.

ashleyfrieze:
Please see pull request #338 for the implementation I produced for this. I also found some orphaned tests which I brought back into the build, increasing code coverage.

We're using these features in my project right now (via a separate library alongside Mockito) and they're relatively useful.

ChristianSchwarz:
[...]

PR #338, opened 2016-01-03
Functional interfaces for Java 8 support in Mockito 2

ashleyfrieze:
Relating to Issue #337 - these are some changes that would allow more usability from Java 8 users. I'd really appreciate feedback on this.

It would make some of my callback use cases easier, and would provide a richer way of expressing assertions for arguments via argThat

One risk introduced by this change is increased use of the type parameter in ArgumentMatcher - this should be a good thing, yet I was left thinking it would surely have been done by someone else if it were easy.


src/main/java/org/mockito/Mockito.java
```diff
@@ -1094,6 +1101,110 @@
...
+ * As the {@link Answer} interface has just one method it is already possible to implement it in Java 8 using
+ * a lambda expression for very simple situations. The more you need to use the parameters of the method call,
+ * the more you need to typecast the arguments from {@link org.mockito.invocation.InvocationOnMock}.
+ *
+ * <p>
+ * Examples:
+ * <p>
+ * <pre class="code"><code class="java">
+ * // answer by returning 12 every time
+ * doAnswer(invocation -> {return 12;})
```
marcingrzejszczak: Are the `{` `}` braces necessary here?

src/main/java/org/mockito/Mockito.java
```diff
@@ -1094,6 +1101,110 @@
...
+ * // Java 8 - style 1
+ * doAnswer(AdditionalAnswers.<String,Callback>answerVoid((operand, callback) -> callback.receive("dummy"))
+ *     .when(mock).execute(anyString(), any(Callback.class));
+ *
+ * // Java 8 - style 2 - assuming static import of AdditionalAnswers
+ * doAnswer(answerVoid((String operand, Callback callback) -> callback.receive("dummy"))
+ *     .when(mock).execute(anyString(), any(Callback.class));
+ *
+ * // Java 8 - style 3 - where mocking function to is a static member of test class
+ * private static void dummyCallbackImpl(String operation, Callback callback) {
```
marcingrzejszczak: maybe just `dummyCallback` ?

src/main/java/org/mockito/Mockito.java
```diff
@@ -1094,6 +1101,110 @@
...
+ *     }})).when(mock).execute(anyString(), any(Callback.class));
+ *
+ * // returning a value is possible with the answer() function
+ * // and the non-void version of the functional interfaces
+ * // so if the mock interface had a method like
+ * String process(String input1, String input2);
+ *
+ * // this could be mocked
+ * // Java 8
+ * doAnswer(AdditionalAnswers.<String,String,String>answer((input1, input2) -> input1 + input2)))
```
marcingrzejszczak: Maybe an example that returns sth else than `String` would be useful? That way the users would immediately know which argument is responsible for the response type

src/main/java/org/mockito/internal/matchers/CompareTo.java
```diff
@@ -18,12 +18,8 @@ public CompareTo(Comparable<T> value) {
...
 
     @SuppressWarnings("unchecked")
-    public boolean matches(Object actual) {
-        if(!(actual instanceof Comparable)) {
-            return false;
-        }
-        
-        return matchResult(((Comparable) actual).compareTo(wanted));
+    public boolean matches(T actual) {
+        return matchResult(((Comparable)actual).compareTo(wanted));
```
marcingrzejszczak: why have you removed the cast check?

src/test/java/org/mockitousage/stubbing/StubbingWithAdditionalAnswersTest.java
```diff
@@ -0,0 +1,225 @@
...
+        given(iMethods.toIntWrapper(anyInt())).will(returnsFirstArg());
+
+        assertThat(iMethods.toIntPrimitive(1)).isEqualTo(1);
+        assertThat(iMethods.toIntWrapper(1)).isEqualTo(1);
+    }
+
+    @Test
+    public void can_return_based_on_strongly_types_one_parameter_function() throws Exception {
+        given(iMethods.simpleMethod(anyString()))
+                .will(answer(new AnswerFunctionalInterfaces.Answer1<String, String>() {
```
marcingrzejszczak: Is there any particular reason why you haven't used lambdas in this test?

marcingrzejszczak:
I like this change - how about you guys @bric3 @TimvdLippe ?

src/main/java/org/mockito/Mockito.java
```diff
@@ -1094,6 +1101,110 @@
...
+ * <p>
+ * <pre class="code"><code class="java">
+ * // answer by returning 12 every time
+ * doAnswer(invocation -> {return 12;})
+ *     .when(mock).doSomething();
+ *
+ * // answer by using one of the parameters - converting into the right
+ * // type as your go - in this case, returning the length of a string parameter
+ * // as the answer
+ * doAnswer(invocation -> invocation.getArgumentAt(2, String.class).length())
```
TimvdLippe: If I am not mistaken, `String.class` can be omitted here: http://site.mockito.org/mockito/docs/current/org/mockito/invocation/InvocationOnMock.html#getArgument(int)

src/main/java/org/mockito/internal/stubbing/answers/AnswerFunctionalInterfaces.java
```diff
@@ -0,0 +1,326 @@
...
+     * @param <T> return type
+     * @param <A> input parameter 1 type
+     * @return a new answer object
+     */
+    public static <T, A> Answer<T> toAnswer(final Answer1<T, A> answer) {
+        return new Answer<T>() {
+            @SuppressWarnings("unchecked")
+            public T answer(InvocationOnMock invocation) throws Throwable {
+                return answer.answer(
+                        (A)invocation.getArguments()[0]);
```
TimvdLippe: Can't we use `invocation.getArgument(x)` for these retrievals? Now we are retrieving the same array multiple times, but just with a different index.

TimvdLippe:
Yes I like it. It does require a rebase though to make sure we have enough coverage. Thanks @ashleyfrieze for spending time on improving Java 8 usability!

src/main/java/org/mockito/internal/stubbing/answers/AnswerFunctionalInterfaces.java
```diff
@@ -0,0 +1,326 @@
...
+                        (B)invocation.getArguments()[1],
+                        (C)invocation.getArguments()[2],
+                        (D)invocation.getArguments()[3],
+                        (E)invocation.getArguments()[4]);
+            }
+        };
+    }
+
+    /**
+     * Construct an answer from a three parameter answer interface
```
ChristianSchwarz: > Construct an answer from a ~~three~~ **five** parameter answer interface

src/test/java/org/mockitousage/stubbing/StubbingWithAdditionalAnswersTest.java
ChristianSchwarz: @marcingrzejszczak i guess cause mockito compiles against java 1.6

src/main/java/org/mockito/Mockito.java
ChristianSchwarz: @marcingrzejszczak even better we can reduce it to `doAnswer(invocation->12)`

src/main/java/org/mockito/internal/stubbing/answers/AnswerFunctionalInterfaces.java
```diff
@@ -0,0 +1,326 @@
+package org.mockito.internal.stubbing.answers;
+
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+
+/**
+ * Functional interfaces to make it easy to implement answers in Java 8
+ */
+public class AnswerFunctionalInterfaces {
+    /**
```
ChristianSchwarz: a private constructor would be nice since it make no sense to instanciate this class

src/main/java/org/mockito/internal/stubbing/answers/AnswerFunctionalInterfaces.java
marcingrzejszczak: Why `five` ? There are 3 params here or I'm blind :P

marcingrzejszczak: Construct an answer from a `five` parameter answer interface

src/main/java/org/mockito/Mockito.java
ChristianSchwarz: In the provided samples the parameter `operation` is unused. What about passing it to the receive-method instead of _"dummy"_ ?

src/main/java/org/mockito/Mockito.java
ChristianSchwarz: Providing type information is not neccessary here. A static import is possible if the parameter types are specified.

`doAnswer(answer((String a, String b)->a+b))`

It is shorter and better readable.

TimvdLippe:
@ashleyfrieze are you available to update this PR with the feedback or should we take it over from here? Thanks for your contribution!

ashleyfrieze:
I should be able to do these code review comments - or at the least have a crack at them - in the next 24 hours or so. Sorry to drop out the loop. Blame a combination of GitHub notifications and google's Inbox for filtering them.

TimvdLippe:
@ashleyfrieze No problem. Whenever you have time :)

ashleyfrieze:
Mockito2 is intended for version of Java pre-java 8, right?

It would be great to have a module that has some Java 8 examples of the usage of this expressed as unit tests - as per the feedback earlier on - is there a home for that?

TimvdLippe:
@ashleyfrieze yes it compiles with Java 5. @raphw are you aware of a method to run unit tests on a separate jdk version?

ashleyfrieze:
We could create a module in the build which requires Java 8. Arguably, we could even make it somehow optional for those people still compiling in Java 5... depends on how well the build tool will let us do that. Or I could create a demo project in Java 8 that demonstrates these features and we could put it outside the main build...?

ashleyfrieze:
One potentially larger impact change within this pull request is to the matcher interface, making it completely generic typed - this has raised a question (above) about the removing of an instanceof test (which i'll double check). It may also have an effect on migration/compatibility. Is everyone cool with this part of the change?

    public boolean matches(Object actual)

is now

    public boolean matches(T actual)

Which makes Java8 happier, but is also better than the weird thing that hamcrest was allowing...

We may be able to relax it to allow subclasses of T if that's preferred?

src/main/java/org/mockito/internal/matchers/CompareTo.java
ashleyfrieze: `<T extends Comparable<T>>` and I've modified the function to have an input parameter of type `T`, so the cast check is not necessary.

src/main/java/org/mockito/internal/stubbing/answers/AnswerFunctionalInterfaces.java
ashleyfrieze: Easily added if that's the house style. I find it unnecessarily drops the code coverage - who unit tests an unreachable private constructor? 

All those in favour of adding it?

src/main/java/org/mockito/Mockito.java
ashleyfrieze: I think it can't in this situation as there's no reason for the surrounding context to imply String over anything else. I think you can "detect" the context if you had a nearby type - in the way that `Collections.emptyList()` does

e.g. `List<MyThing> list = Collections.emptyList(); // will be a List<MyThing>` 

But in the above, the invocation is going to return an "int" which it's going to derive from the length of the String that happens to be the parameter at index 2 - nothing in the code can be used to derive that. Not in this example.

The idea of the functional interfaces is to make writing the above less cumbersome, so the above is intentionally a negative example.

src/main/java/org/mockito/Mockito.java
ashleyfrieze: What I'm trying to show here is what I think is a common test scenario - i.e. when you replace code which would pass the real input to the callback in some way, with something that just hardcodes the answer.

I guess the example didn't speak to you :)

Needs rework.

src/main/java/org/mockito/Mockito.java
ashleyfrieze: Will do.

The challenge I'm going to have to overcome here is that I think I should road test these examples in a Java 8 project and Mockito is currently refusing to build on my Java 8 set up. I'll install JDK6 or 7 shortly and do something useful with it.

ashleyfrieze:
I'm going to have to rebase this against master for it to work on merge. One issue is that the removal of `InvocationOnMock::<T> T getArgumentAt(int index, Class<T> clazz)` could cause some issues. I'll try it out. It's going to be a few days before I get there.

src/main/java/org/mockito/exceptions/stacktrace/StackTraceCleaner.java
```diff
@@ -17,10 +17,10 @@
...
 
     /**
-     * Decides if element is excluded.
+     * Decides if element is included.
      *
      * @param candidate element of the actual stack trace
      * @return whether the element should be excluded from cleaned stack trace.
      */
-    boolean isOut(StackTraceElement candidate);
+    boolean isIn(StackTraceElement candidate);
```
ashleyfrieze: Was implicated in a broken build at my end. The code that was causing it was suffering from a double-negative, so I refactored this into the positive.

src/main/java/org/mockito/internal/exceptions/stacktrace/DefaultStackTraceCleaner.java
```diff
@@ -8,13 +8,13 @@
 public class DefaultStackTraceCleaner implements StackTraceCleaner {
 
 	@Override
-	public boolean isOut(StackTraceElement e) {
+	public boolean isIn(StackTraceElement e) {
```
ashleyfrieze: Refactored into the positive - so switched trues to falses.

src/main/java/org/mockito/internal/exceptions/stacktrace/StackTraceFilter.java
```diff
@@ -30,7 +30,7 @@
         //TODO: investigate "keepTop" commit history - no effect!
         final List<StackTraceElement> filtered = new ArrayList<StackTraceElement>();
         for (StackTraceElement element : target) {
-            if (!CLEANER.isOut(element)) {
+            if (CLEANER.isIn(element)) {
```
ashleyfrieze: This reads nicer IMHO

src/test/java/org/mockitousage/verification/VerificationWithTimeoutTest.java
```diff
@@ -121,7 +121,7 @@ public void shouldAllowMixingOnlyWithTimeout() throws Exception {
         verify(mock, timeout(40).only()).clear();
     }
 
-    @Test
+    @Test(expected=NoInteractionsWanted.class)
```
ashleyfrieze: This test was flickering, so I extended its timeout (40 -> 50). In so doing, I had to understand what the test was up to, so I added comments to aid future readers and took out the counter intuitive exception expectation, replacing it with the default JUnit method.

subprojects/extTest/src/test/java/org/mockitousage/plugins/stacktrace/MyStackTraceCleanerProvider.java
```diff
@@ -13,11 +13,11 @@
     public StackTraceCleaner getStackTraceCleaner(final StackTraceCleaner defaultCleaner) {
         return new StackTraceCleaner() {
 			@Override
-			public boolean isOut(StackTraceElement candidate) {
+			public boolean isIn(StackTraceElement candidate) {
 				if (ENABLED && candidate.getMethodName().contains("excludeMe")) {
                     return false;
```
ashleyfrieze: It seems impossible that this could have ever worked before. The test wanted an exclusion of "excludeMe" functions when ENABLED was true. `isOut()` should return true to exclude something in the original. This would return false in the case that was being tested. It was hard to understand what this was getting at when you have to return true to say whether something is NOT included in a list.

Unless I've totally misunderstood this, then the refactoring and modification should be on the side of making things better. What I can't understand is why this isn't failing the main build!

TimvdLippe:
Are the stacktrace related changes required for this pull request? I would like to reduce the number of side-effect changes in this pull-request as it is already growing quite rapidly. I have no problem with a separate pull request to discuss these changes, just want to minimize the risk of merging in incorrect changes in an otherwise correct PR.

ashleyfrieze:
They were stopping me from building the code.

I know what you mean.

I could pull them into a separate pull request and then rebase this change over that...?

TimvdLippe:
Let's leave them in for now, but I would like to see the comments of @bric3 on these change. Are you fine with changing the `StackTraceFilter` functionality?

ashleyfrieze:
Just pushed some unit tests for some matchers - mainly to remove the complaint by the coverage report - but also to verify that my change to the interface has worked.

ashleyfrieze:
@TimvdLippe - please note, it's a refactor, rather than a change in outcome.. The code of the test HAS been changed, the runtime code should have exactly the same behaviour it ever did... though I can't see how that test EVER passed. It's possible that in some configurations that test is not run, or that it has no impact on some JVMs. I'm building on a JDK8.

TimvdLippe:
Ah okay, if it is backwards compatible then that is fine by me. I thought we were making breaking changes here. Then let's leave the changes here :heart:

ashleyfrieze:
@TimvdLippe thanks. My fear, as with any noob contributing to an OS project, is that ANY change could surprisingly be a breaking one :)

src/main/java/org/mockito/internal/stubbing/answers/AnswerFunctionalInterfaces.java
ashleyfrieze: `invocation.getArgument(x)` can be used. It doesn't change much and it's probably just a minor optimisation.

It might be better to get the array as a local variable and use it directly... but that's such a minor performance improvement, it's probably better not to do so, just for clarity.

src/main/java/org/mockito/internal/stubbing/answers/AnswerFunctionalInterfaces.java
TimvdLippe: Yes let's add it.

ashleyfrieze:
I've one thing left to deal with - the JavaDoc for `doAnswer` with the new functional interfaces for composing answers. I can refine that, or I can go a step further and add things like

````
public static Stubber doAnswer(Answer1 returningAnswerWith1Parameter) {}
public static Stubber doAnswer(VoidAnswer1 voidAnswerWith1Parameter) {}
.
.
.
````

which has been suggested and would make it a lot easier to use.

Can people give me thumbs up or down for whether this is a good thing to do or not. I've made a lot of line changes in the sacred `Mockito` class... dare I go further :)

src/main/java/org/mockito/internal/stubbing/answers/AnswerFunctionalInterfaces.java
ashleyfrieze: Done.

In a parallel universe, there's a mockito2 with _slightly_ higher coverage :p

src/main/java/org/mockito/Mockito.java
TimvdLippe: I think if we change it to `boolean isSameString(String input1, String input2)` the intent is clear enough.

TimvdLippe:
One last comment on the `doAnswer`. Other than that I think this PR LGTM. Thank you for your continued effort @ashleyfrieze !

raphw:
This looks really good. Let us please merge this soon.

As for writing tests: In Byte Buddy, I added some precompiled class files to the `resource` folder of the project such that they are not compiled during build. For each test I then add a regular test that first checks if the current JVM supports Java 8 or higher and if so, I explicitly run the test using `Class:forName` and then triggering a `Runnable::run` that the Java 8-specific test implements.

TimvdLippe:
Landed this pull request in https://github.com/mockito/mockito/commit/14617cc9fb43d2ad4b4eb622b4c720367615608a Thank you for your continued effort @ashleyfrieze (and also fixing that nasty test as reported in #435!)

PR #317, opened 2015-11-10
Modify StackTraceFilter to not exclude "good" stack trace elements

astafev:
currently StackTraceFilter filters in this way (example from javadoc):
[a+, b+, c-, d+, e+, f-, g+] -> [a+, b+, g+]

this patch makes it work:
[a+, b+, c-, d+, e+, f-, g+] -> [a+, b+, d+, e+, g+]

Not sure why it was implemented first time in this strange manner.

------------------------------------
**EDIT by Mockito team** : Fixes #316

PR #344, opened 2016-01-25
Show correct location of unwanted interaction with mock when using MockitoJUnitRule

UrsMetz:
This pull request by @indy5858 and me changes the stacktrace filtering so that when using `MockitoJUnitRule` the correct location of an unwanted interaction with a mock.

Before this commit the location on an unwanted interaction was not
reported properly. Instead of

```
No interactions wanted here:
-> at <package>.<TestClass>.<testMethod>(<TestClass.java>:<lineNumber>)
```
the assertion message said:
```
No interactions wanted here:
-> at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
```

The issue with the current behaviour in master is that all stacktrace elements above an element containing "org.mockito." are removed. As `org.mockito.internal.junit.JUnitRule` fullfills this condition and comes below the "payload" (i.e. test code and code under test) stack elements the interesting part of the stack is filtered.
In case the pull request #317 (which we only found after hacking on this) is merged this pull becomes obsolete except for the two tests that are added:
* checking that the "payload" stack elements aren't removed from the stack and
* a test for removal of the Byte Buddy stack elements.

### DIFF
--- a/src/main/java/org/mockito/internal/exceptions/stacktrace/DefaultStackTraceCleaner.java
+++ b/src/main/java/org/mockito/internal/exceptions/stacktrace/DefaultStackTraceCleaner.java
@@ -12,6 +12,8 @@ public class DefaultStackTraceCleaner implements StackTraceCleaner {
         boolean fromOrgMockito = e.getClassName().startsWith("org.mockito.");
         boolean isRunner = e.getClassName().startsWith("org.mockito.runners.");
         boolean isInternalRunner = e.getClassName().startsWith("org.mockito.internal.runners.");
-        return (fromMockObject || fromByteBuddyMockObject || fromOrgMockito) && !isRunner && !isInternalRunner;
+        boolean isJUnitRule = e.getClassName().startsWith("org.mockito.internal.junit.JUnitRule");
+        return (fromMockObject || fromByteBuddyMockObject || fromOrgMockito)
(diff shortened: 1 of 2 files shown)


raphw:
Looks good, thanks!

UrsMetz:
Thanks for the fast merge @raphw :-)
Are there any plans for new Mockito 1.x releases before the 2.0 release? If yes I would send a pull request backporting this to the 1.x branch (is there any 1.x branch?).

raphw:
I am not sure what the state of 2.0 is. I wished we would have done it already but it seems like nobody has the time right now.

Maybe @bric3 can say?

bric3:
Hi, 

Regarding 1.x we kinda dropped support, and development has been made on master thereby focusing on 2.x. By the way this version is pretty stable right now, but there are still a few things to polish here and there ; I don't have the remaining bits to do for a final version 2 in my head right now and indeed I don't have much free time as I'm quite busy with Devoxx Fr and in my private life (free time is getting scarce every year passing by ;)).

So I you can I advise you to switch to the latest beta, we're not far away anyway. Keep in mind however that other libraries may not be up to date regarding 2.x internal changes (thinking about powermock for example).

bric3:
By the way nice fix ;)

Issue #345, opened 2016-01-26
Verification using After and ArgumentCaptor returns unexpected size of captured values list


johhnny:
Hi, I found weird behavior of Mockito (ver. 1.10.19) when verifying method invocation with after() and ArgumentCaptor. The list of captured values returned by ArgumentCaptor has size much bigger than I expect. Moreover this behavior in connection with long timeout leads to exceeding memory limit. Example below.

```java
package org.mockito.test;

import org.junit.Before;
import org.junit.Test;
import org.mockito.ArgumentCaptor;

import static org.junit.Assert.assertEquals;
import static org.mockito.Mockito.after;
import static org.mockito.Mockito.mock;
import static org.mockito.Mockito.verify;

public class MockitoAfterFeatureTest {

    private RequestSender requestSender;
    private RequestProcessor requestProcessor;
    private ArgumentCaptor<String> stringArgumentCaptor;

    @Before
    public void doBeforeTest() {
        requestSender = mock(RequestSender.class);
        requestProcessor = new RequestProcessor(requestSender);
        stringArgumentCaptor = ArgumentCaptor.forClass(String.class);
    }

    @Test
    public void shouldVerifyUsingAfterAndArgumentCaptor() {
        // when
        requestProcessor.process("first_request");
        requestProcessor.process("second_request");

        // then
        verify(requestSender, after(2500).times(2)).send(stringArgumentCaptor.capture());
        assertEquals("first_request", stringArgumentCaptor.getAllValues().get(0));
        assertEquals("second_request", stringArgumentCaptor.getAllValues().get(1));

        assertEquals(2, stringArgumentCaptor.getAllValues().size());
    }

    public static class RequestProcessor {
        private final RequestSender requestSender;

        public RequestProcessor(RequestSender requestSender) {
            this.requestSender = requestSender;
        }

        public void process(String request) {
            requestSender.send(request);
        }
    }

    public interface RequestSender {
        void send(String request);
    }
}
```

thesnowgoose:
Issue #345 resolves this for atMost verification. What about other verification modes (i.e. times, atLeast)? Are you working on that, @tarnowskijan? Was the issue only for the atMost case? Because in the code from the description you used times instead of atMost.

tarnowskijan:
[...]

PR #349, opened 2016-02-04
Issue #345

tarnowskijan:
Solves #345 by removing already verified invocations before redundant capture of their arguments in AtMost class.

### DIFF
--- a/src/main/java/org/mockito/internal/verification/AtMost.java
+++ b/src/main/java/org/mockito/internal/verification/AtMost.java
@@ -5,6 +5,7 @@
 
 package org.mockito.internal.verification;
 
+import java.util.Iterator;
 import java.util.List;
 
 import org.mockito.exceptions.Reporter;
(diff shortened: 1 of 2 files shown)


tarnowskijan:
Could you send me details of the failure? I'm not able to reproduce the problem.

szpak:

It's not your fault. The Gradle process was killed. There could be a few reasons (like to large memory usage). The build should be rerun - @tarnowskijan, you can force it by pushing (with `--force`) amended version of the last commit (or someone being a project contributor could click a rerun button in Travis).

tarnowskijan:
@szpak, thanks for your explanation.

raphw:
I have rerun the build and there is a test error caused by your changes:
org.concurrentmockito.ThreadsRunAllTestsHalfManualTest > shouldRunInMultipleThreads FAILED

tarnowskijan:
Weird, I've run that test couple times and it always passed (other ones too). Are there some details available about what actually failed?

bric3:
About the build error it seems to be a travis-ci issue, there's multiple reports, this one for example travis-ci/travis-ci#5582

It does match what we currently see in the log (`128` + `9` (`SIGKILL`))

```
* What went wrong:
Execution failed for task ':test'.
```

https://s3.amazonaws.com/archive.travis-ci.org/jobs/107002559/log.txt

bric3:
Restarting the build multiple times made it work

PR #367, opened 2016-03-02
#365 Simplified the API to get an argument in custom Answer's

### LINKED ISSUE #365: Simplify the InvocationOnMock-API to get a casted argument
The 2nd argument of  `InvocationOnMock.getArgumentAt(int,Class<T>)` can be removed cause it is not neccessary.

Intuitively on would write this to get a "auto casted" argument:
`String text = invocation.getArgumentAt(1,String.class) `

The type can be ommited without side effects (the implementation of InvocationOnMock discards it anyway).
`String text = invocation.getArgumentAt(1,null)`

Therefore Mockito 2.0 should consider to simplify the API and ommit the `Class`-argument here. Maybe it is also a good chance to simplify the method name too. What about  `getArgument(int index)` ?

This is how the 'new' API could look like:
`String text = invocation.getArgument(1) `

ChristianSchwarz:
This pull request fixes #365 by introducing `T getArgument(int)` and deprecating `T getArgumentAt(int,Class<T>)` in type `InvocationOnMock`. This improves the readabilty of custom Answers. 

Now we can write: `String text = invocation.getArgument(1) ` 
Instead of: `String text = invocation.getArgumentAt(1,String.class) `

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -5,8 +5,8 @@
 
 **If looking for support**
 
-* Go to the [mockito mailing-list](http://groups.google.com/group/mockito) (moderated)
 * Search / Ask question on [stackoverflow](http://stackoverflow.com)
+* Go to the [mockito mailing-list](http://groups.google.com/group/mockito) (moderated)
 * Issues should always have a [Short, Self Contained, Correct (Compilable), Example](http://sscce.org) (same as any question on stackoverflow.com)
 
 # Contributing to Mockito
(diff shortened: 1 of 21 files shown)


ChristianSchwarz:
@raphw What is going wrong here? Locally everything works as expected, no failures appear.

raphw:
I am not entirely sure what the problem is. I suspect Travis CI to kill the build for some reason, we have had this before. Normally, I can restart the build at night where Travis is less busy and the build works. It is on my list of things to investigate and fix.

So far, thank you for your pull request. I think I have some free time this weekend to look at it!

bric3:
I can confirm this is a travis issue, see my comment on another PR https://github.com/mockito/mockito/pull/349#issuecomment-194295335

bric3:
@ChristianSchwarz I'm ok with the change, however could you rebase on the latest master, otherwise there is other commits that I don't want to merge.

ChristianSchwarz:
I reverted all unintended commits and merged them. So effectivly we have 1 commit containing all changes.

PR #369, opened 2016-03-09
Travis / CI improvements

bric3:
Travis CI jobs were failing inconsistently, 

The gradle forked process that executed tests was killed by the travis VM, hence the build failed with the following log without much clue, but the exit status code `137`. 
```
* What went wrong:
Execution failed for task ':test'.
```

_On the JVM it is the same as bash or zsh, when a program is "closed" with a signal, the exit code is `128` + the signal code._ There `137 = 128 + 9`, in other words the process ha received the `SIGKILL` signal. Other project have experienced that since a month (e.g. https://github.com/travis-ci/travis-ci/issues/5582).

**Indeed removing the container mode made the build run correctly.**

I took the time to tweak a bit more the build

- by leveraging the travis caches for gradle downloads
- by using the compressed oops for the 64bit JVMs (it should help only a little bit)
- by upgrading to gradle 2.11
- by reactivating code coverage

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,14 @@
 # More details on how to configure the Travis build
 # http://docs.travis-ci.com/user/build-configuration/
 
+# Speed up build by leversging travis caches
+before_cache:
+  - rm -f $HOME/.gradle/caches/modules-2/modules-2.lock
+cache:
+  directories:
+    - $HOME/.gradle/caches/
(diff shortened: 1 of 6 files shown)


Taher-Ghaleb:
Hello,

Sorry, I know it's been a long time ago, but I am curious why caching hasn't been enabled before this (i.e., since December 17, 2014, when it became available to open source projects).

Thank you.

PR #370, opened 2016-03-09
Fix #368 Class cast exception in PluginLoader.loadPlugin

### LINKED ISSUE #368: Class cast exception in PluginLoader.loadPlugin
I've been working on two defects for PowerMock ([first]( https://github.com/jayway/powermock/issues/634) and [second](https://github.com/jayway/powermock/issues/632)). Both has very similar stack trace: 

> java.lang.ExceptionInInitializerError
> at org.mockito.internal.exceptions.stacktrace.ConditionalStackTraceFilter.(ConditionalStackTraceFilter.java:17)
> at org.mockito.exceptions.base.MockitoException.filterStackTrace(MockitoException.java:41)
> at org.mockito.exceptions.base.MockitoException.(MockitoException.java:30)
> at org.mockito.exceptions.misusing.MockitoConfigurationException.(MockitoConfigurationException.java:18)
> at org.mockito.internal.configuration.plugins.PluginLoader.loadImpl(PluginLoader.java:66)
> at org.mockito.internal.configuration.plugins.PluginLoader.loadPlugin(PluginLoader.java:24)
> at org.mockito.internal.configuration.plugins.PluginRegistry.(PluginRegistry.java:12)
> at org.mockito.internal.configuration.plugins.Plugins.(Plugins.java:11)
> at org.mockito.internal.util.MockUtil.(MockUtil.java:24)
> at org.mockito.internal.MockitoCore.(MockitoCore.java:44)
> at org.mockito.Mockito.(Mockito.java:975)
> at org.powermock.modules.test.junit4.rule.objenesis.MockStaticNotPreparedTest.testWhenNotPrepared(MockStaticNotPreparedTest.java:45)
> at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
> at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
> at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
> at java.lang.reflect.Method.invoke(Method.java:498)
> at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:44)
> at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:15)
> at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:41)
> at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:20)
> at org.powermock.modules.junit4.rule.PowerMockStatement$1.run(PowerMockRule.java:65)
> at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
> at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
> at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
> at java.lang.reflect.Method.invoke(Method.java:498)
> at org.powermock.reflect.internal.WhiteboxImpl.performMethodInvocation(WhiteboxImpl.java:1898)
> at org.powermock.reflect.internal.WhiteboxImpl.doInvokeMethod(WhiteboxImpl.java:801)
> at org.powermock.reflect.internal.WhiteboxImpl.invokeMethod(WhiteboxImpl.java:666)
> at org.powermock.reflect.Whitebox.invokeMethod(Whitebox.java:401)
> at org.powermock.classloading.ClassloaderExecutor.execute(ClassloaderExecutor.java:98)
> at org.powermock.classloading.ClassloaderExecutor.execute(ClassloaderExecutor.java:78)
> at org.powermock.modules.junit4.rule.PowerMockStatement.evaluate(PowerMockRule.java:57)
> at org.junit.rules.ExpectedException$ExpectedExceptionStatement.evaluate(ExpectedException.java:110)
> at org.junit.runners.BlockJUnit4ClassRunner.runNotIgnored(BlockJUnit4ClassRunner.java:79)
> at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:71)
> at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:49)
> at org.junit.runners.ParentRunner$3.run(ParentRunner.java:193)
> at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:52)
> at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:191)
> at org.junit.runners.ParentRunner.access$000(ParentRunner.java:42)
> at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:184)
> at org.junit.runners.ParentRunner.run(ParentRunner.java:236)
> at org.junit.runner.JUnitCore.run(JUnitCore.java:157)
> at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:69)
> at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:234)
> at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:74)
> at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
> at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
> at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
> at java.lang.reflect.Method.invoke(Method.java:498)
> at com.intellij.rt.execution.application.AppMain.main(AppMain.java:144)
> Caused by: java.lang.NullPointerException
> at org.mockito.internal.configuration.plugins.Plugins.getStackTraceCleanerProvider(Plugins.java:17)
> at org.mockito.internal.exceptions.stacktrace.StackTraceFilter.(StackTraceFilter.java:21)
> ... 51 more

As @bric3 recommend in issue #355 I had started debug `PluginLoader` and I found class cast exception in `org.mockito.internal.configuration.plugins.PluginLoader#loadImpl`. And this class cast exception is transformed to NPE. Main reason of class class exception, that plugin interface class and plugin implementor could be loaded by different class loader, because for plugin implementor current context class loader is used. 

I've create the [gist](https://gist.github.com/thekingnothing/9674b6196f91b67aa944) with test which demonstrate the issue. 

Issue could be reproduce on both latest version:  1.10.19 2, 2.0.43-beta

I'm going to  raise a pull request soon for Mockito 2, but I think hot-fix also should be release for 1.10.

thekingnothing:
Fix issue #368 

- Use `Class.forName` to get plugin implementer class instead loading via context class loader 
- Use `RuntimeException` instead `MockitoException` because `MockitoException` depends on `PluginLoader` and if something incorrect with loader then stack trace  will be also incorrect. 

I don't understand why context class loader was used, because anyway other classes that use plugin is loaded by class loaded of current class.

### DIFF
--- a/src/main/java/org/mockito/internal/configuration/plugins/PluginLoader.java
+++ b/src/main/java/org/mockito/internal/configuration/plugins/PluginLoader.java
@@ -1,6 +1,5 @@
 package org.mockito.internal.configuration.plugins;
 
-import org.mockito.exceptions.base.MockitoException;
 import org.mockito.exceptions.misusing.MockitoConfigurationException;
 import org.mockito.internal.util.collections.Iterables;
 import org.mockito.plugins.PluginSwitch;
@@ -21,10 +20,17 @@ class PluginLoader {
      * Scans the classpath for given pluginType. If not found, default class is used.
      */
     <T> T loadPlugin(Class<T> pluginType, String defaultPluginClassName) {
+
         T plugin = loadImpl(pluginType);
         if (plugin != null) {
             return plugin;
         }
+
+        return loadDefaultPlugin(pluginType, defaultPluginClassName);
+    }
+
+
+    private <T> T loadDefaultPlugin(Class<T> pluginType, String defaultPluginClassName) {
 
         try {
             // Default implementation. Use our own ClassLoader instead of the context
@@ -32,9 +38,11 @@ class PluginLoader {
             // Mockito and may not be available via the context ClassLoader.
             return pluginType.cast(Class.forName(defaultPluginClassName).newInstance());
         } catch (Exception e) {
-            throw new MockitoException("Internal problem occurred, please report it. " +
-                    "Mockito is unable to load the default implementation of class that is a part of Mockito distribution. " +
-                    "Failed to load " + pluginType, e);
+            // use RuntimeException instead MockitoException, because MockitoException depends on plugins and can
+            // break stack track
+            throw new RuntimeException("Internal problem occurred, please report it. " +
+                                               "Mockito is unable to load the default implementation of class that is a part of Mockito distribution. " +
+                                               "Failed to load " + pluginType, e);
         }
     }
 
@@ -42,7 +50,7 @@ class PluginLoader {
      * Equivalent to {@link java.util.ServiceLoader#load} but without requiring
      * Java 6 / Android 2.3 (Gingerbread).
      */
-    <T> T loadImpl(Class<T> service) {
+    private <T> T loadImpl(Class<T> service) {
         ClassLoader loader = Thread.currentThread().getContextClassLoader();
         if (loader == null) {
             loader = ClassLoader.getSystemClassLoader();
@@ -51,20 +59,22 @@ class PluginLoader {
         try {
             resources = loader.getResources("mockito-extensions/" + service.getName());
         } catch (IOException e) {
-            throw new MockitoException("Failed to load " + service, e);
+            // use RuntimeException instead MockitoException, because MockitoException depends on plugins and can
+            // break stack track
+            throw new RuntimeException("Failed to load " + service, e);
         }
 
         try {
             String foundPluginClass = new PluginFinder(pluginSwitch).findPluginClass(Iterables.toIterable(resources));
             if (foundPluginClass != null) {
-                Class<?> pluginClass = loader.loadClass(foundPluginClass);
+                Class<?> pluginClass = Class.forName(foundPluginClass);
                 Object plugin = pluginClass.newInstance();
                 return service.cast(plugin);
             }
             return null;
         } catch (Exception e) {
             throw new MockitoConfigurationException(
-                    "Failed to load " + service + " implementation declared in " + resources, e);
+                                                           "Failed to load " + service + " implementation declared in " + resources, e);
         }
     }
 }
--- a/src/main/java/org/mockito/internal/configuration/plugins/PluginRegistry.java
+++ b/src/main/java/org/mockito/internal/configuration/plugins/PluginRegistry.java
@@ -30,7 +30,7 @@ class PluginRegistry {
     /**
      * Returns the implementation of the mock maker available for the current runtime.
      *
-     * <p>Returns {@link org.mockito.internal.creation.cglib.CglibMockMaker} if no
+     * <p>Returns {@link org.mockito.internal.creation.bytebuddy.ByteBuddyMockMaker} if no
      * {@link org.mockito.plugins.MockMaker} extension exists or is visible in the current classpath.</p>
      */
     MockMaker getMockMaker() {
--- a/src/test/java/org/mockito/internal/configuration/plugins/PluginLoaderTest.java
+++ b/src/test/java/org/mockito/internal/configuration/plugins/PluginLoaderTest.java
@@ -1,0 +1,77 @@
+package org.mockito.internal.configuration.plugins;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.mockitoutil.ClassLoaders;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Test PluginLoader. This test check that loaded plugin can be assigned to fields of classes what loaded by
+ * application class loader
+ */
+public class PluginLoaderTest {
+
+    private static final String PLUGIN_RESOURCE_CONTENT = PluginImpl.class.getName();
+    private static final String PLUGIN_RESOURCE_PATH = "mockito-extensions/org.mockito.internal.configuration.plugins.PluginInterface";
+    private PluginLoader pluginLoader;
+
+    @Before
+    public void setUp() {
+        pluginLoader = new PluginLoader(new DefaultPluginSwitch());
+    }
+
+    @Test
+    public void should_load_custom_plugin_without_context_classloader() {
+
+        Object pluginInterface = loadPlugin();
+
+        assertThatCustomPluginLoaded(pluginInterface);
+    }
+
+
+    @Test
+    public void should_load_custom_plugin_with_custom_context_classloader() throws Exception {
+
+        ClassLoader classLoader = ClassLoaders.inMemoryClassLoader()
+                                              .withClassDefinition(PLUGIN_RESOURCE_PATH, PLUGIN_RESOURCE_CONTENT.getBytes("UTF-8"))
+                                              .build();
+
+        Thread.currentThread().setContextClassLoader(classLoader);
+
+        Object plugin = loadPlugin();
+
+        assertThatCustomPluginLoaded(plugin);
+    }
+
+    @Test
+    public void should_load_default_plugin_with_custom_context_classloader() throws Exception {
+
+        ClassLoader classLoader = ClassLoaders.inMemoryClassLoader().build();
+
+        Thread.currentThread().setContextClassLoader(classLoader);
+
+        Object plugin = loadPlugin();
+
+        assertThatDefaultPluginLoaded(plugin);
+    }
+
+    private Object loadPlugin() {
+        return pluginLoader.loadPlugin(PluginInterface.class, DefaultPlugin.class.getName());
+    }
+
+    private void assertThatCustomPluginLoaded(Object plugin) {
+
+        assertThat(plugin.getClass())
+                .describedAs("Loaded plugin class '%s' is not assignable to local class '%s'", plugin.getClass(), PluginImpl.class)
+                .isAssignableFrom(PluginImpl.class);
+    }
+
+    private void assertThatDefaultPluginLoaded(Object plugin) {
+        assertThat(plugin.getClass())
+                .describedAs("Loaded plugin class '%s' is not assignable to  default plugin local class '%s'", plugin.getClass(), DefaultPlugin.class)
+                .isAssignableFrom(DefaultPlugin.class);
+    }
+
+
+}
--- a/src/test/resources/mockito-extensions/org.mockito.internal.configuration.plugins.PluginInterface
+++ b/src/test/resources/mockito-extensions/org.mockito.internal.configuration.plugins.PluginInterface
@@ -1,0 +1,2 @@
+org.mockito.internal.configuration.plugins.PluginImpl
+# Used for test : org.mockito.internal.configuration.plugins.PluginLoaderTest
--- a/src/test/java/org/mockito/internal/configuration/plugins/DefaultPlugin.java
+++ b/src/test/java/org/mockito/internal/configuration/plugins/DefaultPlugin.java
@@ -1,0 +1,8 @@
+package org.mockito.internal.configuration.plugins;
+
+/**
+ *
+ */
+class DefaultPlugin implements PluginInterface {
+
+}
--- a/src/test/java/org/mockito/internal/configuration/plugins/PluginImpl.java
+++ b/src/test/java/org/mockito/internal/configuration/plugins/PluginImpl.java
@@ -1,0 +1,8 @@
+package org.mockito.internal.configuration.plugins;
+
+/**
+ *
+ */
+public class PluginImpl implements PluginInterface {
+
+}
--- a/src/test/java/org/mockito/internal/configuration/plugins/PluginInterface.java
+++ b/src/test/java/org/mockito/internal/configuration/plugins/PluginInterface.java
@@ -1,0 +1,7 @@
+package org.mockito.internal.configuration.plugins;
+
+/**
+ *
+ */
+interface PluginInterface {
+}


src/main/java/org/mockito/internal/configuration/plugins/PluginRegistry.java
```diff
@@ -8,16 +8,19 @@
...
 
     private final PluginSwitch pluginSwitch
-            = new PluginLoader(new DefaultPluginSwitch()).loadPlugin(PluginSwitch.class, DefaultPluginSwitch.class.getName());
+            = new PluginLoader(new DefaultPluginSwitch()).loadPlugin(PluginSwitch.class, DefaultPluginSwitch.class
+                                                                                                  .getName());
 
     private final MockMaker mockMaker
-            = new PluginLoader(pluginSwitch).loadPlugin(MockMaker.class, "org.mockito.internal.creation.bytebuddy.ByteBuddyMockMaker");
+            = new PluginLoader(pluginSwitch).loadPlugin(MockMaker.class, "org.mockito.internal.creation.bytebuddy" +
+                                                                                  ".ByteBuddyMockMaker");
```
bric3: Is it possible to avoid the split string ?

src/main/java/org/mockito/internal/configuration/plugins/PluginLoader.java
```diff
@@ -51,20 +59,22 @@ public PluginLoader(PluginSwitch pluginSwitch) {
...
-                Class<?> pluginClass = loader.loadClass(foundPluginClass);
+                Class<?> pluginClass = Class.forName(foundPluginClass);
                 Object plugin = pluginClass.newInstance();
                 return service.cast(plugin);
             }
             return null;
         } catch (Exception e) {
             throw new MockitoConfigurationException(
-                    "Failed to load " + service + " implementation declared in " + resources, e);
+                                                           "Failed to load " + service + " implementation declared in " + resources, e);
```
bric3: Unnecessary code reformat is Eclipse(/other IDE) acting a fool.

src/test/java/org/mockito/internal/configuration/plugins/PluginLoaderTest.java
```diff
@@ -0,0 +1,88 @@
...
+        Object pluginInterface = pluginLoader.loadPlugin(PluginInterface.class, "org.mockito.internal" +
+                                                                                                  ".configuration.plugins.DefaultPlugin");
+
+        assertEquals(PluginImpl.class, pluginInterface.getClass());
+    }
+
+    @Test
+    public void if_custom_context_class_loader_is_used_then_plugin_should_be_load() throws Exception {
+
+        ClassLoader classLoader = ClassLoaders.inMemoryClassLoader().withParent(getParentClassLoader()).withClassDefinition("", readResource()).build();
```
bric3: This doesn't work as expected, it is like creating a noname class, whose class byte code is the string found in the mockito-extensions file ;)

src/main/java/org/mockito/internal/configuration/plugins/PluginLoader.java
thekingnothing: IDE is configured for another project formatting, in contribute guide I found only mention about spaces. So I've continued use my configuration

src/test/java/org/mockito/internal/configuration/plugins/PluginLoaderTest.java
```diff
@@ -0,0 +1,88 @@
...
+    @Mock
+    private PluginSwitch pluginSwitch;
+
+    @Before
+    public void setUp() {
+        given(pluginSwitch.isEnabled(anyString())).willReturn(true);
+    }
+
+    @Test
+    public void without_custom_context_class_loader_plugin_should_be_load() {
```
bric3: Could be renamed to `should_load_custom_plugin_without_context_classloader`

src/test/java/org/mockito/internal/configuration/plugins/PluginLoaderTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+
+        Object pluginInterface = pluginLoader.loadPlugin(PluginInterface.class, "org.mockito.internal" +
+                                                                                                  ".configuration.plugins.DefaultPlugin");
+
+        assertEquals(PluginImpl.class, pluginInterface.getClass());
+    }
+
+    @Test
+    public void if_custom_context_class_loader_is_used_then_plugin_should_be_load() throws Exception {
```
bric3: Could be renamed to `should_load_custom_plugin_with_custom_context_classloader`

src/test/java/org/mockito/internal/configuration/plugins/PluginLoaderTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+        ClassLoader classLoader = ClassLoaders.inMemoryClassLoader().withParent(getParentClassLoader()).withClassDefinition("", readResource()).build();
+
+        Thread.currentThread().setContextClassLoader(classLoader);
+
+        Object pluginInterface = pluginLoader.loadPlugin(PluginInterface.class, "org.mockito.internal" +
+                                                                                                  ".configuration.plugins.DefaultPlugin");
+
+        assertEquals(PluginImpl.class, pluginInterface.getClass());
```
bric3: Asserts could be more useful when asserting on classloader themself in this case :
```java
assertEquals(classLoader.loadClass(PluginImpl.class.getName()).getClassLoader(), pluginInterface.getClass().getClassLoader());
```

src/main/java/org/mockito/internal/configuration/plugins/PluginLoader.java
bric3: Yep that's usually an issue ;)
I've myself too in such issue ;) Now I setup my IDE never reformat non modified lines unless I ask it for.

src/test/java/org/mockito/internal/configuration/plugins/PluginLoaderTest.java
thekingnothing: If you look at source code of  `InMemoryClassLoader`

 ```
@Override
        public Enumeration<URL> getResources(String ignored) throws IOException {
            return inMemoryOnly();
        }

        private Enumeration<URL> inMemoryOnly() {
            final Set<String> names = inMemoryClassObjects.keySet();
            return new Enumeration<URL>() {
                private final MemHandler memHandler = new MemHandler(InMemoryClassLoader.this);
                private final Iterator<String> it = names.iterator();

                public boolean hasMoreElements() {
                    return it.hasNext();
                }

                public URL nextElement() {
                    try {
                        return new URL(null, SCHEME + ":" + it.next(), memHandler);
                    } catch (MalformedURLException rethrown) {
                        throw new IllegalStateException(rethrown);
                    }
                }
            };
        }
```

The `InMemoryClassLoader` returns all class as resources and `PluginLoader` reads these resources. So if I remove `readResource` then `PluginLoader` will not find any plugin and then create default instance. 

I can replace `readResource` via `"org.mockito.internal.configuration.plugins.PluginImpl".getBytes()`, but I prefer use same arrangement as for another test.

File: src/test/java/org/mockito/internal/configuration/plugins/PluginLoaderTest.java (hunk 6 above)
Comment (thekingnothing) [inline src/test/java/org/mockito/internal/configuration/plugins/PluginLoaderTest.java]: Suggested assert will be incorrect.  The `PluginImpl`should be loaded via `classLoader` (who plays role context class loader). It should be loaded with `AppClassLoader` to be assignable to fields of `PluginRegistry` even if context class loader != `AppClassLoader`. I change assert to more readable.

File: src/main/java/org/mockito/internal/configuration/plugins/PluginLoader.java (hunk 2 above)
Comment (thekingnothing) [inline src/main/java/org/mockito/internal/configuration/plugins/PluginLoader.java]: I've also  setup my IDE to never reformat non modified code, but I face issue then I change code, revert changes, then IDE format code.

**Comment by thekingnothing** [comment]
@bric3, 

I modified test. I renamed test methods, change comment so it now more correspond to my goal and what I really test. I change assert and give meaningful name. Also I fixed another mistake - I used Mockito Annotation to test PluginLoader then annotation injector uses PluginLoader. I don't use Mockito any more to test Mockito :)

File: src/test/java/org/mockito/internal/configuration/plugins/PluginLoaderTest.java [hunk 7]
```diff
@@ -0,0 +1,87 @@
...
+
+        assertThatLoadedPluginAssignable(pluginInterface);
+    }
+
+
+    @Test
+    public void should_load_custom_plugin_with_custom_context_classloader() throws Exception {
+
+        ClassLoader classLoader = ClassLoaders.inMemoryClassLoader().withParent(getParentClassLoader())
+                                          .withClassDefinition("", readResource()).build();
```
bric3: This is still incorrect use of the in-memory classloader ;)
`.withClassDefinition("", readResource())` has no impact on test

thekingnothing: If I remove `.withClassDefinition("", readResource())` then test will fail, because default plugin implementer will be loaded. Without class definition the in-memory classloader returns empty resource `Enumeration` . So I need a easy way to create class loader which will return as resource that I need to return and failed to load any class. Any suggestion?

bric3: This is definitely wrong. The `.withClassDefinition("", readResource())` directive registers a the class name `""` whose class byte definition is the content of the file `mockito-extensions/org.mockito.internal.configuration.plugins.PluginInterface`. If this code was used then the JVM would raise an error.

thekingnothing: The `.withClassDefinition("", readResource())` doesn't register the class. It simple put value of `
readResource()` to map with key empty string as key. 
`private Map<String , byte[]> inMemoryClassObjects = new HashMap<String , byte[]>();`

And it's used only in to cases: 

-  find class with name, but empty string is invalid class name value, so it should never be called and if it called then `Exception`is correct behaviour for this case 
- getResources - then `Enumeration` is return. 

I could use not empty string, but `"mockito-extensions/org.mockito.internal.configuration.plugins.PluginInterface"`, but meaning will be same.

bric3: I understand the idea, but the test pass with or without `.withClassDefinition("", readResource())`.

The `ClassLoaders` is just a basic tool that wasn't designed to support something else but class bytes. It may be something to improve. Anyway it's just coincidental that file is supported.

Anyway for better readability it would have been clearer to write it like that, and then it's possible to discard `readResource()` :

```java
ClassLoader classLoader = ClassLoaders.inMemoryClassLoader()
                                      .withParent(getParentClassLoader())
                                      .withClassDefinition("mockito-extensions/org.mockito.internal.configuration.plugins.PluginInterface", 
                                                           PluginImpl.class.getName().getBytes("UTF-8"))
                                      .build();
```

thekingnothing: Test passed without because assertion was a bit incorrect after last refactoring :(
I fixed assertion and added one more test which check loading default plugin.

src/main/java/org/mockito/internal/configuration/plugins/PluginLoader.java
```diff
@@ -51,20 +59,22 @@ public PluginLoader(PluginSwitch pluginSwitch) {
...
+            // use RuntimeException instead MockitoException, because MockitoException depends on plugins and can
+            // break stack track
+            throw new RuntimeException("Failed to load " + service, e);
         }
 
         try {
             String foundPluginClass = new PluginFinder(pluginSwitch).findPluginClass(Iterables.toIterable(resources));
             if (foundPluginClass != null) {
-                Class<?> pluginClass = loader.loadClass(foundPluginClass);
+                Class<?> pluginClass = Class.forName(foundPluginClass);
```
bric3: What do you think of trying first `loader.loadClass(found)` then if it doesn't work try the `Class.forName(found)` ?
I'm wary of the plain `Class.forName` as it will always choose the mockito current classloader.

src/test/resources/mockito-extensions/org.mockito.internal.configuration.plugins.PluginInterface
```diff
@@ -0,0 +1 @@
+org.mockito.internal.configuration.plugins.PluginImpl
```
bric3: For documentation maybe add 

```
# Used for test : org.mockito.internal.configuration.plugins.PluginLoaderTest
```

src/test/java/org/mockito/internal/configuration/plugins/PluginLoaderTest.java
```diff
@@ -0,0 +1,76 @@
...
+        boolean assertion = PluginImpl.class.isAssignableFrom(plugin.getClass());
+        String message = String.format("Plugin class %s is not assignable to %s", plugin.getClass(), PluginImpl.class);
+        assertTrue(message, assertion);
+    }
+
+    private void assertThatDefaultPluginLoaded(Object plugin) {
+        boolean assertion = DefaultPlugin.class.isAssignableFrom(plugin.getClass());
+        String message = String.format("Plugin class %s is not assignable to default %s", plugin.getClass(), DefaultPlugin.class);
+        assertTrue(message, assertion);
+    }
```
bric3: If you want there's AssertJ on the classpath, that could be rewritten to 

```java
Assertions.assertThat(plugin.getClass())
          .describedAs("Loaded plugin class '%s' is not assignable to local class '%s'", plugin.getClass(), PluginImpl.class)
          .isAssignableFrom(PluginImpl.class);
```

src/main/java/org/mockito/internal/configuration/plugins/PluginLoader.java
bric3: After playing a bit with the code, all the tests can still pass with the old statement `loader.loadClass(found)`

thekingnothing: I wonder how  old statement could work in this case: 

1. The `MockUtil` is created first time from class which is loaded by Application class loader.  Important that no other Mockito classes haven been used before creating `MockUtil`. So it means that the `MockUtil` class will be loaded by this class loader. 
2. During class initialisation the `Plugins.getMockMaker` is  called, an `Plugins` class also loaded by application class loader.
3. Next step `PluginRegistry` - and again application class loader is used. 
4. And now the `PluginLoader` is called, which uses completed different context class loader to load plugin. Okay, plugin can be loaded and may even cast and return, but how it could be assign to field? 

By application class loader I mean that it's class loader which is used to load first class with `main()`method. A  context class loader - is a new class loader which is create via `new` and doesn't have application class loader as a parent

src/test/resources/mockito-extensions/org.mockito.internal.configuration.plugins.PluginInterface
thekingnothing: Done

src/main/java/org/mockito/internal/configuration/plugins/PluginLoader.java
raphw: Depends on the context class loader which could be just any class loader. (Even though tests are mainly run from the class path.) I do however agree that this should be `loader.findClass` at this place.

raphw:
I do not think that this PR is correct. The basic change is to disable that a plugin is loaded via a class loader that determines the plugin but via Mockito's class loader. If Mockito is loaded via an OSGi bundle, for example, this would break visibility constraints. 

We need to require that a class loader that identifies a plugin can also load this plugin and can also see Mockito. For the tests, this is broken.  

It is however true that the exceptions should not rely on the plugin architecture. I will fix this part manually. Thank you for opening this PR, please get in touch if you need to discuss this issue further.

PR #371, opened 2016-03-09
speedup travis build a bit by downloading gradle-bin instead of gradl…

PascalSchumacher:


bric3:
I think gradle-all was necessary for something, maybe IDE support at some point.
Anyway the gradle dist is supposedly cached now which should help

raphw:
I do not see why this would be required after looking at the build. I will give this a try.

PR #373, opened 2016-03-11
Fixes #365 Simplify the InvocationOnMock-API to get a casted argument

### LINKED ISSUE #365: Simplify the InvocationOnMock-API to get a casted argument

ChristianSchwarz:
This pull request fixes #365 by introducing `T getArgument(int)` and deprecating `T getArgumentAt(int,Class<T>)` in type `InvocationOnMock`. This improves the readabilty of custom Answers. 

This...
`String text = invocation.getArgumentAt(1,String.class) ` 
can be replaced by... 
`String text = invocation.getArgument(1) `

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/InterceptedInvocation.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/InterceptedInvocation.java
@@ -113,9 +113,14 @@ class InterceptedInvocation implements Invocation, VerificationAwareInvocation {
     }
 
     @Override
-    @SuppressWarnings("unchecked")
+    @Deprecated
     public <T> T getArgumentAt(int index, Class<T> clazz) {
-        return (T) arguments[index];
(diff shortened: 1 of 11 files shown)


raphw:
Looks good, makes implementing custom answers easier indeed.

PR #375, opened 2016-03-15
Fixes #358 Added custom capturing matcher to support conditional capture

### LINKED ISSUE #358: Enhancement: ArgumentCaptor.captureIf(ArgumentMatcher)
Assume we have the following test.

```
 @Mock
 private Consumer<Collection<?>> consumer;

 @Test
       
    consumer.accept(new HashSet<>());
    consumer.accept(new ArrayList<>());
       
    ArgumentCaptor<ArrayList<?>> captor = ArgumentCaptor.forClass(ArrayList.class);
    verify(consumer).accept(captor.capture());
 }
```

The test fails with:
```
org.mockito.exceptions.verification.TooManyActualInvocations:
consumer.accept(<Capturing argument>);
Wanted 1 time:
-> at foo.TestClass.testName(TestClass.java:64)
But was 2 times. Undesired invocation:
-> at agh.TestClass.testName(TestClass.java:61)

    at foo.TestClass.test(TestClass.java:64)
    [..]
```

The test fails because the ArgumentCaptor stores every type, because the type information is erased a runtime. How can we capture the type that is specified in the captor, in this case `ArrayList`?  It is possible to implement an `Answer `that stores the argument, and use
`Matchers.isA(ArrayList.class) `but thats ugly->

` doAnswer(captureArg1).when(consumer).accept(isA(ArrayList.class));`

Maybe it is a good idea to create a pull req and extend the ArgumentCaptor API by adding a method like `<T> T captureIf(ArgumentMatcher<? extends T>) `?

Then we could write:
`verify(consumer).accept(captor.captureIf(isA(ArrayList.class)));`

terebesirobert:
This fixes #358 

I added a `CustomCapturingMatcher` to `ArgumentCaptor`. This will store the data based on a `ArgumentMatcher`.
For example:

``` java
Processor mock = ...

mock.process(new ArrayList<>);
mock.process(new HashSet<>);

ArgumentCaptor captor = ...

verify(mock).process(captor.capture());
```

the `CapturingMatcher` inside the captor will capture the list and set too (the implementation does not perform any type check on the arguments).

I added a new conditional capture method `mock.captureIf(ArgumentMatcher)` that will store data based on the matcher received as argument.

This will save a lot of time when writing unit tests.

check list


ChristianSchwarz:
@terebesirobert Thank for your pull request! Would you mind to edit the first comment and title according to the [Contibution Guide](https://github.com/mockito/mockito/blob/master/.github/CONTRIBUTING.md). That way the issue and the pull request are linked automatically.

ChristianSchwarz:
I have some questions about the user experience/expectation of the `captureIf(..)` API,  that lead me to the conclusion that it is better have an API like `getValue(ArgumentMatcher)`. What do you think of this uses cases:

Use Case 1:
```
ArgumentCaptor<Collection> captor = ...
Consumer<Collection> consumer = mock(Consumer.class)

consumer.consume(new ArrayList());
consumer.consume(new HashSet());

verify(consumer).consume(captor.captureIf(instanceOf(List.class)));
verify(consumer).consume(captor.captureIf(instanceOf(Set.class)));

List<Collection> values = captor.getValues(); 
```

What does the user expect is the content of `values` here?

-----

Use Case 2:
```
ArgumentCaptor<Number> captor = ...
Consumer<Number> consumer = mock(Consumer.class)

consumer.consume(1);
consumer.consume(-2);

verify(consumer).consume(captor.captureIf(gt(0)));
verify(consumer).consume(captor.capture());

List<Collection> values = captor.getValues(); 
```
What does the user expect is the content of `values` here?

src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
```diff
@@ -314,4 +330,53 @@ public void captures_correctly_when_captor_used_on_pure_vararg_method() throws E
...
+        //when
+        mock.simpleMethod(FOO, 1);
+        mock.simpleMethod(BAR, 2);
+        mock.simpleMethod(FOO, 3);
+        mock.simpleMethod(BAR, 4);
+        
+        mock.simpleMethod(FOO, 5);
+
+        //then
+        verify(mock, times(2)).simpleMethod(anyString(), captor.captureIf(new LessThan<Integer>(3)));
```
ChristianSchwarz: For better readabilty `captureIf(new LessThan<Integer>(3))` should be `captureIf(lt(3))` (see  ` AdditionalMatchers.lt( int)`)

src/main/java/org/mockito/ArgumentCaptor.java
```diff
@@ -100,12 +82,34 @@ private ArgumentCaptor(Class<? extends T> clazz) {
...
+     * 
+     * @return null or default values
+     */
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+	public T captureIf(final ArgumentMatcher matcher) {
+    	createCustomMatcher(customCapturingMatcher = new CustomCapturingMatcher<T>(matcher));
+    	return handyReturnValues.returnFor(clazz);
+    }
+
+    public T createCustomMatcher(final ArgumentMatcher<T> matcher) {
```
ChristianSchwarz: Can this method be `private` or in other words: why is it `public`?

src/test/java/org/mockito/internal/matchers/CustomCapturingMatcherTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+	@Test
+	public void verify_scenario() {
+		assertThat(expected, is(equalTo(actual)));
+	}
+
+	@Parameters
+	public static List<Object[]> scenarios() {
+		final List<Object[]> scenarios = new ArrayList<Object[]>();
+		scenarios.add(getParams(CustomCapturingMatcher.is(new InstanceOf(String.class)).matches("string_type"), true));
```
ChristianSchwarz: For better readability better use `isA(Class)` from MockitoMatchers instead of ` new InstanceOf(String.class)`

ChristianSchwarz:
#358

src/main/java/org/mockito/internal/matchers/CustomCapturingMatcher.java
```diff
@@ -0,0 +1,60 @@
...
+package org.mockito.internal.matchers;
+
+import java.io.Serializable;
+import java.util.LinkedList;
+import java.util.List;
+
+import org.mockito.ArgumentMatcher;
+import org.mockito.exceptions.Reporter;
+
+public class CustomCapturingMatcher<T> implements ArgumentMatcher<T>, ArgumentHolder<T>, CapturesArguments, VarargMatcher, Serializable {
```
ChristianSchwarz: I think i is not neccessary to have a `CustomCapturingMatcher`,  this implementation could be moved to `CapturingMatcher` and use an `any() `Matcher if used in a `ArgumentCaptor.capture()` call.

src/main/java/org/mockito/ArgumentCaptor.java
```diff
@@ -63,28 +64,9 @@
     HandyReturnValues handyReturnValues = new HandyReturnValues();
 
     private final CapturingMatcher<T> capturingMatcher = new CapturingMatcher<T>();
```
ChristianSchwarz: Is it possible to use a `new CustomCapturingMatcher(any())` here?  If yes we don't need a `CapturingMatcher ` at all.

src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
terebesirobert: Here we can use any Matcher that extends ArgumentMatcher. I used LessThan just to show how it is used

src/main/java/org/mockito/ArgumentCaptor.java
terebesirobert: It should be private. By mistake is public ( if you see the method implementation, you will see that somehow the used matcher is registered. Because we have an CapturingMatcher and a CustomCapturingMatcher, I considered that it's better to use method if something changes in the future.)

src/test/java/org/mockito/internal/matchers/CustomCapturingMatcherTest.java
terebesirobert: I tried to implement as generic as possible, so you can register a InstanceOf matcher, a LessThan, etc. So you can read, is this instance of, is this less than, is this equal to, etc. What do you think?

src/main/java/org/mockito/internal/matchers/CustomCapturingMatcher.java
terebesirobert: I think it should work, I have to see the implementation of any. Good idea. Thanks

src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
ChristianSchwarz: `LessThan` is not part of the public api and should not be referenced in production code or tests, thus it lifes in the org.mockito.**internal**.matchers  package. Matchers implementations should never be created by clients, they are provided by the public api's of [Matchers ](https://github.com/mockito/mockito/blob/master/src/main/java/org/mockito/Matchers.java)and [AdditionalMatchers](https://github.com/mockito/mockito/blob/master/src/main/java/org/mockito/AdditionalMatchers.java)

terebesirobert: Hi Christian,
Thanks for the quick response. I check the source code of AdditionalMatchers. With the current implementation I can't use it, because I would need the ArgumentMatcher used in the is, lt, etc methods. For ex. AdditionMatcher.is will return 0 all the time. Somehow I would need the evaluation result of the argument matcher used behind.

src/test/java/org/mockito/internal/matchers/CustomCapturingMatcherTest.java
ChristianSchwarz: `CustomCapturingMatcher.is(ArgumentMatcher)` is fine! I mean you should consider to call eg.:

scenarios.add(getParams(CustomCapturingMatcher.is(isA(String.class)).matches(1), false));

src/main/java/org/mockito/ArgumentCaptor.java
terebesirobert: Thanks for this. I don't need it anymore.

src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
terebesirobert: I tried to implement as much as possible. Still, I have issues's in implementing the `CapturingMatcher` using `AdditionalMatchers`.
What do you think about replacing `captureIf` with `captureIfIsA`, and so on?

An example would be:
```java
Processor mock = ...

mock.process(new ArrayList<>);
mock.process(new HashSet<>);

ArgumentCaptor captor = ...

verify(mock).process(captor.captureIfIs(ArrayList()));
```

and the implementation of this would be

```java
public T captureIfIsA(Class clazz) {
    new CapturingMatcher(new InstanceOf(clazz), arguments);
...
}
```

Thanks for your ideas.

src/main/java/org/mockito/ArgumentCaptor.java
```diff
@@ -59,34 +61,17 @@
  * @since 1.8.0
  */
 public class ArgumentCaptor<T> {
-    
+
     HandyReturnValues handyReturnValues = new HandyReturnValues();
 
-    private final CapturingMatcher<T> capturingMatcher = new CapturingMatcher<T>();
+    private List<Object> arguments;
```
ChristianSchwarz: should be `List<T>`

src/main/java/org/mockito/ArgumentCaptor.java
```diff
@@ -114,12 +120,13 @@ public T capture() {
      * @return captured argument value
      */
     public T getValue() {
-        return this.capturingMatcher.getLastValue();
+        return capturingMatcher.getLastValue();
```
ChristianSchwarz: Now that we have the `arguments`-List in this class the implementation should use it. The methods `getLastValue()` and `getValues()` can then be removed from the `CapturingMatcher` and the `ArgumentHolder` is not neccessary anymore.

src/main/java/org/mockito/ArgumentCaptor.java
```diff
@@ -145,12 +152,26 @@ public T getValue() {
...
+     *   List expected = asList(ArrayList<>(), LinkedList<>());
+     *   assertThat(expected.containsAll(peopleCaptor.getAllValues()), is(true));
+     * </pre>
      * See more examples in javadoc for {@link ArgumentCaptor} class.
      * 
      * @return captured argument value
      */
     public List<T> getAllValues() {
-        return this.capturingMatcher.getAllValues();
+        return capturingMatcher.getAllValues();
```
ChristianSchwarz: return arguments;

src/main/java/org/mockito/ArgumentCaptor.java
```diff
@@ -100,12 +85,33 @@ private ArgumentCaptor(Class<? extends T> clazz) {
...
+     * <p>
+     * Internally, this method registers a special implementation of an {@link ArgumentMatcher}.
+     * This argument matcher stores the argument value that satisfies the matcher so that you can use it later to perform assertions.  
+     * <p>
+     * See examples in javadoc for {@link ArgumentCaptor} class.
+     * 
+     * @return null or default values
+     */
+    public T captureIf(ArgumentMatcher matcher) {
+        capturingMatcher = CapturingMatcher.is(matcher, arguments);
```
ChristianSchwarz: I think the assignment is not needed cause the implementation can access the `arguments`-List to get the captured values.

src/main/java/org/mockito/ArgumentCaptor.java
```diff
@@ -100,12 +85,33 @@ private ArgumentCaptor(Class<? extends T> clazz) {
      * @return null or default values
      */
     public T capture() {
-        Mockito.argThat(capturingMatcher);
+        registerCapturingMatcher(capturingMatcher);
         return handyReturnValues.returnFor(clazz);
```
ChristianSchwarz: WDYT?
return captureIf(Any.ANY)

src/main/java/org/mockito/ArgumentCaptor.java
terebesirobert: Hi, 
I agree. Thanks for the suggestion.

src/main/java/org/mockito/ArgumentCaptor.java
terebesirobert: Sure, good idea. Thanks

src/main/java/org/mockito/ArgumentCaptor.java
terebesirobert: In this case I will have to implement the Report logic in `ArgumentCaptor`. Right now the `Report` logic is at `CapturingMatcher` level.

src/main/java/org/mockito/ArgumentCaptor.java
terebesirobert: Yes, thats true. Thanks.

src/main/java/org/mockito/ArgumentCaptor.java
terebesirobert: I used Object in case the user will add two matcher that will capture incompatible elements. What do you think?

src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
ChristianSchwarz: @terebesirobert You are right we can't call `captureIf(isA(XY.class))`. I totally forgot that `AdditionalMatchers.*` and `Matchers.*` return default values and no `ArgumentMatchers`. Adding specializations like  `captureIfIsA(..)` or `caputureIfGreaterThat(..)` would add too much complexity and would exclude custom ArgumentMatcher implementations.

What do you think of a class like _ArgumentMatcher_**s** that contains all matchers known from AdditionalMatchers and Matcher but return the ArgumentMatcher instead default values. Eg.:

```
class ArgumentMatchers{
   <T> ArgumentMatcher<T> isA(Class<? extends T> type){
        return new InstanceOf(type);
   }
  [..]
}
```

terebesirobert: I agree, great. Thanks for the idea. I will commit the new enhances ASAP.

src/main/java/org/mockito/ArgumentCaptor.java
ChristianSchwarz: Yes thats right for `getLastValue()`, `CapturingMatcher.getValues()` returns the plain arguments-List.

terebesirobert: you suggest that both, `getValues` and `getLastValue` should be implemented in `ArgumentCaptor` (moved Report logic to `ArgumentCaptor`?

src/main/java/org/mockito/ArgumentCaptor.java
ChristianSchwarz: That is (almost) impossible cause the type parameter `<T>` of `ArgumentMatcher `allows only to capture arguments for type `T`. The compile will prevent you from capturing a `Integer `with an `ArgumentCaptor<String>` unless you do some untyped casting.

terebesirobert: let's take a look to Use Case 2: 
if you use two or more matchers that match elements of incompatible type. 

```java

Consumer<List> mock = ...
ArgumentCapture<Array> captor = ...

verfy(mock).consume(captor.captureIf(instanceOf(Array)));
verfy(mock).consume(captor.captureIf(instanceOf(List)));

```

the second matcher will try to add elements instance of `Set` to arguments of type `List`.
This is the reason why I used Object instead of T. What do you think?

src/test/java/org/mockito/internal/matchers/CustomCapturingMatcherTest.java
ChristianSchwarz: My fault again, i forget that isA() doesn't return an ArgumentMatcher

src/main/java/org/mockito/ArgumentCaptor.java
ChristianSchwarz: Exactly, this way `getValues() `, `getLastValue()` and the defining interface `ArgumentHolder` can be removed. This reduces the complexity of the solution and improves the design cause the responsibility of reporting should be part of the Captor not its ArgumentMatcher. 

Thinking loud: Isn't CapturingMatcher an implementation detail of the ArgumentCaptor? If yes it should be an inner class of ArgumentCaptor IMHO.

terebesirobert: I agree. I tried to keep the old architecture unchanged. Thanks

src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
terebesirobert: @ChristianSchwarz Hi Christian. Please take a look to the new implementation. Thanks for the ideas.

src/main/java/org/mockito/internal/matchers/Any.java
```diff
@@ -10,7 +10,7 @@
 import org.mockito.ArgumentMatcher;
 
 @SuppressWarnings("unchecked")
-public class Any implements ArgumentMatcher, Serializable {
```
ChristianSchwarz: Hmm, i am having a bad feeling about changing code that is not related to the solution of the issue. This change should be part of an other pull request. WDYT?

src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
```diff
@@ -1,43 +1,42 @@
...
-    public List<T> getAllValues() {
-        return (List) arguments;
+    @Override
+    public boolean matches(Object argument) {
+        return matcher.matches(argument);
     }
 
-    public void captureFrom(Object argument) {
-        this.arguments.add(argument);
+    public static <T> CapturingMatcher<T> is(ArgumentMatcher<T> argumentMatcher, List<T> arguments) {
```
ChristianSchwarz: Is that static method needed, what is the benefit of it compared to the constructor?

src/main/java/org/mockito/internal/matchers/Equals.java
```diff
@@ -5,16 +5,16 @@
```
ChristianSchwarz: Don't get me wrong, these changes are good, but should be targeted by an other pull request.

src/main/java/org/mockito/internal/matchers/EqualsWithDelta.java
```diff
@@ -5,22 +5,22 @@
```
ChristianSchwarz: Same here...

src/main/java/org/mockito/internal/matchers/InstanceOf.java
```diff
@@ -5,12 +5,12 @@
```
ChristianSchwarz: Same here....

src/test/java/org/mockito/ArgumentMatchersTest.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Test
+    public void verify_scenarios() {
+        assertThat(expected, is(actual));
+    }
+
+    @Parameters
+    public static List<Object[]> scenarios() {
+        return Arrays.asList(
```
ChristianSchwarz: `<JustAMatterOfTaste>` 
  It should be enough to test that the methods don't return null or the expected argument matcher instance. The behaviour of the matchers are tested in there corresponding test classes, e.g: [EqualsTest ](https://github.com/mockito/mockito/blob/master/src/test/java/org/mockito/internal/matchers/EqualsTest.java) <-> [Equals](https://github.com/mockito/mockito/blob/master/src/main/java/org/mockito/internal/matchers/Equals.java).
`</JustAMatterOfTaste>`

src/test/java/org/mockito/internal/invocation/InvocationMatcherTest.java
```diff
@@ -112,21 +112,6 @@ public void should_be_similar_if_is_overloaded_but_used_with_different_arg() thr
     }
 
     @Test
-    public void should_capture_arguments_from_invocation() throws Exception {
```
ChristianSchwarz: That test can be rewritten, no need to delete it.

src/test/java/org/mockito/internal/invocation/InvocationMatcherTest.java
```diff
@@ -140,21 +125,6 @@ public void should_match_varargs_using_any_varargs() throws Exception {
         assertTrue(match);
     }
 
-    @Test
```
ChristianSchwarz: That one can be rewritten too...

src/test/java/org/mockito/internal/matchers/CapturingMatcherTest.java
```diff
@@ -4,50 +4,30 @@
  */
 package org.mockito.internal.matchers;
```
ChristianSchwarz: All the tests can be rewritten like above

terebesirobert: @ChristianSchwarz Commit the last changes too.

src/main/java/org/mockito/ArgumentMatchers.java
```diff
@@ -0,0 +1,189 @@
...
+import org.mockito.internal.matchers.Find;
+import org.mockito.internal.matchers.GreaterOrEqual;
+import org.mockito.internal.matchers.GreaterThan;
+import org.mockito.internal.matchers.InstanceOf;
+import org.mockito.internal.matchers.LessOrEqual;
+import org.mockito.internal.matchers.LessThan;
+import org.mockito.internal.matchers.NotNull;
+import org.mockito.internal.matchers.StartsWith;
+
+public class ArgumentMatchers {
```
ChristianSchwarz: Good idea to change the return type to ArgumentMatcher !

src/main/java/org/mockito/ArgumentMatchers.java
```diff
@@ -0,0 +1,189 @@
...
+
+    /**
+     * String argument that contains a substring that matches the given regular
+     * expression.
+     * 
+     * @param regex
+     *            the regular expression.
+     * @return {@link ArgumentMatcher}.
+     */
+    public static ArgumentMatcher find(String regex) {
```
ChristianSchwarz: ArgumentMatcher<String> ?

src/test/java/org/mockito/internal/matchers/CapturingMatcherTest.java
ChristianSchwarz: @terebesirobert good work !!!

src/main/java/org/mockito/ArgumentCaptor.java
```diff
@@ -100,12 +84,31 @@ private ArgumentCaptor(Class<? extends T> clazz) {
...
+     * See examples in javadoc for {@link ArgumentCaptor} class.
+     * 
+     * @return null or default values
+     */
+    public T captureIf(ArgumentMatcher<T> matcher) {
+        registerCapturingMatcher(new CapturingMatcher(matcher, arguments));
         return handyReturnValues.returnFor(clazz);
     }
 
+    public void registerCapturingMatcher(ArgumentMatcher<T> matcher) {
```
ChristianSchwarz: That one should be inlined.

src/main/java/org/mockito/ArgumentCaptor.java
```diff
@@ -100,12 +84,31 @@ private ArgumentCaptor(Class<? extends T> clazz) {
      * @return null or default values
      */
     public T capture() {
-        Mockito.argThat(capturingMatcher);
+        return captureIf(Any.ANY);
+    }
+
+    /**
+     * Use it to capture the argument that satisfies the matcher. This method <b>must be used inside of verification</b>.
+     * <p>
```
ChristianSchwarz: I think users of that nice new method are not that much interested in how it is implemented under the hood. As a user I want to know how to use it  (sample code) and a link to the new ArgumentMatchers-class would be useful.

terebesirobert: @ChristianSchwarz Added some details to the captureIf java doc.

src/test/java/org/mockito/internal/matchers/CapturingMatcherTest.java
terebesirobert: @ChristianSchwarz Thanks, I had a good reviewer.

src/main/java/org/mockito/internal/matchers/Equals.java
terebesirobert: @ChristianSchwarz 
If you post an issue related to implement a `Equals` and other non-generic mathcers, please let me know, I would like to contribute. It's really bothering me that I have a lot of warnings.
Thanks

src/main/java/org/mockito/ArgumentCaptor.java
```diff
@@ -100,12 +84,41 @@ private ArgumentCaptor(Class<? extends T> clazz) {
      * @return null or default values
      */
     public T capture() {
-        Mockito.argThat(capturingMatcher);
+        return captureIf(Any.ANY);
```
ChristianSchwarz: In the description of  #358 I wrote about the unexpected TooManyActualInvocations-Exception that lead me to open the ticket. We can get rid of it in case the captor is created via `ArgumentCaptor.forClass(Class)` cause we have the class-Object. Something like this shoud do the trick: 
```
if (clazz==null)
   return captureIf( Any.ANY);
return captureIf(isNullOrA(clazz));
```

This would avoid the capturing of arguments that are not an instances of the given type. Obviously we can't  use `isA(Class)` here because _null_ arguments must be captured too. 

WDYT?

terebesirobert: You mean we should rewrite the `forClass` implementation?

ChristianSchwarz: no the implementation of capture()
Am 19.03.2016 22:07 schrieb "terebesirobert" <notifications@github.com>:

terebesirobert: With this we will add to much restriction. Maybe there are users that rely on the fact on the current implementation there is no type check when capturing arguments, so I would leave it like this for backward compatibility.

terebesirobert: @ChristianSchwarz  WDYT?

ChristianSchwarz: Can you write a test (e.g. in [CapturingArgumentsTest](https://github.com/mockito/mockito/blob/master/src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java)) for the use case you have in mind? The only case I see that will behave differently that now is this:

        ArgumentCaptor<String> argumentCaptor = ArgumentCaptor.forClass(String.class);
        mock.forObject(42);
        mock.forObject("capturedValue");
        verify(mock).forObject(argumentCaptor.capture()); //throws a TooManyActualInvocations-Exeception
     
With the proposed change users won't get a TooManyActualInvocations-Exeception anymore.

terebesirobert: Hi, 
For this use case we can use `captureIf`. 
Right now for `capture` we register an `ANY` matcher. We could register a `Null` matcher too. I don't see the need of `IsNullOrA`. What do you think?

ChristianSchwarz: The question is what dit you expect as a user in the above described scenario? What type's  should `ArgumentCaptor.forClass(String.class)` capture:
 - String ?
 - null ?
 - Integer ?

At least I would not expect that an Integer is captured, if not why? If you capture Integers how would you get the values? Lets look a that example:

```
ArgumentCaptor<String> argumentCaptor = ArgumentCaptor.forClass(String.class);
mock.forObject("capturedValue");
mock.forObject(42);
verify(mock,times(2)).forObject(argumentCaptor.capture());//No TooManyActualInvocations-Exeception this time

String lastArg= argumentCaptor.getValue();//ClassCastException cant cast a Integer to String
```

terebesirobert: I expect that only String to be captured. I totally agree with you (I will implement this, but I think we don't need a `isNullOrA` matcher). Do you think it's necessary?

terebesirobert: @ChristianSchwarz I implemented the last suggestion you had. Please take a look and let me know if you don't understand. Thanks.

ChristianSchwarz: Yes! Currently I see no other way, cause clients want to capture instances of the specified type only or _null_ as the current implementation of `capture()` does.

terebesirobert: In the current implementation I use the `and` and `or` matcher to achieve this. I don't see the need of `isNullOrA`. I think it's a better solution

src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
```diff
@@ -314,4 +329,68 @@ public void captures_correctly_when_captor_used_on_pure_vararg_method() throws E
         verify(mock).varargs(eq(42), argumentCaptor.capture());
         Assertions.assertThat(argumentCaptor.getValue()).contains("capturedValue");
     }
+
```
ChristianSchwarz: +1 for adding more test

src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
```diff
@@ -298,7 +313,7 @@ public void captures_correctly_when_captor_used_multiple_times() throws Exceptio
 
         // then
         // this is only for backwards compatibility. It does not make sense in real to do so.
-        verify(mock).mixedVarargs(any(), argumentCaptor.capture(), argumentCaptor.capture(), argumentCaptor.capture());
+        verify(mock).mixedVarargs(any(), argumentCaptor.capture());
```
ChristianSchwarz: Why did you change this verification? I am afraid did it fail?

terebesirobert: @ChristianSchwarz  this will register 3 captors. Since the matchers share the same argument holder, each matcher will add the value to the holder. With the current implementation, we don't need this anymore. What do you think?

ChristianSchwarz: I think we are better off if we keep existing tests, to prove that the implementation works as before. I guess the one who wrote this test had good reasons to use 3 captors.

terebesirobert: Hi,
I think this is caused because we register 3 intances of `ANY` matcher. Maybe we should store the registered matchers in a `Set`, so we don't register 2 or more matchers of the same type. What do you think?

ChristianSchwarz: Hi,
we should leave this test as is it was before your change, to prove that the implementation works as before.

terebesirobert: Sure, but I think we need some enhancements to make it pass. I almost 100% sure that it will fail (reason see before).

ChristianSchwarz: Sadly I can't checkout your branch nor run any tests. Can you provide a Stacktrace of the failure?

terebesirobert: Sure, here is the stacktrace

`java.lang.AssertionError: 
  <9>
while expected size was:
  <3>
Actual was:
  <["a", "b", "c", "a", "b", "c", "a", "b", "c"]>
Expected was:
  <["a", "b", "c"]>

	at org.mockitousage.matchers.CapturingArgumentsTest.captures_correctly_when_captor_used_multiple_times(CapturingArgumentsTest.java:317)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.lang.reflect.Method.invoke(Unknown Source)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:45)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:15)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:42)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:20)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:28)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:30)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:263)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:68)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:47)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:231)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:60)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:229)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:50)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:222)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:300)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:86)
	at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:38)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:459)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:675)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:382)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:192)
`

terebesirobert: I know what problem is, I will come back with a fix ASAP.
Thanks for the heads up.

terebesirobert: @ChristianSchwarz Hi, I have a proof of concept. I added a `hashCode` and `equals` method to each matcher. When we use `capture` or `captureIf` we create a new instance of `CapturingMatcher` only of that matcher was never used. I will commit when it's ready. What do you think?

ChristianSchwarz: @terebesirobert Actually I don't understand what causes the failure. I am looking forward to your solution! 
About adding equals/hashCode to each matcher is a bit dangerous. Imagine a programmer provides customer matcher but simply don't know  that eq/hashCode must be implemented. Than other team member use this custom matcher with success in the most common cases but in the above described scenario they get an failure and don't know why. 
Java 8 Lambda expression are also not possible (`captureIf(String s-> s.length()==5)`) cause they can't implement eq/hashCode.

terebesirobert: @ChristianSchwarz 
Hi,
When you call `capture`, or `captureIf`, the implementation call a `Mockito.argThat`, which stores the matcher in a `ArgumentMatcherStorageImpl`. The code that generates the problem is the following:

```java
    private Set<ArgumentMatcher> uniqueMatcherSet(int indexOfVararg) {
        HashSet<ArgumentMatcher> set = new HashSet<ArgumentMatcher>();
        for (int position = indexOfVararg; position < matchers.size(); position++) {
            ArgumentMatcher matcher = matchers.get(position);
            set.add(matcher);
        }
        return set;
    }
```
Later, this set is used to decide which values are stored in the `arguments` param from `ArgumentCaptor`. On the previous implementation we never had this problem because each time the `capture` method was called, the same instance of `CapturingMatcher` was used (same hashCode, etc.). With the current implementation, each time we call `capture`, we create a new instance of `CapturingMatcher`.
Because of this I decided to implement `equals` and `hashCode` methods, so I don't add to matchers that should capture the same argument (for example, I don't want to add two `isA(String.class)` matchers).

What is the expected result for the following use case:

```java
   ArgumentCaptor<String> argumentCaptor = ArgumentCaptor.forClass(String.class);

        // when
        mock.mixedVarargs(42, "a", "b", "c");

        // then
        // this is only for backwards compatibility. It does not make sense in real to do so.
        verify(mock).mixedVarargs(any(), argumentCaptor.captureIf(isA(String.class)), argumentCaptor.captureIf(isA(String.class)), argumentCaptor.captureIf(isA(String.class)));
```

If we don't implement `equals` and `hashCode`, 9 elements will be captured. Each `isA` matcher will add an element to the `arguments` from `ArgumentCaptor`.
We could decide to keep the old functionality of `capture` and admit the functionality of `captureIf` is the one described above if you use `captureIf` more than once with the same matcher. What do you suggest?

ChristianSchwarz: @terebesirobert Thanks for pointing me to the problematic code! 
As far as I can see this loop in `captureVarargsPart(Invocation invocation)` is responsible for adding every vararg to every AgrumentMatcher, which leads to our failing test:
```
for (int i = 0; i < Array.getLength(rawArgument); i++) {
  ((CapturesArguments) m).captureFrom(Array.get(rawArgument, i));
}
```
Do you see a chance to fix the problem right here? Maybe we don't need to add eq/hashCode to every matcher.

terebesirobert: @ChristianSchwarz we could enhance the CapturingMatcher to add an element to the `arguments` only if that element is missing. If it's already added, we shouldn't add it. What do you think?

terebesirobert: @ChristianSchwarz  Hi, a commited a fix and some unit tests. Please take a look. What do you think?

ChristianSchwarz: @terebesirobert Hi, I actually fail to see what was changed. Do you mind not squashing the commits next time? 

>  If it's already added, we shouldn't add it. What do you think?

That may break somebodys test. Some may rely on the fact that some instances were passed multiple times.

If you like you can tryout this attempt to fix the issue,i sadly can't test it:
```
    private void captureVarargsPart(Invocation invocation) {
        if (!invocation.getMethod().isVarArgs()) {
            return;
        }

        Object[] rawArguments = invocation.getRawArguments();
        int indexOfVararg = rawArguments.length - 1;

        Object varargArray = rawArguments[indexOfVararg];
        int size = Array.getLength(varargArray);

        for (int i = 0; i < size; i++) {
            ArgumentMatcher m = matchers.get(i+indexOfVararg);
            if (m instanceof CapturesArguments) {
                ((CapturesArguments) m).captureFrom(Array.get(varargArray, i));
            }
        }
    }
```

Hope it is self describing, I am currently short at time.

terebesirobert: Hi,
Thanks, this is a good idea. Sorry for squashing the commits. The fix work, I had to add a protection for the case when you have only one matcher for a vararg. 
What do you think, it there a chance that this fix will be merged?

Thanks.

terebesirobert: Hi,
Did you have time to review the last changes? 
Thanks.

ChristianSchwarz: I think the commits need to be merged by you once we are done with the review/improvement cycle. The mockito guys are short in time.
About the new changes: Maybe we can simplify the new code by using getAruments instead of getRawArguments.  We need some more tests for: more ArgumentMatcher than Arguments and less ArgumentMatcher that Arguments. I cant provide Code snippets now, writing on a Smartphone which is a pain.

terebesirobert: When you can, please provide me some code snippets. Thanks for your time.

terebesirobert: @ChristianSchwarz 
Hi,
Please review the last changes. Thanks.

src/main/java/org/mockito/internal/invocation/InvocationMatcher.java
```diff
@@ -134,24 +133,12 @@ private void captureVarargsPart(Invocation invocation) {
...
-        for (int i = 0; i < varArgSize; i++) {
-            int indexOfVarargMatcher = (i + indexOfVarArg > matchers.size() -1 ? indexOfVarArg : i + indexOfVarArg);
-            ArgumentMatcher m = matchers.get(indexOfVarargMatcher);
+        for (int i = 0; i < arguments.length; i++) {
+            ArgumentMatcher m = matchers.get(i > matchers.size() - 1 ? matchers.size() - 1 : i);
             if (m instanceof CapturesArguments) {
-                ((CapturesArguments) m).captureFrom(Array.get(varArg, i));
+                ((CapturesArguments) m).captureFrom(arguments[i]);
             }
         }
```
ChristianSchwarz: Can you add these varargs tests -> [gist](https://gist.github.com/ChristianSchwarz/57b120a4a9d1071766b0b84d763d3604) 
Why does a ArgumentCaptor capture single values of a varargs and not the whole array, I mean from a user perspective? Assume we have a method `void varArgs(String...text)` how can we destingush between `varArgs((String)null)` and `varArgs((String[])null)`??

terebesirobert: From my point of view, I expect that the scenarios described above to capture the same arguments. What would you expect? As we discussed before, `null` arguments will be captured all the time. This is the reason why we added the isNull argument matcher. What do you think?

ChristianSchwarz: `varArgs((String)null)` and `varArgs((String[])null)` are different cases, the question is how should `varArgs((String[])null)` be treaten, maybe its a good idea to handle to is the same way as an empty array cause we have nothing to capture?

terebesirobert: I don't really see why are different? `isNull` matcher will capture both, (String and String[]).

ChristianSchwarz: Hmm okay, does `invocation.getArguments()` return _null_, an empty array or an array with one null-element when `varArgs((String[])null) `is called?

terebesirobert: I see, I have to check. I'll come back with an answer ASAP.

ChristianSchwarz: I think we can now simplify the implementation a bit more. The method `captureRegularArguments(Invocation invocation)` can be removed if we remove the isVarArgs()-check of `captureVarargsPart(Invocation invocation)` i guess.

terebesirobert: @ChristianSchwarz 
HI, I checked, and invocation.getArguments() return an array with one null element when we call varArgs((String[])null). I still don't understand why would you treat `varArgs((String)null)` differently from `varArgs((String[])null)`.

ChristianSchwarz: A `null`-Argument should be captured but a `null`-array not, imho it is like passing an empty array. E.g:

`mock.varArgs(null,null,null)`  <- should capture 3x null, it is equivalent to 
`mock.varArgs(new String[]{null,null,null})` 
`mock.varArgs((String[]) null)` <- should capture nothing cause there is nothing to capture

What do you think?

---

Are you sure that `mock.varArgs((String[]) null)` will not throw a _NullPointerException_ with the current implementation? Do we have a JUnit-Test?

terebesirobert: @ChristianSchwarz 
Hi Christian,
I tried to fix this. I think the problem is in `ArgumentProcessor` -> `expandVarArgs`. If I try to fix here, I brake the backward compatibility. Do you have any ideas how to fix this?


```java
 public static Object[] expandVarArgs(final boolean isVarArgs, final Object[] args) {
        if (!isVarArgs || new ArrayUtils().isEmpty(args) || args[args.length - 1] != null && !args[args.length - 1].getClass().isArray()) {
            return args == null ? new Object[0] : args;
        }

        int nonVarArgsCount = args.length - 1;
        Object[] varArgs;
        if (args[nonVarArgsCount] == null) {
            // in case someone deliberately passed null varArg array
            varArgs = new Object[] { null };
            nonVarArgsCount = 0;
        } else {
            varArgs = ArrayEquals.createObjectArray(args[nonVarArgsCount]);
        }
        final int varArgsCount = varArgs.length;
        Object[] newArgs = new Object[nonVarArgsCount + varArgsCount];
        System.arraycopy(args, 0, newArgs, 0, nonVarArgsCount);
        System.arraycopy(varArgs, 0, newArgs, nonVarArgsCount, varArgsCount);
        return newArgs;
    }
```

ChristianSchwarz: @terebesirobert  Hi Robert, 
I'm not sure what would break the backward compatibility. Do you mean this case: `mock.varArgs((String[]) null)`  _<- should capture nothing_ ?

```
if (args[nonVarArgsCount] == null) {
            // in case someone deliberately passed null varArg array
            varArgs = new Object[0]; // an empty array should do the trick...
```

terebesirobert: @ChristianSchwarz 
Hi Christian,
I already tried that and a lot of unit test are failing, for ex. 

```java
    @Test
    public void shouldStubUsingAnyVarargs() {
        when(mock.run((String[]) anyVararg())).thenReturn("foo");
        
        assertEquals("foo", mock.run("a", "b"));
    }
```

ChristianSchwarz: @terebesirobert Arrrgh those **w**ar-args.  I currently can't imagine what goes wrong. Can you post the error/failure message? 
Maybe there is an bug in `expandVarArgs(..)` that method is damn cryptic, bugs can hide easily.

terebesirobert: @ChristianSchwarz 
Hi,
Sure, this is the error:

`org.mockito.exceptions.misusing.InvalidUseOfMatchersException: 
Invalid use of argument matchers!
0 matchers expected, 1 recorded:
-> at org.mockitousage.bugs.varargs.VarargsNotPlayingWithAnyObjectTest.shouldStubUsingAnyVarargs(VarargsNotPlayingWithAnyObjectTest.java:51)

This exception may occur if matchers are combined with raw values:
    //incorrect:
    someMethod(anyObject(), "raw String");
When using matchers, all arguments have to be provided by matchers.
For example:
    //correct:
    someMethod(anyObject(), eq("String by matcher"));

For more info see javadoc for Matchers class.

	at org.mockito.exceptions.Reporter.invalidUseOfMatchers(Reporter.java:239)
	at org.mockito.internal.invocation.MatchersBinder.validateMatchers(MatchersBinder.java:37)
	at org.mockito.internal.invocation.MatchersBinder.bindMatchers(MatchersBinder.java:23)
	at org.mockito.internal.handler.MockHandlerImpl.handle(MockHandlerImpl.java:63)
	at org.mockito.internal.handler.InvocationNotifierHandler.handle(InvocationNotifierHandler.java:37)
	at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.doIntercept(MockMethodInterceptor.java:36)
	at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.access$0(MockMethodInterceptor.java:32)
	at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor$DispatcherDefaultingToRealMethod.interceptAbstract(MockMethodInterceptor.java:113)
	at org.mockitousage.bugs.varargs.VarargsNotPlayingWithAnyObjectTest$VarargMethod$MockitoMock$502966141.run(Unknown Source)
	at org.mockitousage.bugs.varargs.VarargsNotPlayingWithAnyObjectTest.shouldStubUsingAnyVarargs(VarargsNotPlayingWithAnyObjectTest.java:51)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.lang.reflect.Method.invoke(Unknown Source)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:45)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:15)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:42)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:20)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:28)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:30)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:263)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:68)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:47)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:231)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:60)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:229)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:50)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:222)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:300)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:86)
	at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:38)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:459)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:675)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:382)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:192)
`
WDYT?

ChristianSchwarz: @terebesirobert I have no idea what's the causes, i have to debug it. I hope to find some time in the next 3 weeks. If you have an idea, let me know. 

@bric3  & @szczepiq  It would be good to know if this PR has a chance of integration. Maybe we waste a lot of time for nothing?

terebesirobert: @ChristianSchwarz 
Hi Christian,

I debug-ed it, and it's using this null value to have a different type of invocation. If we enhance it to use `String[0]` instead of `String[] { null }`

Here is the code that will match this invocations:

```java
    public void check(List<Invocation> invocations, InvocationMatcher wanted) {
        List<Invocation> actualInvocations = finder.findInvocations(invocations, wanted);
        
        if (actualInvocations.isEmpty()) {
            Invocation similar = finder.findSimilarInvocation(invocations, wanted);
            if (similar != null) {
                ArgumentMatchingTool argumentMatchingTool = new ArgumentMatchingTool();
                Integer[] indexesOfSuspiciousArgs = argumentMatchingTool.getSuspiciouslyNotMatchingArgsIndexes(wanted.getMatchers(), similar.getArguments());
                SmartPrinter smartPrinter = new SmartPrinter(wanted, similar, indexesOfSuspiciousArgs);
                reporter.argumentsAreDifferent(smartPrinter.getWanted(), smartPrinter.getActual(), similar.getLocation());
            } else {
                reporter.wantedButNotInvoked(wanted, invocations);
            }
        }
    }
``` in `MissingInvocationChecker`.

terebesirobert: @ChristianSchwarz 
any updated on this?

terebesirobert: @ChristianSchwarz 
Hey,
any chance to integrate this?

raphw: LGTM, but since this is an additional feature I would not want to merge it without approval by @bric3 or @szczepiq.

terebesirobert: @raphw 
Hey,

Any updates on this?

raphw: Hi, sorry for the slow follow-up. I think everybody in our team is quite busy in the moment. Don't worry, it might take some time but eventually, we always look through it. Thank you for contributing!

ChristianSchwarz:
@terebesirobert 
Can you rebase this PR?  I hope the mocktio core devs have a bit time to review and eventually merge it.  Once it is merged I would like to check if this PR will fix #439 partially.

bric3:
@ChristianSchwarz @terebesirobert Not sure it will happen before mockito 2

terebesirobert:
@ChristianSchwarz 
Done. Thanks

ChristianSchwarz:
@terebesirobert Seems like you squashed the commits? Github says: _"This branch is 1 commit ahead, 185 commits behind mockito:master."_ and _"This branch has conflicts that must be resolved"_.

terebesirobert:
@ChristianSchwarz 
Please check now. Yes, I only squashed the changes. Sorry.

Robert

ChristianSchwarz:
@terebesirobert 
No need to say sorry! God job! :+1:

ChristianSchwarz:
I guess the build failed cause the source level changed to java 1.6.

This should work:
`ArgumentCaptor<List> captor = ArgumentCaptor.forClass((List)null);`
or
`ArgumentCaptor<List> captor = ArgumentCaptor.<List>forClass(null);`

terebesirobert:
@ChristianSchwarz 
It's not working. I think the signature of forClass it's not correct.

ChristianSchwarz:
@terebesirobert 
Strange, `ArgumentCaptor<List<Person>> c= ArgumentCaptor.forClass(null);` compiles with1.6 for me under Eclipse.

>  I think the signature of forClass it's not correct.
Yes it looks a over complicated. I tried`<T> T forClass(Class<? extends T>)` but it makes it even worse.

terebesirobert:
@ChristianSchwarz 
I commented the problematic test. I will fix it next week. Thanks for your help

raphw:
What is the status of this ticket? Are we delaying this to 2.1?

TimvdLippe:
Yes this is delayed to 2.1

TimvdLippe:
@terebesirobert There are sadly quite some merge conflicts with this PR, but I think we do want to land this. Could you update this PR so we can take another look? Thanks for your effort!

terebesirobert:
@TimvdLippe Sure, I will rebase this ASAP.
Thanks

TimvdLippe:
Friendly ping @terebesirobert 😄

TimvdLippe:
I am closing this for now as development has halted. If you have time again to update this PR we can reopen. Thank you in advance for your time and contribution!

terebesirobert:
@TimvdLippe Hi, I just tried to rebase it. I think you can reopen it, I will update it.

TimvdLippe:
@terebesirobert as you opened PR based on your master branch, I am unable to. Please branch of your master and open a PR with that branch. Thank you again for putting more time in this PR :)

terebesirobert:
@TimvdLippe I created a new branch and pushed from there. Can you give me some hints what I'm doing wrong?

TimvdLippe:
Your original PR was from your master, see the information at the top of this PR: `terebesirobert  wants to merge 1 commit into mockito:master from terebesirobert:master`. Apparently GitHub disallows me from reopening this PR stating that you force-pushed the master branch. To resolve that, create a new branch, push to origin (as you did) and open a new PR with that branch. Then it should be good to go.

paulduffin:
Did work move to another PR? I am really interested in this change.

TimvdLippe:
It moved to #957

Issue #379, opened 2016-03-24
Verification using After and ArgumentCaptor with Times or AtLeast methods returns unexpected size of captured values list.


thesnowgoose:
Hi!
I was looking issue #345 and found that the test "shouldReturnListOfArgumentsWithSameSizeAsGivenInAtMostVerification" (located in VerificationAfterDelayTest.java) fails if you use atLeast() or times() instead of atMost().
I created these tests:

![screenshot from 2016-03-24 16 50 29](https://cloud.githubusercontent.com/assets/17257307/14034443/8a3d8b98-f1e0-11e5-856a-65dc20a4fa6b.png)

ChristianSchwarz:
@thesnowgoose 
That a bug in VerificationOverTimeImpl#verify(VerificationData). The passed `VerificationData ` contains all invocations, in your case "0","1" and "2". During the specified 200ms these invocations are verified in a loop without delay. That means if your CPU is able to verify the VerificationData 10,000 times during that 200ms the captor captures 30,000 arguments (10000 * 3 arguments ). 

The implementation of `After` uses `VerificationOverTimeImpl ` and that class performs the verify-loop until the time is elapsed.  This behaviour is fine for `Timeout` but violates the specification of  `Mockito.after(long)`, this is what the JavaDoc says:
> Allows verifying over a given period. It causes a verify to *wait for a specified period of time* for a desired interaction

In other words the `After` implementation must use some kind of `VerificationAfterDelayImpl `, that simply waits and than verify the invocations.

I have a solution and will provide a PR.

ChristianSchwarz:
PR #380 relates to this issue.

bric3:
Fixed by #380

jmborer:
Why is this closed? I ran into the issue and lost about an hour before realizing it is a Mockito bug. What is the current status? It is not fixed in the 1.x branch. I am using 1.10.19 which seems the most recent version.

ChristianSchwarz:
@jmborer can you test it with the latest 2.0.0-beta build, it should be fixed

jmborer:
OK. Thanx for the info, but I don't feel confortable to use a beta version to test operational sofware...

TimvdLippe:
The beta is almost over, I hope to publish a release candidate in the
coming days. You should be able to upgrade without much problems, since the
breaking changes are kept at a minimum.

bric3:
[...]

PR #380, opened 2016-03-25
Fixes the issue #379 by removing previously verified invocations, some tests were added to verify that works

thesnowgoose:
check list

### DIFF
--- a/src/main/java/org/mockito/internal/verification/checkers/AtLeastXNumberOfInvocationsChecker.java
+++ b/src/main/java/org/mockito/internal/verification/checkers/AtLeastXNumberOfInvocationsChecker.java
@@ -5,6 +5,7 @@
 
 package org.mockito.internal.verification.checkers;
 
+import java.util.Iterator;
 import java.util.List;
 
 import org.mockito.exceptions.Reporter;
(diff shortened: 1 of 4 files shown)


src/main/java/org/mockito/internal/verification/checkers/AtLeastXNumberOfInvocationsChecker.java
```diff
@@ -28,7 +29,17 @@ public void check(List<Invocation> invocations, InvocationMatcher wanted, int wa
             Location lastLocation = finder.getLastLocation(actualInvocations);
             reporter.tooLittleActualInvocations(new AtLeastDiscrepancy(wantedCount, actualCount), wanted, lastLocation);        
         }
-        
+
+        removeAlreadyVerified(actualInvocations);
         invocationMarker.markVerified(actualInvocations, wanted);
     }
+
+    private void removeAlreadyVerified(List<Invocation> invocations) {
```
TimvdLippe: Since this method is duplicate from the method added in the class below, maybe we can extract an superclass `InvocationsChecker`. This class abstracts away all the logic regarding obtaining and reporting the failures, the constructor as well as this new method.

TimvdLippe:
@thesnowgoose apart from 1 small comment and a rebase, this PR LGTM. Could you update your PR?

bric3:
Closing #379 as well then

Issue #383, opened 2016-04-06
Mockito JAR is not a valid OSGi bundle because illegal version format


SourcePond:
Hi,

The current Mockito JAR is not a valid OSGi-bundle because it uses an illegal version for exporting its packages. I get the following error when I run my PAX-EXAM tests with the latest Mockito version:

org.ops4j.pax.exam.TestContainerException: Problem starting test container.
Caused by: org.osgi.framework.BundleException: Could not create bundle object.
Caused by: java.lang.IllegalArgumentException: invalid version "2.0.44-beta": non-numeric "44-beta"
Caused by: java.lang.NumberFormatException: For input string: "44-beta"

The version should be **2.0.44.beta** (with a dot instead a dash). This is since 2.0.28-beta, the last valid bundle was 2.0.27-beta.

szczepiq:
Can you help us out? If you build locally a regular version, e.g. 2.1.0 or something (version can be changed in 'version.properties' file), you can test it out locally and let us know.

I think I will not change the -beta notation now but I take the feedback will update the automation after 2.x final release.

THANKS for reporting!

seanrohead:
We are having the same problem. This was working in 2.0.27-beta and was broken in 2.0.28-beta. Why are you not planning to fix it until the final release?

szczepiq:
I'm sorry the changes in beta are causing troubles!

In order to be able to maintain and evolve the code base, we had to start using latest OSGi Gradle plugin (we couldn't sustain using ant-based manifest generation that was 5 years old and coupled us with ant). We do not use OSGi and rely on the external contributors to fix any problems with the bundle.

I'd love to fix this problem ASAP - we need help from you to make it happen.

Issue #385, opened 2016-04-08
OSGi metadata is incorrect


seanrohead:
The manifest is incorrectly exporting the org.objenesis package. This package is not included in the mockito jar, so this package needs to be imported rather than exported.

szczepiq:
Is it a duplicate of #383? 

We need help fixing OSGi support. See my comment in #383 

Thanks for reporting!

PascalSchumacher:
@seanrohead Which version are you referring to? I just checked the manifest of the latest version (`2.0.63-beta`) and the only mention of `org.objenesis` in the import package section. I guess https://github.com/mockito/mockito/pull/388 fixed this issue. I would be nice if you could give `2.0.63-beta` a try and let us know if it works for you. Thanks!

seanrohead:
It looks like the metadata is correct, however when running tests in an osgi container, I am unable to mock classes (see error below). I built a custom version of the mockito jar and embedded the byte-buddy and objenesis classes inside the mockito jar (instead of loading them as separate bundles) and the error went away. I'm happy to send a copy of the jar file that I was able to get to work.

This is the error I get if I use the mockito jar file in maven (org.mockito:mockito-core:2.0.70-beta):

`com.proofpoint.governance.agent.tagging.file.sync.test.TaggedFileSyncServiceIntegrationTest STANDARD_ERROR
    org.mockito.exceptions.base.MockitoException:
    Mockito cannot mock this class: interface com.proofpoint.governance.agent.tagging.file.sync.api.ServerTaggedFileSyncService

    Mockito can only mock visible & non-final classes.
    If you're not sure why you're getting this error, please report to the mailing list.

    Underlying exception : java.lang.IllegalStateException: Error invoking java.lang.ClassLoader#findClass

        at com.proofpoint.governance.agent.tagging.file.sync.test.osgi.FileTaggingSyncTestActivator.init(FileTaggingSyncTestActivator.java:14)
        at com.proofpoint.governance.common.osgi.api.BaseActivator.init(BaseActivator.java:36)
        at org.apache.felix.dm.DependencyActivatorBase.start(DependencyActivatorBase.java:75)
        at org.apache.felix.framework.util.SecureAction.startActivator(SecureAction.java:697)
        at org.apache.felix.framework.Felix.activateBundle(Felix.java:2226)
        at org.apache.felix.framework.Felix.startBundle(Felix.java:2144)
        at org.apache.felix.framework.Felix.setActiveStartLevel(Felix.java:1371)
[...]

PR #388, opened 2016-04-09
Fixed OSGi metadata generation

SourcePond:
This pull-request fixes https://github.com/mockito/mockito/issues/383 and https://github.com/mockito/mockito/issues/385

Have a nice weekend
> Hey, 
> 
> Thanks for the contribution, this is awesome.
> As you may have read, project members have somehow an opinionated view on what and how should be
> Mockito, e.g. we don't want mockito to be a feature bloat.
> There may be a thorough review, with feedback -> code change loop.
> 
> _This block can be removed_
> _Something wrong in the template fix it here `.github/PULL_REQUEST_TEMPLATE.md`

check list

 - [ ] Read the [contributing guide](https://github.com/mockito/mockito/blob/master/.github/CONTRIBUTING.md)

### DIFF
--- a/gradle/osgi.gradle
+++ b/gradle/osgi.gradle
@@ -9,14 +9,14 @@ afterEvaluate {
 
             name = 'Mockito Mock Library for Java. Core bundle requires Hamcrest-core and Objenesis.'
             symbolicName = 'org.mockito.mockito-core'
-            version = project.version
+            version = project.version.replace('-', '.')
 
             instruction 'Export-Package',
                     '!org.mockito.asm.*',
                     '!org.mockito.cglib.*',
                     '!org.mockito.internal.*',
                     'org.objenesis;version=2.1',
-                    "org.mockito.*;version=${project.version}"
+                    "org.mockito.*;version=${version}"
 
             instruction 'Import-Package',
                     'net.bytebuddy.*;version=1.2.3',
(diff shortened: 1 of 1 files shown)


gradle/osgi.gradle
```diff
@@ -9,14 +9,13 @@ afterEvaluate {
 
             name = 'Mockito Mock Library for Java. Core bundle requires Hamcrest-core and Objenesis.'
             symbolicName = 'org.mockito.mockito-core'
-            version = project.version
+            version = project.version.replace('-', '.')
```
szczepiq: It does not seem to be correct that the version of the bundle is different than the version of the jar (pom.xml). What is the consequence of this?

szczepiq:
Thanks a lot!!! I commented on the ticket. Let's get OSGi fixed.

SourcePond:
Firstly, this patch has absolutely no effect when Mockito is used in a non-OSGi environment because it modifies Manifest headers which are only used by OSGi frameworks. The patch fixes the following Manifest headers as follows:
**Export-Package**: org.mockito;version="**2.0.45.beta**" instead of "**2.0.45-beta**"
The same for **Bundle-Version**: **2.0.45.beta** instead of **2.0.45-beta**

See http://www.aqute.biz/Bnd/Versioning

Mainly, it adapts the Maven- versioning scheme, which uses a dash to separate the qualifier, to OSGi which uses a dot to to this.

Additionally, the patch removes an illegal export.

PR #389, opened 2016-04-10
Lazily verify without calling collector.verify()

TimvdLippe:
As originally requested in #287: lazily verify all `Mockito.verify(...)` calls inside JUnit test cases.

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -7,7 +7,6 @@ package org.mockito;
 import org.mockito.internal.MockitoCore;
 import org.mockito.internal.creation.MockSettingsImpl;
 import org.mockito.internal.debugging.MockitoDebuggerImpl;
-import org.mockito.internal.stubbing.answers.*;
 import org.mockito.internal.stubbing.defaultanswers.ReturnsEmptyValues;
 import org.mockito.internal.stubbing.defaultanswers.ReturnsMoreEmptyValues;
 import org.mockito.internal.verification.VerificationModeFactory;
--- a/src/main/java/org/mockito/internal/MockitoCore.java
+++ b/src/main/java/org/mockito/internal/MockitoCore.java
@@ -72,7 +72,8 @@ public class MockitoCore {
         } else if (!mockUtil.isMock(mock)) {
             reporter.notAMockPassedToVerify(mock.getClass());
         }
-        mockingProgress.verificationStarted(new MockAwareVerificationMode(mock, mode));
+        VerificationMode actualMode = this.mockingProgress.maybeVerifyLazily(mode);
+        mockingProgress.verificationStarted(new MockAwareVerificationMode(mock, actualMode));
         return mock;
     }
     
@@ -182,5 +183,4 @@ public class MockitoCore {
     public MockingDetails mockingDetails(Object toInspect) {
         return new DefaultMockingDetails(toInspect, new MockUtil());
     }
-
 }
--- a/src/main/java/org/mockito/junit/VerificationCollectorImpl.java
+++ b/src/main/java/org/mockito/junit/VerificationCollectorImpl.java
@@ -1,26 +1,26 @@
 package org.mockito.junit;
 
-import org.junit.rules.TestRule;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
 import org.mockito.exceptions.base.MockitoAssertionError;
-import org.mockito.internal.MockitoCore;
+import org.mockito.internal.progress.MockingProgress;
+import org.mockito.internal.progress.MockingProgressImpl;
+import org.mockito.internal.progress.ThreadSafeMockingProgress;
+import org.mockito.verification.VerificationStrategy;
 import org.mockito.internal.verification.api.VerificationData;
 import org.mockito.verification.VerificationMode;
-
-import static org.mockito.Mockito.times;
 
 /**
  * Mockito implementation of VerificationCollector.
  */
 class VerificationCollectorImpl implements VerificationCollector {
 
-    private static final MockitoCore MOCKITO_CORE = new MockitoCore();
+    private final MockingProgress mockingProgress = new ThreadSafeMockingProgress();
 
     private StringBuilder builder;
     private int numberOfFailures;
 
-    public VerificationCollectorImpl() {
+    VerificationCollectorImpl() {
         this.resetBuilder();
     }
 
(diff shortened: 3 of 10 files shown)


TimvdLippe:
I have to say that the code based on Java 8 would be so much cleaner. This could deter all verbose wrapping and setters with using `VerificationWrapper::new`.

TimvdLippe:
Ping @szczepiq

src/main/java/org/mockito/internal/MockitoCore.java
```diff
@@ -82,7 +96,8 @@ public IOngoingStubbing stub() {
         } else if (!mockUtil.isMock(mock)) {
             reporter.notAMockPassedToVerify(mock.getClass());
         }
-        mockingProgress.verificationStarted(new MockAwareVerificationMode(mock, mode));
+        VerificationMode actualMode = this.verificationState.verifyLazily(mode);
+        mockingProgress.verificationStarted(new MockAwareVerificationMode(mock, actualMode));
```
szczepiq: How about we move the setting of a thread local state to the JUnit rule for consistency? It's kind of awkward that getter method 'isTypeMockable' actually updates the thread local state.

src/main/java/org/mockito/junit/VerificationCollectorImpl.java
```diff
@@ -20,26 +21,36 @@
...
+                    VerificationCollectorImpl.this.collectAndReport();
+                } finally {
+                    // Guard against an exception thrown inside the testcase which was not from the collector
+                    Mockito.setVerificationState(MOCKITO_CORE.getDefaultVerificationStrategy());
+                }
             }
         };
     }
 
     public <T> T verify(T mock) {
```
szczepiq: needed?

src/main/java/org/mockito/junit/VerificationCollectorImpl.java
```diff
@@ -20,26 +21,36 @@
...
-        return MOCKITO_CORE.verify(mock, new VerificationWrapper(times(1)));
+        return MOCKITO_CORE.verify(mock, times(1));
     }
 
     public <T> T verify(T mock, VerificationMode mode) {
```
szczepiq: needed?

src/main/java/org/mockito/junit/VerificationCollectorImpl.java
```diff
@@ -52,6 +63,15 @@ public void collectAndReport() throws MockitoAssertionError {
         }
     }
 
+    public VerificationCollector assertAllLazily() {
```
szczepiq: Needed?

src/main/java/org/mockito/internal/MockitoCore.java
szczepiq: Ignore my comment. I suck in understanding Git review UI.

src/main/java/org/mockito/Mockito.java
```diff
@@ -2568,4 +2568,13 @@ public static VerificationMode description(String description) {
     static MockitoDebugger debug() {
         return new MockitoDebuggerImpl();
     }
+
+    /**
+     * Set the verification strategy (direct or lazily) when invoking {@link #verify(Object)}.
+     *
+     * @param verificationState The type of verification.
+     */
+    public static void setVerificationState(VerificationStrategy verificationState) {
```
szczepiq: -1 to this new public method.

I am very cautious about introducing new public methods to the main entry point API. In this particular instance, I'm not convinced we need it :)

src/main/java/org/mockito/internal/MockitoCore.java
```diff
@@ -39,6 +40,19 @@
     private final Reporter reporter = new Reporter();
     private final MockUtil mockUtil = new MockUtil();
     private final MockingProgress mockingProgress = new ThreadSafeMockingProgress();
+    private VerificationStrategy verificationState;
+
+    public MockitoCore() {
+        this.verificationState = getDefaultVerificationStrategy();
```
szczepiq: I'm not sure about keeping state in Mockito core instance. Why not MockingProgress?

TimvdLippe:
So the general way this PR implements the requested feature is by setting a global static state. Since `Mockito.verify` should be changed depending on the context, my implementation adds a state to the core implementation. I was not aware of `MockingProgress`, so I will take a look at how to use that class.

szczepiq:
I REALLY like where this is going. In general, it feels that if someone uses JUnit, lazy verify should be the default, no?

I’m not happy about extra public on Mockito class and the VerificationStrategy. Both APIs seem to be ‘internal’, not really for regular users. Also, I feel uneasy about adding mutable state MockitoCore object. If we need such state, MockingProgress is the way to go (thread safe to the level sufficient/required in Mockito).

I’m wondering if we should start implementing some kind of listeners to the Mockito frameworks so that it is possible to implement features such you have suggested easily.

(brainstorming)

Mockito.framework().addListener(x);

VerificationListener {
  verificationStarted
}

I need this kind of API/listeners, too. However, I’m not sure how well it fits your use case. I’m working on adding automated validation of unused stubs to the JUnit runner and the Mockito JUnit rule. In order to fail the test when it has unusued stubs, I need to a) set a listener that gets notified when stubbing is done / stubbing is used b) run test c) ensure that all stubs were used out.

Thoughts?

If that works better, we can brainstorm & discuss before you cut code. I don’t want to discourage you by my endless grumbling about how things should be implemented 😃

THANKS for pushing for this feature. I think it’s cool.

src/main/java/org/mockito/junit/VerificationCollectorImpl.java
TimvdLippe: The previous implementation still required you to call `collector.verify`. That is this method. Same as method below.

src/main/java/org/mockito/junit/VerificationCollectorImpl.java
TimvdLippe: Laziness is not turned on by default. If we want to make this default with the rule, this method would not be needed.

TimvdLippe:
My gut feeling (have not tested it yet) is that `MockingProgress` is sufficient for this usecase. When I have time to fiddle with it, I will update this PR :smile:

TimvdLippe:
I have updated the pull request for using `MockingProgress` instead of `MockitoCore`.

Given that this works quite cleanly, I do not think a listening mechanism is needed right now. However, I do think it is a very valuable addition to Mockito 3.0, so let's do it then? Imo we really should publish Mockito 2.0 asap and not continue working on small features.

TimvdLippe:
Rebased the PR to resolve merge conflict.

TimvdLippe:
Ping @szczepiq with the new structure.

raphw:
I like this and I would also like to see this in 2.0, why not merge it?

TimvdLippe:
@raphw The latest version should still be reviewed. If you review it and it's LGTM I would say merge it :)

PR #412, opened 2016-05-17
Fixed #407 Vararg method call on mock object fails

### LINKED ISSUE #407: Vararg method call on mock object fails when used org.mockito.AdditionalAnswers#delegatesTo
I try to mock interface with varargs method and default implementation in final class and get this error:
```
java.lang.IllegalArgumentException: wrong number of arguments

	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.mockito.internal.stubbing.defaultanswers.ForwardsInvocations.answer(ForwardsInvocations.java:31)
	at org.mockito.internal.handler.MockHandlerImpl.handle(MockHandlerImpl.java:93)
	at org.mockito.internal.handler.InvocationNotifierHandler.handle(InvocationNotifierHandler.java:38)
	at org.mockito.internal.creation.MethodInterceptorFilter.intercept(MethodInterceptorFilter.java:51)
```

I run the following code:
```java
public class DefaultAnswerDemo {

    public interface Foo {
        void bar(String baz, Object... args);
    }

    public static final class FooImpl implements Foo {

        @Override
        public void bar(String baz, Object... args) {
            System.out.println("bar");
        }

    }

    @Test
    public void defaultAnswerTest() {
        Foo fooImpl = new FooImpl();

        Foo foo = mock(Foo.class, withSettings()
                .defaultAnswer(delegatesTo(fooImpl)));

        foo.bar("baz", 12, "45", 67.8);
    }

}
```

This is happens because org.mockito.internal.invocation.InvocationImpl by default expands varargs arguments to flat list. And I have no idea, how correctly fix this.

andreyrmg:


ChristianSchwarz:
@andreyrmg Can you add your test too?

andreyrmg:
Oh, yes, of course! I'm sorry. But only tomorrow...

andreyrmg:
Added test.

src/test/java/org/mockito/internal/stubbing/defaultanswers/ForwardsInvocationsTest.java
```diff
@@ -0,0 +1,27 @@
...
+        }
+    }
+
+    @Test
+    public void should_call_method_with_varargs() throws Throwable {
+        Foo foo = new FooImpl();
+        ForwardsInvocations forwardsInvocations = new ForwardsInvocations(foo);
+
+        assertEquals(3, forwardsInvocations.answer(invocationOf(Foo.class, "bar", "baz", new Object[] {})));
+        assertEquals(6, forwardsInvocations.answer(invocationOf(Foo.class, "bar", "baz", new Object[] {12, "34", 5.67})));
```
ChristianSchwarz: - Looks good, can you split the test cases (empty array/ non empty array) ?
- Can you also add a test for `invocationOf(Foo.class, "bar", "baz", new Object[] {null})` and `invocationOf(Foo.class, "bar", "baz", null)` 
- Would you mind using AssertJ's Assertions.assertThat(..).is(..), it is more readable IMHO?

src/test/java/org/mockito/internal/stubbing/defaultanswers/ForwardsInvocationsTest.java
```diff
@@ -0,0 +1,27 @@
+package org.mockito.internal.stubbing.defaultanswers;
+
+import org.junit.Test;
+import org.mockitoutil.TestBase;
+
+public class ForwardsInvocationsTest extends TestBase {
```
ChristianSchwarz: There is no need to extend from `TestBase`, right?

ChristianSchwarz:
+1 for adding unit tests!

Can you also add your `DefaultAnswerDemo` test case to the usage tests (package org.mockitousage.stubbing)? 
Name proposal: StubbingWithDelegateVarArgsTest

andreyrmg:
Sorry for delay. I added usage test case and squash commits into one.
And there is need to extend from `TestBase` for calling help method `org.mockitoutil.TestBase#invocationOf(java.lang.Class<?>, java.lang.String, java.lang.Object...)`.

TimvdLippe:
LGTM thanks for figuring this one out and adding an extensive test suite, luckily an easy fix.

PR #415, opened 2016-05-18
Fixes #379 Mockito.after(..) doesn't verify in a loop anymore

### LINKED ISSUE #379: Verification using After and ArgumentCaptor with Times or AtLeast methods returns unexpected size of captured values list.

ChristianSchwarz:
Fixes #379

The implementation of Mockito.after(..) doesn't loop anymore but waits until the specified time is elapsed, as specified in the JavaDoc of `Mockito.after(long)`.

### DIFF
--- a/src/test/java/org/mockitousage/verification/VerificationAfterDelayTest.java
+++ b/src/test/java/org/mockitousage/verification/VerificationAfterDelayTest.java
@@ -4,14 +4,23 @@
 
 package org.mockitousage.verification;
 
+import static java.lang.System.currentTimeMillis;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.fail;
 import static org.mockito.Mockito.after;
-import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 
-import java.util.LinkedList;
 import java.util.List;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 
 import org.junit.After;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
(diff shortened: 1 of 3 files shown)


src/test/java/org/mockitousage/verification/VerificationAfterDelayTest.java
```diff
@@ -4,152 +4,190 @@
...
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Captor;
 import org.mockito.Mock;
 import org.mockito.exceptions.base.MockitoAssertionError;
-import org.mockitoutil.TestBase;
-
-public class VerificationAfterDelayTest extends TestBase {
```
TimvdLippe: Why did you reformat this file? I think the previous version was more readible and clear for me. Also the indentation seems to be off.

TimvdLippe:
Could you rebase+squash and update the coverage so the codecov check does not fail? Also there was a previous PR solving the same issue (#380). Should we maybe check which one is better?

ChristianSchwarz:
@TimvdLippe Thanks for reviewing this PR.I completely overlooked  #380, it is not referenced in issue #379. 
This PR fixes the `after(..)` implementation by waiting the specified time and verifing it afterwards no polling is needed. The implementation follows the JavaDoc of Mockito.after(long):
> Allows verifying over a given period. It causes a verify to **wait for a specified period of time** for a desired interaction

On the other hand I learned that e.g: Mockito.after(long).atMost() must fail fast, that behaviour is not implemented in this PR. Currently we have no tests for fail-fast behaviour in the master-branch that should be fixed first.

After working on this PR I think we should re-think the current API for concurrent invocation verification. I see no reason to have both `after(long)` and `timeout(long)` but that should be discussed in an other issue.

ChristianSchwarz:
Should I close this PR?

TimvdLippe:
I think it is wise to reach a concensus in the issue first, but in the mean time we can keep the PRs open. I will mark it as blocked for resolution of #379.

raphw:
Can we close this ticket since #379 is closed?

bric3:
Yep that makes sense

PR #417, opened 2016-05-20
Add new WithinRange VerificationMode

karolzmuda:
Add new `WithinRange` `VerificationMode` which allows to verify that method use count is within given range. For example all situations below will be verified successfully:

```
verify(mockedList, withinRange(1, 3)).add("once")
verify(mockedList, withinRange(1, 3)).add("twice");
verify(mockedList, withinRange(1, 3)).add("three times");
```

And those situations will fail:

```
verify(mockedList, withinRange(1, 3)).add("never")
verify(mockedList, withinRange(1, 3)).add("four times");
```

The need for this `VerificationMode` appeared in our project when we tested web service methods which can fail at first time, then application will login and then web service method will retry. But this web service method can as well be successful at first time. So in this situation we wanted to verify if method was called once or twice, but not less or more. It wasn't possible with available `VerificationMode`s and that's why I decided to create additional one.

I hope that you will find this new feature useful :)

Best regards,
Karol Żmuda

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -7,15 +7,15 @@ package org.mockito;
 import org.mockito.internal.MockitoCore;
 import org.mockito.internal.creation.MockSettingsImpl;
 import org.mockito.internal.debugging.MockitoDebuggerImpl;
-import org.mockito.internal.stubbing.answers.*;
 import org.mockito.internal.stubbing.defaultanswers.ReturnsEmptyValues;
 import org.mockito.internal.stubbing.defaultanswers.ReturnsMoreEmptyValues;
 import org.mockito.internal.verification.VerificationModeFactory;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
 import org.mockito.mock.SerializableMode;
 import org.mockito.runners.MockitoJUnitRunner;
 import org.mockito.stubbing.*;
 import org.mockito.verification.*;
-import org.mockito.junit.*;
 
 /**
  * <p align="left"><img src="logo.png" srcset="logo@2x.png 2x" alt="Mockito logo"/></p>
@@ -250,6 +250,11 @@ import org.mockito.junit.*;
  * verify(mockedList, atLeastOnce()).add("three times");
  * verify(mockedList, atLeast(2)).add("five times");
  * verify(mockedList, atMost(5)).add("three times");
+ *
+ * //verification using withinRange()
+ * verify(mockedList, withinRange(1, 3)).add("once")
+ * verify(mockedList, withinRange(1, 3)).add("twice");
+ * verify(mockedList, withinRange(1, 3)).add("three times");
  *
  * </code></pre>
  *
@@ -2345,6 +2350,25 @@ public class Mockito extends Matchers {
      */
     public static VerificationMode atMost(int maxNumberOfInvocations) {
         return VerificationModeFactory.atMost(maxNumberOfInvocations);
+    }
+
+    /**
+     * Allows verifying number of invocations within range. E.g:
+     * <pre class="code"><code class="java">
+     *   verify(mock, withinRange(1, 3)).someMethod("some arg");
+     * </code></pre>
+     *
+     * See examples in javadoc for {@link Mockito} class
+     *
+     * @param minNumberOfInvocations minimum number of invocations
+     * @param maxNumberOfInvocations max number of invocations
+     *
+     * @return verification mode
+     *
+     * @since 2.0.0
+     */
+    public static VerificationMode withinRange(int minNumberOfInvocations, int maxNumberOfInvocations) {
+        return VerificationModeFactory.withinRange(minNumberOfInvocations, maxNumberOfInvocations);
     }
 
     /**
--- a/src/main/java/org/mockito/internal/verification/VerificationModeFactory.java
+++ b/src/main/java/org/mockito/internal/verification/VerificationModeFactory.java
@@ -36,6 +36,10 @@ public class VerificationModeFactory {
     public static VerificationMode atMost(int maxNumberOfInvocations) {
         return new AtMost(maxNumberOfInvocations);
     }
+
+    public static VerificationMode withinRange(int minNumberOfInvocations, int maxNumberOfInvocations) {
+        return new WithinRange(minNumberOfInvocations, maxNumberOfInvocations);
+    }
     
     /**
      * Verification mode will prepend the specified failure message if verification fails with the given implementation.
--- a/src/main/java/org/mockito/internal/verification/WithinRange.java
+++ b/src/main/java/org/mockito/internal/verification/WithinRange.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2007 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+
+package org.mockito.internal.verification;
+
+import org.mockito.exceptions.base.MockitoException;
+import org.mockito.internal.invocation.InvocationMatcher;
+import org.mockito.internal.verification.api.VerificationData;
+import org.mockito.internal.verification.checkers.MissingInvocationChecker;
+import org.mockito.internal.verification.checkers.NumberOfInvocationsWithinRangeChecker;
+import org.mockito.invocation.Invocation;
+import org.mockito.verification.VerificationMode;
+
+import java.util.List;
+
+public class WithinRange implements VerificationMode {
+
+    private final int minNumberOfInvocations;
+    private final int maxNumberOfInvocations;
+
+    private final MissingInvocationChecker missingInvocationChecker = new MissingInvocationChecker();
+    private final NumberOfInvocationsWithinRangeChecker invocationsWithinRangeChecker = new NumberOfInvocationsWithinRangeChecker();
+
+    public WithinRange(int minNumberOfInvocations, int maxNumberOfInvocations) {
+        if (minNumberOfInvocations < 0 || maxNumberOfInvocations < 0) {
+            throw new MockitoException("Negative values are not allowed here");
+        } else if (minNumberOfInvocations > maxNumberOfInvocations) {
+            throw new MockitoException("Minimum number of invocations cannot be higher than max number of invocations");
+        }
+        this.minNumberOfInvocations = minNumberOfInvocations;
+        this.maxNumberOfInvocations = maxNumberOfInvocations;
+    }
+
+    @Override
+    public void verify(VerificationData data) {
+        List<Invocation> invocations = data.getAllInvocations();
+        InvocationMatcher wanted = data.getWanted();
+
+        if (minNumberOfInvocations > 0) {
+            missingInvocationChecker.check(invocations, wanted);
+        }
+        invocationsWithinRangeChecker.check(invocations, wanted, minNumberOfInvocations, maxNumberOfInvocations);
+    }
+
+    @Override
+    public String toString() {
+        return "Wanted invocations count: within range from " + minNumberOfInvocations + " to " + maxNumberOfInvocations;
+    }
+
+    @Override
+    public VerificationMode description(String description) {
+        return VerificationModeFactory.description(this, description);
+    }
+}
--- a/src/main/java/org/mockito/internal/verification/checkers/NumberOfInvocationsChecker.java
+++ b/src/main/java/org/mockito/internal/verification/checkers/NumberOfInvocationsChecker.java
@@ -5,46 +5,26 @@
 
 package org.mockito.internal.verification.checkers;
 
-import java.util.List;
-
 import org.mockito.exceptions.Reporter;
 import org.mockito.internal.invocation.InvocationMatcher;
-import org.mockito.internal.invocation.InvocationMarker;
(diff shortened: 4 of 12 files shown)


ChristianSchwarz:
This reminds me of [#77 Getting an absolute count of calls to a mocked method](../issues/77). 

@karolzmuda Would the proposed solution of #77 fit your needs? Here is a usage sample:

```
@Mock
private Function<String, Integer> callable;

AtomicInteger counter= new AtomicInteger(); //here is the actual invocation count stored

countInvocations(callable,counter).apply( anyString());

assertThat(counter.get()).isBetween(1,3));
```

karolzmuda:
Your proposed solution would work, but it requires use of custom `VerificationMode` which is not included in Mockito by default ;). Also the mentioned case is less typical than mine and maybe that's why it needs custom solution.

I think that `WithinRange` `VerificationMode` complements `AtLeast` and `AtMost` and it would be quite useful to have it in library and not to be forced to use custom `VerificationMode` in order to check such simple case :)

TimvdLippe:
This proposal seems fine. @karolzmuda could you please rebase this PR on current master? Also please minimize the number of import reordering and formatting changes to ease the review process.

Since #380 applies fixes in the same classes, we have to make sure the PRs are compatible.

ChristianSchwarz:
@karolzmuda Does this API addition works also with `InOrder`, `After` and `Timeout` and `ArgumentCaptor` ?

karolzmuda:
@ChristianSchwarz `WithinRange` VerificationMode currently is not compatible with all mentioned API functionalities so of course I'm going to fix it.

I'm not convinced to `atLeast(3).atMost(5)` API. Mainly because there aren't any other VerificationModes which allow chaining and it won't be natural for users to use it like this. Moreover this solution is more error prone, for example what about using `atMost(5).atLeast(3)`? Should we allow it or not?

I would prefer to have one separate VerificationMode than chaining other VerificationModes.

@ChristianSchwarz 
@TimvdLippe 
What do you think about it?

bric3:
@karolzmuda I don't mind having both API

* `atLeast(3).atMost(5)`
* `within(3, 5)`

Actually I even think that within can be an alias of _pre-built_ verification mode i.e. `within(3, 5)` invoke behind `atLeast(3).atMost(5)`

However I feel that we should put this additional API to another class, like we did with `AddtionalAnswers`, `AdditionalMatchers`. Like @szczepiq I don't want to grow the huge `Mockito` class unless the API is considered mandatory to discover and use Mockito, which is not the case here.

ChristianSchwarz:
@karolzmuda 

> I'm not convinced to atLeast(3).atMost(5) API. Mainly because there aren't any other VerificationModes which allow chaining and it won't be natural for users to use it like this.

Chaining VerificationModes is possible: `verify(mock,after(100).atLeast(3).description("should be called 3x after 100ms")).methodCall();`

> Moreover this solution is more error prone, for example what about using atMost(5).atLeast(3)? Should we allow it or not?
I am with you! As i've written above your post it should not be supported!

@bric3 
Allowing 2 equal API's for the same use case is not ideal IMHO. Imagine a developer looks at the following 2 lines of a single test method (yes it will happen):
```
@Test
public void testFoo(){
  Bar f = new Bar(mock);
  f.doFoo("a");
  verify(mock,atLeast(2).atMost(4)).foo();
  f.doFoo("b");
  verify(mock,within(2,4)).foo();
}
```
The poor guy that look at this test, wonders what the difference is between the 2 verification modes. Someone else must did that with a special intend. So he starts comparing the javadocs and if that not help he tries to reverse engineer what is happening. At the end he will find out that he wasted 3h of his life cause one of is collegues knows only the atLeast()+atMost() API and the other how extended the test only knows the within(..)-API.

bric3:
@ChristianSchwarz I see your point, but there's javadoc and implementation available, looking at those is the minimum any developer should do. Anyway I'm thinking about other concerns, when we speak of a range we have boundaries included or excluded. Suppose we have : 

```java
withinRange(2, 5); // is it closed or left-open, right-closed or left-closed, right-open ?
```

While the above question can be answered with javadoc, the signature cannot be changed. While a builder style API provides more flexibility. For example chaining appropriate method it is possible to build a verification mode within an interval

```java
atLeast(2).atMost(4)
```

While current API does not allow yet to express full opened/closed bound possibilities, this can be changed at a later time.

However I agree this builder style API can be introduced later too.
-------

Regarding the `within` naming, I'm ok with `withinRange` as well. What do you think of this ?

```java
within(time());
within(interval());
within(range());
```

ChristianSchwarz:
>I see your point, but there's javadoc and implementation available, looking at those is the minimum any developer should do.

I disagree here respectfully. A developer should not be forced to look into the source. 95% of all question about a API should be answer by descriptive naming. Corner cases and questions like "Why?" should be documented in the JavaDocs

> Regarding the within naming, I'm ok with withinRange as well. What do you think of this ?

The _"within"_ of PR #472 is about a time constraint where the _"within"/"withinRange"_ of this PR is about a number of invocations. Having the same name for different domains may be confusing to the user. I think name should be destinct to express that we verify different things here. WDYT?

bric3:
> A developer should not be forced to look into the source. 95% of all question about a API should be answer by descriptive naming.
> Corner cases and questions like "Why?" should be documented in the JavaDocs

I agree, that's what I wanted to say, both style work without looking at the javadoc, but if one doesn't understand the difference between the two style he can look at the documentation and/or the implementation.

>> Regarding the within naming, I'm ok with withinRange as well. What do you think of this ?

> The "within" of PR #472 is about a time constraint where the "within"/"withinRange" of this PR is about a number of invocations. Having the same name for different domains may be confusing to the user. I think name should be destinct to express that we verify different things here. WDYT?

I agree, I explored other way to express this semantic difference with the _argument method_ `range(...)`, or `interval` which has a time connotation (that may come from my natural tongue though). Because `within` do fit as well for a _count range_.
I do not have strong opinion there, since I don't see any real advantages in any naming form.

ChristianSchwarz:
@bric3 I think i found a solution to avoid the name clash.  The method `within(long,TimeUnit)` of PR  #472 should be renamed to `await(long,TimeUnit)`, "await" is well known in the context of async operations e.g. `Future.await(..),`, `Condition.await(..)`, `CountDownLatch.await()`, etc.

I prefer the API`atLeast(min).atMost(max)` API  over `within(min,max)`/`within(interval(min,max))`, cause it is well readable, adapts the well known verification modes `atLeast`/`atMost` and it communicates that is is a closed interval. No matter which proposal will be chosen I would appreciate if only one API is added not both.
@all WDYT?

karolzmuda:
After reading this discussion I have to admit that I'm convinced to `atLeast(min).atMost(max)` API. If @bric3 will agree as well I would start to implement it like this soon.

This time I will also make sure that this API addition works also with `InOrder`, `After `and `Timeout `and `ArgumentCaptor`

TimvdLippe:
@karolzmuda Yes go ahead and implement the new API :smile:

bric3:
Yep that looks fine to me, I like the `await` method too :)

karolzmuda:
@ChristianSchwarz @TimvdLippe @bric3
I started to implement new API as `atLeast(min).atMost(max)` but it turns out that it breaks the existing one!

For this API to work i had to add `atMost()` method to `AtLeast` class. But in order to use this new method I had to change return type of the static method `Mockito.atLeast()` from `VerificationMode` to `AtLeast` - in other case I would not be able to use `atMost()` method. The same change goes for the `VerificationModeFactory.atLeast()`'. It is not nice but possible to do.

It gets much worse in the case of `After` which implements `VerificationAfterDelay`. After calling `after(100).atLeast(1)` it returns new instance of class `After` which is cast to `VerificationMode` so there is no way to call `after(100).atLeast(1).atMost(3)`, because `atMost()` will never be available in this case. Currently i have no idea how to implement it without adding explicit `atMost()` method to After and casting everything to explicit classes, not Interfaces (in my opinion it is unacceptable solution).

Am I missing something and it is possible to implement `atLeast(min).atMost(max)` in compliance with existing API?

TimvdLippe:
Since `AtlLeast` is a `VerificationMode`, by Liskov substitution principle this is not a breaking change on the return type. Therefore I think you can safely alter the return type without breaking existing tests.

karolzmuda:
Yes it will work but it will be strange that all other static methods in `Mockito` returns `VerificationMode` while only this one returns concrete class `AtLeast`. It is not nice but it would work.

But the bigger problem is in the case of `After` because after using `after(100).atLeast(1)` you got `After` class instance, not `AtLeast`. I wrote it in detail in previous post.

bric3:
@karolzmuda I think the `AtLeast` verification mode could implement a `VerficationWithinRange`, any other method should then declare this interface.

karolzmuda:
I like the idea of this interface but this won't fix the problem with `After`.

After calling `after(100).atLeast(1)` you won't get the `AtLeast` instance but 'After' instance instead. So after calling `after(100).atLeast(1).atMost(3)` you will get just the regular `atMost` verification mode, not the one with the range.

How to call the `atMost()` method which we need here?

ChristianSchwarz:
@karolzmuda To better understand the difficulties you face with the new API i am trying to implement a dummy. Setting up the API is easy, i can write a verification like `after(millis).atLeast(min).atMost(max)`. like you did. 
The current problem is the fact that the signature of After#copySelfWithNewVerificationMode(VerificationMode) has the return type VerificationMode but it cant be a specialized to VerificationModeAtLeast. 

I found a solution. It would take to long to explain the details, so feel free to look at the dummy code:
https://github.com/ChristianSchwarz/mockito/tree/atLeastAtMost
You only need to fill [AtLeastAndAtMost](https://github.com/ChristianSchwarz/mockito/blob/atLeastAtMost/src/main/java/org/mockito/internal/verification/AtLeastAndAtMost.java) with life.

karolzmuda:
Thanks @ChristianSchwarz :)

I'll look at this soon.

karolzmuda:
@ChristianSchwarz 
I really like your idea with specialized Interface but the problem is that if in your solution somebody calls
`after(300).atLeast(1).atMost(3)`
it works exactly the same as:
`after(300).atMost(3)`

In both cases you got the regular `atMost()` verification, not the `atLeastAtMost()` one. That is exactly the case which I had the most trouble with. Do you have any idea how to fix it?

ChristianSchwarz:
@karolzmuda I can't reproduce the described behaviour. I added an implementatio, can you provide some failing test? 

 * Tests [AtLeastXVerificationTest](https://github.com/ChristianSchwarz/mockito/blob/atLeastAtMost/src/test/java/org/mockitousage/verification/AtLeastXVerificationTest.java#L63)
 * Implementation[ AtLeastAndAtMost](https://github.com/ChristianSchwarz/mockito/blob/atLeastAtMost/src/main/java/org/mockito/internal/verification/AtLeastAndAtMost.java)

karolzmuda:
Your tests pass because you didn't use `after().atLeast().atMost()` but just regular `atLeast().atMost()` verification :). Here is the test which should definitely pass, but fails:

[Failing test](https://github.com/karolzmuda/mockito/commit/0e75d82465686ee29e8b4a7e9768bd5e3818f7f3) (I used your atLeastAtMost branch in order to keep it clear)

ChristianSchwarz:
@karolzmuda 
Sorry for the long delay! I can reproduce it thanks to your test case. I have no quick answer that one is tricky. I will look into it during the next week. I hope the VerificationWrapper is flexible enough for this API.

szczepiq:
I scanned over this thread (wow, lots of work, thanks guys!!!!!). Some feedback:

- the use case can be accomplish even with Mockito 1.10, by making 2 verifications, one with atMost(), one with atLeast() (not ideal, however the use cases is pretty edgy).
- VerificationWrapper seems like an internal API that leaked to public? (see #551).
- +1 to atMost().atLeast()
- -1 to within() (too much public API, within seems like a very edge case that even indicates potential code smells).
- -1 to a new class with rare verification modes (hard to discover API, also it seems we don't have a many rare verification modes we want to remove from main public API)
- +1 to finally exposing VerificationMode as public API (I know it's a lot of work :()

Nice discussion. Thanks guys!!!

bric3:
> * -1 to within() (too much public API, within seems like a very edge case that even indicates potential code smells)

I see your point but this is premature to rule that verification out. I've seen test involving time and parallelism where the actual invocation count cannot be determined with certainty. Some CI environment may differ and make these tests hard to be precise, while this kind of verification mode allows one to relax a bit the test.

szczepiq:
@bric3, I agree that we should _not_ discount this verification style completely (Mockito should be able to address such use cases). I do maintain my POV that an extra new method just for this use case seems off. Anyhow, I respect your POV if you opt in for within()

ChristianSchwarz:
@karolzmuda 
Ok i looked into it for 2h now without finding a solution. As I expected VerificationWrapper is "change resistant". The abstract method `copySelfWithNewVerificationMode(..)`  can only return a new instance of the implementing class, but in this usecase we need to return a new verification mode `AtLeastAndAtMost`. Also the self recursiv generic type I needed to add is absolutly not maintainable.
 
The best would be a complete redesign of the VerificationMode chaning. That is a huge work but other issues like #472 could benefit from it.

TimvdLippe:
I am closing this PR as it was targeting `master`, which we no longer use. We also need to have a discussion on the API before we start discussing the implementation.

PR #419, opened 2016-05-22
BDDMockito: rename willNothing to willDoNothing

alb-i986:
.. and keep the original signature as deprecated

### DIFF
--- a/src/main/java/org/mockito/BDDMockito.java
+++ b/src/main/java/org/mockito/BDDMockito.java
@@ -346,8 +346,16 @@ public class BDDMockito extends Mockito {
         /**
          * See original {@link Stubber#doNothing()}
          * @since 1.8.0
+         * @deprecated please use {@link #willDoNothing()} instead
          */
+        @Deprecated
         BDDStubber willNothing();
+
+        /**
+         * See original {@link Stubber#doNothing()}
+         * @since 1.10.20
+         */
+        BDDStubber willDoNothing();
 
         /**
          * See original {@link Stubber#doReturn(Object)}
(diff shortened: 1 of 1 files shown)


src/main/java/org/mockito/BDDMockito.java
```diff
@@ -346,10 +346,18 @@ public void shouldHaveNoMoreInteractions() {
...
          * See original {@link Stubber#doNothing()}
          * @since 1.8.0
+         * @deprecated please use {@link #willDoNothing()} instead
          */
+        @Deprecated
         BDDStubber willNothing();
 
         /**
+         * See original {@link Stubber#doNothing()}
+         * @since 1.10.20
```
alb-i986: Please notice this line.
I think this change should go in 1.x releases.
Or please let me know what version I should enter.

szpak:
It linguistically sounds better to me.

I wonder if we need regression tests to stay in place if original method will be removed one day.

Btw, @alb-i986 Travis failed, but it looks like a flaky test. Could you trigger it once again (e.g. with `git commit --amend`).

alb-i986:
> I wonder if we need regression tests to stay in place if original method will be removed one day.

Looks like there are no tests in place for `willNothing`. Probably because it is not doing much but just forwarding to `Mockito#doNothing`.

alb-i986:
What's the story with this?
Maybe you want me to first open an issue as per contributing guidelines?
Cheers

TimvdLippe:
@alb-i986 thanks for the ping. I think this is fine, so LGTM. The deprecated method can be removed in Mockito 3.0.

PR #423, opened 2016-05-31
Fixes #422 : Add support for setting private static final fields to Whitebox

### LINKED ISSUE #422: Whitebox.setInternalState for private static final fields
Hello,

Can you please add the ability for [Whitebox.setInternalState](https://github.com/mockito/mockito/blob/74e312b7a9236a43eb543bd3184607b75375bc40/src/main/java/org/mockito/internal/util/reflection/Whitebox.java#L22-L31) to be invokable on classes with private static final field?

```java
Whitebox.setInternalState(ClassToTest.class, "LOG", mockLog);
```

I did not understand where is the best place to add a feature request. I hope this is the best place.

Thank you

andreicristianpetcu:
Please see issue #422.
This pull request allows setting `private static final` fields and reading for static fields.

```java
Whitebox.setInternalState(ClassToTest.class, "LOG", mockLog);
Logger privateStaticLogger = Whitebox.getInternalState(ClassToTest.class, "LOG");
```

Please accept my pull request :)

### DIFF
--- a/src/main/java/org/mockito/internal/util/reflection/Whitebox.java
+++ b/src/main/java/org/mockito/internal/util/reflection/Whitebox.java
@@ -5,11 +5,18 @@
 package org.mockito.internal.util.reflection;
 
 import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
 
 public class Whitebox {
 
     public static Object getInternalState(Object target, String field) {
-        Class<?> c = target.getClass();
+        Class<?> c;
+        if(target instanceof Class){
+            c = (Class<?>) target;
+            target = null;
+        } else {
+            c = target.getClass();
+        }
         try {
             Field f = getFieldFromHierarchy(c, field);
             f.setAccessible(true);
@@ -20,13 +27,31 @@ public class Whitebox {
     }
 
     public static void setInternalState(Object target, String field, Object value) {
-        Class<?> c = target.getClass();
+        Class<?> c;
+        if(target instanceof Class){
+            c = (Class<?>) target;
+            target = null;
+        } else {
+            c = target.getClass();
+        }
         try {
             Field f = getFieldFromHierarchy(c, field);
             f.setAccessible(true);
+            removeFinalModifierIfPresent(f);
             f.set(target, value);
         } catch (Exception e) {
             throw new RuntimeException("Unable to set internal state on a private field. Please report to mockito mailing list.", e);
+        }
+    }
+
+    private static void removeFinalModifierIfPresent(Field fieldToRemoveFinalFrom) throws IllegalAccessException, NoSuchFieldException {
+        int fieldModifiersMask = fieldToRemoveFinalFrom.getModifiers();
+        boolean isFinalModifierPresent = (fieldModifiersMask & Modifier.FINAL) == Modifier.FINAL;
+        if (isFinalModifierPresent) {
+            Field modifiersField = Field.class.getDeclaredField("modifiers");
+            modifiersField.setAccessible(true);
+            int fieldModifiersMaskWithoutFinal = fieldModifiersMask & ~Modifier.FINAL;
+            modifiersField.setInt(fieldToRemoveFinalFrom, fieldModifiersMaskWithoutFinal);
         }
     }
 
--- a/src/main/java/org/mockito/internal/util/reflection/Whitebox.java
+++ b/src/main/java/org/mockito/internal/util/reflection/Whitebox.java
@@ -9,36 +9,38 @@ import java.lang.reflect.Modifier;
 
 public class Whitebox {
 
-    public static Object getInternalState(Object target, String field) {
-        Class<?> c;
-        if(target instanceof Class){
-            c = (Class<?>) target;
-            target = null;
-        } else {
-            c = target.getClass();
-        }
+    public static Object getInternalState(Object targetInstance, String fieldName) {
+        return getInternalState(targetInstance.getClass(), targetInstance, fieldName);
+    }
+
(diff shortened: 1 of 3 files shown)


TimvdLippe:
Before reading #422 I was not aware of the existence of `Whitebox`. I am more inclined to completely remove this class and not support/allow such (in my opinion) hacks instead of expanding the features in this class. This is my 2 cents, but I am curious what the others are thinking.

andreicristianpetcu:
You cannot get 100% code coverage without it. The [Stack Overflow](http://stackoverflow.com/questions/2474017/using-reflection-to-change-static-final-file-separatorchar-for-unit-testing/2474242#2474242) question and answer on setting private static final fields is quite popular. I used this class on several projects and I find it quite useful. I really hope you do not remove existing functionality.

raphw:
I am not a big fan of this class either. And to be honest, 100% test coverage is not really something that should be attempted in my opinion. Note that the class is in an internal package and is not intended to be exposed to outside users.

marcingrzejszczak:
100% test coverage is not a goal in itself. I fully agree with @raphw . @andreicristianpetcu are you testing getters and setters too? If anybody asks me for an opinion - I'm strongly against merging this PR.

andreicristianpetcu:
We have 100% code coverage. We either use the getters and setter or we don't. If we do, we do both in production and in the tests the same way and if we don't we delete them. 100% code coverage is not that hard to do. We are required to code like this.

My code does not look like the one in [your comment](https://github.com/mockito/mockito/issues/422#issuecomment-222663244) @marcingrzejszczak.

Mine is more like 

```java
class YourClass {
    private final LoggerProxy proxy;

    YourClass(LoggerProxy proxy) {
        this.proxy = proxy;
    }

    void someMethod() {
        loggerProxy.debug(...); // <-- this you can stub
    }
}
```
```java
class LoggerProxy {

    private static final Logger log ....

    void debug(...) { 
    if(log.isDebugEnabled()){
        Object computedObject = computeComplicatedDebugInfo();
        log.debug(..., computedObject);
    }
}
```
I need to cover the `if(log.isDebugEnabled()){` branch since I am building complex business around logging. Logging for me is the core domain. In any case I still need to mock the proxy `private static final Logger log`. I don't just use `log.debug("my object {}", computedObject);` since when logging complex objects it is better to check if it is debug or not before building the object.

Take the example in the [StackOverflow thread](http://stackoverflow.com/questions/2474017/using-reflection-to-change-static-final-file-separatorchar-for-unit-testing/2474242#2474242). That user want to mock the file system of the operating system so tests can be run on one platform but they will test logic of multiple platforms.

I tried to fix my issue with Powermock but it does not work either since it does bytecode manipulation and I can only mock `LogFactory.getLog(ClassToTest.class)` which is called exactly once for each class and I need to reuse the mock if I have more tests and this really feels dirty.

@TimvdLippe You might not be aware of the existence of the Whitebox class but Mockito's users are. 

I'm sorry for wasting your time with this. I was thinking this was a simple merge request with no huge debate. If you don't want to accept it, I will not insist. I was hoping I would improve the project and also fix my issue. Thank you for your time.

raphw:
Andrei, I can very much understand your use case and it is a legitimate approach to use reflection for it. The problem is simply that Mockito is concerned with logging, not with reflection which you can rather represent in a dedicated utility within your test code. 

The `Whitebox` class is meant for internal use and not as an exposed API that Mockito supports. The problem with adding is that by adding this to the public API would require us to support it. In the future, the scope of the reflection API might also change with Jigsaw which would make this even harder in the future.

Thank you for understanding, we really appreciate your effort!

andreicristianpetcu:
@TimvdLippe you can find it on Github [4,602 code results](https://github.com/search?l=java&q=org.mockito.internal.util.reflection.Whitebox&type=Code&utf8=%E2%9C%93) for Java language and the package + class.

szpak:
As far as I remember that in the past [PowerMock](https://github.com/jayway/powermock/) was suggested to use with private/static/final elements.

andreicristianpetcu:
@szpak Iwill try to use Powermock tomorrow and see exactly what issues I bump into then and post a similar example with the stack trace. I will post a issue on their repo and link it from here so you guys can follow. It might be that I was using Powermock wrong but I have been using it for quite a while now and I doubt that I was using it wrong :)
But who knows, I might learn a thing or two about Powermock.

andreicristianpetcu:
Apparently they have a test for this [exact thing](https://github.com/jayway/powermock/blob/master/reflect/src/test/java/org/powermock/reflect/WhiteBoxTest.java#L272-L276). I'm starting to question my knowledge of Powermock. :)

TimvdLippe:
@andreicristianpetcu I think [this method](http://javadox.com/org.powermock/powermock-reflect/1.5/org/powermock/reflect/Whitebox.html#setInternalStateFromContext(java.lang.Object, java.lang.Class, java.lang.Class...)) is the one you are looking for.

Thanks btw for providing the GitHub source code link. Most of the results are actually using PowerMock already, so I think it is safe to assume PowerMock suits such usecases.

@mockito/developers shall we open an issue to discuss the future of the `Whitebox` class. We might be able to deprecate it still in Mockito 2.0 if we deem necessary?

andreicristianpetcu:
I guess this pull request will never land in Mocktio so I will just close it.
I'm waiting for Powermock developers to take a look at my other [pull request](https://github.com/jayway/powermock/pull/674)

TimvdLippe:
@andreicristianpetcu Thank you again for your pull request. Hopefully Powermock will accept your contribution!

PR #440, opened 2016-06-19
remove override annotations for interface methods as this requires at…

PascalSchumacher:
… least java 1.6

`build.gradle` claims that mockito can be compiled with java 1.5, so these annotations have to go. If mockito moves to 1.6+ (imho there is no need to support java 1.5 anymore), this pull request makes no sense.

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/ByteBuddyMockMaker.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/ByteBuddyMockMaker.java
@@ -22,7 +22,6 @@ public class ByteBuddyMockMaker implements MockMaker {
         cachingMockBytecodeGenerator = new CachingMockBytecodeGenerator();
     }
 
-    @Override
     public <T> T createMock(MockCreationSettings<T> settings, MockHandler handler) {
         Class<T> mockedProxyType = createProxyClass(mockWithFeaturesFrom(settings));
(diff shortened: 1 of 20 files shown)


PascalSchumacher:
By the way the `assertj-core:2.1.0` dependency requires at least java 7.

TimvdLippe:
All right I have hit the same issue with features that were Java 7 and Travis did not fail on those. Since assertj is a test dependency, so for running the test suite of Mockito, Java 7 is required. Therefore we do not have any hint if we are Java 1.5 compatible. @mockito/developers what version of Java do we want to support with Mockito 2.0? If we want to stick with our Java 1.5 methodology, we have to make sure that the current codebase actually compiles on it.

ChristianSchwarz:
I don't know but i think mockito must not be source compatible with 1.5. It should be possible to cross compile from source 1.7 to target 1.5.

raphw:
Byte Buddy requires Java 6, so I suggest that we increment to this version both for the source and the target level. Java 6 is also one of the oldest releases that is still used quite a lot.

TimvdLippe:
All right closing this PR for now. We do have to investigate our compatibility with Java 6, but that is a separate issue.

PascalSchumacher:
Sure. :+1:  I will take a look regarding java 6 compatibility.

PR #442, opened 2016-06-19
enable some ignored tests of BridgeMethodsHitAgainTest and DetectingF…

### LINKED ISSUE #304: Problem verifying bridge methods
Hello, I think there might be a bug in mockito when verifying bridge methods.
The bug is present both in version 2.0.31-beta, and in version 1.10.19.

```java
import java.io.Serializable;
import org.junit.Test;

public class MockitoBridgeTest {

    @Test
    public void test() {
        Extended ext = mock(Extended.class);
        ext.test("123");
        verify(ext).test("123"); // Passes
        ((Base<String>)ext).test("456");
        verify(ext).test("456"); // Fails
    }

    public interface Base<T extends Serializable> {
        int test(T value);
    }

    public interface Extended extends Base<String> {
        @Override
        int test(String value);
    }

}

PascalSchumacher:
…inalMethodsTest

This worked locally, so let's see if this also works on the ci server.

Fixes #304

### DIFF
--- a/src/test/java/org/mockitousage/bugs/BridgeMethodsHitAgainTest.java
+++ b/src/test/java/org/mockitousage/bugs/BridgeMethodsHitAgainTest.java
@@ -5,7 +5,6 @@
 
 package org.mockitousage.bugs;
 
-import org.junit.Ignore;
 import org.junit.Test;
 import org.mockito.Mock;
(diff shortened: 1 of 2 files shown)


TimvdLippe:
Thanks, this has fixed issue #304 !

PR #444, opened 2016-06-19
Add PARAMETER ElementType to @Mock

### LINKED ISSUE #438: Add ElementType.PARAMETER to @Mock
Parameterized tests are becoming more common - JUnit 5 explicitly allows parameters and the sample [MockitoExtension](https://github.com/junit-team/junit5-samples/tree/master/junit5-mockito-extension/src/main/java/com/example/mockito) currently includes an @InjectMock to trigger the creation of a mock parameter.  It would be so much cleaner to use @Mock in both places.

This shouldn't be an issue for most legacy test systems since they don't allow test parameters (by default).

TimvdLippe:
TODO: Add a regression test to also show how this functionality can be used. CC @smoyer64

Fixes #438

### DIFF
--- a/src/main/java/org/mockito/Mock.java
+++ b/src/main/java/org/mockito/Mock.java
@@ -9,6 +9,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
 import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.PARAMETER;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/mockito/Mock.java
```diff
@@ -58,7 +59,7 @@
  * @see MockitoAnnotations#initMocks(Object)
  * @see org.mockito.runners.MockitoJUnitRunner
  */
-@Target(FIELD)
+@Target({FIELD, PARAMETER})
```
sbrannen: Awesome! 👍

Issue #433, opened 2016-06-15
Fix swinging concurrent tests


ChristianSchwarz:
The following concurrent tests fail currently with a >50% chance. This should be fixed!

Here is a build with multiple failed test that are related to concurrent issues #763 https://travis-ci.org/mockito/mockito/builds/128143294
```
VerificationAfterDelayTest.shouldVerifyNormallyWithSpecificTimes 
VerificationWithTimeoutTest.shouldAllowMixingOnlyWithTimeoutAndFail
VerificationWithTimeoutTest.shouldAllowMixingOtherModesWithTimeout
VerificationWithTimeoutTest.shouldAllowMixingOnlyWithTimeout 
VerificationWithTimeoutTest.shouldAllowMixingOtherModesWithTimeoutAndFail
```

raphw:
Those are time-out-dependant test. Those are generally a bit hard to test but I have never seen them fail. Maybe it depends on how busy Travis is at the moment.

TimvdLippe:
Oh right, the original PR was closed so Travis could not pull the branch :stuck_out_tongue_closed_eyes:

Looking at our build history it seems to be happening very rarely. So if it happens again we should restart the build. If it then fails again a manual inspection should be done.

PR #446, opened 2016-06-19
Make tests which test for timeouts with Thread#sleep more lenient.

PascalSchumacher:
I took a look at the recently builds on travis ci and noticed a lot of failed builds. These failures where mostly unrelated to the actual changes, but caused by these strict timeout tests.

This pull request will not completely fix these random failures, but it should reduce their occurrences. 

The ideal solution would be to completely remove usage of `Thread#sleep` from the test, but I have no idea how to do this.


ChristianSchwarz:
relates to #433

PR #449, opened 2016-06-19
Java6

PascalSchumacher:


PascalSchumacher:
Should I squash the commits?

.travis.yml
```diff
@@ -13,9 +13,18 @@ cache:
...
+  - sudo hostname "$(hostname | cut -c1-63)"
+  - sed -e "s/^\\(127\\.0\\.0\\.1.*\\)/\\1 $(hostname | cut -c1-63)/" /etc/hosts | sudo tee /etc/hosts
+  - cat /etc/hosts # optionally check the content *after*
+
 language: java
 jdk:
-- oraclejdk7
+  - openjdk6
+  - oraclejdk7
+  - oraclejdk8
```
TimvdLippe: Maybe we can switch to `openjdk7` and `openjdk8` instead of relying on Oracle.

src/test/java/org/mockitousage/basicapi/MocksSerializationTest.java
```diff
@@ -394,6 +395,13 @@ public void private_constructor_currently_not_supported_at_the_moment_at_deseria
 
     @Test
     public void BUG_ISSUE_399_try_some_mocks_with_current_answers() throws Exception {
+        // for some reason this test fails with NullPointerException java 6, but work on java 7+
+        // java.lang.NullPointerException
+        // at java.util.LinkedList.writeObject(LinkedList.java:942)
+        // ...
+        // at org.mockitoutil.SimpleSerializationUtil.serializeMock(SimpleSerializationUtil.java:34)
```
TimvdLippe: We should extract an issue for this to investigate (and potentially fix for Java 6)

TimvdLippe:
LGTM with 2 small comments.

szpak:
Doesn't the CD mechanism have to be disabled for Java 7 and 8?

src/test/java/org/mockitousage/basicapi/MocksSerializationTest.java
PascalSchumacher: https://github.com/mockito/mockito/issues/450

PascalSchumacher:
@szpak I guess you are right. I reduced the build to Java 6 for now.

raphw:
@PascalSchumacher I could not reproduce the problems you found with the tests but that concerning serialization. Otherwise, I spent quite some time removing remaining deprecation etc where the essential rollback to Java 6 is already implemented.

Sorry for the concurrent fix.

PascalSchumacher:
Thinking about it I guess the line number issue were caused by my first/wrong attempt to get `StringJoiner`working on Java 6.

PascalSchumacher:
As for the compile issues I used jdk-6u45 for testing, maybe these issues were fixed in a later (non-public) release or in open-jdk.

PR #454, opened 2016-06-21
Modified JavaDoc for ArgumentMatcher

### LINKED ISSUE #453: Either revert recent Functional Interfaces change or change JavaDoc
I've been reading the Javadoc regarding `ArgumentMatcher` and I noticed a couple of things. Relating back to the changes of #338 

- The example of how to override it is still based around `Object` rather than `T`
- Someone has very deliberately decided not to make it use `T` in the past and now we've changed that.

> The argument is not using the generic type in order to force explicit casting in the implementation. This way it is easier to debug when incompatible arguments are passed to the matchers. You have to trust us on this one. If we used parametrized type then ClassCastException would be thrown in certain scenarios.

My view at the time was that changing it to `T` would remove the class casting and that this would be a good thing. It was kind of necessary to have it as `T` so that the lambda expressions in Java 8 could infer the right types.

The rationale about a class cast exception giving weird results kind of makes sense, but does that really really happen? It seems it's either solve one problem or the other.

What do folks think? I'll happily amend the code either way.

ashleyfrieze:
To fix #453 - this is the amendment to the `ArgumentMatcher` JavaDoc.


TimvdLippe:
LGTM.

PR #451, opened 2016-06-19
Refactored Timeout and After concurrent test

ChristianSchwarz:
Refactored concurrent test in order to fix #433. The implementation now uses  an `Executor `to perform async calls instead of extending `Thread` and doesn't extend TestBase anymore. 

I also added a Stopwatch which can be used for further time dependent testing. In the future I want to use this class to write more tests for After/Timeout.

PR #424, opened 2016-06-01
Inorder timeouts

Jazzepi:
*Fixes issue https://code.google.com/archive/p/mockito/issues/292
  *Updated code provided by https://github.com/sirchia in PR #20

Closes #20

PR #455, opened 2016-06-22
activate VerificationWithTimeoutTest#shouldAllowTimeoutVerificationIn…

PascalSchumacher:
…Order

This test work since https://github.com/mockito/mockito/pull/424 but was deactivated by mistake with https://github.com/mockito/mockito/pull/451


TimvdLippe:
Note to myself when reviewing PR's: be more precise. Yes I missed this case too and it should not happen. Thank you for correcting me!

PR #460, opened 2016-06-23
delete disabled test for removed objenesis missing reporting feature

PascalSchumacher:


TimvdLippe:
Assigning to @raphw as he spend more time with Objenesis than me.

PR #461, opened 2016-06-23
add test for "Add ElementType.PARAMETER to @Mock" #438 as suggested b…

PascalSchumacher:


TimvdLippe:
I did not include a test yet since the specs for the new MockitoExtensions are not fleshed out. This test alone is not really useful I think.

PascalSchumacher:
Yes, the test does not add much value so feel free to just close the pull request.

TimvdLippe:
Closing per above comment.

PR #463, opened 2016-06-23
Fixes #457 CompareTo compares now only agains instances of Comparable.

### LINKED ISSUE #457: Stubbing with some AdditionalMatchers can NPE with null actuals
```java
import static org.hamcrest.MatcherAssert.assertThat;
import static org.hamcrest.Matchers.is;
import static org.mockito.AdditionalMatchers.leq;
import static org.mockito.Matchers.eq;
import static org.mockito.Mockito.when;

import org.junit.Test;

public class BlahTest
{
	@Test
	public void test()
	{
		final BoolResult mockBoolResult = mock(BoolResult.class);

		when(mockBoolResult.getResult(eq(5))).thenReturn(true);  // Succeeds as expected
		when(mockBoolResult.getResult(leq(5))).thenReturn(true); // Fails with an NPE - as of v2.0.64 - unexpected!

		assertThat(mockBoolResult.getResult(null), is(false));
	}

	interface BoolResult
	{
		boolean getResult(Integer value);
	}
}
```

ChristianSchwarz:
This PR...
 * fixes the NPE described in #457 
 * fixes cases were CCE were thrown if an `ArgumentMatcher `is not type-complatible with an argument to match.
 * fixes parameter types in all constructors of `CompareTo `sub classes 
 * added tests to `HashCodeAndEqualsSafeSetTest `in order to pass the codecov checks

### DIFF
--- a/src/main/java/org/mockito/internal/invocation/ArgumentsComparator.java
+++ b/src/main/java/org/mockito/internal/invocation/ArgumentsComparator.java
@@ -4,11 +4,13 @@
  */
 package org.mockito.internal.invocation;
 
+import static java.lang.reflect.Modifier.isPublic;
+
+import java.lang.reflect.Method;
+import java.util.List;
 import org.mockito.ArgumentMatcher;
 import org.mockito.internal.matchers.VarargMatcher;
 import org.mockito.invocation.Invocation;
-
-import java.util.List;
 
 @SuppressWarnings("unchecked")
 public class ArgumentsComparator {
@@ -22,21 +24,24 @@ public class ArgumentsComparator {
             return false;
         }
         for (int i = 0; i < actualArgs.length; i++) {
-            if (!invocationMatcher.getMatchers().get(i).matches(actualArgs[i])) {
+            ArgumentMatcher<Object> argumentMatcher = invocationMatcher.getMatchers().get(i);
+            Object argument = actualArgs[i];
+
+            if (!matches(argumentMatcher, argument)) {
                 return false;
             }
         }
         return true;
     }
 
-    //ok, this method is a little bit messy but the vararg business unfortunately is messy...      
-    private boolean varArgsMatch(InvocationMatcher invocationMatcher, Invocation actual) {
+    // ok, this method is a little bit messy but the vararg business unfortunately is messy...
+    private static boolean varArgsMatch(InvocationMatcher invocationMatcher, Invocation actual) {
         if (!actual.getMethod().isVarArgs()) {
-            //if the method is not vararg forget about it
+            // if the method is not vararg forget about it
             return false;
         }
 
-        //we must use raw arguments, not arguments...
+        // we must use raw arguments, not arguments...
         Object[] rawArgs = actual.getRawArguments();
         List<ArgumentMatcher> matchers = invocationMatcher.getMatchers();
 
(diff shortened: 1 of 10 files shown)


ChristianSchwarz:
Currently I have no idea why the following test fails before the `NOT_COMPARABLE` instance is passed to `CompareTo.matches(T)`

```
private static final Object NOT_COMPARABLE = new Object();

/** Should not throw an {@link ClassCastException} */
@Test
public void compareToNonCompareable() {
     final IMethods mock = mock(IMethods.class);
     when(mock.forObject(leq(5))).thenReturn("");
     assertThat(mock.forObject(NOT_COMPARABLE)).isNull();//a default value must be returned
}
```

ChristianSchwarz:
Please restart the build, timing issues again....

ChristianSchwarz:
I found a fix for the failing test described above. It is related to the latest changes of `ArgumentMatcher.matches(Object)` to `matches(T)`. Now the deleted JavaDoc of PR #454 makes sense.

I will squash the commits later.

TimvdLippe:
Luckily we now have a testcase which we can use to debug this issue. I would prefer to find a solution where we remain with type `T`, but catch the ClassCastException somewhere. I will spend some time this weekend to investigate what is going on.

ChristianSchwarz:
The ClassCastException is thrown [here ](https://github.com/mockito/mockito/blob/196ff979da156caa07e19f57e4849637d8bede1a/src/main/java/org/mockito/internal/invocation/ArgumentsComparator.java#L25) in `ArgumentsComparator.argumentsMatch(..)`. 

I try to explain what I found so far:
 -  The type paramter `T` of `CompareTo<T extends Comparable<T>>` will not be erased to `Object `but `Comparable ` when the class is extended. E.g. the signature of matches(T)  is not `matches(Object)` but `matches(Comparable)`. When an instance is passed that is not a Comparable the ClassCastException is throw.
 - My current workaround is to change the implements declaration from` ArgumentMatcher<T>` to `ArgumentMatcher<Object>`, this solves the problem.
 - An other idea is to keep the implements declaration `ArgumentMatcher<T>` and try to retrieve the type information from the class object of the given matcher, [here is a sample](https://gist.github.com/ChristianSchwarz/9b45dfccd3680405250b73de22e3e956). It works to a limited degree but it is a pain and hard to maintain.

ChristianSchwarz:
Ok i think the way to go is to retrieve the parameter type of the matches-method and check if it is assignable form the given argument type. This way we don't need to catch ClassCastException or drop the type arguments for ArgumentMatcher. 

-> [sample implementation](https://gist.github.com/ChristianSchwarz/9b45dfccd3680405250b73de22e3e956#file-argumentscomparator_methodtype-java)

raphw:
This will only work if the parameterized type is explicit in the generated class, wouldn'it? If I create a type extending `ArgumentMatcher` that is itself generic, we cannot retreive the parameterized type.

ChristianSchwarz:
@raphw 

> This will only work if the parameterized type is explicit in the generated class, wouldn'it? 

Right cause the current implementation of this PR uses the type of the parameter of the passed ArgumentMatcher.matches(T) implementation. 

> If I create a type extending ArgumentMatcher that is itself generic, we cannot retreive the parameterized type.

For e.g. LessOrEqual it works fine. Do you see a problem here, for other cases?

src/main/java/org/mockito/internal/invocation/ArgumentsComparator.java
```diff
@@ -46,18 +52,71 @@ private boolean varArgsMatch(InvocationMatcher invocationMatcher, Invocation act
...
+                // varargs), just do the ordinary check
+            } else if (!m.matches(rawArgs[i])) {
                 return false;
             }
         }
 
         return true;
     }
+
+    private static boolean matches(ArgumentMatcher<Object> argumentMatcher, Object argument) {
```
TimvdLippe: Shouldn't this type be `ArgumentMatcher<?>`?

src/test/java/org/mockito/internal/matchers/ComparableMatchersTest.java
```diff
@@ -5,14 +5,13 @@
 
 package org.mockito.internal.matchers;
 
-import org.junit.Test;
-import org.mockitoutil.TestBase;
-
```
TimvdLippe: Unecessary reordering of imports and whitespace.

src/main/java/org/mockito/internal/invocation/ArgumentsComparator.java
ChristianSchwarz: I thougth the same, but it doesn't compile.

src/main/java/org/mockito/internal/matchers/CompareTo.java
```diff
@@ -9,24 +9,27 @@
...
         this.wanted = value;
     }
 
-    @SuppressWarnings("unchecked")
     public boolean matches(T actual) {
-        return matchResult(((Comparable)actual).compareTo(wanted));
+        if (actual == null) {
+            return false;
+        }
+
```
ChristianSchwarz: To avoid a `ClassCastException` it must be ensured that _wanted_ is compatible to _actual_ : 
```
if (!actual.getClass().isInstance(wanted)){ 
   return false
}
```

src/main/java/org/mockito/internal/invocation/ArgumentsComparator.java
```diff
@@ -46,18 +52,70 @@ private boolean varArgsMatch(InvocationMatcher invocationMatcher, Invocation act
...
+     * {@link ClassCastException}.
+     */
+    private static boolean isCompatible(ArgumentMatcher<?> argumentMatcher, Object argument) {
+        if (argument == null)
+            return true;
+
+        Class<?> expectedArgumentType = getArgumentType(argumentMatcher);
+        Class<?> actualArgumentType = argument.getClass();
+
+        return expectedArgumentType.isAssignableFrom(actualArgumentType);
```
ChristianSchwarz: should be simplified to `return expectedArgumentType.isInstance(argument)`

ChristianSchwarz:
@raphw 

> This will only work if the parameterized type is explicit in the generated class, wouldn'it? If I create a type extending ArgumentMatcher that is itself generic, we cannot retreive the parameterized type.

Yes thats both true! But that should not be a problem. At least I can't imaging how it could cause a `ClassCastException` when `matches(T)` is called. Can you provide a test that might fail ?

This test case succeed and throws no CCE:

```
class IsNegative<T extends Number> implements ArgumentMatcher<T> {
    @Override
    public boolean matches(T argument) {
        return argument.floatValue()<0;
    }
}

when(mock.forObject(argThat(new IsNegative()))).thenReturn("matched");

assertThat(mock.forObject("i'am not a number")).isNull();
```

ChristianSchwarz:
The NPE-Bug #457 plus several other `ClassCastExceptions` are fixed and no new bugs was introduced I hope.
What is needed to complete this PR?

TimvdLippe:
I will have some time this evening to review and merge this :smile:

TimvdLippe:
Landed in https://github.com/mockito/mockito/commit/118bed80d2dc42153ea3218a9698d65de073724a

Added the two tests to verify the correctness when subclassing `ArgumentMatcher` with a generic type.

Thanks a ton @ChristianSchwarz !

PR #470, opened 2016-06-26
Refactored mock cache to be non-blocking.

raphw:
The previous cache used a lock to unify concurrency and weak references. Instead, keys are now explicitly wrapped using weak references and similarly to the functioning of a weak hash map, the entries are polled upon accessing the cache. This avoids blocking.

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/CachingMockBytecodeGenerator.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/CachingMockBytecodeGenerator.java
@@ -1,70 +1,95 @@
 package org.mockito.internal.creation.bytebuddy;
 
-import static org.mockito.internal.util.StringJoiner.join;
-import java.lang.ref.WeakReference;
-import java.lang.reflect.Modifier;
-import java.util.HashSet;
-import java.util.Set;
-import java.util.WeakHashMap;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReentrantLock;
 import org.mockito.exceptions.base.MockitoException;
 
-class CachingMockBytecodeGenerator {
+import java.lang.ref.Reference;
+import java.lang.ref.ReferenceQueue;
+import java.lang.ref.SoftReference;
+import java.lang.ref.WeakReference;
+import java.lang.reflect.Modifier;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 
-    private final Lock avoidingClassLeakCacheLock = new ReentrantLock();
-    public final WeakHashMap<ClassLoader, CachedBytecodeGenerator> avoidingClassLeakageCache =
-            new WeakHashMap<ClassLoader, CachedBytecodeGenerator>();
+import static org.mockito.internal.util.StringJoiner.join;
+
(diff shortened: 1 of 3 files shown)


src/main/java/org/mockito/internal/creation/bytebuddy/CachingMockBytecodeGenerator.java
```diff
@@ -1,70 +1,94 @@
...
-                    new CachedBytecodeGenerator(mockBytecodeGenerator)
-            );
+    private CachedBytecodeGenerator mockCachePerClassLoaderOf(ClassLoader classLoader) {
+        classLoader = classLoader == null ? BOOT_LOADER : classLoader;
+        CachedBytecodeGenerator generator = avoidingClassLeakageCache.get(new StrongKey<ClassLoader>(classLoader)), newGenerator;
+        if (generator == null) {
+            generator = avoidingClassLeakageCache.putIfAbsent(new WeakKey<ClassLoader>(classLoader, this), newGenerator = new CachedBytecodeGenerator(mockBytecodeGenerator, weak));
+            if (generator == null) {
+                generator = newGenerator;
+            }
```
bric3: I'd change that code to one variable declaration per line. And I would avoid inline assignment too (the first `newGenerator` is not needed before the `if`).
Those too may confuse most readers.

src/main/java/org/mockito/internal/creation/bytebuddy/CachingMockBytecodeGenerator.java
```diff
@@ -1,70 +1,94 @@
...
     }
 
-    private <T> CachedBytecodeGenerator mockCachePerClassLoaderOf(Class<T> mockedType) {
-        if (!avoidingClassLeakageCache.containsKey(mockedType.getClassLoader())) {
-            avoidingClassLeakageCache.put(
-                    mockedType.getClassLoader(),
-                    new CachedBytecodeGenerator(mockBytecodeGenerator)
-            );
+    private CachedBytecodeGenerator mockCachePerClassLoaderOf(ClassLoader classLoader) {
+        classLoader = classLoader == null ? BOOT_LOADER : classLoader;
```
bric3: Nifty !

src/main/java/org/mockito/internal/creation/bytebuddy/CachingMockBytecodeGenerator.java
```diff
@@ -1,70 +1,94 @@
...
-                    params
-            );
+        pollAndClearReferences();
+        return (Class<T>) mockCachePerClassLoaderOf(params.mockedType.getClassLoader()).getOrGenerateMockClass(params);
+    }
 
-            return (Class<T>) generatedMockClass;
-        } finally {
-          avoidingClassLeakCacheLock.unlock();
+    void pollAndClearReferences() {
```
bric3: I would like a comment on this, or maybe rename it better (I'm thinking loudly).
Not everyone knows `ReferenceQueue` or how it works, and thus why we need to invoke this method upon `get(...)`.

Ideas welcome.

src/main/java/org/mockito/internal/creation/bytebuddy/CachingMockBytecodeGenerator.java
```diff
@@ -133,4 +163,60 @@ public int hashCode() {
             }
         }
     }
+
+    private interface Key<T> {
+
+        T get();
+    }
+
+    private static class StrongKey<T> implements Key<T> {
```
bric3: I think we could force `T` to be `ClassLoader`, as this is the only usage of those key objects.
Also `StrongKey` is only used for lookup, maybe we could rename it this way.

What do you think ?

bric3:
Looks good codewise.
I wonder if coverage can be increased however.

bric3:
LGTM

Issue #425, opened 2016-06-02
doThrow(), Wrong error message on final method (Checked exception is invalid for this method!)


Maragues:
Trying to force an Exception on a final method throws a misleading error message

```
org.mockito.exceptions.base.MockitoException: 
Checked exception is invalid for this method!
Invalid: java.lang.Exception
```

where it should inform you that it can't be done because the method is *final*

This a self contained demonstration

```
public class DoThrowOnFinalMethodTest {

  private static class MyClass {
    void run() throws Exception {

    }

    final void walk() throws Exception {

    }

    void executeRun() {

    }

    void executeWalk() {

    }
  }

  @Test
  public void executeRun_throwsException() throws Exception {
    MyClass myClass = spy(new MyClass());

    doThrow(new Exception()).when(myClass).run();

    myClass.executeRun();
  }

  @Test
  public void executeWalk_throwsException() throws Exception {
    MyClass myClass = spy(new MyClass());

    doThrow(new Exception()).when(myClass).walk();

    myClass.executeWalk();
  }
}
```

Mockito version: 1.10.19
JUnit version: 4.12
OS: Mac OS X 10.11.4

Executing from Android Studio using *Gradle-aware make*

ChristianSchwarz:
The JavaDoc of `spy(..)` contains the section _**"Important gotcha on spying real objects!"**_  here is the part for your specific case:
>3 . Watch out for final methods. Mockito doesn't mock final methods so the bottom line is: when you spy on real objects + you try to stub a final method = trouble. Also you won't be able to verify those method as well. 

Mockito has no chance here to detect that you sub a final method. Mockito can't use the call  to the final method to finish he stubbing,so the subbing get finnished on the next call to the same mock.  

To illustrate the problem we can rewrite this failure case like this:

```
Mock m = doThrow(new Exception()).when(mock); //note m == mock

m.callToFinalMethod(); //cause it is final the call can't be recorded by mockito 

mock.callNonFinalMethodThatDeclaresNoException(); //now Mockito records this call and tries to finish the stubbing but detects that this method can't throw Exception's!!
```

I am afraid the only way to provided better error messages for this case is to analyse the byte-code of the test method to detect the misuse or to provide an IDE-PlugIn that does this kind of analysis at complie time.

Issue #194, opened 2015-04-05
Tweaks Matchers.any family matchers behavior


bric3:
This issue is a follow up of the work started in #141 by @pimterry. Also here's an extract of a message on [this thread](https://groups.google.com/forum/#!topic/mockito/8_WGBB3Jbtk) of the mailing list :

-------------

The origin of these methods is they come from `anything` i.e. anything matches, later for shortness and cast avoidance the aliases grew, but the API naming thus became inconsistent with what a human would expect. So this behavior is being changed in mockito 2 beta, to be precise here's the status on these API in the version 2.0.5-beta :

- `any`, `anyObject`, `any(Class)` won't check anything (at first they were just aliases for _anything_ and for cast avoidance), `null` is a valid value
- `anyX` like `anyString` will check the arg is not `null` and that has the correct type
- `anyList` will check the argument is not null and a `List` instance
- `anyListOf` (and the likes) at the moment are just aliases to their non generic counter part like `anyList`

Note this is work in progress (started here in [#141](https://github.com/mockito/mockito/pull/141)), these new behavior can / will change in the beta phase. I'm especially wondering if the `any` family should allow `null` and if not do a type check. For example with these matchers : 

- `any`, `anyObject` stay the same, they currently allow `null` and don't have to do type check anyway
- `any(Class)` currently allows `null` and doesn't do type check => allows `null` and if not checks for the given type
- `any<Collection>Of` currently doesn't allow `null` and does a type check of the collection, not elements => allows `null`, if not checks collection type, if not empty checks element type

Maybe extend/create a _symmetric_  `isA` family API that won't allow any `null` arguments.

-------------

bric3:
`any` family is currently _(at his time up to 2.0.5-beta)_ inconsistent on the way they allow or not `null`s. Type checking is ok, but `null` check behavior consistency is also important.

szczepiq:
AnyX gramtaically tends to imply to return false for null because any
object is not null. I agree completely the api should be consistent.

bric3:
> AnyX gramtaically tends to imply to return false for null because any object is not null.

I'm not anymore of sure of that. Anyway the changes introduced in this API make it inconsistent.
That's why I propose any validates `null`, and if not `null` then validates the type

Also on the mentioned thread _Francisco Olarte_ was proposing a `anyNotNull` family, this could clarify the behavior on `null` of these APIs.

szczepiq:
Agreed. Any should return false for null and false for wrong type (instance
[...]

PR #474, opened 2016-06-28
ArgumentCaptor: Enhance public API

### LINKED ISSUE #473: Feature Request ArgumentCaptor.getValueAt(index)
Instead of `argumentCaptor.getAllValues().get(1)`, `argumentCaptor.getValueAt(1)` would be pretty handy.

If this is something you'd want, I'm more than happy to post a PR.

vanniktech:
Once we agree on the public API I'd start adding tests + documentation. @ChristianSchwarz @TimvdLippe 

Any feedback is welcome. Also we might want to revisit the decision of renaming `getValue` since this is a breaking change.

This will fix #473

### DIFF
--- a/src/main/java/org/mockito/ArgumentCaptor.java
+++ b/src/main/java/org/mockito/ArgumentCaptor.java
@@ -40,9 +40,9 @@ import org.mockito.internal.matchers.CapturingMatcher;
  *
  * <p>
  * In a way ArgumentCaptor is related to custom argument matchers (see javadoc for {@link ArgumentMatcher} class).
- * Both techniques can be used for making sure certain arguments where passed to mocks. 
+ * Both techniques can be used for making sure certain arguments where passed to mocks.
  * However, ArgumentCaptor may be a better fit if:
- * <ul>  
+ * <ul>
  * <li>custom argument matcher is not likely to be reused</li>
  * <li>you just need it to assert on argument values to complete verification</li>
  * </ul>
@@ -60,7 +60,7 @@ import org.mockito.internal.matchers.CapturingMatcher;
  * @since 1.8.0
  */
 public class ArgumentCaptor<T> {
-    
+
 
     private final CapturingMatcher<T> capturingMatcher = new CapturingMatcher<T>();
     private final Class<? extends T> clazz;
@@ -73,10 +73,10 @@ public class ArgumentCaptor<T> {
      * Use it to capture the argument. This method <b>must be used inside of verification</b>.
      * <p>
      * Internally, this method registers a special implementation of an {@link ArgumentMatcher}.
-     * This argument matcher stores the argument value so that you can use it later to perform assertions.  
+     * This argument matcher stores the argument value so that you can use it later to perform assertions.
      * <p>
      * See examples in javadoc for {@link ArgumentCaptor} class.
-     * 
+     *
      * @return null or default values
      */
     public T capture() {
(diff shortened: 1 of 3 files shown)


TimvdLippe:
I do agree with the renaming of `getValue` to `getLastValue` and the other methods seem logical to me too. Public API LGTM.

Also I would like to hear the thoughts of @raphw and @bric3 and if they also agree tests can be fixed.

vanniktech:
Yup I just spent the minimal time right now to get the API straight afterwards I'll fix all of those things. Also what about `getFourthValue` and `getFifthValue`, do you think that one need them?

src/main/java/org/mockito/ArgumentCaptor.java
```diff
@@ -93,10 +93,26 @@ public T capture() {
      * 
      * @return captured argument value
      */
-    public T getValue() {
+    public T getLastValue() {
```
szpak: Wouldn't it be worth to keep `getValue()` as `@Deprecated` for a month or two?

szpak:
I don't think `getFourthValue` and `getFifthValue` is so often used.

src/main/java/org/mockito/ArgumentCaptor.java
vanniktech: Yes that's what I also thought. Make it deprecated and internally call the new one (which also has the same javadoc)

vanniktech:
@TimvdLippe  & @szpak I updated this. Is the public API for ArgumentCaptor good now? If so I'd start adding unit tests for the added methods.

TimvdLippe:
@vanniktech yes that looks good!

vanniktech:
@TimvdLippe I created some tests however they're failing for a reason I can't find out. Can you guys help me? Also before there were no tests for ArgumentCaptor so I had no example to look at.

ChristianSchwarz:
@vanniktech The tests fail cause the methods in `AtomicInteger ` are final, see #425 for more details. 
Hint: It is common for mockito test to mock `org.mockitousage.IMethods`.

local.properties
```diff
@@ -0,0 +1,11 @@
+## This file is automatically generated by Android Studio.
```
ChristianSchwarz: can you add this file to gitignore

src/main/java/org/mockito/ArgumentCaptor.java
```diff
@@ -90,25 +90,92 @@ public T capture() {
...
+    /**
+     * Returns the captured value of the argument. When capturing varargs use {@link #getAllValues()}.
+     * <p>
+     * If verified method was called multiple times then this method it returns the latest captured value at the given index.
+     * <p>
+     * See examples in javadoc for {@link ArgumentCaptor} class.
+     *
+     * @return captured argument value
+     */
+    public T getValueAt(int index) {
```
ChristianSchwarz: `getValue(int index)` ?

vanniktech:
Ups what a rookie mistake. I changed it now to `IMethods`. Tests are also now running. Thanks @ChristianSchwarz

src/main/java/org/mockito/ArgumentCaptor.java
vanniktech: Personally I don't care whether its `getValueAt` or `getValue`. @TimvdLippe wants `getValueAt`. I'll leave it up to you guys

TimvdLippe: I am still convinced the At suffix is better, but since I voiced my opinion already let's wait for the response of the other @mockito/developers

vanniktech:
@bric3 is there any chance to get this into the current 2.0.x beta?

bric3:
@vanniktech I'd rather not include this now, I'd like to focus on the remaining API change of #194. I think argument captor need some overhaul but there's many pull request and ideas, that I don't want to introduce yet in the master without delaying mockito release / rc.

szczepiq:
Thanks for PR!!!

Can you describe the motivation behind this PR? E.g. what's the use case that drives this? It is really helpful if we can clearly understand the reasons you want to change the API :)

If this change is to drive clarity, we could also consider:
get()
get(int)
getAll()

getArgument()
getArgument(int)
getArguments()

I vote -1 for the getFirst(), getSecond(), etc. It feels adding public methods to the API, thus making it bigger, for little value (I don't think those methods are used often enough). Also it will inconsistent with other places in the API where we access multiple values (like invocation.getArgument(x))

vanniktech:
Well mostly to enhance the public API and make the tests more readable.

Consider `captor.getAllValues().get(1)` against `captor.getFirstValue()` or `captor.getValueAt(1)`. The first one feels really clunky in comparison.

`get()`, `get(int)` & `getAll()` also do seem a lot better than the current API.

Also in the current API `getValue()` seems a bit off in my opinion. Up until the point of this PR I always thought I'd get the first value back and not the last one.

Now thinking of it I like `get()`, `get(int)` & `getAll()` a lot more than the proposal here. Also with `get(int)` there's really no need for `getFirst()` etc.

szczepiq:
I also like get/getAll

Is it worth changing (the cost is churn to usesrs) ? Is it worth changing
in 2.0?

I am leaning towards not changing it.

vanniktech:
I'd like the enhanced API to be in 2.0

bric3:
Nope, `ArgumentCaptor` has some issues that we won't address in 2.0. They will be in 2.1 which will be released not too long after 20, **it will be the one of the main focus of 2.1**.

@szczepiq This proposition doesn't allow to expose vararg in a clean way, also many time I found out that people that needed that with multiple invocation didn't have the APIs to so.

szczepiq:
ok, thanks for info.

vanniktech:
Any update(s) on this one?

bric3:
@vanniktech Strong disagreement on the release model in #618 have delayed this work and related issues.

ChristianSchwarz:
bump...

vanniktech:
Anything I can do to speed this up?

szczepiq:
Thanks for submitting!!! In the interest of progress I will summarize this PR.

Use cases:
 1. "captor.getAllValues().get(1)" is too long
 2. getValue() does not indicate which value it is (first or last)

My feedback to the use cases:
 1. Reasonable, +1
 2. Subjective, -1. In most scenarios, there is only single invocation on mock so qualifier (last, first, etc) would clutter the most common use case for captor.

Summary of public API changes of ArgumentCaptor in this PR:

```java
//New methods
getValueAt(index)
getFirstValue()
getSecondValue()
getThirdValue()
getLastValue()

//Deprecate
getValue() /* -> */ getLastValue()
```

My feedback:
1. (-1) to getFirst/Second/Third - unnecessary
2. (-1) to deprecation 
3. (+1) getValueAt(x)

Public API I would like most:

```java
//New methods
get()
get(int)
getAll() /* or */ all()

//Deprecate (or keep as alias)
getValue() /* -> */ get()
```

Timing re deprecated method:
1. Avoid deprecation now, deprecate in Mockito 3, delete in future version giving enough time to clients to fix.

Guys, feedback/opinions about the API above?

bric3:
Actually I'd rather way on this yet. I don't agree with these APIs in context they are.

**Context :**

> I've found found `ArgumentCaptor` obscure to understand and to use regarding the its behaviour or its API, especially when varargs are involved.

* `getFirstValue() / getSecondValue() / getThirdValue() / getLastValue()` :-1: they add no real value.
* `getValueAt` is interesting however when used with varargs then users don't know if the index is the position of the value or if the index is the invocation number.

* `getValue()` mildly ok for deprecation, it suffers from the same issues mentioned above. It is not explicit which value will be returned when there's several invocation unless one read the javadoc. Yet it's not that problematic.
* `getAllValues()` :+1: for deprecation, it's the worse offender here. This methods has hidden non explicit features, it accumulates values form both vararg and invocation in a single flat list, it is impossible to distinguish them. Making impossible to assert properly values coming from this method.

`ArgumentCaptor` is ok for simple cases (only using `getValue()`) but begin to be **counter-productive** when when not so edge cases appear i.e. varargs / multiple invocations.

The refactoring (with proper deprecation) of `ArgumentCaptor` was scheduled fo 2.2 as a continuation of the work on 2.x. I would have loved user feedback on this feature in some beta form, but it's not possible with the actual release model. Given the impact of this API change I'd prefer to delay this work in 3.x now.

szczepiq:
@bric3, great points. Regardless if this is for 2.x or 3.x, what is your proposal for the API?

bric3:
@szczepiq I had made some API a while back but unfortunately I lost the fork at some point, so now it's in my head. I'll find some time to lay it out code wise.

vanniktech:
What about this now? Should I close this PR and just leave the issue open?

ChristianSchwarz:
I would like to see the getLast/getFirst/getSecond API as you implemented it, but I am no core member! 

-1 for getValueAt(int) cause it is not clear if index 0 is the last or first captured value

+1 deprecate ArgumentCaptor and provide a new API that supports AtomicReference/AtomicPRIMITVE and Collections:
  ```java
  T captureTo(Collection<T>) 
  T captureTo(AtomicReference<T>)
  PRIMITIVE captureTo(AtomicPRIMITIVE)

  //sample
 Set<String> userId;
 AtomicInteger pin;
  verify(mock).login(captureTo(userId),captureTo(pin));
  ```

TimvdLippe:
@vanniktech Since there is no concensus yet on the API, this PR became stale. In order to continue on this PR we need to resolve that.

Given the good arguments of @szczepiq and @bric3 I am now in favor of `getValueAt()`, but the point of @ChristianSchwarz 's point is fair. Therefore can we come up with a naming that is unambiguous?

vanniktech:
Gentle push.

vanniktech:
Closing since Kotlin Extensions functions can get me where I want to get. I'll leave the issue open for further discussion.

PR #476, opened 2016-07-03
refactored ThreadSafeMockingProgress to a singleton

ChristianSchwarz:
The actual implementation of [ThreadSafeMockingProgress](https://github.com/mockito/mockito/blob/master/src/main/java/org/mockito/internal/progress/ThreadSafeMockingProgress.java) disguise that it is effectively a singleton cause the ThreadLocal member is static. This makes it hard to see that different instances  share the same state.

This PR...
 * refactores `ThreadSafeMockingProgress `to provider of MockingProgress instances
 * uses the standard way of instantiating the initial `ThreadLocal `value this also avoids possible race conditions

### DIFF
--- a/src/main/java/org/mockito/internal/MockitoCore.java
+++ b/src/main/java/org/mockito/internal/MockitoCore.java
@@ -13,11 +13,11 @@ import static org.mockito.exceptions.Reporter.notAMockPassedWhenCreatingInOrder;
 import static org.mockito.exceptions.Reporter.nullPassedToVerify;
 import static org.mockito.exceptions.Reporter.nullPassedToVerifyNoMoreInteractions;
 import static org.mockito.exceptions.Reporter.nullPassedWhenCreatingInOrder;
+import static org.mockito.internal.progress.ThreadSafeMockingProgress.mockingProgress;
 import static org.mockito.internal.verification.VerificationModeFactory.noMoreInteractions;
 
 import java.util.Arrays;
 import java.util.List;
-
 import org.mockito.InOrder;
 import org.mockito.MockSettings;
 import org.mockito.MockingDetails;
@@ -25,7 +25,6 @@ import org.mockito.exceptions.misusing.NotAMockException;
 import org.mockito.internal.creation.MockSettingsImpl;
 import org.mockito.internal.invocation.finder.VerifiableInvocationsFinder;
 import org.mockito.internal.progress.MockingProgress;
-import org.mockito.internal.progress.ThreadSafeMockingProgress;
 import org.mockito.internal.stubbing.InvocationContainer;
 import org.mockito.internal.stubbing.OngoingStubbingImpl;
 import org.mockito.internal.stubbing.StubberImpl;
@@ -47,7 +46,6 @@ import org.mockito.verification.VerificationMode;
 public class MockitoCore {
 
     private final MockUtil mockUtil = new MockUtil();
-    private final MockingProgress mockingProgress = new ThreadSafeMockingProgress();
 
     public boolean isTypeMockable(Class<?> typeToMock) {
         return mockUtil.typeMockabilityOf(typeToMock).mockable();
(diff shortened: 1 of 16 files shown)


src/main/java/org/mockito/internal/progress/ThreadSafeMockingProgress.java
```diff
@@ -10,79 +10,112 @@
...
+    private  static MockingProgress SINGLETON = new Impl();
 
-    public OngoingStubbing<?> pullOngoingStubbing() {
-        return threadSafely().pullOngoingStubbing();
-    }
-    
-    public void verificationStarted(VerificationMode verify) {
-        threadSafely().verificationStarted(verify);
+    public final static MockingProgress threadSafeMockingProgress() {
+        return SINGLETON;
```
TimvdLippe: I think `SINGLETON` should be a `ThreadLocal` and this call should be `SINGLETON.get()` instead. That also gets rid of the weird `transient` call in `Impl`.

ChristianSchwarz: Thats a good idea cause this would avoid  the delegations. Unfortunatly `MockingProgressImpl` is not serializable and I found no other work around for the serialization issues.

ChristianSchwarz: @TimvdLippe 
Okay now I got it! Your idea is to use `Impl` as a serializable wrapper right?

ChristianSchwarz:
@TimvdLippe That was ticky but I found a solution. The only restriction is not to cache the MockingProgress in an instance or static field other wise thread safety get lost.

src/main/java/org/mockito/internal/progress/ThreadSafeMockingProgress.java
```diff
@@ -5,84 +5,26 @@
...
-    public void setVerificationStrategy(VerificationStrategy strategy) {
-        threadSafely().setVerificationStrategy(strategy);
-    }
-
-    public VerificationMode maybeVerifyLazily(VerificationMode mode) {
-        return threadSafely().maybeVerifyLazily(mode);
+    /**
+     * Returns the serializable {@link MockingProgress} for the current Thread.
+     * 
+     * IMPORTANT: Never cache the return {@link MockingProgress} as it might
```
TimvdLippe: Unfinished documentation?

ChristianSchwarz: uups, i fix it

ChristianSchwarz: @TimvdLippe fixed!

TimvdLippe:
I am fine with these changes, but would like to have a second opinion of either @bric3 or @raphw .

src/main/java/org/mockito/internal/progress/ThreadSafeMockingProgress.java
```diff
@@ -5,84 +5,32 @@
...
+/**
+ * Provides access to the {@link MockingProgress} of a corresponding {@link Thread}. Every {@link Thread} in Mockito has
+ * it s own {@link MockingProgress} to avoid data races while stubbing. 
+ */
+public class ThreadSafeMockingProgress {
 
-    static MockingProgress threadSafely() {
-        if (mockingProgress.get() == null) {
-            mockingProgress.set(new MockingProgressImpl());
+    private static final ThreadLocal<MockingProgress> SINGLETON = new ThreadLocal<MockingProgress>() {
```
bric3: I don't like the _singleton_ word in the context of thread, I'd prefer something more explicit `THREAD_LOCAL_MOCKING_PROGRESS`

src/main/java/org/mockito/internal/progress/ThreadSafeMockingProgress.java
```diff
@@ -5,84 +5,32 @@
...
+    /**
+     * Returns the {@link MockingProgress} for the current Thread.
+     * <p>
+     * <b>IMPORTANT</b>: Never assign and access the returned {@link MockingProgress} to an instance or static field.
+     * Thread safety can not be guaranteed in this case, cause the Thread that wrote the field might not be the same
+     * that read it. In other words multiple threads will access the same {@link MockingProgress}.
+     * 
+     * @return never <code>null</code>
+     */
+    public final static MockingProgress currentThreadMockingProgress() {
```
bric3: What do you think of just `mockingProgress()`

src/main/java/org/mockito/internal/MockitoCore.java
```diff
@@ -47,26 +46,24 @@
 public class MockitoCore {
 
     private final MockUtil mockUtil = new MockUtil();
-    private final MockingProgress mockingProgress = new ThreadSafeMockingProgress();
```
bric3: Why not using using `currentThreadMockingProgress`, it is not a static field, and MockitoCore is not serializable anyway

src/main/java/org/mockito/internal/progress/ThreadSafeMockingProgress.java
ChristianSchwarz: Yes I think it is better, I will change it!

src/main/java/org/mockito/internal/progress/ThreadSafeMockingProgress.java
ChristianSchwarz: I see your point!  The name `THREAD_LOCAL_MOCKING_PROGRESS` is redundant to to the type `ThreadLocal<MockingProgress>`, it smells. What about  `MOCKING_PROGRESS_PROVIDER`  ?

ChristianSchwarz:
@bric3 fixed the review issues

bric3:
OK looks good

bric3:
Thank you @ChristianSchwarz for this work, this is much appreciated

PR #481, opened 2016-07-06
Exclude mockito internal packages from the Javadoc

### LINKED ISSUE #226: Exclude internal package from javadoc
Internal package don't need to be included in the javadoc generation, just like the jdk doesn't have `com.sun` package

In the javadoc that may be done via a doclet since `exclude '**/internal/**'` raise javadoc errors

```groovy
doclet = "qualified.name.of.DocletExclude"
docletpath = [rootProject.file('./gradle/doclet-exclude.jar')]
```

TimvdLippe:
This one was a toughy, but I managed to write a Doclet that post-processes the generated classes. It filters all internal classes. This way, there are no compilation failures when generating the Javadoc if you would use `exclude "**/internal/**"`.

Fixes #226

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -8,6 +8,7 @@ dependencies {
     //TODO SF use jcabi to edit issues after the release so that they have the milestone attached
     //compile "com.jcabi:jcabi-github:0.17"
     compile "com.googlecode.json-simple:json-simple:1.1.1@jar"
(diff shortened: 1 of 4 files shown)


raphw:
Great, work! Could you get the code level down to Java 6?

TimvdLippe:
Ugh apparently IntelliJ reset my project settings. It should be fixed now :tada:

bric3:
Fantastic !

Issue #482, opened 2016-07-06
Restore depecated Answers.get() method


philwebb:
Commit da5e750957b494e7fa0548bf1286d67b8b0386d5 removed the deprecated `Answers.get()` method, however the method was only deprecated in the 2.0 BETA line and never made it into a GA release. It would be helpful if the method could remain in the deprecated form in 2.0 to give people an opportunity to move away from it.

(originally raise by a Spring Boot user [here](https://github.com/spring-projects/spring-boot/pull/6323))

philwebb:
On a related note adding a "as of" to the deprecated description would also be useful:

```java
/**
 * @deprecated as of 2.0 in favor of using the enum-constant directly. This method will be removed in a future release<br> 
 * E.g. instead of <code>Answers.CALLS_REAL_METHODS.get()</code> use <code>Answers.CALLS_REAL_METHODS</code> .
 */
```

szpak:
I would agree that that method had been made deprecated in 2.0 (-beta.2) not in 1.10.x (in fact at first it was [removed](https://github.com/mockito/mockito/commit/34f4436d988f04eaf7635b0497067ee8df7971a6) completely and one month later [restored](https://github.com/mockito/mockito/commit/32dce3ea56fb2891b54be48c16a13d2a9eb0cd33) as deprecated), but what is more important `Answers` started implementing `Answer` then. Because of that when I was fixing that issue in Spring Boot it was hard to keep code compatible with both versions at runtime (for 1.x casting using it as Answer does not work, for never 2.x `get()` is not available). In the end I had to [do](https://github.com/spring-projects/spring-boot/pull/6323/files) ugly casting through `Object` to fool the compiler. With that Spring Boot should work fine (at least with changes related to `Answers` :) ), but there could be other tools (PowerMock?) that after removing this method cannot be easily compatible with both Mockito 1.x and 2.x. IMHO it would be good to keep this method in 2.x and remove in 3.x.

TimvdLippe:
Yes that seems like a good idea. Meant want to open a pull request?

PR #486, opened 2016-07-07
Cleanup: removed dead/unnecessary classes

ChristianSchwarz:
Removed dead/unnecessary classes:

 * VarargCapturingMatcher and Test
 * AnyVararg removed in Favor of Any, Any can now used as varargs matcher too 
 
 removed test [shouldNotAllowAnyObjectMatchEntireVararg()](https://github.com/mockito/mockito/blob/master/src/test/java/org/mockito/internal/invocation/ArgumentsComparatorTest.java#L110) cause it was not clear why ANY should not be allowed as varargs matcher


src/main/java/org/mockito/Matchers.java
```diff
@@ -198,9 +198,12 @@ public static short anyShort() {
      * See examples in javadoc for {@link Matchers} class
      *
      * @return <code>null</code>.
+     * 
+     * @deprecated use {@link #any()}
```
TimvdLippe: Per #482 please update this to `@deprecated as of 2.0, use {@link #any()}`

TimvdLippe:
1 small documentation update, other than that LGTM

src/main/java/org/mockito/Matchers.java
ChristianSchwarz: fixed

ChristianSchwarz:
@TimvdLippe fixed

Issue #390, opened 2016-04-11
Support for JUnit 5 (JUnit Lambda)


raphw:
I think we should support JUnit 5 with Mockito 2. Currently, we maintain an optional dependency on JUnit four but the JUnit 5 API is completely incompatible to the old one. I already explored the extension mechanism a bit and for our use case, it is quite trivial.

There are some caveats with JUnit 5, for example the possibility to reuse a test instance what would require us to extend the rule for reseting mocks between runs but it is more than doable. Actually, I suggest making `@Mock` a meta annotation, this way, we can completely drop the MockitoJUnitRunner runner and the MockitoJUnit rule and make the setup even easier.

The JUnit 5 API is still in alpha. How do we do this?

szpak:
JUnit 5 seems to support only Java 8. Is its API Java 7 compatible? Or maybe you meant Mockito 3 (incoming just after Mockito 2, but with official Java 8 support :) )?

raphw:
That is why I would suggest to put it into a side-project. JUnit 5 is compiled to Java 8 byte code but we could deliver the integration in Java 6 nevertheless.

ChristianSchwarz:
JUnit 5 looks very promising, extensions replacing [Runner, Rule & ClassRule](http://junit.org/junit5/#extension-model). 

Surprise surprise :sparkles: , here is a sample -> the [MockitoExtension](https://github.com/junit-team/junit5-samples/blob/master/junit5-mockito-extension/src/main/java/com/example/mockito/MockitoExtension.java)

TimvdLippe:
While I do think it is worth exploring, in my eyes pushing this to Mockito 3.0 with Java 8 seems more safe. This also gives time for the JUnit team to tinker everything and push to beta, to reduce the compatibility burden for jumping on the train too soon.

TimvdLippe:
Merging this into #445

Issue #401, opened 2016-04-23
JUnit runner detects unused stubs


szczepiq:
It is useful to know that stubbing in the test was not used. Perhaps the stubbing is not needed? Removal of unnecessary code from the test is important for code clarity.

More details:
- we detect unused stubs only for passing tests (adding different failure to an existing failure could be confusing)
- need to have something similar for JUnit rules (will create separate ticket)
- stubbing in @Before or class constructor should be used in at least one test
- stubbing in test needs to be used in that test

szczepiq:
The current impl is not thread safe. Reopening.

Issue #384, opened 2016-04-07
JUnit rule logs warnings about unsued / misused stubs


szczepiq:
### Why

When test fails, the failure might be caused by misused stubs. Hence, it might be worth to log out debugging information to the System out (e.g. misused stubs, unused stubs).

### Plan

- [ ] document 2.x change
- [x] tweak the warning message
- [ ] add behavior to the runner
- [ ] add silent runner
- [ ] add silent setting to the rule (or warnings level for all / exception only / none)
- [ ] deprecate/remove console spamming runner

### Impl

The JUnit rule or the runner will potentially include following info in the output (somehow): https://gist.github.com/szczepiq/38619fd8766c66c46dff4dc2b6db06ec

I'm wondering if "Mockito.validateMockitoUsage();" should automatically print warnings. We could provide boolean parameter to control printing of the warnings.

szczepiq:
On it.

ChristianSchwarz:
I like the feature! What do you think about a `rule().strict()` that doesn't emit logs but fails asap?

A note to the API:
 * How do we prevent misuse like -> `rule().strict().strict()` ? What about `rule(LOG_WARNINGS)` aka `rule()`, `rule(SILENT)`, `rule(FAIL_FAST)`

szczepiq:
Christian, very cool ideas!!! 

I'm not sure I'll have time for it before 2.0. However, I want the new default behavior in rule / runner in 2.0 so that users see what it is doing. Then we can grow and add strict mode.

t1:
@szczepiq: I like this feature very much. Also that the `MockitoJUnitRunner` fails by default is good. I think a note in the javadoc of `UnnecessaryStubbingException` to use `MockitoJUnitRunner.Silent` would be helpful.

We often have quite complex test fixtures, and it's cumbersome to switch absolutely each and every stub on or off... some methods simply have to be there; it's also not necessary to verify if they've been invoked. So it would be great to be able to disable it for some stubs, but leave it on for others.

TimvdLippe:
@t1 That seems like a reasonable idea. Want to make a PR for that?

szczepiq:
Good feedback. I think that MockitoHint class should also mention that it is possible to silence the new mechanism.

szczepiq:
#609

Sarev0k:
[...]

PR #491, opened 2016-07-10
Deprecate whitebox and corresponding verboserunner and junitfailureha…

### LINKED ISSUE #489: Remove Whitebox class
The original question stems from #422 where additional changes were requested. This class seems to stimulate bad testing practices. Mockito only uses this class in [JUnitFailureHacker](https://github.com/mockito/mockito/blob/196ff979da156caa07e19f57e4849637d8bede1a/src/main/java/org/mockito/internal/util/junit/JUnitFailureHacker.java) which consequently is only used in [VerboseMockitoJUnitRunner](https://github.com/mockito/mockito/blob/196ff979da156caa07e19f57e4849637d8bede1a/src/main/java/org/mockito/runners/VerboseMockitoJUnitRunner.java). Given the nature of this class and only 1 usage in the library, I think we should remove it to prevent users from obtaining bad testing habits.

TimvdLippe:
A quick google showed that there are almost zero usages of this runner. Moreover it was relying on a very hackish implementation. I think we can do a better job by creating a new runner with JUnit 5 as part of #390. Therefore the intent is to deprecate these classes and remove them in Mockito 3.0.

Fixes #489


bric3:
Don't remove this runner, a few people were relying on this. It may not show up on open source project.

In my opinion `Whitebox` should be removed though. So it could be possible to use other internal reflection utils, like `FieldGetter`

src/main/java/org/mockito/runners/VerboseMockitoJUnitRunner.java
```diff
@@ -34,7 +34,11 @@
  * Do you think it is useful or not? Drop us an email at mockito@googlegroups.com
  * <p>
  * Experimental implementation - will change in future!
+ *
+ * @deprecated as of 2.0. Runner will likely be replaced with a JUnit 5 compliant runner. This runner relied
+ * on a hacky implementation as was not used widely.
```
Xaerxess: "as" -> "and"?

bric3: Also sync with @szczepiq on this

ChristianSchwarz:
@TimvdLippe @bric3 
What do you think about deprecating all Mockito-Runners in order to promote the uses of `MockitoRule`? 

Motivation: 
 * Mockito should promote only one API to initialize annotated fields, cause it may confuse users what to choose when (Runners vs Rule). 
 * `MockitoRule `does the same as the Runners but is more flexible cause it allows to uses other Runners e.g. for parameterized tests which is a often requested feature

raphw:
I do not think we should deprecate them. JUnit (4) still has runners as a non-deprecated feature. We should deprecate the runners once they do the same. A lot of people still rely on old JUnit versions where runners are the only option. We can however add proper documentation that indicates that runners should be the second choice.

ChristianSchwarz:
@raphw 
Even if I disagree at some points I accept the core-team decision. 

> A lot of people still rely on old JUnit versions where runners are the only option.

I see you point, but i think that people that update to mockito 2.x can also update there junit version.

> We can however add proper documentation that indicates that runners should be the second choice.

Good idea, here we go #531

TimvdLippe:
The reason to deprecate this runner is because the new Mockito runner is strict and is a lot more userfriendly than this runner. To promote the strict runner, we should deprecate this one and use the strict version imo.

Regarding runners vs rules: we can state our preference, but we should keep the possibility.

szczepiq:
I'll tidy up the Verbose runner documentation. I don't think it is useful given the work on #384 and #401.

bric3:
OK deprecation sounds good to me, however I'd remove `Whitebox it's internal after all.

Let's just replace the methods in `JUnitFailureHacker` at [line 19](https://github.com/mockito/mockito/pull/491/files#diff-a041e60ee2d1a736352af1c34c19b1f4R19) and at [line 26](https://github.com/mockito/mockito/pull/491/files#diff-a041e60ee2d1a736352af1c34c19b1f4R26) by private methods.

src/main/java/org/mockito/runners/VerboseMockitoJUnitRunner.java
TimvdLippe: Documentation updated :smile:

TimvdLippe:
Rebased and updated pull request with feedback. PTAL :smile:

src/main/java/org/mockito/internal/util/junit/JUnitFailureHacker.java
```diff
@@ -6,26 +6,73 @@
 
 import org.junit.runner.notification.Failure;
 import org.mockito.internal.exceptions.ExceptionIncludingMockitoWarnings;
-import org.mockito.internal.util.reflection.Whitebox;
 
+import java.lang.reflect.Field;
+
+@Deprecated
```
ChristianSchwarz: It should be documented what should be used instead.

TimvdLippe: It's an internal class and is only used in the VerboseMockitoRunner. Therefore I do not think documentation is required.

src/main/java/org/mockito/internal/util/junit/JUnitFailureHacker.java
```diff
@@ -6,26 +6,73 @@
...
+        } catch (Exception e) {
+            throw new RuntimeException("Unable to get internal state on a private field. Please report to mockito mailing list.", e);
+        }
+    }
+
+    private static void setInternalState(Object target, String field, Object value) {
+        Class<?> c = target.getClass();
+        try {
+            Field f = getFieldFromHierarchy(c, field);
+            f.setAccessible(true);
```
ChristianSchwarz: @TimvdLippe Does it make sense to uses the AccessabilityChanger here?

src/main/java/org/mockito/runners/VerboseMockitoJUnitRunner.java
```diff
@@ -34,7 +34,10 @@
  * Do you think it is useful or not? Drop us an email at mockito@googlegroups.com
  * <p>
  * Experimental implementation - will change in future!
+ *
+ * @deprecated as of 2.0. Use the new {@link org.mockito.runners.MockitoJUnitRunner.Strict} runner instead.
```
ChristianSchwarz: @TimvdLippe cosmetics: 2.0.0 is used in other places, it should be used here too

src/main/java/org/mockito/internal/util/junit/JUnitFailureHacker.java
TimvdLippe: It's more code with object creation, so I think this one-liner is better.

src/main/java/org/mockito/runners/VerboseMockitoJUnitRunner.java
TimvdLippe: Done

src/main/java/org/mockito/internal/util/junit/JUnitFailureHacker.java
bric3: And especially for deprecated code

bric3:
Indeed the timeout test is picky, especially with openjdk 6, I restarted the job 2 times

szczepiq:
Verbose runner and console spamming runner are now both deprecated. Let's kill them in 3.0 :)

PR #493, opened 2016-07-10
Code cov on releases

### LINKED ISSUE #465: Delivery drone breaks codecov
Since the commits of the delivery drone have `[ci skip]`, there is no coverage report for that specific commit. As a result, when a pull request has as base such a commit, CodeCov reports that it could not generate a coverage report.

Maybe we can depend executing `gradle release` if the commit is from a user other than the delivery drone, to not break codecov.

TimvdLippe:
Releases are now triggering CI builds to generate the code coverage reports.
An example of a normal build can be found at https://travis-ci.org/mockito/mockito/builds/143765346#L334-L374
An example of a build ignored as the commit message contained `[ci skip-release]` can be found at https://travis-ci.org/mockito/mockito/builds/143765834#L311-L314

Fixes #465

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,9 @@ branches:
 install:
  - true
 script:
-  - ./gradlew ciBuild release
+  - TRAVIS_MESSAGE="$(git log --format=%B -n 1)"
+  - ./gradlew ciBuild
+  - if [[ $TRAVIS_MESSAGE != *"[ci skip-release]"* ]] ; then ./gradlew release; fi
 after_success:
   - ./gradlew --stacktrace coverageReport && cp build/reports/jacoco/mockitoCoverage/mockitoCoverage.xml jacoco.xml || echo "Code coverage failed"
   - bash <(curl -s https://codecov.io/bash) || echo "Codecov did not collect coverage reports"
--- a/dummy-commit.txt
+++ b/dummy-commit.txt
@@ -1,1 +1,1 @@
-Change this if you need a dummy commit. Push build.
+Change this if you need a dummy commit. Push build
--- a/gradle/release.gradle
+++ b/gradle/release.gradle
@@ -86,7 +86,7 @@ release {
 
 releaseSteps {
     String currentVersion = project.version //the version loaded when Gradle build has started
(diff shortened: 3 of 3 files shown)


.travis.yml
```diff
@@ -39,7 +39,9 @@ branches:
 install:
  - true
 script:
-  - ./gradlew ciBuild release
+  - TRAVIS_MESSAGE="$(git log --format=%B -n 1)"
+  - ./gradlew ciBuild
+  - if [[ $TRAVIS_MESSAGE != *"[ci skip-release]"* ]] ; then ./gradlew release; fi
```
bric3: This can be a _oneliner_ I believe 

```
if [[ $(git log --format=%B -n 1) != *"[ci skip-release]"* ]] ; then ./gradlew release; fi
```

bric3:
From the changes I don't understand why the code coverage is mentioned in this PR. Code coverage should be triggered by any travis build.

TimvdLippe:
Yes it is triggered on every Travis build. However, as explained in #465 the delivery drones creates a release commit with ci skip. Therefore Travis is not triggered and codecov fails, see the first comment on this PR.

To prevent a release loop on travis, instead of a full ci skip it only does a ci skip-release. This is handled in the .travis.yml.

.travis.yml
TimvdLippe: The git log command was from a travis issue as there is an intent to add the new environment variable. The command is kinda magic, so naming the variable makes it imo easier to comprehend.

raphw:
Is this PR still relevant? If yes, is it ready to be merged?

TimvdLippe:
Yes it is still relevant. Releases do not trigger code cov and therefore result in empty comparisons.

TimvdLippe:
I will merge it, if it seems to break stuff (I do not expect it to) then I will revert.

mureinik:
@TimvdLippe continuing our discussion on PR #980 - can this fix be backported to `release/2.x`? IIUC, the majority of the work is done there, and CodeCov will be very helpful there too.

TimvdLippe:
@mureinik Back then we did not divert yet with `release/2.x` so I think it recently regressed in `release/2.x` instead. Have to investigate :(

PR #495, opened 2016-07-11
Moves Reporter friendly exception factory to internal package

bric3:
The `Reporter` friendly exception factory class lies in `org.mockito.exceptions.Reporter`, however it's API is subject to change without really impacting user experience.

Thus, I propose to move `org.mockito.exceptions.Reporter` to `org.mockito.internal.exceptions.Reporter`.

I didn't provided a dummy class for two reasons :

1. I would like to schedule this for Mockito 2 a major version
2. This class shouldn't be used outside framework development


TimvdLippe:
LGTM but also requires a rebase

PR #496, opened 2016-07-11
Renames Matchers to ArgumentMatchers to avoid name clash with Hamcrest Matchers class

bric3:
Renames `Matchers` to `ArgumentMatchers` to avoid name clash with Hamcrest `org.hamcrest.Matchers` class.

This plagued usage of mockito and hamcrest together since the introduction of matchers. The work in Mockito 2 beta allowed to avoid relying on hamcrest internally and thus split the Hamcrest matcher factory `MockitoHamcrest`. Yet this is not enough as using both mockito matchers and hamcrest matchers in the same class, made the developer life not comfortable when using both API.

This change helps in that regard, as it renames the import class. Not this change tweaks only the class hierarchy, so current code still compiles, thanks to the javac compiler that can lookup static method in parent classes. (That is why I didn't updated tests yest, they could be upgraded in a late PR, a good timing would probably be when `Matchers` is removed)

`Matchers` class can be removed in the next version of mockito, likely 3.0.


bric3:
Note I'm open on the name of `ArgumentMatchers`.

TimvdLippe:
LGTM!

TimvdLippe:
Ah but it requires a rebase :sob:

PR #499, opened 2016-07-23
Fixes #401 - makes the JUnit runner detect unused stubs

### LINKED ISSUE #401: JUnit runner detects unused stubs

szczepiq:
Fixes #401

This is a breaking change because JUnit runner will now fail the test class if there is an unused stubbing.

Feedback is welcome. Sorry for working on it for a long time and big number of commits.


szczepiq:
investigating the failure on travis. Works for me locally! :D

src/main/java/org/mockito/internal/progress/MockingProgressImpl.java
```diff
@@ -143,4 +146,18 @@ public void setVerificationStrategy(VerificationStrategy strategy) {
...
     }
-}
+
+    public void setStubbingListener(StubbingListener stubbingListener) {
+        this.stubbingListener = stubbingListener;
+    }
+
+    public StubbingListener getStubbingListener() {
+        if (this.stubbingListener == null) {
+            return new StubbingListener() {
```
TimvdLippe: I would rather set this in the constructor of this class, to prevent needless if checks if there are no listeners (which is the majority of the cases)

src/test/java/org/mockitousage/matchers/InvalidUseOfMatchersTest.java
```diff
@@ -6,23 +6,21 @@
...
 import org.mockitousage.IMethods;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.fail;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.when;
 
-@RunWith(MockitoJUnitRunner.class)
+//TODO make this working with MockitoJUnitRunner with opt-out unnecessary stubs validation
+//@RunWith(MockitoJUnitRunner.class)
```
TimvdLippe: This can be the silent version right?

TimvdLippe:
I would rather not default to strict right away. This will break a lot of user test suites which significantly increasing the upgrade process to Mockito 2.0.

If we let developers know our intention is to makr the strict version default later, they can slowly make the changes for each of their runners. Then we can change the default in Mockito 3.0 so developers had some time to adopt.

szczepiq:
Yes, we can change the default if that's the decision!

Personally, I think we should go ahead with this change at the cost of upgrade pain. Why?

- it actually helps the codebase - dead code removed, tests are cleaner. Arguably, it could make the users _want_ to upgrade because they can make the codebase better. It's nice if the upgrade effort actually brings immediate value to the codebase.
- because of above, it could be easier to justify the upgrade to the team / manager.
- it shouldn't be that much work. I doubt that users have that many unused stubs. In Mockito codebase (huge amount of mockito tests) we had just few examples.
- not changing the default will make this feature not used by most users because most devs don't care unless forced to care :P

Thoughts?

src/test/java/org/mockitousage/matchers/InvalidUseOfMatchersTest.java
szczepiq: great idea

src/main/java/org/mockito/internal/progress/MockingProgressImpl.java
szczepiq: great idea

TimvdLippe: Actually, there might be a case of multiple listeners being attached (just like with observers) so lets turn this into a list? Then it can be initialized on the field as empty arraylist and a simple for loop to process all listeners. Later wirh streams it can be enhanced to provide a function callback and turn it into a oneliner.

PascalSchumacher:
+1 for making it the default

Cleaner test because of removed dead code are a big plus. I guess it should not be much work to delete some code.

szczepiq:
Thanks for feedback! If there is no more feedback, I'm merging it this weekend.

src/main/java/org/mockito/internal/progress/MockingProgressImpl.java
szczepiq: great idea. Single listener is simpler, so I would go ahead and merge it as is. However, I think a list of listeners makes much more sense. So I would just do it sometime before 2.0 :)

src/main/java/org/mockito/internal/creation/bytebuddy/InterceptedInvocation.java
```diff
@@ -129,6 +129,7 @@ public Object callRealMethod() throws Throwable {
 
     @Override
     public int hashCode() {
+        //TODO SF we need to provide hash code implementation so that there are no unexpected, slight perf issues
```
bric3: hashcode without equals is wrong.
For the TODO the hashCode can be lazy

src/main/java/org/mockito/internal/runners/StrictRunner.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private final Class<?> testClass;
+    private final RunnerImpl runner;
+    private boolean filterRequested;
+
+    /**
+     * @param runner - the runner to wrap around
+     * @param testClass - for reporting purposes
+     */
+    public StrictRunner(RunnerImpl runner, Class<?> testClass) {
```
bric3: TODO cleanup dead javadoc

bric3:
Late in the review game, but nice feature.

Issue #426, opened 2016-06-11
Refactor instance based utility classes to static utility classes


ChristianSchwarz:
The internal implementation of Mockito contains numerous helper classed that serve as a container for useful methods. By its nature these classes don't have a state. Currently some them are instance based in other word you need to create the utility class to call a helper method. This not only pollutes the heap but also the code cause instance methods can't be imported statically. 

Here is an example:

```java
Helper helper = new Helper();
if (helper.isInputValid(input)){
  [...] 
}
```

vs. static import of Helper.isInputValid

```java
if (isInputValid(input)){
  [...] 
}
```

The aim of this ticket is to identify canidates that can be refactored to static utility classes. If you like to refactoring and mockito feel free to send a PR and reference this issue.

Refactoring canidates:
- [ ] `AccessibilityChanger`
- [ ] `BeanPropertySetter`
- [ ] `ConditionalStackTraceFilter`
- [ ] `FieldCopier`
- [ ] `FieldReader`
- [ ] `GenericMaster` should be integrate into `GenericTypeResolver`
- [ ] `JUnitFailureHacker` can be removed when the deprecated `VerboseMockitoJUnitRunner` is removed
- [ ] `LenientCopyTool`
- [ ] `MatcherBinder`
- [ ] `MockitoCore` should better be a singleton
- [ ] `MockCreationValidator`
- [ ] `RemoveFirstLine`
- [x] #591 `ArgumentMatchingTool`
- [x] #515 `AllInvocationsFinder`
- [x] #502 `ArgumentsComparator`
- [x] #540 `ArrayUtils`
- [x] #490 `AtLeastXNumberOfInvocationsChecker`
- [x] #490 `AtLeastXNumberOfInvocationsInOrderChecker`
- [ ] #912 `Constructors`
- [x] #427 `FieldSetter`
- [x] #908 `FriendlyExceptionMaker`
- [x] #431 `HandyReturnValues` 
- [x] #432 `InvocationMarker`
- [x] #462 `InvocationsFinder`
- [x] #908 `JUnitDetecter`
- [x] #490 `MissingInvocationChecker`
- [x] #490 `MissingInvocationInOrderChecker`
- [x] #514 `MockUtil`
- [x] #503 `NonGreedyNumberOfInvocationsInOrderChecker`
- [x] #907 `NumberOfInvocationsInOrderChecker`
- [x] #907 `NumberOfInvocationsChecker`
- [x] #547 `ObjectMethodsGuru`
[...]

PR #501, opened 2016-07-24
refactored TestMethodsFinder to a static utility class (#426)

PascalSchumacher:


src/main/java/org/mockito/internal/runners/util/TestMethodsFinder.java
```diff
@@ -9,7 +9,7 @@
 import java.lang.reflect.Method;
 
 public class TestMethodsFinder {
-    public boolean hasTestMethods(Class<?> klass) {
```
ChristianSchwarz: Can you add a private constructor here?

TimvdLippe:
LGTM

src/main/java/org/mockito/internal/runners/util/TestMethodsFinder.java
PascalSchumacher: Sure. I have updated the pull request.

PR #505, opened 2016-07-24
removed useless empty string at the end of exception message in ByteBuddyMockMaker

PascalSchumacher:


TimvdLippe:
Hm I think this is necessary because it requires a new line for the next console message?

PascalSchumacher:
I now that you mention it I check the code for other usages of `MockitoException` with mulitline messages. Some have the blank line a the end of the message and some do not.

I have never seen this before. Shouldn't it be left up to decide to whatever is outputting the messages how to display them?

bric3:
The new line is to separate the message from the stack, this is purely _UI_.

PR #510, opened 2016-07-28
Fixes #194 tweaks any matchers

### LINKED ISSUE #194: Tweaks Matchers.any family matchers behavior

bric3:
Details is #194 

I took the opportunity to 

* replace `TestBase.assertThat` by more powerful AssertJ equivalent. There's still some work to be done in this class.
* add iterable matchers

Before merging

* [ ] I ponder to make `anyObject()` not to accept `null`, currently `any()` is an alias of `anyObject()`
* [x] I need to update `anyInt` family description.


bric3:
Here's some poll on twitter :

https://twitter.com/BriceDutheil/status/758967944881639424
https://twitter.com/BriceDutheil/status/758969135489388544

bric3:
30 votes and I merge the PR, I'll monitor the poll after my holidays

Stephan202:
If I understand correctly `List<T> anyListOf(Class<T> clazz)` was deprecated in favour of `List anyList()`. Shouldn't the latter then be made generic? When I replace one with the other I get "unchecked conversion" warnings.

bric3:
@Stephan202 thanks for the feedback, yes it was deprecated for people that use Java8 as documented in the javadoc. And the deprecation announces removal in Mockito 3 which will be Java 8 minimum.

Stephan202:
@bric3: I am on Java 8 and did migrate as suggested. I agree this is a change for the better :).

My point is that this replaces a deprecation warning with an unchecked conversion warning. The question is, shouldn't the following type of change be applied to all aforementioned methods?
```diff
--- a/src/main/java/org/mockito/ArgumentMatchers.java
+++ b/src/main/java/org/mockito/ArgumentMatchers.java
@@ -474,9 +474,9 @@ public class ArgumentMatchers {
      * @see #isNull()
      * @see #isNull(Class)
      */
-    public static List anyList() {
+    public static <T> List<T> anyList() {
         reportMatcher(new InstanceOf(List.class, "<any List>"));
-        return new ArrayList(0);
+        return new ArrayList<T>(0);
     }
```

bric3:
> shouldn't the following type of change be applied to all aforementioned methods?

What do you mean by that ?

> this replaces a deprecation warning with an unchecked conversion warning

Do you mean that IDEs show an unchecked conversion when using `anyList()`

Stephan202:
> > shouldn't the following type of change be applied to all aforementioned methods?

> What do you mean by that ?

I'm not sure what more to explain: the diff shows the addition of a generic type argument to `ArgumentMatchers#anyList`. I propose this change is made to said method, as well as `ArgumentMatchers#{anySet,anyMap,anyCollection,anyIterable}`.

> > this replaces a deprecation warning with an unchecked conversion warning

> Do you mean that IDEs show an unchecked conversion when using anyList()

My compiler does, and we compile with `-Werror`. Sure, I can suppress the warning on our side, but the changes I'm proposing is in line with `ArgumentMatchers#any`, so it seems to "make sense".

bric3:
Oh thanks it seems I forgot to stash these lines. I thought it was in the commit !
I'm fixing that right now.

Stephan202:
Cool :)

bric3:
See #528

PR #513, opened 2016-07-29
replace TestBase#assertContainsIgnoringCase(String sub, String string…

PascalSchumacher:
…) with AssertJ #containsIgnoringCase

Other methods of TestBase could be replaced in a similar way.


szpak:
LGTM

bric3:
:+1:

PR #515, opened 2016-07-29
refactored AllInvocationsFinder and VerifiableInvocationsFinder to st…

PascalSchumacher:
…atic utility classes (#426)


ChristianSchwarz:
@PascalSchumacher 
These 2 classes have only one public method, they can be merged into:
* `VerificationFinder::findVerified(..)` /  `::findAll(..)`

What do you think?

PascalSchumacher:
@ChristianSchwarz Sure

PR #516, opened 2016-07-29
replace TestBase#assertNotContains(String sub, String string) with As…

PascalSchumacher:
…sertJ #doesNotContain(CharSequence sequence)


bric3:
LGTM

PR #517, opened 2016-07-30
Make ciBuild depends also on subprojects state

szpak:
Before those changes CI build was silently ignoring compilation erros in Mockito submodules. Because of that `gw check` was failing locally (e.g. due to missing `assertNotContainsError()`).

In addition to enable building those modules (I don't know what are the plans about testng module by the way) I fixed Java 6 compatibility in them.

### DIFF
--- a/subprojects/testng/src/main/java/org/mockito/testng/MockitoBeforeTestNGMethod.java
+++ b/subprojects/testng/src/main/java/org/mockito/testng/MockitoBeforeTestNGMethod.java
@@ -15,7 +15,7 @@ import static org.mockito.internal.util.reflection.Fields.annotatedBy;
 
 public class MockitoBeforeTestNGMethod {
 
-    private final WeakHashMap<Object, Boolean> initializedInstances = new WeakHashMap<>();
+    private final WeakHashMap<Object, Boolean> initializedInstances = new WeakHashMap<Object, Boolean>();
 
     /**
      * Initialize mocks.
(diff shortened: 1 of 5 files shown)


subprojects/testng/src/main/java/org/mockito/testng/MockitoBeforeTestNGMethod.java
```diff
@@ -15,7 +15,7 @@
 
 public class MockitoBeforeTestNGMethod {
 
-    private final WeakHashMap<Object, Boolean> initializedInstances = new WeakHashMap<>();
+    private final WeakHashMap<Object, Boolean> initializedInstances = new WeakHashMap<Object, Boolean>();
```
marcingrzejszczak: Why?

marcingrzejszczak:
LGTM (I had one question though)

subprojects/testng/src/main/java/org/mockito/testng/MockitoBeforeTestNGMethod.java
szpak: Because of that: https://github.com/mockito/mockito/blob/master/build.gradle#L14-L15

szpak: Before those changes the modules were not built on Travis at all.

PascalSchumacher:
+1 and sorry for breaking the build :(

szpak:
@PascalSchumacher In fact it was a good thing. Without that we wouldn't know that submodules are skipped :).

PR #519, opened 2016-07-30
replace TestBase#assertContains(String sub, String string) with Asser…

PascalSchumacher:
…tJ #contains and #hasMessageContaining

### DIFF
--- a/src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsSmartNullsTest.java
+++ b/src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsSmartNullsTest.java
@@ -4,6 +4,8 @@
  */
 package org.mockito.internal.stubbing.defaultanswers;
 
+import static org.assertj.core.api.Assertions.assertThat;
+
 import org.junit.Test;
 import org.mockito.exceptions.verification.SmartNullPointerException;
 import org.mockito.stubbing.Answer;
@@ -51,8 +53,9 @@ public class ReturnsSmartNullsTest extends TestBase {
 
         Foo smartNull = (Foo) answer.answer(invocationOf(Foo.class, "get"));
 
-        assertContains("SmartNull returned by", smartNull + "");
-        assertContains("foo.get()", smartNull + "");
+        assertThat(smartNull.toString())
+            .contains("SmartNull returned by")
+            .contains("foo.get()");
     }
 
     @Test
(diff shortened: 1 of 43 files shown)


src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsSmartNullsTest.java
```diff
@@ -51,8 +53,9 @@ public void should_return_an_object_that_allows_object_methods() throws Throwabl
 
         Foo smartNull = (Foo) answer.answer(invocationOf(Foo.class, "get"));
 
-        assertContains("SmartNull returned by", smartNull + "");
-        assertContains("foo.get()", smartNull + "");
+        assertThat(smartNull.toString())
```
szpak: I wonder if a risk of NPE is real in those tests. With `smartNull + ""` the nicer error message would be displayed when `smartNull` is `null`.

src/test/java/org/mockitousage/stacktrace/ClickableStackTracesTest.java
```diff
@@ -34,8 +35,7 @@ public void shouldShowActualAndExpectedWhenArgumentsAreDifferent() {
             verifyTheMock(1, "not foo");
             fail();
         } catch (ArgumentsAreDifferent e) {
-            assertContains("callMethodOnMock(", e.getMessage());
-            assertContains("verifyTheMock(", e.getMessage());
+            assertThat(e).hasMessageContaining("callMethodOnMock(").hasMessageContaining("verifyTheMock(");
```
szpak: To keep consistency the second `hasMessageContaining` should be in the another line.

src/test/java/org/mockitousage/verification/DescriptiveMessagesWhenVerificationFailsTest.java
```diff
@@ -138,8 +138,7 @@ public void should_print_actual_and_wanted_when_actual_method_name_and_wanted_me
             verify(mock).simpleMethod(10);
             fail();
         } catch (ArgumentsAreDifferent e) {
-            assertContains("simpleMethod(10)", e.getMessage());
-            assertContains("simpleMethod()", e.getMessage());
+            assertThat(e).hasMessageContaining("simpleMethod(10)").hasMessageContaining("simpleMethod()");
```
szpak: The same here.

szpak:
LGTM.

It was a very ungrateful work, thanks Pascal!

src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsSmartNullsTest.java
bric3: For me that's ok, the NPE is straightforward in this case.

PR #523, opened 2016-07-31
made some timeouts in VerificationAfterDelayTest and VerificationWith…

PascalSchumacher:
…TimeoutTest more lenient to prevent frequent build failures on travis ci


bric3:
Good idea thanks!

PR #524, opened 2016-07-31
replace TestBase#assertContains(String sub, String string) with Asser…

PascalSchumacher:
…tJ #contains and #hasMessageContaining

include feedback from review of #519

### DIFF
--- a/src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsSmartNullsTest.java
+++ b/src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsSmartNullsTest.java
@@ -64,7 +64,7 @@ public class ReturnsSmartNullsTest extends TestBase {
 
         Foo smartNull = (Foo) answer.answer(invocationOf(Foo.class, "withArgs", "oompa", "lumpa"));
 
-        assertThat(smartNull.toString())
+        assertThat(smartNull + "")
             .contains("foo.withArgs")
             .contains("oompa")
             .contains("lumpa");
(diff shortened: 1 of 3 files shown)


szpak:
LGTM

src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsSmartNullsTest.java
```diff
@@ -64,7 +64,7 @@ public void should_print_the_parameters_when_calling_a_method_with_args() throws
 
         Foo smartNull = (Foo) answer.answer(invocationOf(Foo.class, "withArgs", "oompa", "lumpa"));
 
-        assertThat(smartNull.toString())
+        assertThat(smartNull + "")
```
raphw: I do not like this change. This code is less clear than calling toString.

raphw:
I do not think that this PR really improves anything. It only adds line breaks and obfuscates a toString call.

src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsSmartNullsTest.java
szpak: I proposed that change to the previous PR (that was merged in the meantime - preventing Pascal to amend it). `smartNull + ""` is more tricky, but provides a little bit better error message in case of `smartNull` being `null`. However, it is a matter of taste, so I will not insist on a change if you don't like it.

PR #527, opened 2016-08-02
Fixes #384 - JUnit rule provides stubbing hint to help debugging tests

### LINKED ISSUE #384: JUnit rule logs warnings about unsued / misused stubs

szczepiq:
Recreated the branch / recreated the PR. This is ready, I want to improve the docs (MockitoHint) before merging.

See the use case in issue #384 

I'm growing the StubbingListener and getting JUnit rule take advantage of it to report hints to the user that hopefully help debugging tests.


szczepiq:
If there is no feedback, I'll merge in the next 48 hours.

src/main/java/org/mockito/internal/junit/JUnitRule.java
```diff
@@ -14,25 +13,74 @@
...
-                try {
-                    base.evaluate();
-                } catch(Throwable t) {
-                    logger.log(c.getWarnings());
-                    throw t;
-                }
-                Mockito.validateMockitoUsage();
+        if (silent) {
+            return new SilentStatement(target, base);
+        } else {
```
ChristianSchwarz: the else-block is unnecessary here and can be removed

src/main/java/org/mockito/internal/junit/JUnitRule.java
```diff
@@ -14,25 +13,74 @@
 public class JUnitRule implements MockitoRule {
 	
 	private final MockitoLogger logger;
+    private final boolean silent;
 
-	public JUnitRule(MockitoLogger logger) {
+    public JUnitRule(MockitoLogger logger, boolean silent) {
```
ChristianSchwarz: At least the parameter `silent` should be documented. It is not clear at this point what effect the argument has.

src/main/java/org/mockito/internal/junit/JUnitRule.java
```diff
@@ -14,25 +13,74 @@
...
+        private final Object target;
+        private final Statement base;
+
+        public SilentStatement(Object target, Statement base) {
+            this.target = target;
+            this.base = base;
+        }
+
+        public void evaluate() throws Throwable {
+            MockitoAnnotations.initMocks(target);
```
ChristianSchwarz: this can be a static import

src/test/java/org/mockitousage/junitrule/StubbingWarningsJUnitRuleTest.java
```diff
@@ -0,0 +1,257 @@
+package org.mockitousage.junitrule;
```
ChristianSchwarz: Why is this test class in the `mockitousage` package? Isn't it a unit test for JUnitRule?

src/test/java/org/mockitousage/junitrule/StubbingWarningsJUnitRuleTest.java
```diff
@@ -0,0 +1,257 @@
...
+    private FrameworkMethod dummy = mock(FrameworkMethod.class);
+
+    @After public void after() {
+        //so that the validate framework usage exceptions do not collide with the tests here
+        resetState();
+    }
+
+    @Test
+    public void no_unused_stubs_reported_on_failure() throws Throwable {
+        try {
```
ChristianSchwarz: Better use the Rule [ExpectedException ](http://junit.org/junit4/javadoc/4.12/org/junit/rules/ExpectedException.html)to verify the expected exception.

src/test/java/org/mockitousage/junitrule/StubbingWarningsJUnitRuleTest.java
```diff
@@ -0,0 +1,257 @@
...
+                    declareStubbing(mock);
+                    throw new AssertionError("x");
+                }
+            }, dummy, new DummyTestCase()).evaluate();
+
+            //then
+            fail();
+        } catch (AssertionError e) {
+            assertEquals("x", e.getMessage());
+            assertTrue(logger.getLoggedInfo().isEmpty());
```
ChristianSchwarz: Here you test 2 different things, they should be seperated in 2 test methods.

src/test/java/org/mockitousage/junitrunner/StrictRunnerTest.java
```diff
@@ -55,6 +58,17 @@
         JUnitResultAssert.assertThat(result).fails(1, MyAssertionError.class);
     }
 
+    @Test public void runner_can_coexists_with_rule() {
```
ChristianSchwarz: I have a different opinion on this. I would like to see mockito fail here. It indicates that the dev made something unintended and it is easy to fix.

src/test/java/org/mockitousage/junitrule/StubbingWarningsJUnitRuleTest.java
```diff
@@ -0,0 +1,257 @@
...
+                filterLineNo(logger.getLoggedInfo()));
+    }
+
+    @Test
+    public void warns_about_unused_stubs_when_passed() throws Throwable {
+        jUnitRule.apply(new Statement() {
+            public void evaluate() throws Throwable {
+                IMethods mock = mock(IMethods.class);
+                declareStubbing(mock);
+            }
```
ChristianSchwarz: This anonymous `Statement` should be moved to a nested type or method cause it is duplicated.

src/test/java/org/mockitousage/junitrule/StubbingWarningsJUnitRuleTest.java
```diff
@@ -0,0 +1,257 @@
...
+
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class StubbingWarningsJUnitRuleTest extends TestBase {
+
+    private SimpleMockitoLogger logger = new SimpleMockitoLogger();
+    private JUnitRule jUnitRule = new JUnitRule(logger, false);
+    private FrameworkMethod dummy = mock(FrameworkMethod.class);
```
ChristianSchwarz: `dummy` should be renamed to `mock` otherwise it is not clear in your tests that dummy is a mock

ChristianSchwarz:
@szczepiq can you squash the commits?

src/main/java/org/mockito/internal/runners/UnnecessaryStubbingsReporter.java
```diff
@@ -45,13 +48,20 @@ public void report(Class<?> testClass, RunNotifier notifier) {
...
+            // we cannot safely rerun validateUnusedStubs() method.
+            // However, we get good performance and simpler code.
             stubbings.remove(u);
         }
 
         if (stubbings.isEmpty()) {
             return;
         }
 
+        if (stubbings.isEmpty()) {
```
ChristianSchwarz: This will very unlikely to happen, see 3 lines above  :smile:

szczepiq:
Woah, that's what I'm calling feedback :) Thanks, I'll get onto it!!!

szczepiq:
Some update: I will update this PR very soon. I'm making nice progress.

src/main/java/org/mockito/internal/progress/MockingProgressImpl.java
```diff
@@ -142,15 +145,20 @@ public ArgumentMatcherStorage getArgumentMatcherStorage() {
...
+            if (listener instanceof MockCreationListener) {
+                ((MockCreationListener) listener).mockCreated(mock, settings);
+            }
         }
         validateMostStuff();
     }
 
-    public void setListener(MockitoListener listener) {
-        this.listener = listener;
+    public void addListener(MockitoListener listener) {
```
ChristianSchwarz: null arguments should not be allowed, they may cause an NPE later when listeners are notified

src/main/java/org/mockito/internal/junit/ArgMismatchFinder.java
```diff
@@ -0,0 +1,28 @@
+package org.mockito.internal.junit;
+
+import org.mockito.internal.invocation.finder.AllInvocationsFinder;
+import org.mockito.invocation.Invocation;
+import org.mockito.internal.invocation.Stubbing;
+
+/**
+ * For given mocks, finds stubbing arg mismatches
+ */
+class ArgMismatchFinder {
```
ChristianSchwarz: @szczepiq 
Why do we need an instance of `ArgMismatchFinder`? Is this class intended to be subclassed so `getStubbingArgMismatch(..)` can be overriden?

src/main/java/org/mockito/internal/junit/ArgMismatchFinder.java
```diff
@@ -0,0 +1,28 @@
...
+
+    StubbingArgMismatches getStubbingArgMismatches(Iterable<Object> mocks) {
```
ChristianSchwarz: @szczepiq 
This method should be moved to class `RuleStubbingHintsReporter` since it is only used there if I am not mistaken.

src/main/java/org/mockito/internal/framework/DefaultMockitoFramework.java
```diff
@@ -0,0 +1,32 @@
...
+        MockingProgress p = mockingProgress();
+        if (listener == null) {
+            p.setStubbingListener(null);
+        } else {
+            p.setStubbingListener(new ThreadSafeStubbingListener(listener));
+        }
+    }
+
+    public void addListener(MockitoListener listener) {
+        if (listener instanceof MockCreationListener) {
```
ChristianSchwarz: @szczepiq 

Is this really needed? [MockingProgressImpl.mockingStarted()] (https://github.com/mockito/mockito/pull/527/files#diff-df53f45103af61fe77db9216591b7ad5R146) filters already for `MockCreationListener`.

ChristianSchwarz:
@szczepiq 
Register / remove MockitoListeners should be independent of the calling Thread. I wonder if a listener will be notified if it was registered by Thread-A and a mock is created by Thread-B.

szczepiq:
Hey Christian!!! Good feedback! I'm scratching the idea of StubbingListener, because I cannot get this working for multi-threaded scenarios.

I made final changes, please check it out if you can. Javadoc left TODO for one of the class, otherwise I'm pretty happy :). Will get it done tomorrow. Feedback welcome (unless you want to pick on the stateless classes that could be static!!! :))). I'm merging very soon.

szczepiq:
I'm done. Currently, the listeners need to be per thread and it's documented in the javadoc. I don't know if it is the best idea but given trade-offs, I chose this implementation.

src/main/java/org/mockito/internal/progress/MockingProgressImpl.java
```diff
@@ -128,15 +143,21 @@ public ArgumentMatcherStorage getArgumentMatcherStorage() {
...
+                ((MockCreationListener) listener).mockCreated(mock, settings);
+            }
         }
         validateMostStuff();
     }
 
-    public void setListener(MockingProgressListener listener) {
-        this.listener = listener;
+    public void addListener(MockitoListener listener) {
+        this.listeners.add(listener);
```
ChristianSchwarz: This method should reject null listeners

ChristianSchwarz:
@szczepiq 

The threading problem should be avoided before release. I looked into it worked on a fix proposal that you can find here -> [Gist](https://gist.github.com/ChristianSchwarz/26c2dc118d407a1347ea523d52cffcc1)

The idea is to store the listeners is in `ThreadSafeMockingProgress` that is used as global access instead of the `MockingProgressImpl` that is used only for the calling thread.

ChristianSchwarz:
@szczepiq 

More more thing that I noticed. The listener methods should be prefixed with "on" like `onMockCreated(..)` and the notifing method should be prefixed with "fire" like `fireMockCreated()`. Currently both methods (the notifying and listener method) can't be distinguished they cause both have the name `mockCreated(..)`.

ChristianSchwarz:
@szczepiq 

`StubbingListener` should implement `MockitoListener` this way it can be registered using  `MockingProgress.addListener(..)`. After that set-/ removeStubbingListener(..) can be removed.

src/main/java/org/mockito/internal/progress/MockingProgressImpl.java
szczepiq: +1

szczepiq:
@ChristianSchwarz 

Great feedback, thanks!

+1 to the rename on the listeners
-1 StubbingListener changes - I'm getting rid of StubbingListener, I reopened #401 

StubbingListener does not work in multi-threaded environment where shared mock is interacted with from multiple threads. 'stubbing used' event can be triggered from multiple threads. I would have to make StubbingListener completely static to handle this scenario (which would introduce set of complexities).

src/main/java/org/mockito/internal/framework/DefaultMockitoFramework.java
szczepiq: Done. I merged those listeners.

src/main/java/org/mockito/internal/junit/ArgMismatchFinder.java
szczepiq: I like it seperate. Cleaner OO, separation of complexities into objects/classes. easy testing.

src/main/java/org/mockito/internal/junit/ArgMismatchFinder.java
szczepiq: No subclassing. I hate violently inheritance :D (love interfaces, though).

I prefer objects than static methods (as we are already discussing elsewhere ;)

szczepiq:
Feedback incorporated :) Any final feedback?

I'll recreate PR with squashed commits tomorrow & merge. Next step for me: fix #401

bric3:
Massive change (52 files) give us a bit more time !

src/main/java/org/mockito/Mockito.java
```diff
@@ -2581,7 +2581,7 @@ public static VerificationMode description(String description) {
     }
 
     /**
-     * Helps debugging failing tests. Experimental - use at your own risk. We're not sure if this method will stay in public api.
+     * TODO 543 move to MockingDetails
```
bric3: Issue #543 is created, let's revert this javadoc

src/main/java/org/mockito/exceptions/misusing/UnnecessaryStubbingException.java
```diff
@@ -1,10 +1,27 @@
...
+ * Notice that one of the stubbed methods were never realized in the code under test, during test execution.
+ * The stray stubbing might be an oversight of the developer, the artifact of copy-paste
+ * or the effect not understanding the test/code.
+ * Either way, the developer ends up with unnecessary test code.
+ * In order to keep the codebase clean & maintainable it is necessary to remove unnecessary code.
+ * Otherwise tests are harder to read and reason about.
+ * <p>
+ * To find out more about detecting unused stubbings see {@link org.mockito.quality.MockitoHint}.
  */
 public class UnnecessaryStubbingException extends RuntimeException {
```
bric3: Why not extending `MockitoException` ?

szczepiq:
>Massive change (52 files) give us a bit more time !

@bric3, absolutely. In return, can you commit to review ETA? (Reopened issue #401 needs this PR merged).

src/main/java/org/mockito/internal/framework/DefaultMockitoFramework.java
```diff
@@ -0,0 +1,40 @@
...
+    public void removeListener(MockitoListener listener) {
+        assertNotNull(listener);
+        mockingProgress().removeListener(listener);
+    }
+
+    private void assertNotNull(MockitoListener listener) {
+        if (listener == null) {
+            throw new IllegalArgumentException("listener cannot be null");
+        }
+    }
```
bric3: This can be replaced by calling instead `org.mockito.internal.util.Checks.checkNotNull(listner, "mockito listener")`

bric3:
@szczepiq Doing it right now ;)

src/main/java/org/mockito/internal/invocation/finder/AllInvocationsFinder.java
```diff
@@ -32,9 +34,33 @@ private AllInvocationsFinder() {}
         return new LinkedList<Invocation>(invocationsInOrder);
     }
 
+    /**
+     * Gets all stubbings from mocks. Invocations are ordered earlier first.
+     *
+     * @param mocks mocks
+     * @return stubbings
+     */
+    public static Set<StubbedInvocationMatcher> findStubbings(Iterable<?> mocks) {
```
bric3: Idea or later improvement, `StubbedInvocationMatcher` is a class, what do you think of introducing a sub-interface of `InvocationOnMock` that could called `StubbedInvocation` in order to hide the

src/main/java/org/mockito/internal/junit/JUnitRule.java
bric3: Maybe use an enum instead of a boolean, the benefit would allow self documenting.

src/main/java/org/mockito/internal/junit/JUnitRule.java
```diff
@@ -14,25 +13,74 @@
...
+        private void performEvaluation(RuleStubbingHintsReporter reporter) throws Throwable {
+            MockitoAnnotations.initMocks(target);
+            try {
+                base.evaluate();
+            } catch(Throwable t) {
+                reporter.getStubbingArgMismatches().format(testName, logger);
+                throw t;
             }
-        };
+            reporter.getUnusedStubbings().format(testName, logger);
```
bric3: Would it be possible to let the reporter print everything i.e. stubbing arg mismatches and/or unused stubbings, instead of making two calls ?
This would allow to keep the stubbing reporting concerns in the reporting concerns instead of having them in the statement evaluator.

src/main/java/org/mockito/internal/junit/StubbingArgMismatches.java
```diff
@@ -0,0 +1,49 @@
...
+        matchingInvocations.add(invocation);
+    }
+
+    public void format(String testName, MockitoLogger logger) {
+        if (mismatches.isEmpty()) {
+            return;
+        }
+
+        StubbingHint hint = new StubbingHint(testName);
+        //TODO 544 it would be nice to make the String look good if x goes multiple digits (padding)
```
bric3: Use `java.lang.String#format(java.lang.String, java.lang.Object...)`, something like :

```java
String.format("%4d. %s %s %n", x++, ". Unused... ", m.getKey().getLocation())
```

`4` is hard coded here, but that would work up to 9999 hint. It's still possible to compute that value if hard coded value don't work.

src/main/java/org/mockito/internal/progress/MockingProgressImpl.java
```diff
@@ -111,6 +115,17 @@ public void stubbingCompleted(Invocation invocation) {
         stubbingInProgress = null;
         getStubbingListener().newStubbing(invocation);
     }
+
+    /**
```
bric3: With 2 stars this comment appear as a javadoc comment of the above method (`toString`)

src/main/java/org/mockito/junit/MockitoJUnit.java
```diff
@@ -12,20 +13,23 @@
...
-        return new JUnitRule(new ConsoleMockitoLogger());
+        return new JUnitRule(new ConsoleMockitoLogger(), false);
     }
 
     /**
      * Creates a rule instance that can perform lazy verifications.
      *
      * @see VerificationCollector
      * @return the rule instance
+     * @since 2.*
```
bric3: For 2.0 feature, let's use `@since 2.0.0`

src/main/java/org/mockito/junit/MockitoRule.java
```diff
@@ -36,4 +46,30 @@
...
+     * <pre class="code"><code class="java">
+     * public class ExampleTest {
+     *
+     *     &#064;Rule
+     *     public MockitoRule rule = MockitoJUnit.rule().silent();
+     *
+     * }
+     * </code></pre>
+     *
+     * @since 2.*
```
bric3: For 2.0 feature, let's use @since 2.0.0

src/main/java/org/mockito/junit/MockitoRule.java
```diff
@@ -3,19 +3,29 @@
 import org.junit.rules.MethodRule;
 
 /**
+ * Since 2.*, JUnit rule emits stubbing warnings and hints to System output
```
bric3: Let's have a real version like `2.0.0`

src/main/java/org/mockito/junit/VerificationCollector.java
```diff
@@ -26,15 +27,19 @@
  *
  * @see org.mockito.Mockito#verify(Object)
  * @see org.mockito.Mockito#verify(Object, VerificationMode)
+ * @since 2.*
```
bric3: `2.0.0`

src/main/java/org/mockito/quality/MockitoHint.java
```diff
@@ -0,0 +1,66 @@
...
+ * a) stubbing declaration and b) the method call on a stub with mismatched argument.
+ * <p>
+ * Note that it is just a warning, not an assertion.
+ * The test fails on assertion because it is the assertion's duty
+ * to document what the test stands for and what behavior it proves.
+ * Hints just makes it quicker to figure out if the test fails for the right reason.
+ * <p>
+ * Feedback very welcome at <a href="https://github.com/mockito/mockito/issues/384">issue 384</a>.
+ *
+ * @since 2.*
```
bric3: `2.0.0`

src/main/java/org/mockito/internal/framework/DefaultMockitoFramework.java
```diff
@@ -0,0 +1,40 @@
...
+            p.setStubbingListener(new ThreadSafeStubbingListener(listener));
+        }
+    }
+
+    public void addListener(MockitoListener listener) {
+        assertNotNull(listener);
+        if (listener instanceof MockCreationListener) {
+            mockingProgress().addListener(listener);
+        }
+    }
```
bric3: Why the instance check ?

bric3:
OK reviewed, nice work !

src/main/java/org/mockito/internal/framework/DefaultMockitoFramework.java
szczepiq: good catch. I'll remove it.

src/main/java/org/mockito/quality/MockitoHint.java
szczepiq: +1

src/main/java/org/mockito/internal/junit/StubbingArgMismatches.java
szczepiq: Neat, I'll try that.

src/main/java/org/mockito/internal/junit/JUnitRule.java
szczepiq: I think so. 

For now I'm happy with this design. Please suggest something else. I hope to revisit this soon and refactor to something like you say (I have a rough plan how to do it).

src/main/java/org/mockito/internal/junit/JUnitRule.java
szczepiq: +1 but I want to wait for some feedback. @ChristianSchwarz already had a good idea for "strict" mode that fails when unusued stubbing is detected.

src/main/java/org/mockito/internal/invocation/finder/AllInvocationsFinder.java
szczepiq: Great idea. Let's call it "Stubbing". I created a ticket: #542

src/main/java/org/mockito/internal/framework/DefaultMockitoFramework.java
szczepiq: sweeeeeet. I was looking for just something like it!!!

szczepiq: @bric3, I'd like to change checkNotNull() to throw IAE instead of NPE. See #554

src/main/java/org/mockito/exceptions/misusing/UnnecessaryStubbingException.java
szczepiq: Possibly.

Since we don't need to filter stack trace in this case, I'll keep the RuntimeException hierarchy.

szczepiq:
Recreated with squash at #555. Addressed code feedback selectively, making subjective judgements :)

bric3:
You can push force on the same branch github will figure out the difference and will close comments that have been fixed.

ChristianSchwarz:
@szczepiq 
Did you see the comments were not referenced your name? Most of the them are uncommented.

PR #528, opened 2016-08-02
Missing generics info on collection matchers

bric3:
Follow up on #510 (issue #194) following this comment https://github.com/mockito/mockito/pull/510#issuecomment-236877717

Basically it tweaks the collection matchers with generic support. It should have bee in abf9851.

### DIFF
--- a/src/main/java/org/mockito/ArgumentMatchers.java
+++ b/src/main/java/org/mockito/ArgumentMatchers.java
@@ -474,9 +474,9 @@ public class ArgumentMatchers {
      * @see #isNull()
      * @see #isNull(Class)
      */
-    public static List anyList() {
+    public static <T> List<T> anyList() {
         reportMatcher(new InstanceOf(List.class, "<any List>"));
-        return new ArrayList(0);
+        return new ArrayList<T>(0);
     }
 
     /**
@@ -532,9 +532,9 @@ public class ArgumentMatchers {
      * @see #isNull()
      * @see #isNull(Class)
      */
-    public static Set anySet() {
+    public static <T> Set<T> anySet() {
         reportMatcher(new InstanceOf(Set.class, "<any set>"));
-        return new HashSet(0);
+        return new HashSet<T>(0);
     }
 
     /**
@@ -592,9 +592,9 @@ public class ArgumentMatchers {
      * @see #isNull()
      * @see #isNull(Class)
      */
-    public static Map anyMap() {
+    public static <K, V> Map<K, V> anyMap() {
         reportMatcher(new InstanceOf(Map.class, "<any map>"));
-        return new HashMap(0);
+        return new HashMap<K, V>(0);
     }
 
     /**
@@ -653,9 +653,9 @@ public class ArgumentMatchers {
      * @see #isNull()
      * @see #isNull(Class)
      */
-    public static Collection anyCollection() {
+    public static <T> Collection<T> anyCollection() {
         reportMatcher(new InstanceOf(Collection.class, "<any collection>"));
-        return new ArrayList(0);
+        return new ArrayList<T>(0);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/test/java/org/mockitousage/bugs/DiamondInheritanceIsConfusingMockitoTest.java
```diff
@@ -0,0 +1,44 @@
...
+    public class Sub
+            extends Super<String>
+            implements iInterface {
+
+        public Sub(String s) {
+            super(s);
+        }
+    }
+
+    public interface iInterface {
```
ChristianSchwarz: Should be renamed to `Interface`

src/test/java/org/mockitousage/bugs/DiamondInheritanceIsConfusingMockitoTest.java
```diff
@@ -0,0 +1,44 @@
...
+public class DiamondInheritanceIsConfusingMockitoTest {
+
+    @Test
+    public void should_work() {
+        Sub mock = Mockito.mock(Sub.class);
+        // The following line results in
+        // org.mockito.exceptions.misusing.MissingMethodInvocationException:
+        // when() requires an argument which has to be 'a method call on a mock'.
+        // Presumably confused by the interface/superclass signatures.
+        Mockito.when(mock.getFoo()).thenReturn("Hello");
```
ChristianSchwarz: static imports should be used were possible to improve readability

src/test/java/org/mockitousage/bugs/DiamondInheritanceIsConfusingMockitoTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+        Assert.assertEquals("Hello", mock.getFoo());
```
ChristianSchwarz: What about AssertJ ? `assertThat(mock.getFoo()).isEqualTo("Hello")`

src/main/java/org/mockito/ArgumentMatchers.java
```diff
@@ -474,9 +474,9 @@ public static String anyString() {
      * @see #isNull()
      * @see #isNull(Class)
      */
-    public static List anyList() {
+    public static <T> List<T> anyList() {
```
ChristianSchwarz: Why can't we simply use `List<?>` here?

src/main/java/org/mockito/ArgumentMatchers.java
```diff
@@ -474,9 +474,9 @@ public static String anyString() {
...
         reportMatcher(new InstanceOf(List.class, "<any List>"));
-        return new ArrayList(0);
+        return new ArrayList<T>(0);
```
ChristianSchwarz: Why don't we return `null `?

src/test/java/org/mockitousage/bugs/DiamondInheritanceIsConfusingMockitoTest.java
bric3: Could be it's an additional test for a previous bug. We can change that in some future commit.

src/test/java/org/mockitousage/bugs/DiamondInheritanceIsConfusingMockitoTest.java
bric3: Same as previous comment. User reported test.

src/test/java/org/mockitousage/bugs/DiamondInheritanceIsConfusingMockitoTest.java
bric3: Same as previous test.

src/main/java/org/mockito/ArgumentMatchers.java
bric3: The wildcard should be avoided in return types, especially this one since it's unbounded it forces the caller to deal with the wildcard (casting with mockito matchers). Also using `T` allows the Java 8 compiler to infer (capture) the wanted type.

src/main/java/org/mockito/ArgumentMatchers.java
bric3: For spies.

PR #536, opened 2016-08-08
Move release skipping logic to Gradle

szpak:
This is an enhancement for #493. I'm not a big fan of putting build logic into bash & .travis.yml, especially that we already have it separated in `release.gradle`.


TimvdLippe:
Fine by me, LGTM!

szczepiq:
Nice change!

PR #540, opened 2016-08-09
inlined ArrayUtils.isEmpty() in ArgumentsProcessor

ChristianSchwarz:
Inlined ArrayUtils.isEmpty() in ArgumentsProcessor and removed ArrayUtils cause it was only used in ArgumentProcessor.
Also isEmpty(..) was renamed to isNullOrEmpty(..) for clarity.

relates to #426


TimvdLippe:
LGTM

PR #550, opened 2016-08-12
Fixed #538 changed error message in case initialization for mock injection fails.

### LINKED ISSUE #538: Improve error message when @InjectMocks is uses on an interface or enum field
By accident a tests delares a field of an interface type instead of the implementing class.  

```
@InjectMocks
public InterfaceType unitUnderTest;
```

Injection is not possible on interfaces or enums. The error message should make this clear, the currenta message that doesn't help much:

```
Cannot instantiate @InjectMocks field named 'configurationManager'.
You haven't provided the instance at field declaration so I tried to construct the instance.
However, I failed because: the type 'ConfigurationManager' is an interface.
Examples of correct usage of @InjectMocks:
   @InjectMocks Service service = new Service();
   @InjectMocks Service service;
   //also, don't forget about MockitoAnnotations.initMocks();
   //and... don't forget about some @Mocks for injection :)
```

The error message should be someting like:
`The field 'unitUnderTest' can not be annotated with @InjectMocks cause the type 'InterfaceType ' not a class!
`

ChristianSchwarz:
Fixed #538

If a filed is annotated with @InjectMocks and the type of a field is an enum, inner class, interface or local class the error message is now: _"Mock injection failed on field 'fieldName' cause the type 'AbstractCollection' is an abstract class._"

### DIFF
--- a/src/main/java/org/mockito/internal/exceptions/Reporter.java
+++ b/src/main/java/org/mockito/internal/exceptions/Reporter.java
@@ -627,16 +627,14 @@ public class Reporter {
                                          ""), details);
     }
 
-    public static MockitoException cannotInitializeForInjectMocksAnnotation(String fieldName, Exception details) {
-        return new MockitoException(join("Cannot instantiate @InjectMocks field named '" + fieldName + "'.",
+    public static MockitoException cannotInitializeForInjectMocksAnnotation(String fieldName, String causeMessage) {
+        return new MockitoException(join("Cannot instantiate @InjectMocks field named '" + fieldName + "'! Cause: "+causeMessage,
                                          "You haven't provided the instance at field declaration so I tried to construct the instance.",
-                                         "However, I failed because: " + details.getMessage(),
                                          "Examples of correct usage of @InjectMocks:",
                                          "   @InjectMocks Service service = new Service();",
                                          "   @InjectMocks Service service;",
-                                         "   //also, don't forget about MockitoAnnotations.initMocks();",
                                          "   //and... don't forget about some @Mocks for injection :)",
-                                         ""), details);
+                                         ""));
     }
 
     public static MockitoException atMostAndNeverShouldNotBeUsedWithTimeout() {
--- a/src/main/java/org/mockito/internal/configuration/injection/PropertyAndSetterInjection.java
+++ b/src/main/java/org/mockito/internal/configuration/injection/PropertyAndSetterInjection.java
@@ -98,7 +98,7 @@ public class PropertyAndSetterInjection extends MockInjectionStrategy {
                 Throwable realCause = e.getCause().getCause();
                 throw fieldInitialisationThrewException(field, realCause);
             }
-            throw cannotInitializeForInjectMocksAnnotation(field.getName(), e);
+            throw cannotInitializeForInjectMocksAnnotation(field.getName(),e.getMessage());
         }
     }
 
--- a/src/main/java/org/mockito/internal/util/reflection/FieldInitializer.java
+++ b/src/main/java/org/mockito/internal/util/reflection/FieldInitializer.java
@@ -7,6 +7,7 @@ package org.mockito.internal.util.reflection;
 import org.mockito.exceptions.base.MockitoException;
 import org.mockito.internal.util.MockUtil;
 
+import static java.lang.reflect.Modifier.isStatic;
 import static org.mockito.internal.util.reflection.FieldSetter.setField;
 
 import java.lang.reflect.Constructor;
@@ -68,7 +69,9 @@ public class FieldInitializer {
             checkNotLocal(field);
             checkNotInner(field);
             checkNotInterface(field);
+            checkNotEnum(field);
             checkNotAbstract(field);
+            
         }
         this.fieldOwner = fieldOwner;
         this.field = field;
@@ -100,8 +103,9 @@ public class FieldInitializer {
     }
 
     private void checkNotInner(Field field) {
-        if(field.getType().isMemberClass() && !Modifier.isStatic(field.getType().getModifiers())) {
-            throw new MockitoException("the type '" + field.getType().getSimpleName() + "' is an inner class.");
+        Class<?> type = field.getType();
+        if(type.isMemberClass() && !isStatic(type.getModifiers())) {
+            throw new MockitoException("the type '" + type.getSimpleName() + "' is an inner non static class.");
         }
     }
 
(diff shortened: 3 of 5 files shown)


ChristianSchwarz:
openjdk6 build failed the others gone clear


Wow! the native world exploded
```
:buildSrc:test*** buffer overflow detected ***: /usr/lib/jvm/java-6-openjdk-amd64/bin/java terminated
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(__fortify_fail+0x37)[0x7fa189635e37]
/lib/x86_64-linux-gnu/libc.so.6(+0x108d30)[0x7fa189634d30]
/usr/lib/jvm/java-6-openjdk-amd64/jre/lib/amd64/libnet.so(Java_java_net_Inet4AddressImpl_getLocalHostName+0x1a0)[0x7fa1843c0d80]
[0x7fa17d0136b0]
```

bric3:
This seem to be an issue with Travis itself : https://github.com/travis-ci/travis-ci/issues/5227

bric3:
From their thread, the workaround in the [.travis.yml](https://github.com/mockito/mockito/blob/master/.travis.yml#L18-L23) doesn't work anymore.

cc @raphw

bric3:
Ha, there's been a change a few days ago : https://github.com/travis-ci/travis-ci/issues/5227#issuecomment-165131913

People should use the hosts addon : 

```yml
addons:
  hosts:
    - myshorthost
  hostname: myshorthost
```

src/main/java/org/mockito/internal/exceptions/Reporter.java
```diff
@@ -627,16 +627,10 @@ public static MockitoException cannotInitializeForSpyAnnotation(String fieldName
...
-    public static MockitoException cannotInitializeForInjectMocksAnnotation(String fieldName, Exception details) {
-        return new MockitoException(join("Cannot instantiate @InjectMocks field named '" + fieldName + "'.",
-                                         "You haven't provided the instance at field declaration so I tried to construct the instance.",
-                                         "However, I failed because: " + details.getMessage(),
-                                         "Examples of correct usage of @InjectMocks:",
-                                         "   @InjectMocks Service service = new Service();",
-                                         "   @InjectMocks Service service;",
-                                         "   //also, don't forget about MockitoAnnotations.initMocks();",
-                                         "   //and... don't forget about some @Mocks for injection :)",
-                                         ""), details);
```
bric3: Please don't remove these indication, lot of people would like to see this.

bric3:
Aside from the help removal that I don't want removed the code looks good.

bric3:
@ChristianSchwarz you can rebase on master I may have pushed a working fix for the buffer overflow issue in #553

ChristianSchwarz:
@bric3 
> Aside from the help removal that I don't want removed the code looks good.

Thanks for reviewing! If you don't mind I would like to have a short discussion on the error message inorder to help devs to fix the issue.



----

full error message:
```
Cannot instantiate @InjectMocks field named 'configurationManager'.
You haven't provided the instance at field declaration so I tried to construct the instance.
However, I failed because: the type 'ConfigurationManager' is an interface.
Examples of correct usage of @InjectMocks:
   @InjectMocks Service service = new Service();
   @InjectMocks Service service;
   //also, don't forget about MockitoAnnotations.initMocks();
   //and... don't forget about some @Mocks for injection :)
```


**Line 1**: `"Cannot instantiate @InjectMocks field named '" + fieldName + "'.",`
Okay that line was replaced with something like: `"@InjectMocks failed on field 'f' cause the type 'InnerClass' is an inner non static class"`. So far so good.

**Line 2**:  `"You haven't provided the instance at field declaration so I tried to construct the instance."`
I would like to remove that line. The dev added @InjectMocks in order to construct  an instance, do  we really need to tell him that?  

**Line 3**: `"However, I failed because: " + details.getMessage(),`
Can be removed, it is now included in the first line.

**Line 4**: `"Examples of correct usage of @InjectMocks:",`
Hmmm I have no idea how a dev can learn the provided examples. The field declaration is not wrong in this case but the provided filed type. That is hard (impossible) to express in an example. If you have an idea I will add it.

**Line 5**: `"   @InjectMocks Service service = new Service();",`
Ouch that hurts! I didn't know that it works! What do users expect to happen in this case, big question? IMHO Mockito should fail here, indicating that either @InjectMocks should be removed cause it have no effect or indicating that the field must be initialised with `null`/ leaved blank. Or is it over written after mockito set the field?

**Line 6**: ` "   @InjectMocks Service service;",`
Same as for Line 4

**Line 7**: `"   //also, don't forget about MockitoAnnotations.initMocks();"`
This line can be removed. The user (or Rule/Runner) didn't forget to call initMocks() other wise he won't see this error message.

**Line 8**:  `"   //and... don't forget about some @Mocks for injection :)", `
That is confusing, we don't need mocked fields if the class has an no arg  constructor.
`

bric3:
Here's my comment. I strongly insist on keeping a message like the current one.

> **Line 1**: `"Cannot instantiate @InjectMocks field named '" + fieldName + "'.",`
Okay that line was replaced with something like: `"@InjectMocks failed on field 'f' cause the type 'InnerClass' is an inner non static class"`. So far so good.

I'd like to keep this introduction, here the issue is the instantiation of the field, not the mock injection.

> **Line 2**:  `"You haven't provided the instance at field declaration so I tried to construct the instance."`
I would like to remove that line. The dev added @InjectMocks in order to construct  an instance, do  we really need to tell him what he did?  

This is debatable but a lot of users don't read the javadoc, explaining what mockito has been trying allows users to give a clue how that works. This is especially important for this auto-injection where the behavior is different than regular DI framework. 

> **Line 3**: `"However, I failed because: " + details.getMessage(),`
Can be removed, it is now included in the first line.

Indeed the proposed change on cause explanation looks better.

**Line 4**: `"Examples of correct usage of @InjectMocks:",`
Hmmm I have no idea how a dev can learn from the provided examples. The field declaration is not wrong in this case but the provided filed type. That is hard (impossible) to express in an example. If you have an idea I will add it.

Maybe the examples can be reworked, but he can compare what is good in this simple example with his code without looking at the javadoc. That's the idea behind the `Reporter` not just throwing exceptions.

**Line 5**: `"   @InjectMocks Service service = new Service();",`
Ouch that hurts! I didn't know that it works! What do users expect to happen in this case, big question? IMHO Mockito should fail here, indicating that either @InjectMocks should be removed cause it have no effect or indicating that the field must be initialised with `null`/ leaved blank. Or is it over written after mockito set the field?

Yes this is voluntary and necessary, users may require an initialized instance, and still perform mock injection after. I've seen that with spring or guice, constructor then other `@Autowired` fields.

Field initialization is a feature that was added after and progressively allowed constructor injection. This was not the primary objective at first and thus it may break code.

**Line 6**: ` "   @InjectMocks Service service;",`
Same as for Line 4

I don't see the issue there to have this line as an example.

**Line 7**: `"   //also, don't forget about MockitoAnnotations.initMocks();"`
This line can be removed. The user (or Rule/Runner) didn't forget to call initMocks() other wise he won't see this error message.

Agreed, this line can be removed. Note that this is the JUnit code, others than not use JUnit.

**Line 8**:  `"   //and... don't forget about some @Mocks for injection :)", `
That is confusing, we don't need mocked fields if the class has an no arg  constructor.

Actually some people believed that mocks where created automatically from the fields of the injected class. his could be reworded though.

ChristianSchwarz:
Okay fixed! The first line contains now the cause message, the rest is the same.

bric3:
LGTM

PR #555, opened 2016-08-14
JUnit rules report unused stubs - fixes #384

### LINKED ISSUE #384: JUnit rule logs warnings about unsued / misused stubs

szczepiq:
- Makes JUnit rules detect unused stubs
- Introduces MockitoListener API
- See the use case in issue #384 
- Recreated PR #527 with squashed commits


szczepiq:
Let's merge this today. This PR is a squash of #527 plus some code changes per code review feedback (I addressed code feedback subjectively and selectively :).

src/main/java/org/mockito/internal/runners/UnnecessaryStubbingsReporter.java
```diff
@@ -45,13 +47,20 @@ public void report(Class<?> testClass, RunNotifier notifier) {
...
         }
 
         if (stubbings.isEmpty()) {
             return;
         }
 
+        if (stubbings.isEmpty()) {
+            return; //whoa!!! All stubbings were used!
+        }
+
```
bric3: Comments in #527 not fixed

szczepiq: Ah, missed that. I'll fix it with next PR.

PR #556, opened 2016-08-15
Move Mockito internal classes to internal package

### LINKED ISSUE #551: Unexpected new public API: VerificationWrapper, VerificationWrapperInOrder, VerificationCollectorImpl
It looks like some classes leaked to the public API: VerificationWrapper, VerificationWrapperInOrder, VerificationCollectorImpl

They seem like internal classes because:
a) they are concrete implementation where public API should be an interface
b) no javadoc
c) looking at their API, they don't seem like something we want to expose to the users :)

We need to fix it before 2.0, perhaps it's just the matter of moving it to internal packages.

TimvdLippe:
Fixes #551


szpak:
LGTM

PR #559, opened 2016-08-16
Enable container based infrastructure on Travis

szpak:
To verify it works and is faster.

DO NOT MERGE (YET)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,8 @@ cache:
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
 
-# Enabling container based infrastructure hoping it will help the build speed <= this didn't work well, so it's reverted
-# see http://docs.travis-ci.com/user/workers/container-based-infrastructure/
-sudo: true
+# Enable container based infrastructure - https://docs.travis-ci.com/user/ci-environment/
+sudo: false
 
 language: java
 
--- a/build.gradle
+++ b/build.gradle
@@ -52,6 +52,10 @@ test {
         exceptionFormat 'full'
         showCauses true
     }
+    //Required for building with older Java version on Travis' container-based infrastructure
+    //to not be killed - exit code 137 - https://github.com/travis-ci/travis-ci/issues/5582
+    //-Xmx512m is enough, but Java 8 build seems to be faster with 768m
+    jvmArgs '-XX:MaxPermSize=256m', '-Xmx768m'
 }
 
 dependencies {


TimvdLippe:
So thats a nope and a bummer :(

bric3:
Yep still the same thing :

https://travis-ci.org/mockito/mockito/jobs/152623206#L354

```
> Process 'Gradle Test Executor 3' finished with non-zero exit value 137
```

bric3:
Nice idea to increase memory.

Not sure it's way faster though

* Before https://travis-ci.org/mockito/mockito/builds/152633515 ~ 2min
* After https://travis-ci.org/mockito/mockito/builds/152638526 ~ 2min

bric3:
Why increasing test even more, 512 MB seems already quite huge for a lib like mockito.

szpak:
@bric3 The build time for Java 8 is much higher. I'm testing if accidentally memory is not a limitation.

bric3:
Oh OK

szpak:
With @marcingrzejszczak we had the same problem with some other open source project. The remedy was to increase the amount of memory available for tests. It seems to help also in Mockito. The last 5 builds in that branch finished successfully - https://travis-ci.org/mockito/mockito/branches.

Btw, it's hard to compare reliably, but with `-Xmx768m` Java 8 build is faster than with 512m (while 512m seems to be enough to be not killed).

Unfortunately the overall build time seems to be similar. However, it seems to be the future of builds on Travis, so IMHO we could make a switch.

szpak:
Created https://github.com/mockito/mockito/issues/563 for flaky test: ThreadsRunAllTestsHalfManualTest > shouldRunInMultipleThreads

szpak:
Btw, https://travis-ci.org/mockito/mockito/jobs/152643498 failed due to my interactive rebase.

bric3:
Maybe try with the tricks in this ticket : https://github.com/travis-ci/travis-ci/issues/3259

```yml
sudo: false
addons:
  apt:
    packages:
      - oracle-java8-installer
```

bric3:
Not sure it's the cause of the problem, but that's a big difference that I'm spotting right now, between container based and bare metal

szpak:
Yup, I've been using `oracle-java8-installer` yesterday. Works, but increased build time by ~20 seconds. It's bad that Travis doesn't provide recent Java 8 version by default.

TimvdLippe:
Superseded by #785

PR #560, opened 2016-08-16
Fixes #554 : Checks.checkNotNull now throws IAE instead of NPE

### LINKED ISSUE #554: Checks.checkNotNull should emit IllegalArgumentException instead of NPE
Checks.checkNotNull emits NPE. I think it is better to emit IAE, because:
 - IAE is more specific whereas NPE is more generic. Specific exception is easier to test.
 - Specific exception gives better experience for users, NPE is often assumed to be developer error.

This change is not backwards compatible.

@bric3, thoughts?

bric3:
Changes NPE to IAE, see #554


TimvdLippe:
LGTM You can merge when Travis passes.

PR #569, opened 2016-08-19
Improved the public API of MockingDetails

### LINKED ISSUE #541: improve Mockito.mockingDetails API
In order to make the API cleaner & expose useful information:

1. MockingDetails.getInvocations() throws meaningful exception when passed object is not a mock and documents this behavior
2. getMockedType() and getExtraInterfaces() are replaced with getMockCreationSettings() that returns MockCreationSettings instance.

szczepiq:
Fixes #541, highlights:

- started exposing full mock creation information instead of just type to mock and extra interfaces
- changed public API but only for methods that were introduced in 2.x. So it is _not_ backwards incompatible change as far as 1.10 is concerned.


src/main/java/org/mockito/internal/util/DefaultMockingDetails.java
```diff
@@ -36,17 +38,25 @@ public boolean isSpy(){
...
-        return getMockHandler(toInspect).getMockSettings().getTypeToMock();
+    public MockCreationSettings<?> getMockCreationSettings() {
+        return mockHandler().getMockSettings();
     }
 
-    @Override
-    public Set<Class<?>> getExtraInterfaces() {
-        return getMockHandler(toInspect).getMockSettings().getExtraInterfaces();
+    private InternalMockHandler<Object> mockHandler() {
+        assertGoodMock();
```
bric3: I think the assertions should be on the constructor. This enable a fail fast behavior.

bric3:
I'd like the comment fixed. But it looks good to me!

src/main/java/org/mockito/internal/util/DefaultMockingDetails.java
szczepiq: We cannot have them on constructor at the moment because isMock / isSpy methods permit null / non-mock arguments. Options:

1. separate isMock / isSpy API somewhere else
2. return empty / null result from getInvocations / getCreationSettings - then we don't need any exception
3. leave things as they are (merge PR :D)
4. ?

My thinking process

1 - API discoverability suffers
2 - poor API, no distinction between non-mock and mock without invocations, null handling required for the client code
3 - late exception not ideal but acceptable IMO

Good feedback!

szczepiq:
Waiting until EOD today and merging if no other feedback. If you're happy with change merge yourself :)

szczepiq:
I needed to merge because it was blocking me. I can address @bric3 comment later if such is the decision.

src/main/java/org/mockito/internal/util/DefaultMockingDetails.java
bric3: > We cannot have them on constructor at the moment because isMock / isSpy methods permit null / non-mock arguments. 

Good point. That was a suggestion only. But that work too !

Discoverability won't suffer that much. Imho

bric3:
@szczepiq Nope your feedback works for me as well ;)

src/test/java/org/mockito/internal/util/DefaultMockingDetailsTest.java
```diff
@@ -42,28 +44,51 @@ public void should_check_that_a_spy_is_also_a_mock() throws Exception {
...
-        return new HashSet<T>(Arrays.asList(items));
+    @Test
+    public void fails_when_getting_invocations_when_null() {
+        try {
+            //when
+            mockingDetails(null).getInvocations();
+            //then
+            fail();
+        } catch (NotAMockException e) {
+            TestCase.assertEquals("Argument passed to Mockito.mockingDetails() should be a mock, but is null!", e.getMessage());
```
ChristianSchwarz: @szczepiq We should use the ExpectedException Rule here. WDYT?

bric3: @ChristianSchwarz I'm not a fan of `ExpectedException` rule, this make the test a bit awkward, try/catch is a bit more bulky but it offers more control.

Also when we switch to Java8 after mockito 2, we can bump to AssertJ 3.x which has interesting assertions like : 

```java
Assertions.assertThatThrownBy(() -> mockingDetails(null).getInvocations())
          .isInstanceOf(NotAMockException.class)
          .hasMessageContaining("Argument passed to Mockito.mockingDetails() should be a mock, but is null!")
```

PR #483, opened 2016-07-07
Update version scheme to publish release candidate

TimvdLippe:
Beta releases now follow the official semver scheme of `MAJOR.MINOR.PATCH-beta.BETAVERSION`. This PR also updates the current version to a release candidate.

We should decide after the release candiate what the version of Mockito 2.0 should be. E.g. should it be `2.0.0`? That would mean we go down on our `PATCH` number which might confuses people. For the Mockito 3.0 beta is to set the version to `3.0.0-beta.BETAVERSION` to not confuse with the `PATCH` number, but it still leaves us with the question regarding the official release of Mockito 2.0 :cry: 

Closes #123

PR #588, opened 2016-08-23
Update version explanation with bad beta versions

bric3:
The previous versioning scheme has been changed in #483, however the previous artefacts still exists, and some pages may mention them, we should still explain what are these artefact and what they mean in regard of other versions.


TimvdLippe:
LGTM

szczepiq:
Nice. Thanks!

PR #589, opened 2016-08-23
Reintroduces javadoc stylesheet

### LINKED ISSUE #552: Fix again javadoc stylesheet
Previously we had a nice javadoc stylesheet. That worked well with JDK6.
Then we used travis that was only allowing OracleJDK7, that came with a better javadoc stylesheet so we dropped ours.

Now since the release of mockito 2.x is using openjdk6, we should reintroduce a better stylesheet.

Current state ❌ 
http://site.mockito.org/mockito/docs/2.0.100-beta/org/mockito/Mockito.html
screen shot 2016-08-13 at 19 41 59

Previous ✅ 
http://site.mockito.org/mockito/docs/2.0.0-beta/org/mockito/Mockito.html
screen shot 2016-08-13 at 19 45 32

bric3:
Fixes #552

Since artifacts are built with openjdk6, we can reintroduce a good looking javadoc stylesheet.

### DIFF
--- a/gradle/javadoc.gradle
+++ b/gradle/javadoc.gradle
@@ -52,7 +52,7 @@ task mockitoJavadoc(type: Javadoc) {
           }
         </script>
     """.replaceAll(/\r|\n/, ""))
-//    options.stylesheetFile file("src/javadoc/stylesheet.css")
+    options.stylesheetFile file("src/javadoc/stylesheet.css")
 //    options.addStringOption('top', 'some html')
     if (JavaVersion.current().isJava8Compatible()) {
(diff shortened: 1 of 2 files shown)


bric3:
This looks like the previous style we had.
screen shot 2016-08-23 at 19 53 55

bric3:
**Note :** _The rendering is good only on the JDK6. Which is used to create the artefacts._

bric3:
Note the javadoc change will be visible in the next published build, `2.0.0-beta.112`.
There wasn't any code change in the mockito jar so the script didn't triggered a release.

PR #600, opened 2016-08-26
suppressed compiler warning and unchecked collections cast

lukasz-szewc:
- suppressed compile time warning
- refactor raw collection cast


TimvdLippe:
LGTM

PR #602, opened 2016-08-27
Generate service-worker after generating JavaDoc for offline access

TimvdLippe:
Service-workers intercept network calls and (if you are offline)
can serve files from the cache. The documentation of Mockito can
therefore be accessed even if you do not have internet access.

All files are pre-cached the first time you open the webpage.
Consecutive openings of the website will result in significantly
faster responses (since we poke the cache) and also work offline.

When new documentation is generated, the service-worker is updated
and a friendly toast is displayed informing the user can refresh
the page to view the new content.

Requirements for generating the service-worker is having npm
(and therefore Node.js) installed. You must also make sure the
gradle daemon has the correct permissions to write to the global
node_modules folder.

This normally is done with (on linux):

``` shell
sudo chown -R $(whoami) /usr/local/lib/node_modules
sudo chown -R $(whoami) /usr/local/bin
sudo chown -R $(whoami) ~/.npm
```

Fixes https://github.com/mockito/mockito.github.io/issues/17

### DIFF
--- a/gradle/javadoc.gradle
+++ b/gradle/javadoc.gradle
@@ -31,7 +31,7 @@ task mockitoJavadoc(type: Javadoc) {
     options.noNavBar = false
     options.noTree = false
     options.links('http://docs.oracle.com/javase/8/docs/api/')
-    options.header("""<em id="mockito-version-header-javadoc7-header"><strong>Mockito ${project.version} API</strong></em>""")
+    options.header("""<script>var l = window.location; if (l.protocol === "http" && l.origin.indexOf("javadoc.io") > -1) { l.protocol = "https"; }</script><em id="mockito-version-header-javadoc7-header"><strong>Mockito ${project.version} API</strong></em>""")
     options.footer("""<em id="mockito-version-header-javadoc7-footer"><strong>Mockito ${project.version} API</strong></em>""")
     options.bottom("""
         <script type="text/javascript" src="{@docRoot}/js/${javadocJdk6Hack}"></script>
--- a/gradle/javadoc.gradle
+++ b/gradle/javadoc.gradle
@@ -52,12 +52,18 @@ task mockitoJavadoc(type: Javadoc) {
             from "src/javadoc"
             into "$buildDir/javadoc"
         }
-        logger.lifecycle("Installing sw-precache globally")
-        exec {
-            commandLine "npm", "install", "-g", "sw-precache"
-        }
-        exec {
-            commandLine "sw-precache", "--config=sw-precache-config.json"
+        try {
+            logger.lifecycle("Installing sw-precache globally")
+            exec {
+                commandLine "npm", "install", "-g", "sw-precache"
+            }
+            exec {
+                commandLine "sw-precache", "--config=sw-precache-config.json"
+            }
+            // Even though there is an option ignoreExitValue for exec, exec keeps on throwing exceptions.
+            // Therefore we just have to catch them and continue on with the build.
+        } catch(Exception ignored) {
+            logger.lifecycle("NPM is not installed. Aborting generating service-worker");
         }
     }
 }
--- a/gradle/javadoc.gradle
+++ b/gradle/javadoc.gradle
@@ -37,20 +37,9 @@ task mockitoJavadoc(type: Javadoc) {
         <script type="text/javascript" src="{@docRoot}/js/${javadocJdk6Hack}"></script>
         <script type="text/javascript" src="{@docRoot}/js/jquery-1.7.min.js"></script>
         <script type="text/javascript" src="{@docRoot}/js/highlight-8.6-java/highlight.pack.js"></script>
-        <link rel="stylesheet" type="text/css" href="{@docRoot}/js/highlight-8.6-java/styles/obsidian.css"/>
 
-        <script type="text/javascript">
-          var usingOldIE = \$.browser.msie && parseInt(\$.browser.version) < 9;
-          if(!usingOldIE) {
-              \$("head").append("<link rel=\\"shortcut icon\\" href=\\"{@docRoot}/favicon.ico?v=cafebabe\\">");
-              \$("head", window.parent.document).append("<link rel=\\"shortcut icon\\" href=\\"{@docRoot}/favicon.ico?v=cafebabe\\">");
-              hljs.initHighlightingOnLoad();
-              \$("pre.code").css("font-size", "0.9em");
-              injectProjectVersionForJavadocJDK6("Mockito ${project.version} API",
-                                                 "em#mockito-version-header-javadoc7-header",
-                                                 "em#mockito-version-header-javadoc7-footer");
-          }
-        </script>
+        <link rel="stylesheet" type="text/css" href="{@docRoot}/js/highlight-8.6-java/styles/obsidian.css"/>
+        <script type="text/javascript" src="{@docRoot}/js/index.js" async defer></script>
     """.replaceAll(/\r|\n/, ""))
     options.stylesheetFile file("src/javadoc/stylesheet.css")
 //    options.addStringOption('top', 'some html')
@@ -62,6 +51,13 @@ task mockitoJavadoc(type: Javadoc) {
         copy {
             from "src/javadoc"
             into "$buildDir/javadoc"
+        }
+        logger.lifecycle("Installing sw-precache globally")
+        exec {
+            commandLine "npm", "install", "-g", "sw-precache"
+        }
+        exec {
+            commandLine "sw-precache", "--config=sw-precache-config.json"
         }
     }
 }
--- a/src/javadoc/js/index.js
+++ b/src/javadoc/js/index.js
@@ -1,0 +1,64 @@
+var usingOldIE = $.browser.msie && parseInt($.browser.version) < 9;
+if(!usingOldIE) {
+    $("head").append("<link rel=\"shortcut icon\" href=\"{@docRoot}/favicon.ico?v=cafebabe\">");
+    $("head", window.parent.document).append("<link rel=\"shortcut icon\" href=\"{@docRoot}/favicon.ico?v=cafebabe\">");
+    hljs.initHighlightingOnLoad();
+    $("pre.code").css("font-size", "0.9em");
+    injectProjectVersionForJavadocJDK6("Mockito ${project.version} API",
+        "em#mockito-version-header-javadoc7-header",
+        "em#mockito-version-header-javadoc7-footer");
+}
+
+if ('serviceWorker' in navigator) {
+    var toast = document.getElementById('sw-toast');
+    if (!toast) {
+        toast = document.createElement('div');
+        toast.id = 'sw-toast';
+        toast.show = function(message) {
+            toast.textContent = message;
+            setTimeout(function() {
+                toast.classList.remove('shown')
+            }, 3000);
+            toast.classList.add('shown');
+        };
+        document.body.appendChild(toast);
+        toast.classList.add('hidden');
+    }
+    // Show a toast with a service-worker-related update.
+    window.showToast = function(message) {
+        toast.show(message);
+    };
+
+    navigator.serviceWorker.register('service-worker.js')
+        .then(function(registration) {
+            registration.onupdatefound = function() {
+                // The updatefound event implies that registration.installing is set; see
+                // https://slightlyoff.github.io/ServiceWorker/spec/service_worker/index.html#service-worker-container-updatefound-event
+                var installingWorker = registration.installing;
+                installingWorker.onstatechange = function() {
+                    switch (installingWorker.state) {
+                        case 'installed':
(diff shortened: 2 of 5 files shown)


TimvdLippe:
Small note on testing. You can run `./gradlew clean javadoc` and then open the website with a local webserver. Then go offline and refresh the page. You should be able to view the javadoc just fine, even though you don't have internet access.

TimvdLippe:
Oh but I discovered one issue. ServiceWorker requires https to function. However, since mockito.org does not have https, this will fail. Moreover, mockito.github.io redirects to site.mockito.org which is incorrect, since github.io DOES support https.

Therefore instead of mockito.github.io redirecting to site.mockito.org, it should be the other way around: mockito.org redirecting to mockito.github.io. I suppose you are already fixing this redirect @szczepiq ?

bric3:
@TimvdLippe Actually as soon as a repository has a `CNAME` github is not able to offer https (yet). One can redirect to `mockito.github.io`, but its a bit sad to lose the `mockito.org` also I'm think google will eventually reduce `mockito.org` ranking, which is not what we want.

Github is nice, fancy and comfortable, but we don't know if it will be for life, I'd rather take a conservatory approach here.

Also I'm sure github will support Let's Encrypt at some point.

bric3:
My opinion is that I'm not sure java developer wants to have a nodejs running in background. Or want to think about starting a NodeJS in background.

TimvdLippe:
@bric3 I am okay with a conservative approach, the only thing I am worried about is that the only developer who has actual access to the server (configuration) is planning to leave after Mockito 2. Using github as server provider is a lot easier for me at least, so that was the reason I was advocating for a different approach.

Not having https is sadly blocking for this change, but I hope we can have https soon on mockito.org.

Regarding Node.js: yes that actually crossed my mind after submitting the PR. It is probably better to make it detect node and only if it is installed run it. I am heavily involved in web dev too atm, so I considered Node as given. Sorry :p

bric3:
No problem, i think this is a cool idea, yet I think it should be developed outside mockito.
Something that can cache javadocs of multiple projects. It would much more interesting as end user.

There's programs that did that on OSX like Dash, this one does much much more than downloading javadoc, anyway I don't think the program is free.

bric3:
About passwords and such we need to find a way to share secrets.

szczepiq:
Can you describe the use case better? What's the benefit?

What are the risks of the implementation? What are the downsides?

TimvdLippe:
The usecase for implementing a service-worker is having offline access to our Javadoc. This means that even if you are developing in an airplane or at a place without internet, the developer can read the Javadoc. Moreover, since files are cached, the time to load the website is significantly reduced. Since files are served from the cache, responses can be in range of ms instead of >300 ms previously.

The serviceWorker itself is generated by the tool [sw-precache](https://github.com/GoogleChrome/sw-precache). This is an official tool developed by Chrome to ease the adoption of ServiceWorker. I configured the tool to pre-cache all files required to view our Javadoc. This means that when you open the website, the files are downloaded and stored in the browser cache. At the moment there are 282 files downloaded with a total size of 3.32 MB. If you are browsing our Javadoc, the developer should experience significantly reduced load times and (if he/her has no internet) offline access, at a very small cost of pre-storing the required files.

TimvdLippe:
Regarding external tools like Dash: a serviceWorker is less intrusive for the enduser. Simply opening the website is enough to experience the improvements. If the developer needs to actively use an external program, we do not have control of the Javadoc versions. With a serviceWorker and the automatic deployments, we are in control of what version is shown to the developers.

szczepiq:
Thanks for describing it thoroughly and for being super patient with me!!! :)

I still have some questions and perhaps it is a good example to list out important components of the use case that really help gauging the usefulness/validity of an idea/PR. It really helps the team make good decisions about Mockito features:

1. By developer you mean: 
 - Mockito user working on his project
 - Mockito user who wants to contribute to Mockito
 - Mockito team member working on Mockito code
2. By working you mean:
 - Working on some project that uses Mockito
 - Working on Mockito codebase
 - Working on Mockito site code
3. The use is driven by:
 - your own experiences/problems
 - user request formal/informal
 - subjective hypothesis

Thanks for again for your time getting the use case described for us!!!

TimvdLippe:
This PR is primarily aimed at developers learning Mockito and using it for
their own projects. E.g. someone who is unfamiliar with how Mockito works
or unfamiliar with new features of Mockito. These developers will be
frequently checking the Javadoc to look something up or read the big doc on
the Mockito class.

For these developers the benefits are faster loading of the javadoc as well
as reduced bandwidth. Moreover if a user is trying out Mockito in his free
time (for example during commute) internet is most likely nonexistent which
means after this change he/she can continue working with Mockito.

So to sum up: this is for primarily for (new) users of the library, not the
contributors. Contributors have the code available in their IDE anyways.

bric3:
That represents only my own work flow but I generally download sources in my IDE so I can have both sources and javadoc. I have bias in this, typically I font totally trust the javadoc of my libraries, however I rely heavily on the behind my IDE to download sources (IntelliJ, others might propose similar actions).

TimvdLippe:
Yes for such a workflow this change does not pose many benefits. I personally do not download sources, but browse Javadoc online. This mostly boils down to personal preferences, but I think this change improves the usage for browsing Javadoc online, while having minimal impact on the website.

szczepiq:
Excellent, this describes the use case very well.

In order for us to make good decision / recommendation, we should know:

1. What are the software requirements for John Java to take advantage of this feature? How likely users will actually use the new feature?
2. How would users know how to set up things so that they can leverage?
3. If the requirements are not met, what's the behavior?

Some of that we have already covered in the discussion:

1) npm
2) ?
3) assume graceful degradation

Stuff to consider:

- John Java does not have npm installed
- John Java has javadoc in IDE, downloaded when project was imported to IDE.
- npm/Javascript modules ecosystem is not very mature. Stuff breaks. Can we ensure graceful degradation of this feature if things don't work as expected? Are we able to fix problems when @TimvdLippe retires :) Do we want to maintain the code?

Votes please! Good discussion! Good framework for making decisions about features we add to Mockito.

TimvdLippe:
1. The browser must implement service-worker (http://caniuse.com/#feat=serviceworkers) and then whenever he/she opens site.mockito.org/mockito/docs/current/ the service-worker will be active
* They don't, it's automatically done just by opening the website
* The behavior as it is currently, e.g. no offline access and no caching. This is in web dev known as "progressive enhancement"

E.g. A developer does not require `npm`, only the generation script does which currently runs on Travis which has `npm` installed by default. Therefore the build already passes. I will alter the script however for those Mockito contributors who run `./gradlew ciBuild` locally to not bail on not having `npm` installed.

If John Java has JavaDoc in IDE, this change will not have any effect. It only effects users who visit the website.

The only dependency introduced is `sw-precache` which is maintained by Google. I do not expect to be deprecated anytime soon. Not sure what you mean with "not very mature", but `npm` has been stable for quite some years now. In other words, almost every web developer can maintain this code, but for now I will be the one :wink:

bric3:
@TimvdLippe Think also about users that are forced to use old browsers. In the past some people couldn't like the code beautification because it was breaking stuff and making the browser super slow.

Unfortunately too many companies still blocks other browser usage and/or enforce vendor and version like IE6 (yes that's true)

TimvdLippe:
@brice3 yes good point, luckily we take care of this due to progressive enhancement (see this line: https://github.com/mockito/mockito/pull/602/files#diff-d07ac501bac0ecff5e37e7006f03bf4eR12). For older browsers which do not support serviceworker, there is NO difference. The site is working exactly the same as before this change. Only the browsers that do support serviceworker (see caniuse) there is a functional (and performance win) change.

Regarding slow: for the browsers that do not support SW, the change is a couple of extra bytes when downloading the website and 1 more if statement. So time and memory wise negligible. For the browsers that do support, impact on bandwidth and response time is huge.

Metanote: I have to admit I overestimated how easy it would be to explain such an abstract technology as serbviceworker is. Personally I have a lot of experiences, but I did not realize Java devs obviously do not. I will make small update on this PR regarding Mockito contributors who do not have npm to not break the build. This to make sure there are no additional dependencies for you :)

szczepiq:
Are we ready to vote?

I see benefits and respect & admire @TimvdLippe passion to get this in!

Despite the benefits, I'm -1 because:
 - it adds maintenance cost in the long run:
  - technology we don't specialize in
  - new dependencies in the build
  - using APIs that may change over time
 - adds risk that we regress behavior for old browsers (bugs, etc)
 - hypothetical use case, not suggested / reported by users
 - javadoc is availabe in IDE

TimvdLippe:
> adds risk that we regress behavior for old browsers (bugs, etc)

I have tried to explain this in my previous comment, but this is not true. There is no regression for old browsers.

> hypothetical use case, not suggested / reported by users

Service-workers are recent, so not much time since they could have. Regardless, users have been wanting offline javadoc for quite a while. These are for example regarding JDK docs: http://stackoverflow.com/questions/6986993/how-to-download-javadoc-to-read-offline, http://stackoverflow.com/questions/10266756/javadoc-location-for-eclipse-offline

> javadoc is availabe in IDE

Do we have any analytics on how much users browser the website JavaDoc? Just voicing that while we might read in IDE, others don't. I personally prefer to read on a website.


Regardless of above points, I can't deny the slight maintenance cost for us. For now I can bear this cost, but of course I can't promise for over 10 years ;). Even though this cost exists, I think it would be sad to not opt-in of this awesome technology :smile: (but this passion was tangible already I suppose :wink: )

bric3:
About maintainance I agree with @szczepiq, however we can still remove this code if we feel that way. 
Also I'm ok to try this new thing as long as it doesn't impact any users / dev that don't have it, have improper setup, or don't want it.

------------------------------------

Outside the scope of this PR but related to my https://github.com/mockito/mockito/pull/602#issuecomment-242931092 could it be something like 

* An electron app, you go on some javadoc URL and the app caches the content of this javadoc site
* Some webiste with a serviceworker iframe. Like javadoc browser, you go on some javadoc URL, and the site rely on the installed serviceworker
* A browser extension, that identify the javadoc site and caches it

This could be much more interesting as a general tool than a mockito specific.
Additional thing to think about is allowing users to control this cache (refresh site, delete site, etc.)

TimvdLippe:
> An electron app, you go on some javadoc URL and the app caches the content of this javadoc site

> Some webiste with a serviceworker iframe. Like javadoc browser, you go on some javadoc URL, and the site rely on the installed serviceworker

Actually, if you would load site.mockito.org inside an iframe, the service-worker is triggered and can work offline :smile: . This means that if you would build a website that loads javadoc in iframes, Mockito is the one working offline.

In the end, there a lot of possible ways to implement it, but I can assure you this is the easiest way to do so. I will update this PR soonish (after I finish some other work) to make sure it has no functional impact on your Mockito contributions.

TimvdLippe:
If you were to merge this PR, the only thing blocking is https on site.mockito.org.

bric3:
Unfortunately HTTPS cannot be activated on Github when there's a CNAME

TimvdLippe:
Yes that is correct. If I have access to the server, I can generate a certificate with LetsEncrypt :smile:

TimvdLippe:
Oh I just saw your tweet and discovered this issue https://github.com/isaacs/github/issues/156 :sob:

bric3:
I'm not sure that's possible with a canonical name DNS record, that have to be the role of the host to serve.

TimvdLippe:
It seems this requires an external static hosting service. Some users suggested netifly.com (https://github.com/isaacs/github/issues/156#issuecomment-243328658). Hm this really sucks for all GitHub pages. Will have to ponder a bit on this issue. To be continued....

bric3:
I'm sure https for github pages is coming though, but when that is the question

bric3:
holding the pr for now, then we can assess later if we see changes in GH hosting.

TimvdLippe:
javadoc.io does use https, so maybe we can merge this in to be used on their website?

TimvdLippe:
If we do so: need to implement a check to not register the sw if the website is not using https. Since javadoc.io does not employ HSTS (e.g. enforce https)

TimvdLippe:
I emailed the developers of javadoc.io to request enabling HSTS.

bric3:
It won't be a problem if HSTS is not enabled, right ? 
javadoc.io is not the only javadoc hosting service in town, and they may or may not have enforced https

TimvdLippe:
Yes I thought about that too. I will implement a check for https

szczepiq:
You have at least 3 options: a) talk to them and ask for a way to test, b) publish a test version to maven central, c) take calculated risk :)

TimvdLippe:
Calculated risk it is 😄

TimvdLippe:
Let's go and see!

TimvdLippe:
Ah ofc it is not a functional change, so it did not trigger the build just yet. Seems I also can't create a commit to force the build 😭

PR #604, opened 2016-08-27
Release notes group improvements by labels

### LINKED ISSUE #603: Release notes group improvements by labels
For clearer release notes overlook we should group improvements by labels. 

This was originally contributed by @szpak, then lost during the early days of continuous integration, now it's back and it will rock!!!

szczepiq:
Fixes #603 

I will generate a demo release notes and check-in to some throw-away branch.

Here's a demo of generated release notes. Note that final notes will have some manual edits (some narrative, announcement, etc.): https://github.com/mockito/mockito/blob/rc-release-notes-demo/doc/release-notes/official.md

IMHO the current level of automation of release notes is enough :) It's time to work on the announcement, etc. In the process we can tweak the labels on tickets so that generation is more accurate.

### DIFF
--- a/gradle/release.gradle
+++ b/gradle/release.gradle
@@ -28,10 +28,10 @@ notes {
     notesFile = file("doc/release-notes/official.md")
     //label captions, in order of importance - this is the order they will appear in the release notes
     labels = [
-        '1.* incompatible': 'Incompatible changes',
+        '1.* incompatible': 'Incompatible changes with previous major version (v1.x)',
         'java-9': "Java 9 support",
         'java-8': "Java 8 support",
-        'new feature': "New Features",
+        'new feature': "New features",
         'BDD': 'Behavior-Driven Development support',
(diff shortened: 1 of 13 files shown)


gradle/release.gradle
```diff
@@ -25,7 +25,19 @@ apply plugin: 'release-notes'
 apply plugin: 'release'
 
 notes {
-    notesFile = project.file("doc/release-notes/official.md")
+    notesFile = file("doc/release-notes/official.md")
+    //label captions, in order of importance - this is the order they will appear in the release notes
+    labels = [
+        '1.* incompatible': 'Incompatible changes',
```
bric3: I'd rename this to `Incompatible changes with v1.x`

bric3:
This looks good !

gradle/release.gradle
szczepiq: Good idea. Thanks!

PR #605, opened 2016-08-27
Fixes #595, going towqard 2.1.0 instead of 2.0.0

### LINKED ISSUE #595: Remove beta release from maven central
Thanks to the odd number scheme of the beta, the maven badge and other systems like **maven central** consider that version `2.0.111-beta` is more recent than `2.0.0-beta.115` or event more dangerous that `2.0.111-beta` is more recent than the actual release `2.0.0`.

That may cause problems for users that don't read the README on the projects page. This is counter intuitive and forces users to search what is the difference between those versions.

screen shot 2016-08-24 at 16 48 21

http://search.maven.org/#search%7Cga%7C1%7Cmockito-core
screen shot 2016-08-24 at 16 53 43

**Since sonatype does not allow public ways to remove artifacts, we need to create a ticket on their [JIRA](https://issues.sonatype.org/) asking for deletion of the betas, whose access is only possible to authenticated users.**

I propose to do that **after** the release of 2.0.0 !

bric3:
Following discussion in #595, about the best way to handle our non semantic versioning of 2.0 betas, we decided to release 2.1.0 instead of 2.0.0.

I'm simply bumping the versions properties from `2.0.0-beta.119` to `2.1.0-beta.119`

### DIFF
--- a/version.properties
+++ b/version.properties
@@ -1,2 +1,2 @@
-version=2.0.0-beta.119
+version=2.1.0-beta.119
 mockito.testng.version=1.0


bric3:
If the team agree, after this ticket I can edit the milestones : 

* from `2.1` to `2.2`
* from `2.0` to `2.1`

version.properties
```diff
@@ -1,2 +1,2 @@
-version=2.0.0-beta.119
+version=2.1.0-beta.119
```
szczepiq: beta.1 ?

bric3: Yeah I pondered on the beta number. But since Rafael said how did I miss 2.0.0. I thought that a continuous beta build may help.

I don't have strong opinion on this.

TimvdLippe: I would keep incrementing instead of resetting.

szczepiq:
Thanks guys!

Issue #584, opened 2016-08-23
ArgumentCaptor can't capture varargs-arrays


ChristianSchwarz:
Relates to: #439, #565, #567, #583 

ArgumentCaptor can't capture varargs-arrays.

```
@Test
public void shouldCaptureVarArgsAsArray() throws Exception {
    ArgumentCaptor<String[]> varargCaptor = ArgumentCaptor.forClass(String[].class);

    mock.varargs("1","2");   

    verify(mock).varargs(varargCaptor.capture());

    assertThat(varargCaptor).containsExactly(new String[]{"1","2"});
}
```

**Expected:** A `ArgumentCaptor<String[]>` should be able to capture the varargs-array as a whole.
**Actual:** The ArgumentCaptor captures no arrays but every single argument passed to the varargs method. The following exception is thrown:

```
java.lang.AssertionError: 
Actual and expected should have same size but actual size was:
  <2>
while expected size was:
  <1>
Actual was:
  <["1", "2"]>
Expected was:
  <[["1", "2"]]>
```

bric3:
For reference this issue is not new ; see #74, #201, #211

paulduffin:
Can you not just use varargCaptor.getAllValues() to access the list of arguments passed as varargs and then compare that?

ChristianSchwarz:
@paulduffin 

> Can you not just use varargCaptor.getAllValues() to access the list of arguments passed as varargs and then compare that?

This is not always possible cause `getAllValues()` returns all ever captured values.

```java
when(mock).addVararg(captor.capture());
mock.addVararg(1,2);
mock.addVararg(3,4);
mock.addVararg(5,6);

captor.getAllValues() // {1,2,3,4,5,6}
```

vs.

```java
when(mock).addVararg(captor.capture());
mock.addVararg(1,2,3);
mock.addVararg(4,5,6);

captor.getAllValues() // {1,2,3,4,5,6}
```

big-andy-coates:
With https://github.com/mockito/mockito/pull/2807 merged, `argumentCaptor(String.class)` will capture a single value passed to a varargs parameter, where as `argumentCaptor(String[].class)` will capture any number of values.

Issue #565, opened 2016-08-17
ArgumentCaptor should be type aware


ChristianSchwarz:
Assume we have the following test.

```
 @Mock
 private Consumer<Collection<?>> consumer;

 @Test
 public void test() throws Exception {

    consumer.accept(new HashSet<>());
    consumer.accept(new ArrayList<>());

    ArgumentCaptor<ArrayList<?>> captor = ArgumentCaptor.forClass(ArrayList.class);
    verify(consumer).accept(captor.capture());
 }
```

The test fails with:

```
org.mockito.exceptions.verification.TooManyActualInvocations:
consumer.accept(<Capturing argument>);
Wanted 1 time:
-> at foo.TestClass.testName(TestClass.java:64)
But was 2 times. Undesired invocation:
-> at agh.TestClass.testName(TestClass.java:61)

    at foo.TestClass.test(TestClass.java:64)
    [..]
```

The problem here is that ArgumentCaptor is not type aware. That should be fixed.

This issue was extracted from #358 and relates to #439.

bric3:
We are aware of that issue, the design of this tool is to rely on the method signature.

szczepiq:
@ChristianSchwarz, guys any ideas how to improve it?

ChristianSchwarz:
The `CapturingMatcher` can use the Class-Object passed to the `ArgumentCaptor` and capture only arguments that are instances of the expected type. This solution has a down side the captor still matches anything which may surprise users, here is an example:

```
stringArgs= ArgumentCaptor.forClass(String.class);
mock.call(123);
mock.call("abc");
verify(mock).call(stringArgs.capture()); //fails with too many invocations cause call(123) matches too.
```

If the `CapturingMatcher` matches only instances of the expected type then existing client tests may break. But if we go this way we  open a "can of worms™" cause `CapturingMatcher` implements `VarargMatcher` and there are many corner case that must be handled in `ArgumentsComparator`. I currently write regression tests to explore these. When I have more details I will post them here.

terebesirobert:
@ChristianSchwarz Hi,
This seems to be related to #358. WDYT?

ChristianSchwarz:
@terebesirobert 
> This seems to be related to #358. WDYT?

Absolutly, I splitted #358 cause it included two toppics:

1. adding a new captor API `captureIf( )`,still ticket #358
1. type safe capture, this ticket

[...]

PR #606, opened 2016-08-28
Fixes #583 fix bug when trying to capture null vararg

### LINKED ISSUE #583: ArgumentCaptor: NPE when an null array is passed to a varargs method
A NullPointerException is thrown if an null array is passed to a varargs method.

```
@Test
public void shouldCaptureVarArgs_nullArrayArg() {
        String[] argArray = null;
        mock.varargs(argArray);

        verify(mock).varargs(captor.capture()); //<- Kaboom NPE
}
```

Since an vararg is simply an array the ArgumentCaptor should reject it silently. This relates to #565.

```
	at java.lang.reflect.Array.getLength(Native Method)
	at org.mockito.internal.invocation.InvocationMatcher.captureVarargsPart(InvocationMatcher.java:142)
	at org.mockito.internal.invocation.InvocationMatcher.captureArgumentsFrom(InvocationMatcher.java:122)
	at org.mockito.internal.invocation.InvocationMarker.markVerified(InvocationMarker.java:24)
	at org.mockito.internal.invocation.InvocationMarker.markVerified(InvocationMarker.java:18)
	at org.mockito.internal.verification.checkers.NumberOfInvocationsChecker.check(NumberOfInvocationsChecker.java:43)
	at org.mockito.internal.verification.Times.verify(Times.java:40)
	at org.mockito.internal.verification.MockAwareVerificationMode.verify(MockAwareVerificationMode.java:21)
	at org.mockito.internal.handler.MockHandlerImpl.handle(MockHandlerImpl.java:73)
	at org.mockito.internal.handler.NullResultGuardian.handle(NullResultGuardian.java:32)
	at org.mockito.internal.handler.InvocationNotifierHandler.handle(InvocationNotifierHandler.java:37)
	at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.doIntercept(MockMethodInterceptor.java:36)
	at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.access$0(MockMethodInterceptor.java:32)
	at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor$DispatcherDefaultingToRealMethod.interceptAbstract(MockMethodInterceptor.java:113)
	at org.mockitousage.IMethods$MockitoMock$81506210.varargs(Unknown Source)
	at org.mockitousage.matchers.VarargsTest.shouldCaptureVarArgs_nullArrayArg(VarargsTest.java:182)
```

lukasz-szewc:
Should fix just one issue - described in #583

### DIFF
--- a/src/main/java/org/mockito/internal/invocation/InvocationMatcher.java
+++ b/src/main/java/org/mockito/internal/invocation/InvocationMatcher.java
@@ -141,11 +141,19 @@ public class InvocationMatcher implements DescribedInvocation, CapturesArguments
         for (ArgumentMatcher m : uniqueMatcherSet(indexOfVararg)) {
             if (m instanceof CapturesArguments) {
                 Object rawArgument = invocation.getRawArguments()[indexOfVararg];
-                for (int i = 0; i < Array.getLength(rawArgument); i++) {
-                    ((CapturesArguments) m).captureFrom(Array.get(rawArgument, i));
+                for (int i = 0; i < getLength(rawArgument); i++) {
+                    ((CapturesArguments) m).captureFrom(getArgument(rawArgument, i));
                 }
             }
         }
+    }
+
+    private int getLength(Object rawArgument) {
+        return rawArgument == null ? 1 : Array.getLength(rawArgument);
+    }
+
+    private Object getArgument(Object rawArgument, int i) {
+        return rawArgument == null ? null : Array.get(rawArgument, i);
     }
 
     private int regularArgumentsSize(Invocation invocation) {
--- a/src/main/java/org/mockito/internal/invocation/InvocationMatcher.java
+++ b/src/main/java/org/mockito/internal/invocation/InvocationMatcher.java
@@ -141,19 +141,15 @@ public class InvocationMatcher implements DescribedInvocation, CapturesArguments
         for (ArgumentMatcher m : uniqueMatcherSet(indexOfVararg)) {
             if (m instanceof CapturesArguments) {
                 Object rawArgument = invocation.getRawArguments()[indexOfVararg];
-                for (int i = 0; i < getLength(rawArgument); i++) {
-                    ((CapturesArguments) m).captureFrom(getArgument(rawArgument, i));
+                if (rawArgument != null) {
+                    for (int i = 0; i < Array.getLength(rawArgument); i++) {
+                        ((CapturesArguments) m).captureFrom(Array.get(rawArgument, i));
+                    }
+                } else {
+                    ((CapturesArguments) m).captureFrom(null);
                 }
             }
         }
-    }
-
-    private int getLength(Object rawArgument) {
-        return rawArgument == null ? 1 : Array.getLength(rawArgument);
-    }
-
-    private Object getArgument(Object rawArgument, int i) {
-        return rawArgument == null ? null : Array.get(rawArgument, i);
     }
 
     private int regularArgumentsSize(Invocation invocation) {
--- a/src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
+++ b/src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
@@ -315,4 +315,17 @@ public class CapturingArgumentsTest extends TestBase {
         verify(mock).varargs(eq(42), argumentCaptor.capture());
         Assertions.assertThat(argumentCaptor.getValue()).contains("capturedValue");
     }
+
+    @Test
+    public void should_capture_null_vararg() {
+        // given
+        ArgumentCaptor<String> argumentCaptor = ArgumentCaptor.forClass(String.class);
+
+        // when
+        mock.varargs(null);
+
+        // then
+        verify(mock).varargs(argumentCaptor.capture());
+        Assertions.assertThat(argumentCaptor.getValue()).isNull();
+    }
 }
--- a/src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
+++ b/src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
@@ -317,12 +317,26 @@ public class CapturingArgumentsTest extends TestBase {
     }
 
     @Test
-    public void should_capture_null_vararg() {
+    public void should_capture_null_single_argument_in_vararg_method() {
         // given
         ArgumentCaptor<String> argumentCaptor = ArgumentCaptor.forClass(String.class);
 
         // when
         mock.varargs(null);
+
+        // then
+        verify(mock).varargs(argumentCaptor.capture());
+        Assertions.assertThat(argumentCaptor.getValue()).isNull();
+    }
+
+    @Test
+    public void should_capture_null_array_in_method_vararg() {
+        // given
+        ArgumentCaptor<String[]> argumentCaptor = ArgumentCaptor.forClass(String[].class);
+
+        // when
+        String[] nullArray = null;
+        mock.varargs(nullArray);
 
         // then
         verify(mock).varargs(argumentCaptor.capture());


src/main/java/org/mockito/internal/invocation/InvocationMatcher.java
```diff
@@ -141,13 +141,21 @@ private void captureVarargsPart(Invocation invocation) {
         for (ArgumentMatcher m : uniqueMatcherSet(indexOfVararg)) {
             if (m instanceof CapturesArguments) {
                 Object rawArgument = invocation.getRawArguments()[indexOfVararg];
-                for (int i = 0; i < Array.getLength(rawArgument); i++) {
-                    ((CapturesArguments) m).captureFrom(Array.get(rawArgument, i));
+                for (int i = 0; i < getLength(rawArgument); i++) {
```
TimvdLippe: Instead of magically changing the length of the array, I would rather introduce an if-else around the for-loop, to make the intent clearer. With this implementation there is a hidden connection between `getLength` and `getArgument`

TimvdLippe:
Small implementation comment, other than that LGTM. Thanks for debugging and fixing this issue!

src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
```diff
@@ -315,4 +315,17 @@ public void captures_correctly_when_captor_used_on_pure_vararg_method() throws E
...
+        // given
+        ArgumentCaptor<String> argumentCaptor = ArgumentCaptor.forClass(String.class);
+
+        // when
+        mock.varargs(null);
+
+        // then
+        verify(mock).varargs(argumentCaptor.capture());
+        Assertions.assertThat(argumentCaptor.getValue()).isNull();
+    }
```
bric3: Could you add another test with casting?

in the end we will have a test with `mock.varargs((Type[]) null)` and a test with `mock.varargs((Type) null)`

This allows thorough testing and documentation of each caseof.

src/main/java/org/mockito/internal/invocation/InvocationMatcher.java
bric3: Agreed

src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
```diff
@@ -315,4 +315,17 @@ public void captures_correctly_when_captor_used_on_pure_vararg_method() throws E
...
         Assertions.assertThat(argumentCaptor.getValue()).contains("capturedValue");
     }
+
+    @Test
+    public void should_capture_null_vararg() {
+        // given
+        ArgumentCaptor<String> argumentCaptor = ArgumentCaptor.forClass(String.class);
+
+        // when
+        mock.varargs(null);
```
ChristianSchwarz: Hi @lukasz-szewc, 

thanks for you PR! Please note this test doesn't test the captoring of an null-array (`(String[])null`) but an null-argument (`new String[]{null})`. 
Can you add a test with `mock.varargs((String[])null)` to verfiy #583 is fixed?

src/main/java/org/mockito/internal/invocation/InvocationMatcher.java
lukasz-szewc: true,
however code with if-else will be longer - but you know - there are no perfect solutions.

src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
lukasz-szewc: Hi @ChristianSchwarz 

I will add that test and let see what happens. Should work :)

src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
```diff
@@ -317,7 +317,7 @@ public void captures_correctly_when_captor_used_on_pure_vararg_method() throws E
     }
 
     @Test
-    public void should_capture_null_vararg() {
+    public void should_capture_null_single_argument_in_vararg_method() {
```
bric3: This one is not explicit about casting.

Sorry I'm picky and on a mobile phone ;)

lukasz-szewc: take a look at second test method I have added : should_capture_null_array_in_method_vararg

src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
lukasz-szewc: Take a look at: should_capture_null_array_in_method_vararg. Is it sufficient ?

src/main/java/org/mockito/internal/invocation/InvocationMatcher.java
```diff
@@ -141,8 +141,12 @@ private void captureVarargsPart(Invocation invocation) {
...
             if (m instanceof CapturesArguments) {
                 Object rawArgument = invocation.getRawArguments()[indexOfVararg];
-                for (int i = 0; i < Array.getLength(rawArgument); i++) {
-                    ((CapturesArguments) m).captureFrom(Array.get(rawArgument, i));
+                if (rawArgument != null) {
+                    for (int i = 0; i < Array.getLength(rawArgument); i++) {
+                        ((CapturesArguments) m).captureFrom(Array.get(rawArgument, i));
+                    }
+                } else {
+                    ((CapturesArguments) m).captureFrom(null);
```
ChristianSchwarz: Should we capture null in case an null-array was passed? How can it be distinguished from an null-element? 
IMO it should be ignored cause `null` is strictly speaken an `<Type[]>` but the captor bound to `<Type>`.

src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
```diff
@@ -315,4 +315,31 @@ public void captures_correctly_when_captor_used_on_pure_vararg_method() throws E
...
+        // given
+        ArgumentCaptor<String[]> argumentCaptor = ArgumentCaptor.forClass(String[].class);
+
+        // when
+        String[] nullArray = null;
+        mock.varargs(nullArray);
+
+        // then
+        verify(mock).varargs(argumentCaptor.capture());
+        Assertions.assertThat(argumentCaptor.getValue()).isNull();
```
ChristianSchwarz: In this case noting should be captured, see [my comment above](https://github.com/mockito/mockito/pull/606/files#r76538541).

lukasz-szewc: @ChristianSchwarz not sure about that, because 'null' has been passed to method, hasn't it ?

If 'noting should be captured' then what getValue method should return in that case?

ChristianSchwarz: Note that null can be the vararg array it self or an element of the vararg array. In this case        we talk about the vararg array. 

getValues() must return an empty List in this case, getValue() throws an exception if nothing was captured.

TimvdLippe:
Regardless if we accept this PR or not, shall we postpone merging for 2.2 since we have a code freeze right now for the release candidate? I think our focus right now must be documentation to release ASAP.

src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
ChristianSchwarz: If we capture a `null` vararg-array it might block #584 and #565, so I vote to keep it out!

lukasz-szewc: if we keep it out there will be NPE in the code right now. 
What is better: allow NPE or return null (even if that is not desired value) ?
I vote: in.

TimvdLippe: My expectation as a user would be that the captor does capture null, just like any other null invocation.

bric3:
I'm postponing this one with the rest of the Argument captor refactoring that will happen after mockito.

src/test/java/org/mockitousage/matchers/CapturingArgumentsTest.java
ChristianSchwarz: @lukasz-szewc 
I think we speak about two different things here. I mean we should remove the capturing part in case the vararg-array is null not the vararg element that is null. It should work if you simply remove the [*else* part in line 149](https://github.com/mockito/mockito/pull/606/files#diff-6bedbf35a65e3a44c8e6c7f033172013R149). This change should not cause an NPE.

----

@TimvdLippe 

> My expectation as a user would be that the captor does capture null, just like any other null invocation.

My expectation is that a ArgumentCaptor<String> only capture Strings, but if Ian String[] is passed even if it is null then it should capture nothing, this relates to #565.

PR #613, opened 2016-08-29
ArgumentCaptor and ArgumentMatchers can now mixed in varargs

### LINKED ISSUE #439: ArgumentCaptor and ArgumentMatcher can't be mixed in varargs
In the given test below the ArgumentCaptor should captor only the char 'c':

```
    @Captor
    private ArgumentCaptor<Character> argumentCaptor;
    
    @Test
    public void capturesVararg() throws Exception {
        mock.varargs(42, 'c');

        verify(mock).varargs(eq(42), argumentCaptor.capture());
       
       assertThat(argumentCaptor.getAllValues()).containsExactly('c');
    }
```

The test fails with this message:

```
  <2>
while expected size was:
  <1>
Actual was:
  <[42, 'c']>
Expected was:
  <['c']>
```

Note that we see here 2 bugs:
 - a wrong argument index was captored
 - an `int `was captored by an  `ArgumentCaptor<Character> ` -> #565

ChristianSchwarz:
fixes #439 
 * ArgumentCaptor and ArgumentMatchers can now mixed in varargs
 * added 25 regression tests in VarargsTest (5 are ignored cause they belong to other issues and fail currently)
 * refactored InvocationMatcher:
  * removed duplicates and unneccesary code
  * simplified the logic as much as possible
  * moved caputing logic to InvocationArgumentCaptor (if you have a better name let me know)

### DIFF
--- a/src/main/java/org/mockito/internal/invocation/InvocationArgumentCaptor.java
+++ b/src/main/java/org/mockito/internal/invocation/InvocationArgumentCaptor.java
@@ -1,0 +1,90 @@
+package org.mockito.internal.invocation;
+
+import static java.lang.reflect.Array.get;
+import static java.lang.reflect.Array.getLength;
+
+import java.lang.reflect.Method;
+import java.util.List;
+import org.mockito.ArgumentMatcher;
+import org.mockito.internal.matchers.CapturesArguments;
+import org.mockito.invocation.Invocation;
+
+
+/**
+ * 
+ * @author Christian Schwarz
+ *
+ */
+public class InvocationArgumentCaptor {
+
+    private InvocationArgumentCaptor() {
+    }
+    
+    public static void captureArguments(Invocation invocation,List<ArgumentMatcher<?>> matchers) {
+        captureRegularArguments(invocation,matchers);
+        if (invocation.getMethod().isVarArgs()) {
+            captureVarargsPart(invocation,matchers);
+        }
+    }
+
+    private static void captureRegularArguments(Invocation invocation,List<ArgumentMatcher<?>> matchers) {
+        for (int position = 0; position < regularArgumentsSize(invocation); position++) {
+            ArgumentMatcher<?> m = matchers.get(position);
+            if (m instanceof CapturesArguments) {
+                ((CapturesArguments) m).captureFrom(invocation.getArgument(position));
+            }
+        }
+    }
+
+    private static void captureVarargsPart(Invocation invocation,List<ArgumentMatcher<?>> matchers) {
+        int indexOfVararg = invocation.getRawArguments().length - 1;
+        
+        ArgumentMatcher<?>[] varargMatchers =varargMatchers(matchers,indexOfVararg);
+        Object varargsArray = invocation.getRawArguments()[indexOfVararg];
+        
+        if (varargMatchers.length>1){
+            captureEachWithItsDedicatedCaptor(varargsArray,varargMatchers);
+        }else if (varargMatchers.length==1){
+            captureAllWithASingleCaptor(varargsArray, varargMatchers[0]);
+            
+        }
+    }
+
+    private static void captureAllWithASingleCaptor(Object varargsArray, ArgumentMatcher<?> varargMatcher) {
+        if (!(varargMatcher instanceof CapturesArguments))
+            return;
+        CapturesArguments capturingMatcher = (CapturesArguments) varargMatcher;
+
+        for (int i = 0; i < getLength(varargsArray); i++) {
+            capturingMatcher.captureFrom(get(varargsArray, i));
+        }
+    }
+
+    private static void captureEachWithItsDedicatedCaptor(Object varargsArray , ArgumentMatcher<?>[] matchers) {
+        for (int i = 0; i < matchers.length; i++) {
+            ArgumentMatcher<?> m = matchers[i];
+            if (m instanceof CapturesArguments) {
+                ((CapturesArguments) m).captureFrom(get(varargsArray, i));
+            }
+        }
+    }
+
+    private static ArgumentMatcher<?>[] varargMatchers(List<ArgumentMatcher<?>> matchers,int indexOfVararg) {
+        int varargsSize = matchers.size()-indexOfVararg;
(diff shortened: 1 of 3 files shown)


szczepiq:
Great PR:
- it neatly describes the high level design - I don't need to look into the code to find out the design
- at adds tangible value to Mockito, making the library more intuitive to use. Library such as Mockito becomes great when we apply consistent effort over time and improve it one step at the time :D
- it's not a refactoring-only PR :))))

Thanks!!! Keep it coming!

src/main/java/org/mockito/internal/invocation/InvocationArgumentCaptor.java
```diff
@@ -0,0 +1,90 @@
...
+import org.mockito.internal.matchers.CapturesArguments;
+import org.mockito.invocation.Invocation;
+
+
+/**
+ * 
+ * @author Christian Schwarz
+ *
+ */
+public class InvocationArgumentCaptor {
```
bric3: Can you remove empty comment ?

src/main/java/org/mockito/internal/invocation/InvocationMatcher.java
```diff
@@ -5,98 +5,102 @@
...
-import java.lang.reflect.Method;
-import java.util.*;
-
-@SuppressWarnings("unchecked")
 /**
- * In addition to all content of the invocation, the invocation matcher contains the argument matchers.
- * Invocation matcher is used during verification and stubbing.
- * In those cases, the user can provide argument matchers instead of 'raw' arguments.
- * Raw arguments are converted to 'equals' matchers anyway.
+ * In addition to all content of the invocation, the invocation matcher contains the argument matchers. Invocation matcher is used during verification and stubbing. In those cases, the user can provide argument matchers instead of 'raw' arguments. Raw arguments are converted to 'equals' matchers anyway.
```
bric3: Could you keep the comment intact ;)

src/main/java/org/mockito/internal/invocation/InvocationMatcher.java
```diff
@@ -5,98 +5,102 @@
...
  */
+@SuppressWarnings("serial")
 public class InvocationMatcher implements DescribedInvocation, CapturesArgumentsFromInvocation, Serializable {
 
     private final Invocation invocation;
-    private final List<ArgumentMatcher> matchers;
+    private final List<ArgumentMatcher<?>> matchers;
```
bric3: Sometime raw type is ok

src/main/java/org/mockito/internal/invocation/InvocationMatcher.java
```diff
@@ -5,98 +5,102 @@
...
-                && argumentsMatch(this, actual);
-    }
-
-    private boolean safelyArgumentsMatch(Object[] actualArgs) {
-        try {
-            return argumentsMatch(this, actualArgs);
-        } catch (Throwable t) {
-            return false;
-        }
+        return invocation.getMock().equals(actual.getMock()) && hasSameMethod(actual) && argumentsMatch(this, actual);
```
bric3: style comment, but the vertical alignement is nicer to the eye when parsing this code

src/test/java/org/mockitousage/matchers/VarargsTest.java
```diff
@@ -0,0 +1,259 @@
...
+
+    @Test
+    public void shouldMatchVarArgs_noArgs() {
+        mock.varargs();
+
+        verify(mock).varargs();
+    }
+
+    @Test
+    @Ignore
```
bric3: That would be nice to describe why it's ignored

src/test/java/org/mockitousage/matchers/VarargsTest.java
```diff
@@ -0,0 +1,259 @@
...
+        mock.varargsbyte();
+
+        verify(mock).varargsbyte((byte[]) isNotNull());
+    }
+
+    @Test
+    public void shouldMatchVarArgs_oneArgIsNotNull() {
+        mock.varargsbyte((byte) 1);
+
+        verify(mock).varargsbyte((byte[]) isNotNull());
```
bric3: When we start Java 8 these casts could be removed, at last :P

bric3:
For me this code is really nice, this is definitely an improvement !
I would like however to see a unit test on `InvocationArgumentCaptor` this way it is easier to measure coverage overthere.

bric3:
I will wait for the release of mockito 2.1 though before merging this one. (As you may be aware we are skipping the number 2.0 in favor of 2.1 due to semantic versioning issues we had)

src/test/java/org/mockitousage/matchers/VarargsTest.java
ChristianSchwarz: I'am afraid java 8 won't help us here. The dev must decide if an element should be passed to the vararg-part or an array.

src/test/java/org/mockitousage/matchers/VarargsTest.java
ChristianSchwarz: Good idea! I opened  #616 to reference it in the `@Ignore` annotation.

src/main/java/org/mockito/internal/invocation/InvocationMatcher.java
ChristianSchwarz: Hmm I don't like warnings...

src/main/java/org/mockito/internal/invocation/InvocationArgumentCaptor.java
ChristianSchwarz: Sure.

ChristianSchwarz:
@szczepiq 
@bric3 
Thanks for the great feedback! I am going to incorporate the requested changes during the coming week end.

src/test/java/org/mockitousage/matchers/VarargsTest.java
bric3: Oh I see what you meant in the code then!

ChristianSchwarz:
I close this PR it is superseded by #635.

PR #617, opened 2016-08-31
Moves arity interfaces of java8 helper answers to public API

### LINKED ISSUE #614: AdditionalAnswers.answer family leaks internal classes
`AdditionalAnswers.answer` family leaks the internal `AnswerFunctionalInterfaces`, I believe these _arity_ interfaces should be public, either in the package `org.mockito.stubbing` or within `org.mockito.stubbing.Answer`

```
public static <T, A, B> Answer<T> answer(AnswerFunctionalInterfaces.Answer1<T, A> answer)
public static <A, B> Answer<Void> answerVoid(AnswerFunctionalInterfaces.VoidAnswer1<A> answer) {
public static <T, A, B> Answer<T> answer(AnswerFunctionalInterfaces.Answer2<T, A, B> answer) {
...
```

bric3:
The Java 8 helper interfaces have been introduced in #338. However the public `AdditionalAnswers.answer` family leak the functional interfaces.

These _parameter arity_ interfaces can be public, hence they are moved next to `Answer` in the `org.mockito.stubbing` package.

I'm however in favor of adding the `@Incubating` annotation on those APIs and related objects. Thoughts ?

Fixes #614

### DIFF
--- a/src/main/java/org/mockito/AdditionalAnswers.java
+++ b/src/main/java/org/mockito/AdditionalAnswers.java
@@ -216,6 +216,7 @@ public class AdditionalAnswers {
      * @return the answer object to use
      * @since 2.0.0
      */
+    @Incubating
     public static <T, A> Answer<T> answer(Answer1<T, A> answer) {
         return toAnswer(answer);
     }
@@ -228,6 +229,7 @@ public class AdditionalAnswers {
      * @return the answer object to use
      * @since 2.0.0
      */
+    @Incubating
     public static <A> Answer<Void> answerVoid(VoidAnswer1<A> answer) {
         return toAnswer(answer);
     }
@@ -242,6 +244,7 @@ public class AdditionalAnswers {
      * @return the answer object to use
      * @since 2.0.0
      */
+    @Incubating
     public static <T, A, B> Answer<T> answer(Answer2<T, A, B> answer) {
         return toAnswer(answer);
     }
@@ -255,6 +258,7 @@ public class AdditionalAnswers {
      * @return the answer object to use
      * @since 2.0.0
      */
+    @Incubating
     public static <A, B> Answer<Void> answerVoid(VoidAnswer2<A, B> answer) {
         return toAnswer(answer);
     }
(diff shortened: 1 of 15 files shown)


src/main/java/org/mockito/Answers.java
```diff
@@ -77,6 +77,7 @@
      * <p>Please see the {@link org.mockito.Mockito#RETURNS_SELF} documentation for more details.</p>
      *
      * @see org.mockito.Mockito#RETURNS_SELF
+     * @see org.mockito.Mockito#RETURNS_SELF
```
TimvdLippe: Hm this seems unnecessary?

src/main/java/org/mockito/BDDMockito.java
```diff
@@ -336,7 +336,10 @@ public void shouldHaveNoMoreInteractions() {
         BDDStubber will(Answer<?> answer);
 
         /**
-         * See original {@link Stubber#doNothing()}
+         * See original {@link Stubber#doNothing()}.
+         *
+         * This method will be removed in future release, probably in version 3.0.0
+         *
          * @since 1.8.0
          * @deprecated please use {@link #willDoNothing()} instead
```
TimvdLippe: `@deprecated as of 2.0.0` while we are at it.

src/main/java/org/mockito/BDDMockito.java
```diff
@@ -336,7 +336,10 @@ public void shouldHaveNoMoreInteractions() {
         BDDStubber will(Answer<?> answer);
 
         /**
-         * See original {@link Stubber#doNothing()}
+         * See original {@link Stubber#doNothing()}.
+         *
+         * This method will be removed in future release, probably in version 3.0.0
```
TimvdLippe: I think we can remove the word probably.

TimvdLippe:
Some very small doc comments. LGTM also with the `@Incubating` :+1:

src/main/java/org/mockito/Answers.java
bric3: Bad key stroke I believe thanks

src/main/java/org/mockito/AdditionalAnswers.java
```diff
@@ -219,8 +229,9 @@
      * @return the answer object to use
      * @since 2.0.0
      */
-    public static <A, B> Answer<Void> answerVoid(AnswerFunctionalInterfaces.VoidAnswer1<A> answer) {
-        return AnswerFunctionalInterfaces.toAnswer(answer);
+    @Incubating
```
szpak: :+1:

Issue #601, opened 2016-08-26
One test fails probably undeterministically 


lukasz-szewc:
Hey,

Check console output for this build: 
https://travis-ci.org/mockito/mockito/jobs/155443990

It seems that one test from time to time for no apparent reason.

```
org.mockitousage.verification.VerificationWithTimeoutTest > canIgnoreInvocationsWithJunit FAILED
    Wanted but not invoked:
    mock.oneArg('1');
    -> at org.mockitousage.verification.VerificationWithTimeoutTest.canIgnoreInvocationsWithJunit(VerificationWithTimeoutTest.java:147)
    Actually, there were zero interactions with this mock.
        at org.mockito.internal.exceptions.Reporter.wantedButNotInvoked(Reporter.java:329)
        at org.mockito.internal.verification.checkers.MissingInvocationChecker.checkMissingInvocation(MissingInvocationChecker.java:37)
        at org.mockito.internal.verification.Times.verify(Times.java:37)
        at org.mockito.internal.verification.VerificationOverTimeImpl.verify(VerificationOverTimeImpl.java:78)
        at org.mockito.internal.verification.VerificationWrapper.verify(VerificationWrapper.java:15)
        at org.mockito.internal.verification.MockAwareVerificationMode.verify(MockAwareVerificationMode.java:21)
        at org.mockito.internal.handler.MockHandlerImpl.handle(MockHandlerImpl.java:73)
        at org.mockito.internal.handler.NullResultGuardian.handle(NullResultGuardian.java:32)
        at org.mockito.internal.handler.InvocationNotifierHandler.handle(InvocationNotifierHandler.java:36)
        at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.doIntercept(MockMethodInterceptor.java:36)
        at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.access$000(MockMethodInterceptor.java:17)
        at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor$DispatcherDefaultingToRealMethod.interceptAbstract(MockMethodInterceptor.java:113)
        at org.mockitousage.IMethods$MockitoMock$1357077317.oneArg(Unknown Source)
        at org.mockitousage.verification.VerificationWithTimeoutTest.canIgnoreInvocationsWithJunit(VerificationWithTimeoutTest.java:147)
```

lukasz-szewc:
Well, I looked closely to this test, and it fails for a very good reason.
[...]

PR #619, opened 2016-09-01
Fixes #548 from now on verification happens always call in other thre…

### LINKED ISSUE #548: VerificationWithTimeoutTest is unstable
See build failure: https://travis-ci.org/mockito/mockito/jobs/151406672

Full stack trace:

```
org.mockitousage.verification.VerificationWithTimeoutTest > shouldAllowTimeoutVerificationInOrder FAILED
    org.mockito.exceptions.verification.VerificationInOrderFailure: 
    Verification in order failure
    Wanted but not invoked:
    mock.oneArg('1');
    -> at org.mockitousage.verification.VerificationWithTimeoutTest.shouldAllowTimeoutVerificationInOrder(VerificationWithTimeoutTest.java:161)
    Wanted anywhere AFTER following interaction:
    mock.oneArg('x');
    -> at org.mockitousage.verification.VerificationWithTimeoutTest.shouldAllowTimeoutVerificationInOrder(VerificationWithTimeoutTest.java:155)
```

lukasz-szewc:
Should fix problems with #548 and #601

### DIFF
--- a/src/test/java/org/mockitousage/verification/DelayedExecution.java
+++ b/src/test/java/org/mockitousage/verification/DelayedExecution.java
@@ -1,0 +1,54 @@
+package org.mockitousage.verification;
+
+import org.mockitousage.IMethods;
+
+import java.util.ArrayList;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+class DelayedExecution {
+    private final ScheduledExecutorService executor;
+    private final IMethods mock;
+    private final int delay;
+    private final TimeUnit timeUnit;
+    private final ArrayList<Character> invocations = new ArrayList<Character>();
+
+    DelayedExecution(ScheduledExecutorService executor, IMethods mock, int delay, TimeUnit timeUnit) {
+        this.executor = executor;
+        this.mock = mock;
+        this.delay = delay;
+        this.timeUnit = timeUnit;
+    }
+
+    void recordAsyncCall(char c) {
+        invocations.add(c);
+    }
+
+    void allAsyncCallsStarted() throws InterruptedException {
+        final CountDownLatch countDownLatch = new CountDownLatch(invocations.size());
+        for (final Character invocation : invocations) {
+            executor.execute(runnable(countDownLatch, invocation));
+        }
+        countDownLatch.await();
+    }
+
+    private Runnable runnable(final CountDownLatch countDownLatch, final Character invocation) {
+        return new Runnable() {
+            @Override
+            public void run() {
+                countDownLatch.countDown();
+                sleep();
+                mock.oneArg(invocation.charValue());
+            }
+        };
+    }
+
+    private void sleep() {
+        try {
+            timeUnit.sleep(delay);
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+    }
+}
--- a/src/test/java/org/mockitousage/verification/VerificationWithTimeoutTest.java
+++ b/src/test/java/org/mockitousage/verification/VerificationWithTimeoutTest.java
@@ -6,7 +6,6 @@
 package org.mockitousage.verification;
 
 import static java.util.concurrent.Executors.newSingleThreadScheduledExecutor;
-import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.junit.rules.ExpectedException.none;
 import static org.mockito.Mockito.inOrder;
@@ -17,8 +16,9 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoMoreInteractions;
 import static org.mockito.junit.MockitoJUnit.rule;
 
+import java.util.ArrayList;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
 
 import org.junit.After;
 import org.junit.Before;
@@ -60,14 +60,16 @@ public class VerificationWithTimeoutTest {
     @Test
     public void shouldVerifyWithTimeout() throws Exception {
         // given
-        callAsyncWithDelay(mock, 'c', 20, MILLISECONDS);
+        DelayedExecution delayedExecution = new DelayedExecution(executor, mock, 20);
+        delayedExecution.recordAsyncCall('c');
+
+        // when
+        delayedExecution.allAsyncCallsStarted();
 
         // then
         verify(mock, timeout(100)).oneArg('c');
-
         verify(mock, timeout(100).atLeastOnce()).oneArg('c');
         verify(mock, timeout(100).times(1)).oneArg('c');
-
         verify(mock).oneArg('c');
         verify(mock, times(1)).oneArg('c');
     }
@@ -75,11 +77,14 @@ public class VerificationWithTimeoutTest {
     @Test
     public void shouldFailVerificationWithTimeout() throws Exception {
         // given
-        callAsyncWithDelay(mock, 'c', 30, MILLISECONDS);
+        DelayedExecution delayedExecution = new DelayedExecution(executor, mock, 30);
+        delayedExecution.recordAsyncCall('c');
+
+        // when
+        delayedExecution.allAsyncCallsStarted();
 
         // then
         verify(mock, never()).oneArg('c');
-
         exception.expect(MockitoAssertionError.class);
         verify(mock, timeout(20).atLeastOnce()).oneArg('c');
     }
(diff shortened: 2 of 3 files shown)


TimvdLippe:
I am going to restart this build a couple of times to test the robustness.

lukasz-szewc:
It could take a  while. Hopefully we have loops.

In the meantime try this gist : https://gist.github.com/lukasz-szewc/f1ca9cece0a8997e898e5b6ba1a65f0c

TimvdLippe:
Got a failed Travis job: https://travis-ci.org/mockito/mockito/jobs/156890987. The culprit seems to be `org.mockitousage.verification.VerificationAfterDelayTest > shouldVerifyNormallyWithSpecificTimes`

lukasz-szewc:
I have not touched `VerificationAfterDelayTest` in this PR...

src/test/java/org/mockitousage/verification/VerificationWithTimeoutTest.java
```diff
@@ -161,13 +193,47 @@ public void shouldAllowTimeoutVerificationInOrder() throws Exception {
...
+                executor.execute(runnable(countDownLatch, invocation));
+            }
+            countDownLatch.await();
+        }
+
+        private Runnable runnable(final CountDownLatch countDownLatch, final Character invocation) {
+            return new Runnable() {
+                @Override
+                public void run() {
+                    countDownLatch.countDown();
```
TimvdLippe: Shouldn't this statement be moved as last statement? If I read the docs correctly, `countDownLatch.await()` will proceed when all Threads have called `countDownLatch.countDown()`.

TimvdLippe:
Yes I know, was merely stating that the Travis failure we experienced was a different test. Nonetheless, I think it is best if we aggregrate all test fixes in a PR to study the robustness. Would you be willing to do so?

lukasz-szewc:
Yup, another commit with `VerificationAfterDelayTest` changes. Take a look.

TimvdLippe:
Great thanks, will be restarting these changes a couple of times too. Really nice to have deterministic results now! :tada:

src/test/java/org/mockitousage/verification/DelayedExecution.java
```diff
@@ -0,0 +1,51 @@
...
+            executor.execute(runnable(countDownLatch, invocation));
+        }
+        countDownLatch.await();
+    }
+
+    private Runnable runnable(final CountDownLatch countDownLatch, final Character invocation) {
+        return new Runnable() {
+            @Override
+            public void run() {
+                countDownLatch.countDown();
```
TimvdLippe: Adding my previous comment again:

Shouldn't this statement be moved as last statement? If I read the docs correctly, `countDownLatch.await()` will proceed when all Threads have called `countDownLatch.countDown()`.

src/test/java/org/mockitousage/verification/VerificationWithTimeoutTest.java
lukasz-szewc: Yes and no.  :)

If I move it as a last line in run method then I guarantee that invocation on mock has been performed and in that case any time verification (later in test) is pretty pointless. It would not make a difference if the time slot is 1 millisecond or 1000 milliseconds. 

On the other hand, If I move the countDown before execution on mock (like it is proposed right now), then verification after some time makes sense because we are executing at least two activities in parallel.

src/test/java/org/mockitousage/verification/DelayedExecution.java
lukasz-szewc: Yes and no. :)

If I move it as a last line in run method then I guarantee that invocation on mock has been performed and in that case any time verification (later in test) is pretty pointless. It would not make a difference if the time slot is 1 millisecond or 1000 milliseconds.

On the other hand, If I move the countDown before execution on mock (like it is proposed right now), then verification after some time makes sense because we are executing at least two activities in parallel.

src/test/java/org/mockitousage/verification/VerificationWithTimeoutTest.java
bric3: @lukasz-szewc But the goal of the test is not to test these invocation but to test the timeout verification mode. So when we want to ensure calls have been finished before the verification is virtually OK.

Otherwise that was pretty much the idea I had in mind for timeout tests. And we can have the same kind of tricks for the after timeout.

bric3:
I'm ok with the change. Even if the fact that async are not completely deterministic, it definitely an improvement, plus the test code is elegant!

src/test/java/org/mockitousage/verification/DelayedExecution.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ScheduledExecutorService;
+
+class DelayedExecution {
+    private final ScheduledExecutorService executor;
+    private final IMethods mock;
+    private final int delay;
+    private final ArrayList<Character> invocations = new ArrayList<Character>();
+
+    public DelayedExecution(ScheduledExecutorService executor, IMethods mock, int delay) {
```
bric3: Could you add the TimeUnit parameter?

bric3:
@lukasz-szewc Also after adding the `TimeUnit` parameter to `DelayedExection` ? Could rebase your commits (`added one missing line` is not really the most expressive message) ;)

src/test/java/org/mockitousage/verification/VerificationWithTimeoutTest.java
```diff
@@ -137,21 +156,32 @@ public void shouldAllowMixingOnlyWithTimeoutAndFail() throws Exception {
...
+        delayedExecution.recordAsyncCall('2');
+
         // when
-        callAsyncWithDelay(mock, '0', 0, MILLISECONDS);
-        callAsyncWithDelay(mock, '1', 0, MILLISECONDS);
-        callAsyncWithDelay(mock, '2', 20, MILLISECONDS);
+        delayedExecution.allAsyncCallsStarted();
 
         // then
+        verify(mock, timeout(50)).oneArg('0');
```
ChristianSchwarz: @lukasz-szewc 
This line must be removed cause this was not tested in the original test case. For the same reason the call with '2' must be delayed by 20ms.

bric3: @ChristianSchwarz Good point. I overlooked this one thinking the countdown latch semantic forbid that, but there is two way to handle this 
* it's possible add a new instance of `DelayedExecution` and insert `allAsyncCallsStarted` before the last verify
* tweak the `DelayedExecution` API to be more fine grained

About the missing verify I noted it as well in the original test as well, but it may be de to a human error so I'm ok with the addition.

ChristianSchwarz:
I like the idea of starting the threads and await there readiness, this might solve the timing issues. An other option is to write an simplified ScheduledExecutorService with the same behavior as the DelayedExecution of this PR. This way the test can remain untouched and we have still the possibiliy run tasks with individual delay as required by one test.

src/test/java/org/mockitousage/verification/DelayedExecution.java
lukasz-szewc: I will

src/test/java/org/mockitousage/verification/VerificationWithTimeoutTest.java
lukasz-szewc: I will keep original intent of that test. However this particular test is at least weird and I have no idea what this test is actually testing

lukasz-szewc:
1. added `TimeUnit` to constructor
2. reworked one of test in `VerificationWithTimeoutTest' mentioned above
3. rebased and pushed with force

take a look.

src/test/java/org/mockitousage/verification/DelayedExecution.java
```diff
@@ -0,0 +1,54 @@
...
+                countDownLatch.countDown();
+                sleep();
+                mock.oneArg(invocation.charValue());
+            }
+        };
+    }
+
+    private void sleep() {
+        try {
+            timeUnit.sleep(delay);
```
bric3: @lukasz-szewc Actually the idea of passing the TimeUnit is to do a `TimeUnit.toMillis(delay)` here ;)

src/test/java/org/mockitousage/verification/VerificationWithTimeoutTest.java
```diff
@@ -137,11 +157,14 @@ public void shouldAllowMixingOnlyWithTimeoutAndFail() throws Exception {
...
+        // given
+        DelayedExecution delayedExecution = createZeroMillisDelayedExecution();
+        DelayedExecution twentyMillisDelayedExecution = createTwentyMillisDelayedExecution();
+
         // when
-        callAsyncWithDelay(mock, '0', 0, MILLISECONDS);
-        callAsyncWithDelay(mock, '1', 0, MILLISECONDS);
-        callAsyncWithDelay(mock, '2', 20, MILLISECONDS);
+        delayedExecution.allAsyncCallsStarted();
+        twentyMillisDelayedExecution.allAsyncCallsStarted();
```
bric3: @lukasz-szewc I'd rather propose

```java
delayedExecution.allAsyncCallsStarted();

verify(mock, timeout(50)).oneArg('1');
twentyMillisDelayedExecution.allAsyncCallsStarted();
verify(mock, timeout(50)).oneArg('2');
```

src/test/java/org/mockitousage/verification/DelayedExecution.java
ChristianSchwarz: @bric3 @lukasz-szewc 
better use `timeUnit.sleep(delay)` here ;-)

bric3: That works too ;)

bric3: I think I bugged there it's already `timeUnit.sleep(delay);`

Issue #596, opened 2016-08-24
Review and update documentation for 2.1.0


szczepiq:
- [x] update the migration guide (should be easy, there are not many breaking changes, if we miss out we can update in following RC)
- [x] add information about bytebuddy in the migration guide
- [x] mention noteworthy 2.0.0 features (should be easy, there are not many, plus they are already documented :)
- [x] fix odd TODO #593
- [x] review/update @since tags so that they say 2.1.0

Final steps I plan doing (by the EOD Wednesday, tomorrow), in order:

- [x] Test release from release/test (in progress, see Travis CI)
- [x] Merge/cherry-pick changes from "release/test" to master
- [x] Move announcement / motivation text (https://github.com/mockito/mockito/blob/rc-release-notes-demo/doc/release-notes/official.md) to top of "What's new in Mockito 2"
- [x] Final pass over "What's new in Mockito 2", make the information exciting and compelling, emphasize use cases, add missing links
- [x] Update the 2.1.0 information in main Mockito class to link to "What's new in Mockito 2". See "TODO 596" in code
- [x] Review Continuous Delivery [wiki page](https://github.com/mockito/mockito/wiki/Mockito-Continuous-Delivery)
- [x] Update the announcement text (https://github.com/mockito/mockito/blob/rc-release-notes-demo/doc/release-notes/official.md) so that instead of tons of text it just links to "What's new in Mockito 2"
- [x] Review release checklist (mostly done) - #620
- [x] Optional - Add some info in the 'What's new in Mockito 2' about the next steps for engs (e.g. ask to use the new version and give feedback).

TimvdLippe:
There were separate issues for these points. For example the last point was already fixed by @brice3 in a recent PR

szczepiq:
This list nicely summarizes the incompatible changes: https://github.com/mockito/mockito/issues?utf8=✓&q=label%3A%221.*%20incompatible%22%20

TimvdLippe:
@bric3 has started the migration guide: https://github.com/mockito/mockito/wiki/What's-new-in-mockito-2.0---2.1

TimvdLippe:
Migration guide has been updated and I did a small spell check. That LGTM.

bric3:
Thanks @TimvdLippe
@raphw @szpak @marcingrzejszczak @szczepiq if you want to add other stuff to this migration guide ?

raphw:
[...]

PR #623, opened 2016-09-06
Updated the javadocs

szczepiq:
Polishing the javadocs, starting with this small PR:

Issue #596
- updated @since to 2.1.0 (instead 2.0.0)
- added more docs around the verification collector

### DIFF
--- a/src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsMoreEmptyValues.java
+++ b/src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsMoreEmptyValues.java
@@ -13,7 +13,7 @@ import java.io.Serializable;
 import java.lang.reflect.Array;
 
 /**
- * It's likely this implementation will be used by default by every Mockito 2.0.0 mock.
+ * It's likely this implementation will be used by default by every Mockito 2.1.0 mock.
  * <p>
  * Currently <b>used only</b> by {@link Mockito#RETURNS_SMART_NULLS}
  * <p>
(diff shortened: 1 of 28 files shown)


TimvdLippe:
LGTM

src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsMoreEmptyValues.java
```diff
@@ -13,7 +13,7 @@
 import java.lang.reflect.Array;
 
 /**
- * It's likely this implementation will be used by default by every Mockito 2.0.0 mock.
+ * It's likely this implementation will be used by default by every Mockito 2.1.0 mock.
```
bric3: In the end this is not the case.

szczepiq: Good catch. 3.0.0? :)

bric3: Why not, you advocated on the google code tracker the change. I forgot about it too. I think this answer is good.

If you agree and I agree too then let's make the change ;)

szczepiq: I agree. It's a good change I think. However, it's a scope creep and needs documentation update. Also many tests may fail :) Do you want to include it in 2.1.0?

bric3: I meant for 3.0.0 ;)

PR #628, opened 2016-09-09
Fixed javadoc documentation in the main class

szczepiq:
Checks out line item in #596

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -30,7 +30,7 @@ import org.mockito.junit.*;
  * <h1>Contents</h1>
  *
  * <b>
- *      <a href="#0">0. Migrating to 2.1.0</a><br/>
+ *      <a href="#0">0. Migrating to Mockito 2</a><br/>
  *      <a href="#1">1. Let's verify some behaviour! </a><br/>
  *      <a href="#2">2. How about some stubbing? </a><br/>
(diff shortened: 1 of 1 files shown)


src/main/java/org/mockito/Mockito.java
```diff
@@ -68,36 +68,25 @@
  *      <a href="#35">35. (new) Custom verification failure message (Since 2.1.0)</a><br/>
  *      <a href="#36">36. (new) Java 8 Lambda Matcher Support (Since 2.1.0)</a><br/>
  *      <a href="#37">37. (new) Java 8 Custom Answer Support (Since 2.1.0)</a><br/>
- *
  * </b>
  *
- * <h3 id="0">0. <a class="meaningful_link" href="#verification">Migrating to 2.1.0</a></h3>
+ * <h3 id="0">0. <a class="meaningful_link" href="#verification">Migrating to Mockito 2</a></h3>
```
bric3: The href is incorrect and duplicated with [line 84](https://github.com/mockito/mockito/pull/628/files#diff-ddbc0d47e06a589778a6efe24c5d71d3R84)

bric3:
Looks good other wise

src/main/java/org/mockito/Mockito.java
szczepiq: well spotted. Thanks!

Issue #520, opened 2016-07-30
Provide Java 9 compatibility


szpak:
Once having Mockito 2.0 out the door it would be good to explore Java 9 compatibility.

Basic todo list:
- [x] - migration to Gradle 3.0 (2.x doesn't support Java 9)
- [x] - Travis build with Java 9
- [x] - successful Mockito project build with Java 9
- [x] - fix all disabled tests when the new ASM version is available - #863

Open Java 9 related [issues](https://github.com/mockito/mockito/issues?q=is%3Aissue+is%3Aopen+label%3Ajava-9).

bric3:
I agree with that however note that mockito 2.1 should still be Java 5 compatible (we still build Java 5 binaries with Java 6).
Before that we need to branch mockito 2.x to support old JDKs and make mockito 3 Java 8 min.

bric3:
Damn I didn't see this change. I'm on my phone.

However that's fine for me.

bric3:
Maybe something to mention in the migration guide though.

bric3:
Otherwise for  2.1 I think we will only include few API refactoring like `ArgumentCaptor`. And bugfixes otherwise it will delay serious work on mockito 3.
Also I don't want to backport java 8 code to mockito 2.
My idea is to have a `2.x` branch

TimvdLippe:
I think this is ready now?

raphw:
ASM 6 is still in alpha: http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22org.ow2.asm%22%20AND%20a%3A%22asm%22

bric3:
Really don't focus too much on Java 9 it's still months away.

ceki:
For your information, logback uses Mockito version 1.10.19 which is apparently not Java 9 compatible. It follows that logback build fails under Java 9.

raphw:
Mockito 2 is Java 9 compatible.

ceki:
Great. Thank you.

sdfelts:
Running on version 1.9.5 with Jigsaw b162 with --permit-illegal-access.

I'm seeing the following warning.  
[...]

Issue #521, opened 2016-07-30
Remove com.sun.* references


szpak:
Currently build with gradle 3.0-milestone-2 fails with:

```
:buildSrc:compileJava/mnt/homeold/szpak/cosie/code/inne/mockito/buildSrc/src/main/java/org/mockito/javadoc/JavadocExclude.java:4: error: package com.sun.tools.doclets.formats.html does not exist
import com.sun.tools.doclets.formats.html.HtmlDoclet;
                                         ^
/mnt/homeold/szpak/cosie/code/inne/mockito/buildSrc/src/main/java/org/mockito/javadoc/JavadocExclude.java:20: error: cannot find symbol
        return HtmlDoclet.optionLength(var0);
               ^
  symbol:   variable HtmlDoclet
  location: class JavadocExclude
/mnt/homeold/szpak/cosie/code/inne/mockito/buildSrc/src/main/java/org/mockito/javadoc/JavadocExclude.java:25: error: cannot find symbol
        return HtmlDoclet.start((RootDoc) Proxy.newProxyInstance(clz.getClassLoader(), clz.getInterfaces(), new ExcludeHandler(root)));
               ^
  symbol:   variable HtmlDoclet
  location: class JavadocExclude
/mnt/homeold/szpak/cosie/code/inne/mockito/buildSrc/src/main/java/org/mockito/javadoc/JavadocExclude.java:29: error: cannot find symbol
        return HtmlDoclet.validOptions(var0, var1);
               ^
  symbol:   variable HtmlDoclet
  location: class JavadocExclude
/mnt/homeold/szpak/cosie/code/inne/mockito/buildSrc/src/main/java/org/mockito/javadoc/JavadocExclude.java:33: error: cannot find symbol
        return HtmlDoclet.languageVersion();
               ^
  symbol:   variable HtmlDoclet
  location: class JavadocExclude
Note: /mnt/homeold/szpak/cosie/code/inne/mockito/buildSrc/src/main/java/org/mockito/javadoc/JavadocExclude.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
5 errors
 FAILED
```

Related to #520.

TimvdLippe:
This doclet is required to skip generating Javadoc for internal classes. Is there a mention in the Java 9 jdk of this breaking change?

TimvdLippe:
This seems the corresponding JIRA ticket: https://bugs.openjdk.java.net/browse/JDK-8154399

[...]

PR #639, opened 2016-09-17
Update Travis configuration to also build on oraclejdk9

TimvdLippe:
Let's see if this actually works. Testing for #520


TimvdLippe:
All right, for this one the `HTMLDoclet` is actually breaking the build, which was apparently the issue in #521. Therefore reopening that one. JDK8 is working, but it seems they have moved some classes in JDK9. Not sure if we can have a configuration which works for both JDKs.

TimvdLippe:
I have started the configuration to only compile certain files on a specific JDK version, to at least not fail compilation on JDK 9. However as I personally do not have JDK 9 installed (still running on JDK 8) I can't develop a valid doclet. Does anyone else want to take over implementing the exclude function?

gradle/javadoc.gradle
```diff
@@ -14,7 +14,7 @@ task mockitoJavadoc(type: Javadoc) {
     title = "Mockito ${project.version} API"
     def javadocJdk6Hack = "jdk6-project-version-insert.min.js"
 
-    options.doclet "org.mockito.javadoc.JavadocExclude"
+    options.doclet "org.mockito.javadoc.JavadocExclude1_${JavaVersion.current().toString()[-1]}"
```
bric3: It is a one-liner yet I find the name ugly ;)
Also this won't work with Java 10.

bric3:
Otherwise it LGTM

TimvdLippe:
**One eternity later**

All right this took quite a while, but we can actually compile on JDK9 now :tada:. This required some "workarounds" to bugs in various versions of our dependencies.

* I updated Gradle to 3.1. This release fixes an internal ClassCastException when compiling Groovy buildSrc code
* I updated Jacoco to the latest snapshot release to get around several exceptions thrown in the bytecode generated by Jacoco. The corresponding PR is https://github.com/jacoco/jacoco/pull/434

I sadly had to remove the doclet for JDK > 8, because the latest release available for xenial release is build 114 which does not expose the actual doclets and internal packages required to write a doclet. This is fixed in later builds, but these are only published for the development version of ubuntu :cry: 

All in all we are now compiling correctly, but sadly we have failing tests: https://travis-ci.org/mockito/mockito/jobs/161332991#L433-L465 These seem to be related to AssertJ failing to understand generics.

I have also seen other test failures, so maybe some other tests are failing still. @szpak could you try this branch and see if it compiles locally for you?

TimvdLippe:
Only 1 more failure with the inline mock maker tests: https://travis-ci.org/mockito/mockito/jobs/164437417#L396-L401

raphw:
ASM 6 (support for Java 9) is still in an alpha state, therefore Java 9 class files are still rejected. The moment ASM 6 is released, this is updated.

raphw:
Optionally, we can use byte-buddy-dep and explicitly use ASM 6 alpha.

bric3:
I think we should keep the actual bytebuddy. We could eventually disable the test with JDK 9 using `assumeThat`

raphw:
I do not think that Remi will release ASM 6 before the class file format is finalized and I do not think it is a good idea to rely on these things yet. Without the inlining, Byte Buddy does not need to process any Java 9 class files which is why I would suggest to add a rule to the inliner tests to not be run on Java 9 (yet).

TimvdLippe:
I will disable the test on jdk9 then, thanks for the explanation!

TimvdLippe:
Ignoring the broken test on JDK9. This PR should now be green. I would propose to merge this PR right now and fix possible incoming Java 9 issues in separate PRs

bric3:
Instead may we could detect the API level of ASM. Not sure what's the best way, to start we can detect if `Opcodes` contains `V1_9` or `ASM6`, see [branch of asm 6](http://websvn.ow2.org/filedetails.php?repname=asm&path=%2Fbranches%2FASM_6_FUTURE%2Fasm%2Fsrc%2Forg%2Fobjectweb%2Fasm%2FOpcodes.java). Bytebuddy repackages it in `net.bytebuddy.jae.asm.Opcodes`

bric3:
Also shall we update jacoco [snapshot](http://www.eclemma.org/jacoco/trunk/doc/changes.html), there's a new build

TimvdLippe:
> Also shall we update jacoco snapshot, there's a new build

I would rather wait for the official 0.7.8 release and upgrade to that.

bric3:
ok fair enough ;)

TimvdLippe:
ShallI merge this one?

bric3:
Yes

PR #641, opened 2016-09-17
Update generated pom.xml with more project information

### LINKED ISSUE #633: Update core dev team in gradle
For example https://mvnrepository.com/artifact/org.mockito/mockito-core/2.1.0-RC.1 only lists @bric3 and @szczepiq. We need to include all @mockito/developers e.g. @timvdlippe, @raphw, @marcingrzejszczak and @szpak 

@szczepiq For the 3.0.0-beta.1 should we remove you from the list or what is the expected timeline for you?

TimvdLippe:
I added some more information based on the specification at https://maven.apache.org/pom.html

Fixes #633

### DIFF
--- a/gradle/pom.gradle
+++ b/gradle/pom.gradle
@@ -1,6 +1,6 @@
 publishing.publications.all {
     //TODO github profile?
-    def devs = ["szczepiq:Szczepan Faber", "bric3:Brice Dutheil"]
+    def devs = ['szczepiq:Szczepan Faber', 'bric3:Brice Dutheil', 'raphw:Rafael Winterhalter', 'TimvdLippe:Tim van der Lippe']
 
     pom.withXml {
         def root = asNode()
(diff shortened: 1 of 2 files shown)


gradle/pom.gradle
```diff
@@ -1,6 +1,6 @@
 publishing.publications.all {
     //TODO github profile?
-    def devs = ["szczepiq:Szczepan Faber", "bric3:Brice Dutheil"]
+    def devs = ['szczepiq:Szczepan Faber', 'bric3:Brice Dutheil', 'raphw:Rafael Winterhalter', 'TimvdLippe:Tim van der Lippe']
```
szczepiq: If you like adventures, we could add more info to every developer (for example, github, twitter, blog, etc.)

szczepiq:
Nice. With every PR, it would nice if the comment described how the change was tested.

TimvdLippe:
Ah yes good point. You can generate the dummy pom with `./gradlew generatePomFileForMockitoCorePublication`

PR #646, opened 2016-09-22
Fixes #640 Warns user to upgrade if Java 8 version is to low

### LINKED ISSUE #640: Improve exception message to hint at upgrading java minor version to latest
Driven by #636

bric3:
Fixes #640 Warns user to upgrade if Java 8 version is to low.

Also took the opportunity to log other platform details if mock class cannot be created.

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/ByteBuddyMockMaker.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/ByteBuddyMockMaker.java
@@ -1,18 +1,17 @@
 package org.mockito.internal.creation.bytebuddy;
 
+import static org.mockito.internal.util.StringJoiner.join;
+import java.lang.reflect.Modifier;
 import org.mockito.exceptions.base.MockitoException;
 import org.mockito.internal.InternalMockHandler;
 import org.mockito.internal.configuration.plugins.Plugins;
 import org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.MockAccess;
 import org.mockito.internal.creation.instance.Instantiator;
+import org.mockito.internal.util.Platform;
 import org.mockito.invocation.MockHandler;
 import org.mockito.mock.MockCreationSettings;
 import org.mockito.mock.SerializableMode;
 import org.mockito.plugins.MockMaker;
-
-import java.lang.reflect.Modifier;
-
-import static org.mockito.internal.util.StringJoiner.join;
 
 public class ByteBuddyMockMaker implements MockMaker {
 
@@ -37,8 +36,8 @@ public class ByteBuddyMockMaker implements MockMaker {
         } catch (ClassCastException cce) {
             throw new MockitoException(join(
                     "ClassCastException occurred while creating the mockito mock :",
-                    "  class to mock : " + describeClass(mockedProxyType),
-                    "  created class : " + describeClass(settings.getTypeToMock()),
+                    "  class to mock : " + describeClass(settings.getTypeToMock()),
+                    "  created class : " + describeClass(mockedProxyType),
                     "  proxy instance class : " + describeClass(mockInstance),
                     "  instance creation by : " + instantiator.getClass().getSimpleName(),
                     "",
@@ -51,7 +50,11 @@ public class ByteBuddyMockMaker implements MockMaker {
     }
 
     <T> Class<T> createProxyClass(MockFeatures<T> mockFeatures) {
-        return cachingMockBytecodeGenerator.get(mockFeatures);
+        try {
+            return cachingMockBytecodeGenerator.get(mockFeatures);
+        } catch (Exception bytecodeGenerationFailed) {
+            throw prettifyFailure(mockFeatures, bytecodeGenerationFailed);
+        }
     }
 
 
@@ -69,6 +72,27 @@ public class ByteBuddyMockMaker implements MockMaker {
         // This allows us to catch earlier the ClassCastException earlier
         Class<T> typeToMock = settings.getTypeToMock();
         return typeToMock.cast(mock);
+    }
+
+    private RuntimeException prettifyFailure(MockFeatures<?> mockFeatures, Exception generationFailed) {
+        if (Modifier.isPrivate(mockFeatures.mockedType.getModifiers())) {
+            throw new MockitoException(join(
+                    "Mockito cannot mock this class: " + mockFeatures.mockedType + ".",
+                    "Most likely it is a private class that is not visible by Mockito",
+                    ""
+            ), generationFailed);
+        }
+        throw new MockitoException(join(
+                "Mockito cannot mock this class: " + mockFeatures.mockedType,
+                "",
+                "Mockito can only non-private & non-final classes.",
+                "If you're not sure why you're getting this error, please report to the mailing list.",
+                "",
+                Platform.isJava8BelowUpdate45() ? "Java 8 early builds have bugs that were addressed in Java 1.8.0_45, please update your JDK!\n" : "",
+                Platform.describe(),
+                "",
+                "Underlying exception : " + generationFailed
+        ), generationFailed);
     }
 
     private static String describeClass(Class<?> type) {
--- a/src/main/java/org/mockito/internal/util/Platform.java
+++ b/src/main/java/org/mockito/internal/util/Platform.java
@@ -1,0 +1,64 @@
+package org.mockito.internal.util;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public abstract class Platform {
+
+    private static final Pattern JAVA_8_RELEASE_VERSION_SCHEME = Pattern.compile("1\\.8\\.0_(\\d+)(?:-ea)?(?:-b\\d+)?");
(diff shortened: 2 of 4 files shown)


bric3:
Before merging not that automatic release system will release a 2.1.0.RC.2.

src/main/java/org/mockito/internal/util/Platform.java
```diff
@@ -0,0 +1,48 @@
+package org.mockito.internal.util;
+
+public abstract class Platform {
```
TimvdLippe: Why would this be abstract? It has a private constructor anyways.

src/main/java/org/mockito/internal/util/Platform.java
```diff
@@ -0,0 +1,48 @@
...
+                             "OS name            : %s\n" +
+                             "OS version         : %s\n"
+                ,
+                             JAVA_VERSION,
+                             JVM_VENDOR,
+                             JVM_VENDOR_VERSION,
+                             JVM_NAME,
+                             JVM_VERSION,
+                             JVM_INFO,
+                             System.getProperty("os.name"),
```
TimvdLippe: To keep it consistent, let's make these constant Strings too.

src/main/java/org/mockito/internal/util/Platform.java
```diff
@@ -0,0 +1,48 @@
...
+                             System.getProperty("os.version"));
+    }
+
+    public static boolean isJava8BelowUpdate45() {
+        if(!JAVA_VERSION.equals("1.8")) {
```
TimvdLippe: Space after `if`?

src/main/java/org/mockito/internal/util/Platform.java
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        int beginIndex = JVM_VERSION.indexOf('_');
+        int endIndex = JVM_VERSION.indexOf('b');
+        if(beginIndex == -1 || endIndex == -1) {
+            return false;
+        }
+
+        String updateVersion = JVM_VERSION.substring(beginIndex + 1, endIndex - 1);
+        return 45 > Integer.valueOf(updateVersion);
```
TimvdLippe: I would reverse this, to make the conditional equivalent to the name of the method. Very small nit though, but I had to think about it for a second :smile:

src/main/java/org/mockito/internal/creation/bytebuddy/ByteBuddyMockMaker.java
```diff
@@ -37,8 +36,8 @@ public ByteBuddyMockMaker() {
         } catch (ClassCastException cce) {
             throw new MockitoException(join(
                     "ClassCastException occurred while creating the mockito mock :",
-                    "  class to mock : " + describeClass(mockedProxyType),
-                    "  created class : " + describeClass(settings.getTypeToMock()),
+                    "  class to mock : " + describeClass(settings.getTypeToMock()),
+                    "  created class : " + describeClass(mockedProxyType),
```
TimvdLippe: It seems that these lines are untested. Are we able to come up with a test for this? E.g. why would we reorder these values.

ChristianSchwarz:
Some test for Platform would be nice. Especially too see if the version number extraction works with all known formats e.g. : 
 * Feature -> 1.8.0-b24
 * Maintainance -> 1.4.1_05-ea-b01
 * Update -> 1.4.0_03-b04 , 1.4.0_03-ea-b01

http://www.oracle.com/technetwork/java/javase/versioning-naming-139433.html

How does it work for other JVM's like JRockit, J9 or Excelsior JET?

bric3:
OK thanks for feedback I'll update the code accordingly

src/main/java/org/mockito/internal/creation/bytebuddy/ByteBuddyMockMaker.java
```diff
@@ -71,6 +74,28 @@ public ByteBuddyMockMaker() {
...
+            throw new MockitoException(join(
+                    "Mockito cannot mock this class: " + mockFeatures.mockedType + ".",
+                    "Most likely it is a private class that is not visible by Mockito",
+                    ""
+            ), generationFailed);
+        }
+        throw new MockitoException(join(
+                "Mockito cannot mock this class: " + mockFeatures.mockedType,
+                "",
+                "Mockito can only mock visible & non-final classes.",
```
szczepiq: replace 'visible' with 'non-private'? Users might not know what 'visible' is :)

src/main/java/org/mockito/internal/creation/bytebuddy/ByteBuddyMockMaker.java
```diff
@@ -71,6 +74,28 @@ public ByteBuddyMockMaker() {
...
+                "If you're not sure why you're getting this error, please report to the mailing list.",
+                "",
+                Platform.isJava8BelowUpdate45() ? "Java 8 early builds have bugs that were addressed in Java 1.8.0_45, please update your JDK!\n" : "",
```
szczepiq: AFAIR, StringJoiner always adds next line for each String argument. This means that if java is < 45, there will be extra EOLs. Perhaps it's fine, just wanted FYI.

szczepiq: This is a fantastic improvement, very Mockito-style :D We always strive to make the exception messages better, even a notch better. In the long run all those small improvements create a compound effect. We'll have fantastic error handling, crisp and clear exception messages and happy users!!!

szczepiq:
Beautiful!!! We can push and generate new RC. Or use the magic incantation in the commit message to prevent release and stack this change to next RC (if there is need for yet another RC).

src/main/java/org/mockito/internal/util/Platform.java
szczepiq: Let's make sure this is always safe, regardless of the content of the system properties. Those properties are mutable AFAIK, and/or might change/be different with future versions of java or perhaps even jvm/jdk vendor.

Otherwise, love this PR :)

src/main/java/org/mockito/internal/creation/bytebuddy/ByteBuddyMockMaker.java
bric3: Yes I wanted an extra empty line before platform information (if the runtime is a java 8 version below update 45)

src/main/java/org/mockito/internal/creation/bytebuddy/ByteBuddyMockMaker.java
bric3: ok

src/main/java/org/mockito/internal/util/Platform.java
```diff
@@ -0,0 +1,48 @@
+package org.mockito.internal.util;
+
+public abstract class Platform {
+
+    public static final String JAVA_VERSION = System.getProperty("java.specification.version");
+    public static final String JVM_VERSION = System.getProperty("java.runtime.version");
```
bric3: At the moment this shouldn't be a problem. the parsing targets Java 8 runtime

http://openjdk.java.net/jeps/223

bric3:
@ChristianSchwarz We dont care about those as we don't have report on those, only Oracle / OpenJDK. Although I don't know about their version scheme.

* BEA (now Oracle) JRockit => Discontinued, Last version is Java 6, and oracle migrated features in Oracle JDK
* IBM J9
* Excelsior JET 
* Azul Zing => ?
* Azul Zulu => Based on openjdk

ChristianSchwarz:
@bric3 What I wanted to say was: The update version mechanism should be aware of version schemes other than the expected.  The current implementation throws an IndexOutOfBoundsException if the JVM_VERSION is eg.  "1.8.0b5_u44" or an NumberFormatException if the JVM_VERSION is "1.4.0_03-ea-b01".

szczepiq:
@ChristianSchwarz, fantastic review. Thanks!

src/main/java/org/mockito/internal/creation/bytebuddy/ByteBuddyMockMaker.java
bric3: Because they are wrong in the message ;) 
The type to mock is not the class that the mockmaker created ;)

src/main/java/org/mockito/internal/util/Platform.java
bric3: yes i forgot that, I was a bit in hurry for this PR, and wanted to push on the remote ;)
This was still a work in progress ;)

src/main/java/org/mockito/internal/util/Platform.java
bric3: It's a usual practice to disallow instantiation.

TimvdLippe: Yes true, but that is what the private constructor guarantees too right? But leaving it in is fine too :)

bric3:
@ChristianSchwarz Yes no problem, that was a valuable input, I didn't want these exceptions to be thrown !
I updated the test case and I did some research as well for future cases like this.

TimvdLippe:
Wow really nice tests!! :tada: This LGTM

Issue #642, opened 2016-09-17
Release 2.1.0 final


szczepiq:
The purpose of this ticket is to collect all actions and requirements for getting 2.1.0 final. Let's not wait too long :)

- [x] investigate #638 
- [x] identify opportunities to promo the RC and get more feedback
- [x] decide how long do we wait for feedback before publishing final version
- [x] @TimvdLippe review the #648
- [x] #648 - include more info the javadoc, e.g. direction of the feature, what we expect from the user, inform it is incubating, etc
- [x] @szczepiq review the #648
- update the release notes re #648
 - [x] inform about the new feature (incubating, motivating, future direction)
- [x] publish another RC and spread the word

TimvdLippe:
I think we can push `2.1.0` next week since RC has been available for a little while now. Also I submitted a (random) PR to Google Guava and 850k+ tests did not fail :tada: https://github.com/google/guava/pull/2567

TimvdLippe:
Shall I publish a RC tomorrow after merging #646 ?

szczepiq:
-1 for merging #646 unless it's safe with unexpected system properties
-1 for RC for #646. If this is the only bug / improvement, I suggest we push out 2.1.0 final, followed by 2.1.1 that contains the improvement. While I like the improvement, it does not strike me as something that needs another release candidate.

Given that we don't have more feedback to investigate, I suggest to set the final release to Monday. Good day to release :)

raphw:
I think we should merge #646 before releasing a final version. We already got several reported issues that were caused by old Java versions and I expect that quite a lot of people might not have updated their JDK in years.

Otherthan that, I agree that we do not need a release candidate for this.

TimvdLippe:
Oh oops I indeed did not mean RC, but the final release :p

So lets merge #646 when the issue is fixed and then publish `2.1.0` on monday :tada:

bric3:
I will tweak the code before the week-end, I was short on time these past two week.

szczepiq:
The final version needs to be exactly the same as the last RC. That's the idea of "release candidates" :)

Essentially, the options are:

a) publish 2.1.0 final now, include the improvement in 2.1.1 (no rc needed, after 2.1.0 is out we won't be releasing candidates for 2.x versions)
b) wait for fixes in the improvement, merge it, publish 2.1.0-rc-2, wait some time (for example, a day), publish 2.1.0 final

@raphw says it's important to include the fix, so let's do option b)

[...]

PR #648, opened 2016-09-23
Added InlineByteBuddyMockMaker which uses the instrumentation API for redefining classes and inlining the mocking logic.

raphw:
Inline mocks allow for creating mocks of final types and final methods within any type. The inline mock engine also makes the mocking type invisible such that `mock(Foo.class).getClass() == Foo.class`. Exceptions need to be made for abstract types and for types with serialization requirements where final methods are still instrumented but where a subclass needs to be created nevertheless. There are however few limitations of the inline mock engine that do not apply for the subclass mock engine:
    
1. It is no longer possible to mock `native` methods which do not provide any byte code that can be manipulated.
2. Synchronized methods remain synchronized as the instrumentation API does not allow the changing of modifiers.
    
The final mock engine is activated just like any other custom mock engine by placing an indicator file into the root folder. To make this more convenient, an alias mechanism was added where the mock type name can be referenced by *inline*.


build.gradle
```diff
@@ -60,7 +60,9 @@ test {
 }
 
 dependencies {
-    compile 'net.bytebuddy:byte-buddy:1.4.20'
+    compile 'net.bytebuddy:byte-buddy:1.4.26'
+    compile 'net.bytebuddy:byte-buddy-agent:1.4.26'
+    compile 'com.blogspot.mydailyjava:weak-lock-free:0.11'
```
szczepiq: I'd rather avoid extra dependencies on mockito-core

src/main/java/org/mockito/internal/configuration/plugins/PluginLoader.java
```diff
@@ -6,13 +6,23 @@
...
     private final PluginSwitch pluginSwitch;
 
+    private final Map<String, String> alias;
+
     public PluginLoader(PluginSwitch pluginSwitch) {
         this.pluginSwitch = pluginSwitch;
+        this.alias = new HashMap<String, String>();
+    }
+
+    PluginLoader withAlias(String name, String type) {
```
szczepiq: javadoc around alias would be good

szczepiq:
Assuming that the direction of this idea is to allow mocking final classes out of the box, I really like this direction :). Now, let's get the party started with all my concerns about this PR :D

- what is the practical implication for the end user, how does the end user opt in for new behavior?
- how does the 'alias' work?
- TODOs and missing documentation in the PR
- from the PR, it's hard to figure out that the test coverage is comprehensive (there seem to be more new complexity than more new tests)
- new dependencies from mockito-core (perhaps there should be separate jar for this functionality?)
- this change is a brand new feature so it should _not_ be merged into the RC or 2.1.0 final. We only publish new RCs for bugs

TimvdLippe:
Shall we schedule this for `2.2.0`? We need some time to tinker with this change and investigate impact on the codebase. Would rather not rush it and it gives a nice marketing opportunity for a good minor release :)

szczepiq:
>Shall we schedule this for 2.2.0?

Absolutely! We are not time-constrained with this feature. It can be any 2.x, so long it is not 2.1.0 :) @raphw, when it is ready, we're happy to consume it!

raphw:
I already wrote down my opinion on #642 about whether to include this in 2.1 or not. Here are some technical details:

- Yes, there were to-dos left. Lesson learned: Never squash commits 3.30 in the morning. Seems like I dropped my last fix doing so. Fortunately, I did not loose to much and everything is good now.
- As for the test coverage: Ideally, we would add an additional test goal to Gradle where we rerun all tests using the `InlineByteBuddyMockMaker`. This is what I did during development of the mock maker and I only had to fix a few tests that implied that a mock was always a subclass of the mocked type. I also tried to capture most generic scenarios in the mock-maker specific tests and I think despite of some corner cases, this is a good approximation of the general requirements of a mock maker.
- I added javadoc for the alias property. In the end, I want to avoid that people have to name the `InlineByteBuddyMockMaker` which resides in an internal package. The alias gives an explicit name to that mock maker `mockito-inline`, which can be used instead.

While inline mocks are a brand new feature, code that is part of this feature is only ever executed if a user enables it explicitly as a plugin. I see this inclusion as more of a preview feature which people with specific requirements should try out. We will need these experiences before we can make it a standard-feature and the only way of gathering this feedback is to get it out to the masses. I think Mockito 2 is the best vehicle for this why I would like to see it merged before going final.

As for the additional dependencies: They are very minimal, I wrote both of them as any code generation library consumer requires these feature over and over. I have written weak, concurrent maps and self-attachment agents so many times that I decided to bundle them in small jars which are well-tested and well-documented. If Mockito was a zero-dependency jar I would agree that we should try to avoid these dependencies but since Mockito already depends on Byte Buddy, I do not see an advantage of duplicating the functionality of those two < 10 kb jar files.

For all of those reasons, let us please include this in another RC for 2.1 and have this feature available. I really think this is an exciting new possibility that puts the *major* into a major release.

TimvdLippe:
After our discussion in Slack I also vote for including in 2.1.0. It is a major selling feature, good job!

It seems that Travis still complains and we need to look into testing. I can take a look tomorrow to build it with gradle :)

szczepiq:
This PR introduces new dependencies which means it is a _breaking_ change. Introducing breaking changes in RCs is a really bad idea.

I like the idea of introducing opt-in experimental features, so long they are backwards compatible. We ship incompatible changes in major versions :)

raphw:
Neither dependency is required by the "standard" mocking mechanism, any application that has worked with RC1 would also work with this feature if it is not activated. In theory, we could make the dependencies optional but this would require users to add them manually when they wanted to use inline mocks. Given the small size of the dependencies and the fact that users with size constrains should minimize their jars anyways, I do however not recommend this.

TimvdLippe:
To be honest, I do not understand how a new dependency is a breaking change. It does not break any existing user code. Why reinvent the wheel if someone built a functional implementation you can use without too much overhead?

build.gradle
bric3: @raphw Can we make the dependency optional ?, or maybe shadow the jar or just the needed classes ?

bric3:
> This PR introduces new dependencies which means it is a breaking change.

I don't agree with that statement.

However we can work around the added dependency, we can for exemple shadow the necessary within mockito classes.

src/test/java/org/concurrentmockito/ThreadsRunAllTestsHalfManualTest.java
```diff
@@ -131,23 +131,23 @@ public void run() {
                     System.err.println(failures.size());
                     for (Failure failure : failures) {
                         System.err.println(failure.getTrace());
-                        failed = true;
+                        failed.add(failure.getDescription().getTestClass());
```
bric3: so +10000 ;)

TimvdLippe:
Can we bump the code coverage a bit too? We should be able to obtain more now since we run the inline MockMaker on Travis.

raphw:
The code coverage is not accurate in this case. The inline mock maker copies code into instrumented classes what is nothing jacoco is aware of. Also, the coverage increases a lot when the test suite is run against the mock maker.

TimvdLippe:
Hm but we are launching two jobs, so CodeCov should merge the diff reports (which it does afaik, since we are getting coverage hits in the new class).

build.gradle
```diff
@@ -55,12 +55,14 @@ test {
     include "**/*Test.class"
     testLogging {
         exceptionFormat 'full'
+        showStandardStreams true
```
TimvdLippe: We should probably disable this again

src/main/java/org/mockito/internal/creation/bytebuddy/ByteBuddyCrossClassLoaderSerializationSupport.java
```diff
@@ -5,36 +5,30 @@
...
+
+import java.io.*;
+import java.lang.reflect.Field;
+import java.util.Set;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+import static org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.ForWriteReplace;
+import static org.mockito.internal.util.StringJoiner.join;
+import static org.mockito.internal.util.reflection.FieldSetter.setField;
```
TimvdLippe: Hm an autoformatter reordering imports?

gradle/osgi.gradle
```diff
@@ -12,14 +12,11 @@ afterEvaluate {
...
 
             instruction 'Export-Package',
-                    '!org.mockito.asm.*',
-                    '!org.mockito.cglib.*',
                     '!org.mockito.internal.*',
                     "org.mockito.*;version=${version}"
 
             instruction 'Import-Package',
                     'net.bytebuddy.*;version=[1.4.17,2.0)',
-                    'junit.*;resolution:=optional',
```
bric3: Mmmh would that be problem for OSGI? Junit still has this package.

src/main/java/org/mockito/internal/creation/bytebuddy/ByteBuddyCrossClassLoaderSerializationSupport.java
bric3: Yes could be. I'd like to set common style at some point without being too strict ; readability and expressive code are the goal not some space by tab debate.

TimvdLippe: Yes agreed, we might want to have a very basic (and lenient) checkstyle config again to have some basic checks. It's just that these reorders clutters the diffs a lot 😭

gradle/osgi.gradle
raphw: Oops. This was not supposed to be removed.

build.gradle
raphw: Disabled it.

src/main/java/org/mockito/internal/creation/bytebuddy/ByteBuddyCrossClassLoaderSerializationSupport.java
raphw: We should, yes. I reverted the import changes in the PR.

bric3: No problem here anyway, maybe my IDE that took the config from another project.

raphw:
I looked into the code coverage changes. Even though Travis runs the tests with the inline mock maker, those results are not merged with the code coverage report. All the lines not covered are either:

1. Lines that are code inlined (not covered as this is runtime code that does not get instrumented).
2. Lines concerning serialization that are only partially covered by the mock maker-specific tests.

bric3:
+1

marcingrzejszczak:
+1

TimvdLippe:
I wanted to test some methods that were not invoked yet (as reported by codecov). I ran debug on the whole test suite and for example `MockMethodAdvice.SuperMethodCall#invoke` was not called at all. Setting a debug statement at the start of the method with my test case, it is actually triggering (hooray, more coverage). Sadly the test case is currently ignored, because there is an endless loop upon invoking `tryInvoke` on the actual delegation method (since I am using `CallsRealMethods`). Hopefully we can fix this issue :smile: 

Other than that still a +1 for me on merging. We might want to focus a little bit more on code coverage since I think the tools are correctly retrieving the code coverage.

szczepiq:
+1 to merging ([I wrote here why](https://github.com/mockito/mockito/issues/642#issuecomment-249752547)). Looks like the saga will have a happy end after all :)

Before it gets merged I'd like to:
- [ ] review the PR
- [ ] update the docs a bit and push to the branch
ETA next 12 hours.

PS.
Adding new dependency is a breaking change.

TimvdLippe:
I want to write some more tests later today to increase coverage and better understand the feature. ETA this afternoon

raphw:
Good catch @TimvdLippe. I built this as a feature. Something along the lines of "great if we can also mock super method calls". But this is not a great default but something we should consider once we add this to the Mockito DSL. Something like: `.withFinals()` or `.withSuperCalls()`.

szczepiq:
Great work!!!

Questions:
1. Given that we're using instrumentation, does it open possibilities to mock the unmockable (like static methods)?
2. What is the direction of this change? Do we want to make it a default behavior or an opt-in / per mock / per project configuration?

Feedback:
1. I didn't get to the review yet, will do later today.
2. I would like to document better what is the direction of this change and what we expect from the users. Later today :)
3. I think it's ok to wait few more days and complete reviews, tweaks, etc.

szczepiq:
Updated steps to get 2.1.0 final out at #642.

raphw:
Hi @szczepiq, 

yes, I absolutely think about mocking static methods at some point. However, I would like to avoid mocking them globally. Instead, I would rather mock static calls from within a mock (I think this is the general use case).

My plan is to enable something like the following construct at some point:
```java
class Foo {
  String foo() { return "foo" + Bar.value(); }
}
class Bar {
  static String value() { return "bar"; }
}

Foo foo = spy(new Foo());
doReturn("qux").when(foo).callsStatic(Bar.class).value();
```

What I would like to do is to refine `Foo` to replace all static callsites within the class with `invokedynamic` instructions such that calls that are done by this particular mock instance are treated as stubs. What I want to avoid is some form of *global* stubbing of `Bar::value` but only instrument calls from within the mock instance.

As for the direction, I think we should make this feature either the default once we collected sufficient feedback or to make it the default for final and non-visible types which could not be mocked otherwise; Also, we might want to add a DSL feature such as:
```java
Foo foo = mock(Foo.class, withSettings().mockType(MockType.INLINE));
```
The big drawback of inline mocks is that this approach does not allow to instrument native methods and it is not possible to strip synchronized keywords from methods so there is still a reason to use subclass-type mocks.

raphw:
FYI: I do not think the coverage reports are rerun upon updating the PR. Locally, the coverage is much higher now.

TimvdLippe:
@raphw It actually updates, since my new commits did trigger a new coverage report. However, codecov is currently down it seems.

TimvdLippe:
Code coverage is pretty decent now and I think this feature is fairly stable. Therefore LGTM.

szpak:
LGTM. However, IMHO it would be good to have another RC before releasing the final version (more feedback could be useful). In general very nice feature, @raphw.

Regarding being a breaking change, I agree with @szczepiq that adding a new dependency (**in general**) can, **in some very specific circumstances**, behave like a breaking change. Lets, assume that a client application uses FooLib 1.0.0 directly (in addition to Mockito 5.6.0). After upgrading Mockito  to 5.6.1 or 5.7.0 providing FooLib 2.0.0 (as a transitive dependency) the FooLib version used by Gradle will be 2.0.0. The client application can have compilation issues or even worse work incorrectly due to incompatible changes between FooLib 1.0.0 and 2.0.0. Usually it will not happen, but there is a potential risk.

src/main/java/org/mockito/Mockito.java
```diff
@@ -1215,6 +1216,63 @@
  * <p>
  * When using Java 8, Mockito now also preserves type annotations. This is default behavior and might not hold <a href="#28">if an
  * alternative {@link org.mockito.plugins.MockMaker} is used</a>.
+ *
+ * <h3 id="39">39. <a class="meaningful_link" href="#Mocking_Final">Mocking final types, enums and final methods</a> (Since 2.1.0)</h3>
+ * <p>
+ * Mockito new offers an alternative mock maker which uses the Java instrumentation API for inlining the mocking logic
```
szpak: `Mockito new offers` - it sounds weird to me.

szczepiq:
@szpak, this PR does not change the version.properties so it will trigger another RC. Absolutely new RC is needed.

szczepiq:
I didn't get to this today, either. I will do it tomorrow, and if not, don't wait for me and ship this baby! (e.g. new RC :D)

bric3:
@szczepiq You raced me to javadoc, I was gonna commit a lighter version, I'll inlcude your change then  ;)

szczepiq:
@bric3, my pleasure :)

szczepiq:
FYI: For extra dependencies, there's also a "runtime check" approach. Code checks the presence of required dependency at runtime and informs the user that he needs to add extra dependencies onto the classpath. Not pushing for this, only mentioning for completeness.

raphw:
Classes of the additional dependency are only attempted to be loaded if the new, optional mock maker is used. There is no scenario where an application has worked with RC1 but would fail with these changes. 

As the additional dependency is part of Byte Buddy and since Byte Buddy Agent has never introduced binary incompatible changes, there is no risk in implicit upgrades either.

Is this good to merge by this?

szczepiq:
>Is this good to merge by this?

Yes.

Thanks so much!!!

The final mocks journey does not end here. There are many interesting features we can built on top of it :) Also, we want to implement #657

raphw:
Great, let us merge this then and get out another RC. Once we collected feedback, we make the mock maker available via the Mockito API.

TimvdLippe:
After my lecture has finished I can push out a new one :)

bric3:
Merge it @raphw ;)

raphw:
Hurray!

lidordv:
Hi, After adding the file src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker With text: mock-maker-inline, The tests stop runnings from command line ("./gradlew testDebugUnitTest” and “./gradlew connectedDebugAndroidTest". Any suggestions?

raphw:
The inline mock maker does not work on Android.

raphw:
Its already part of the stack trace but probably barried one or two ticks down in the stack.

santoki-darshan:
> The inline mock maker does not work on Android.

Any update on this?

raphw:
There are some efforts by alternative byte code providers to offer such functionaltity: https://github.com/linkedin/dexmaker

PR #660, opened 2016-09-29
Adds a code style section to CONTRIBUTING document

bric3:
This PR proposes a code style for mockito.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,7 +1,8 @@
 - [Contributing to Mockito](#)
-	- [Pull request criteria](#)
-	- [General info](#)
-	- [More on pull requests](#)
+  - [Pull request criteria](#)
+  - [General info](#)
+  - [More on pull requests](#)
+  - [Coding style](#)
 
 **If looking for support**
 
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -58,3 +59,111 @@ Things we pay attention in a PR :
 
 
 _If you are unsure about git you can have a look on our [git tips to have a clean history](https://github.com/mockito/mockito/wiki/Using git to prepare your PR to have a clean history)._
+
+
+## Coding style
+
+This section is not about some kind of fruitless tabs vs spaces debate. It's about having the code readable, the project grows and it is not rare to read contributions from many different individuals. Each one of us has a different writing style, we are fine with this. Without enforcing we may be picky about it, however this we think this may improve helps readability for everyone.
```
TimvdLippe: > however this we think this may improve helps readability for everyone.

Seems some words were incorrectly placed? `however we think that this improves the readibility of the code for everyone`.

.github/CONTRIBUTING.md
```diff
@@ -58,3 +59,111 @@ Things we pay attention in a PR :
...
+
+Imports must be sorted in the following order :
+
+1. `import java.*`
+2. `import javax.*`
+3. `import all other imports`
+4. blank line
+5. `import static java.*`
+6. `import static javax.*`
+7. `import static all other imports`
```
TimvdLippe: I would actually place static imports before normal imports. Any particular reason for this order?

.github/CONTRIBUTING.md
```diff
@@ -58,3 +59,111 @@ Things we pay attention in a PR :
...
+
+Make sure that
+* One blank lines before imports.
+* One blank lines after imports.
+* Never import with wildcard `*` (_use a very high count threshold_).
```
TimvdLippe: The example above shows `*` :wink: Also IDEA automatically transforms >5 imports to a * import, maybe mention that you need to disable this option as well.

.github/CONTRIBUTING.md
```diff
@@ -58,3 +59,111 @@ Things we pay attention in a PR :
...
+    1. For parameter `Method declaration parameters` choose : `Do not wrap`
+    2. For sub-parameter `Align when multiline` tick the checkbox
+
+3. When using a function with several parameters and when multiple lines make sense,
+    we want to align vertically the method parameters and arguments
+
+    ```java
+    given(mock.action()).willReturn("a very long param",
+                                    "b another very long parameter",
+				    "c yet another");
```
TimvdLippe: The preview seems off for the last String, need another space?

.github/CONTRIBUTING.md
```diff
@@ -58,3 +59,111 @@ Things we pay attention in a PR :
...
+    1. For parameter `Method call parameters` choose : `Do not wrap`
+    2. For sub-parameter `Align when multiline` tick the checkbox
+
+4. When declaring an annotation with several parameters and when multiple lines make sense,
+    we want to align vertically the annotation parameters
+
+    ```java
+    @Mock(answer = Answers.RETURNS_DEFAULTS,
+          serializable = true, 
+	  extraInterfaces = { List.class, YetAnotherInterface.class })
```
TimvdLippe: Needs an extra space as well I think

.github/CONTRIBUTING.md
bric3: haha possibly ;)

ChristianSchwarz:
Nice write up. This should be done by the IDE formatter. The eclipse formatter can be imported by intellij, mockito should provide such a formatter. It makes it easier to stick to the rules.

.github/CONTRIBUTING.md
bric3: yup

bric3:
@ChristianSchwarz Per se it's not rules, I don't want to enforce them as I have bad memories of such enforcement. Also I tend to be wary of automatic formatting, that's why I prefer to rely on human. I could be wrong.

Anyway if you think I should mention something else.

.github/CONTRIBUTING.md
bric3: yup

.github/CONTRIBUTING.md
bric3: I have found that there is usually more conflicts in static methods, that's why I place them closer to the code.

bric3:
@TimvdLippe OK with the changes ? :P

TimvdLippe:
I do think that without enforcement of these rules, contributors will likely miss it. Most devs (including me) usually miss these rules and by making them concrete everyone can adhere. However that could be fixed later in a separate PR

Issue #357, opened 2016-02-16
ClassCastExceptions with JDK9 javac


cushon:
JDK 9 fixes a javac bug ([JDK-8058199](https://bugs.openjdk.java.net/browse/JDK-8058199)) that was causing checkcast intructions to be skipped. Previously javac used the parameter types of a method symbol's erased type as targets when translating the arguments. In JDK 9, javac has been fixed to use the inferred types as targets. The fix causes additional checkcasts to be generated if the inferred types do not have the same erasure.

The fix breaks Mockito answer strategies that pick types based on the erased method signature's return type, and causes tests to fail with ClassCastExceptions when compiled with the JDK 9 javac.

---

Example 1

``` java
class Super<T> {
  T g() {
    return null;
  }
}

class Sub extends Super<Boolean> {}
```

``` java
@Mock Sub s;
when(s.g()).thenReturn(false);
```

compiled with javac 8

```
INVOKEVIRTUAL Sub.g ()Ljava/lang/Object;
INVOKESTATIC org/mockito/Mockito.when (Ljava/lang/Object;)Lorg/mockito/stubbing/OngoingStubbing;
```

compiled with javac 9

```
INVOKEVIRTUAL Sub.g ()Ljava/lang/Object;
CHECKCAST java/lang/Boolean
INVOKESTATIC org/mockito/Mockito.when (Ljava/lang/Object;)Lorg/mockito/stubbing/OngoingStubbing;
```

The erased return type of `Super.g` is `Object`, but the expected return type of `Sub.g` is `Boolean`. If the answer strategy returns `Object` the checkcast fails.

---

Example 2

``` java
class Foo {
  <T> T getFirst(Iterable<T> xs) { return xs.iterator().next(); }
}
```

``` java
@Mock Foo f;
Iterable<Boolean> it = Arrays.asList(false);
when(f.getFirst(it)).thenReturn(false)
```

compiled with javac 8

```
INVOKEVIRTUAL Foo.getFirst (Ljava/lang/Iterable;)Ljava/lang/Object;
INVOKESTATIC org/mockito/Mockito.when (Ljava/lang/Object;)Lorg/mockito/stubbing/OngoingStubbing;
```

compiled with javac 9

```
INVOKEVIRTUAL Foo.getFirst (Ljava/lang/Iterable;)Ljava/lang/Object;
CHECKCAST java/lang/Boolean
[...]

PR #664, opened 2016-10-03
Add tests to demonstrate #357

cushon:


src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsSmartNullsTest.java
```diff
@@ -85,4 +90,36 @@ public void should_print_the_parameters_on_SmartNullPointerException_message() t
...
+    // https://github.com/mockito/mockito/issues/357
+    @Test
+    public void generic_return_type_erasure() throws Throwable {
+        class S<T> {
+            public T f() {
+                return null;
+            }
+        }
+
+        S<Boolean> s = Mockito.mock(S.class, org.mockito.Answers.RETURNS_SMART_NULLS);
```
raphw: This cannot be fixed without providing the generic argument to Mockito. We would need to add something like:
```java
mockGeneric(S.class, Boolean.class, org.mockito.Answers.RETURNS_SMART_NULLS)
```
Byte Buddy is capable of subclassing a generic type where anything down the chain would automatically adapt the actual type.

Currently, this is however a case of heap-pollution which the Java compiler warns against.

src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsSmartNullsTest.java
```diff
@@ -85,4 +90,36 @@ public void should_print_the_parameters_on_SmartNullPointerException_message() t
...
+
+        Mockito.when(s.f()).thenReturn(true);
+        assertTrue(s.f());
+    }
+
+    // https://github.com/mockito/mockito/issues/357
+    @Test
+    public void inferred_generic_type() throws Throwable {
+        class F {
+            <T> T g(Iterable<T> xs) {
```
raphw: I do not think that we can even fix this. The inferred type is not available from within the method. This would require to parse the byte code from the method's call site and even with advanced techniques like rebinding the call site to invokedyanmic, it is not always possible to decide the correct type for the smart nulls.

raphw:
Thanks for providing these. I do not think there is much we can do to prevent those but we can maybe extend the API to ask a user to provide additional arguments. We do however need to update our docs in any case.

TimvdLippe:
I wonder though why they are not failing on jdk8 but are on jdk9. Am I missing some changes?

cushon:
@TimvdLippe there's a javac change in JDK9, see #357.

src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsSmartNullsTest.java
cushon: Agreed. Maybe it should skip the smart nulls in this case?

src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsSmartNullsTest.java
cushon: I think the TypeToken/TypeLiteral approach might be better, especially for more complicated types:

```
mockGeneric(new TypeToken<S<Boolean>>() {}, RETURNS_SMART_NULLS)
```

raphw: I thought about type tokens here, too. One problem is that they are not 100% intuitive, i.e. people might use non-resolvable types such as `TypeToken<S<? extends Number>>` or wonder about the required braces. Maybe we can offer both, it is easy enough to overload.

src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsSmartNullsTest.java
raphw: I think this is the way to go. Maybe we should even warn when mocking methods with a generic type argument when using smart nulls.

TimvdLippe:
@cushon Ah missed the link, thanks!

TimvdLippe:
@cushon Could you change this PR to point to `base:oraclejdk9`? I will take a look and potential catch the exception in the branch which is targeted at jdk9. Thanks!

TimvdLippe:
@cushon This PR also needs a rebase onto `mockito:oraclejdk9`

cushon:
> This PR also needs a rebase onto mockito:oraclejdk9

It'll cherry-pick cleanly. I'm not sure you actually want to merge this, I'm just trying to make it easy to repro #357.

TimvdLippe:
Yes I want to merge it so I can fix it locally. The branch is failing anyways so I can fix them all at once.

cushon:
The tests pass in master, so you could also merge the PR and rebase the jdk9 branch. But you can also check out the PR and merge it anywhere that's convenient: `git fetch origin pull/664/head:664`

TimvdLippe:
@raphw I updated this branch and pushed at https://github.com/mockito/mockito/tree/issue-357 Could you take a look, if we can even do something?

TimvdLippe:
This PR is open for a while now and atm I see no possible action. Therefore I am closing this for now. We will revisit Java 9 compatibility when ASM 6 has been officially released.

PR #672, opened 2016-10-04
Update Byte buddy to latest version

TimvdLippe:


bric3:
You should mention in the description why you didn't bump agent to 1.4.29 ;)

Also we should note that the agent has a race condition in [1.4.27](https://github.com/raphw/byte-buddy/blob/master/release-notes.md#29-september-2016-version-1427). Could it be fixed with the lock added in [1.4.30](https://github.com/raphw/byte-buddy/blob/master/release-notes.md#4-october-2016-version-1430)

TimvdLippe:
Oh maven central these versions as latest, so I assumed there was no more recent version.

szczepiq:
Shoudn't both versions be consistent?

raphw:
This lock does not affect Mockito, we do not use the agent.

TimvdLippe:
image
This is what maven central reported as latest version and therefore I updated to these versions. Probably I should have checked the official repository and not trusted their numbers...

raphw:
mvnrepository has some strange caching. I recommend using the official repository search: http://search.maven.org/#search%7Cga%7C1%7Cg%3A%22net.bytebuddy%22

PR #673, opened 2016-10-04
Removed duplicate test

hansjoachim:
Hello again :p 

I originally requested this test to be added in #525 after the underlying problem had been resolved. While browsing the code now, I discovered that it had already been added (as https://github.com/mockito/mockito/blob/master/src/test/java/org/mockitousage/bugs/DiamondInheritanceIsConfusingMockitoTest.java). This probably doesn't need to be tested twice, so I've removed the duplicate test.


TimvdLippe:
LGTM. Travis failure is unrelated to this removal.

PR #676, opened 2016-10-05
New API: Mockito.mockingDetails(mock).getStubbings()

szczepiq:
Preview PR only. Heads-up.


src/main/java/org/mockito/internal/invocation/InvocationComparator.java
```diff
@@ -0,0 +1,14 @@
...
+import org.mockito.invocation.Invocation;
+
+import java.util.Comparator;
+
+/**
+ * Compares invocations based on the sequence number
+ */
+public class InvocationComparator implements Comparator<Invocation> {
+    public int compare(Invocation o1, Invocation o2) {
+        return Integer.valueOf(o1.getSequenceNumber()).compareTo(o2.getSequenceNumber());
```
ChristianSchwarz: @szczepiq 
This should be simplified...
 * for Mockito 3.x / Java 7+ to: `return Integer.compare(o1.getSequenceNumber(),o2.getSequenceNumber())`
 * for Mockito 2.x / Java 5+ to: `return ((Integer)o1.getSequenceNumber())compareTo(o2.getSequenceNumber())`

Issue #678, opened 2016-10-07
OSGI bundle problem - incorrect version range syntax in manifest file


mgrafl:
The mockito-core version 2.1.0 has an incorrect version range syntax for the Byte Buddy dependencies  in the _META-INF/MANIFEST.MF_ file. Consequently, the bundle cannot be loaded by OSGi (tested with equinox R-4.6). 

Actual: 
> net.bytebuddy;version="[1.4.26",...

Expected: 
> net.bytebuddy;version="[1.4.26,2.0)",...

On the other hand, the version ranges for Objensis dependencies are specified correctly. 

The problem originates from missing double quotes around the version specification in https://github.com/mockito/mockito/blob/v2.1.0/gradle/osgi.gradle#L19 

Stacktrace: 
> !MESSAGE Error installing bundle: reference:file:plugins/mockito-core.jar
> !STACK 0
> org.osgi.framework.BundleException: Error occurred installing a bundle.
	at org.eclipse.osgi.storage.Storage.install(Storage.java:562)
	at org.eclipse.osgi.internal.framework.BundleContextImpl.installBundle(BundleContextImpl.java:146)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.installBundles(EclipseStarter.java:974)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.loadBasicBundles(EclipseStarter.java:583)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.startup(EclipseStarter.java:319)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:239)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.lang.reflect.Method.invoke(Unknown Source)
	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:673)
	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:610)
	at org.eclipse.equinox.launcher.Main.run(Main.java:1519)
	at org.eclipse.equinox.launcher.Main.main(Main.java:1492)
Caused by: java.lang.IllegalArgumentException: invalid range "[1.4.26": invalid format
	at org.osgi.framework.VersionRange.<init>(VersionRange.java:173)
[...]

PR #679, opened 2016-10-07
Correct version specification syntax for bytebuddy

mgrafl:
Proposed fix for #678: 
Use double quotes around version specification for Byte Buddy dependencies. 
The version for Objenesis dependencies is already specified correctly.

### DIFF
--- a/gradle/osgi.gradle
+++ b/gradle/osgi.gradle
@@ -16,7 +16,7 @@ afterEvaluate {
                     "org.mockito.*;version=${version}"
 
             instruction 'Import-Package',
-                    'net.bytebuddy.*;version=[1.4.26,2.0)',
+                    'net.bytebuddy.*;version="[1.4.26,2.0)"',
                     'junit.*;resolution:=optional',
                     'org.junit.*;resolution:=optional',
                     'org.hamcrest;resolution:=optional',


gradle/osgi.gradle
```diff
@@ -16,7 +16,7 @@ afterEvaluate {
                     "org.mockito.*;version=${version}"
 
             instruction 'Import-Package',
-                    'net.bytebuddy.*;version=[1.4.26,2.0)',
+                    'net.bytebuddy.*;version="[1.4.26,2.0)"',
```
ChristianSchwarz: @mgrafl 
this can be reduced to [1.4.26) which has an inclusive lower limit and an exclusive upper limit, denoting version 1.4.26 and any version after this, up to, but not including, version 2.0.0.

bric3:
@mgrafl could you rebase your change on top of `release/2.x` and change the base to `release/2.x` as well ?

mgrafl:
@bric3, Unfortunately, I cannot rebase the pull request from the website without write access to the repo. 
Would it be acceptable that I make another pull request with the proper base?

bric3:
@mgrafl you can from your own repository ;)

If you want I can cherry pick your commit.

mgrafl:
@bric3 yes, please do cherry pick it. Thanks.

gradle/osgi.gradle
ChristianSchwarz: @bric3 nobody want's to introduce an infinite boundary, i requested to use the short form

bric3:
Fixed via 615f7b4

PR #683, opened 2016-10-09
Updated release process so that Mockito can continuously deliver high quality features

szczepiq:
When merged, this will produce new version 2.2.0 and set release/2.x to continuously publish 2.2.1, 2.2.2, ... as new high quality features and enhancements are merged. Hurray!

Highlights:

- set next version to 2.2.0 (arbitrary number)
- reworked the conditional complexity to avoid accidental releases
- removed custom release version logic because it does not work (not opposed to this feature, just cleaned up the code. This feature should be implemented in version.gradle file)
 - for the time being, custom version can be achieved by using Gradle env variable: ORG_GRADLE_PROJECT_version=2.3.0 (it looks like magic but it is a real feature from the Gradle book ;)
- added basic docs around testing the release process (it's way too hard now)

### DIFF
--- a/gradle/release.gradle
+++ b/gradle/release.gradle
@@ -137,7 +137,7 @@ releaseSteps {
     step("ensure good chunk of recent commits is pulled for release notes automation") {
         //Travis default clone is pretty shallow
         //Loading extra 1K commits (1.x -> 2.x had ~700 commits)
-        run "git", "pull", "--depth", "1000"
+        run "git", "pull", "--depth", "1000", "origin", System.env.TRAVIS_BRANCH
     }
 
     def gitAuthor
--- a/gradle/release.gradle
+++ b/gradle/release.gradle
@@ -44,7 +44,8 @@ def dryRun = project.hasProperty('dryRun')
 if (dryRun) {
     //TODO SF lets not use excludedTaskNames to model this
     logger.lifecycle "Automatically scheduling 'rollbackRelease' task"
-    rollbackRelease.dependsOn release
+    //mustRun instead of dependsOn to avoid running 'release' task every time use 'dryRun' property
+    rollbackRelease.mustRunAfter release
     gradle.startParameter.taskNames += "rollbackRelease"
 }
 
--- a/gradle/release.gradle
+++ b/gradle/release.gradle
@@ -9,6 +9,11 @@ In case the release process from Travis CI fails, please try:
     preventing version.properties to be bumped.
   3. delete top release notes paragraph(s) so that they are recreated
     - useful for recreating/merging release notes
+
+Testing release locally:
+  (this section of docs is in progress)
+  1. Configure env TRAVIS_BRANCH, GH_TOKEN, MOCKITO_BINTRAY_API_KEY
+  2. run './gradlew release -PdryRun'
 """
 
 def skipReleaseCommitMessage = '[ci skip-release]'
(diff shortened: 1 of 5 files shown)


gradle/release.gradle
```diff
@@ -9,12 +9,15 @@ In case the release process from Travis CI fails, please try:
...
+
+Testing release locally:
+  (this section of docs is in progress)
+  1. Configure env TRAVIS_BRANCH, GH_TOKEN, MOCKITO_BINTRAY_API_KEY
+  2. run './gradlew release -PdryRun'
 """
 
 def skipReleaseCommitMessage = '[ci skip-release]'
 
-def customReleaseVersionPattern = /.*\[ci-release ([^\]]+)\].*/
```
szpak: I liked that feature. Do you plan to restore it later?

version.properties
```diff
@@ -1,2 +1,2 @@
-version=2.2.0-beta.2
+version=2.2.0
```
szpak: Wouldn't it release 2.2.0 to JCenter (in addition to Mockito's Bintray repo)?

szpak: Things in JCenter shouldn't be rather removed as well (it's the primary artifacts repo for Gradle projects by default).

szpak:
> ORG_GRADLE_PROJECT_version=2.3.0

Hmm, I haven't tried. However, AFAIR `-Pversion=2.3.0` didn't work (out-of-box) some time ago.

version.properties
szpak: Things in JCenter shouldn't be rather removed as well (it's the primary artifacts repo for Gradle projects by default).

szczepiq: I intend to push the button manually from BinTray Web interface once I smoke test the release. After few successful releases, I will put back automatic maven central.

szczepiq: I removed central only temporarily. After the release, I have 24 hours to validate the binary in Bintray. During this time I can delete it if needed. I intend to sync to central manually from BinTray Web interface once I smoke test the release. After few successful releases, I will put back automatic maven central.

gradle/release.gradle
szpak: I'm tweaking my tomorrow presentation about Continuous Delivery right now, so not today, but in general why not :).

version.properties
szczepiq: It's better to reenable automatic maven central sync once this PR is out ;)

TimvdLippe:
I did not review this PR and therefore am saddened to see it is already merged. I do not agree with some of the decisions made in this PR

szpak:
Hmm, I had some comments, but in general I accepted the changes. There were no strong objection against merging, so I understand why Szczepan did it (taking into account that he felt this change as "needed").

Maybe we should clarify the workflow when PR can be merged (e.g. how long to stay in a case more "controversial " changes), but I propose to do it offline.

TimvdLippe:
Hm I thought I was pretty clear in the corresponding issues what my opinion was. Nonetheless I will discuss with the others to come to a conclusion and to resolve the disagreement we are currently having

bric3:
@szczepiq Don't forget to delete the topic branch ;)

szczepiq:
>@szczepiq Don't forget to delete the topic branch once merged ;)

Thanks for reminder and deleting it! Is this a standard practice?

bric3:
yes it is, git / github allows to revive the branch if needed

In `git` a branch is not a really branch, it's rather a reference that points to a commit, since this has been merged there's no reason to keep the _branch reference_.

PR #690, opened 2016-10-13
Enabled automated sync to central repository

szczepiq:
After validating 2 releases manually, we can enable back automated central sync.


bric3:
Thanks @szczepiq

Issue #542, opened 2016-08-10
new API MockingDetails.getStubbings for advanced users and integrations


szczepiq:
One of the principles of designing Mockito is "opinionated features on top of unopinionated toolkit". This ticket is an example of this promise. Opinionated feature: unused stubs detection - new feature of Mockito 2. Unopinionated toolkit: API to get stubbing information from a mock object - this very ticket.

Building high level features on top of on top of low level public API makes the framework flexible and extensible for unidentified use cases and scenarios.

New MockingDetails.getStubbings() -> Collection<Stubbing> api provides information about the stubbings declared for given mock. We use this internally to implement #384, let's expose it for advanced users / framework integrators.

szczepiq:
Created a preview PR for this feature: #676

szczepiq:
I'm about to finish off this feature and roll out 2.2.0. This way, we avoid showing 2.2.0-beta.

TimvdLippe:
Um I do not think that was the plan, given our milestone 2.2.0. There are other issues we want to fix in 2.2.0. Let's not publish the release until we have fixed these issues. See https://github.com/mockito/mockito/milestone/3 Most of these issues have a PR already open

szczepiq:
Let's adjust the plans. Including multiple changes/features into single compatible version is an anti-pattern.

TimvdLippe:
I have to disagree. We published 2.1 less than a week ago, releasing 2.2.0 now would be way too soon. Moreover, we have discussed the feature set for 2.2.0 targeting to improve the ArgumentCaptor. Releasing a minor version with this 1 new api seems redundant, annoying on the user site and will result in an overload of minor version upgrades. Therefore I would like to stick to our plan and include this extra api in to 2.2.0.

szczepiq:
The reason we need to release 2.2.0 is to remove 2.2.0-beta version from being the latest in central.

I'm with you that it feels too small feature for minor version. However, it solves the problem above and it is consistent with semantic versioning :)

Mockito release process (continuous delivery) does not include multiple features in a single version, with the exception for major version releases. I consider the 2.2.0 milestone as rough plan for what's after 2.1.0 but not a feature set to include in a specific version. Every new feature / change will ship a separate version, like 2.2.1, 2.2.2, etc.

bric3:
> The reason we need to release 2.2.0 is to remove 2.2.0-beta version from being the latest in central.

I don't agree. We don't have to release a new version just not see 2.2.0-beta.1.

[...]

PR #691, opened 2016-10-13
New API: Mockito.mockingDetails(mock).getStubbings()

szczepiq:
Added new API. For motivation, see #542


szczepiq:
ping

src/test/java/org/mockitousage/basicapi/MockingDetailsTest.java
```diff
@@ -1,54 +0,0 @@
...
-import org.mockito.Mock;
-import org.mockito.Mockito;
-import org.mockito.Spy;
-import org.mockitoutil.TestBase;
-
-import static junit.framework.TestCase.assertFalse;
-import static junit.framework.TestCase.assertTrue;
-import static org.mockito.Mockito.*;
-
-public class MockingDetailsTest extends TestBase {
```
bric3: This test was documentation for usage, while the other is a unit test.

src/main/java/org/mockito/MockingDetails.java
```diff
@@ -4,8 +4,10 @@
  */
 package org.mockito;
 
+import org.mockito.internal.invocation.Stubbing;
```
szczepiq: Thanks, Stubbing must move to public package. Well spotted!

src/main/java/org/mockito/internal/stubbing/StubbingComparator.java
```diff
@@ -0,0 +1,15 @@
...
+
+import org.mockito.internal.invocation.InvocationComparator;
+import org.mockito.internal.invocation.Stubbing;
+
+import java.util.Comparator;
+
+/**
+ * Compares stubbings based on sequence number
+ */
+public class StubbingComparator implements Comparator<Stubbing> {
```
szczepiq: +1

src/test/java/org/mockitousage/basicapi/MockingDetailsTest.java
szczepiq: I found this separation confusing, e.g. both of the tests are documentation of usage. It also introduces duplicate coverage and creates unnecessary decision points (where to write tests for mocking details).

Do you want to keep this separation?

src/test/java/org/mockito/internal/util/DefaultMockingDetailsTest.java
```diff
@@ -39,6 +50,31 @@ public void should_check_that_a_spy_is_indeed_a_spy() throws Exception {
...
+    }
+
+    @Test
+    public void should_handle_non_mocks() {
+        assertFalse(mockingDetails("non mock").isSpy());
+        assertFalse(mockingDetails("non mock").isMock());
+
+        assertFalse(mockingDetails(null).isSpy());
+        assertFalse(mockingDetails(null).isMock());
+    }
```
szczepiq: Correct.

src/test/java/org/mockitousage/basicapi/MockingDetailsTest.java
bric3: Actually I liked it as it allows to drive the user public API design and they can serve as _acceptance tests_ too. While other tests are better suited for unit tests. In this case there's a single class, but the distinction still apply since it's a public API.

szczepiq: Let's move on. I merge those tests because the distinction between public API testing and internal implementation testing for this piece of functionality is unclear. All tests in this class demonstrate public API and are de facto acceptance tests.

PR #704, opened 2016-10-20
DelayedExecution uses now a more precise approach to call a mock async.

ChristianSchwarz:
New approach to fix flickering async tests.

Detail description, coming soon...

### DIFF
--- a/src/test/java/org/mockitousage/verification/DelayedExecution.java
+++ b/src/test/java/org/mockitousage/verification/DelayedExecution.java
@@ -1,54 +1,78 @@
 package org.mockitousage.verification;
 
-import org.mockitousage.IMethods;
+import static java.lang.System.currentTimeMillis;
+import static java.lang.Thread.MAX_PRIORITY;
+import static java.util.concurrent.Executors.newScheduledThreadPool;
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static java.util.concurrent.locks.LockSupport.parkUntil;
 
-import java.util.ArrayList;
-import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 
 class DelayedExecution {
+    private static final int CORE_POOL_SIZE = 3;
+    /**
+     * Defines the number of milliseconds we expecting a Thread might need to unpark, we use this to avoid "oversleeping" while awaiting the deadline for  
+     */
+    private static final long MAX_EXPECTED_OVERSLEEP_MILLIS = 50;
+    
     private final ScheduledExecutorService executor;
-    private final IMethods mock;
-    private final int delay;
-    private final TimeUnit timeUnit;
-    private final ArrayList<Character> invocations = new ArrayList<Character>();
 
-    DelayedExecution(ScheduledExecutorService executor, IMethods mock, int delay, TimeUnit timeUnit) {
-        this.executor = executor;
-        this.mock = mock;
-        this.delay = delay;
-        this.timeUnit = timeUnit;
+    public DelayedExecution() {
+        this.executor = newScheduledThreadPool(CORE_POOL_SIZE, maxPrioThreadFactory());
     }
 
-    void recordAsyncCall(char c) {
-        invocations.add(c);
+    public void callAsync(long delay, TimeUnit timeUnit, Runnable r) {
+        long deadline = timeUnit.toMillis(delay) + currentTimeMillis();
+
+        executor.submit(delayedExecution(r, deadline));
     }
 
-    void allAsyncCallsStarted() throws InterruptedException {
-        final CountDownLatch countDownLatch = new CountDownLatch(invocations.size());
-        for (final Character invocation : invocations) {
-            executor.execute(runnable(countDownLatch, invocation));
+    public void close() throws InterruptedException {
+        executor.shutdownNow();
+
+        if (!executor.awaitTermination(5, SECONDS)) {
+            throw new IllegalStateException("This delayed excution did not terminated after 5 seconds");
         }
-        countDownLatch.await();
     }
 
-    private Runnable runnable(final CountDownLatch countDownLatch, final Character invocation) {
+    private static Runnable delayedExecution(final Runnable r, final long deadline) {
         return new Runnable() {
             @Override
             public void run() {
-                countDownLatch.countDown();
-                sleep();
-                mock.oneArg(invocation.charValue());
+                //we park the current Thread till 50ms before we want to execute the runnable 
+                parkUntil(deadline - MAX_EXPECTED_OVERSLEEP_MILLIS);
+                //now we closing to the deadline by burning CPU-time in a loop
+                burnRemaining(deadline);
+
+                System.out.println("[DelayedExecution] exec delay = "+(currentTimeMillis() - deadline)+"ms");
+
+                r.run();
+            }
+
+            /**
+             * Loop in tight cycles until we reach the dead line. We do this cause sleep or park is very not precise, 
+             * this can causes a Thread to under- or oversleep, sometimes by +50ms.  
+             */
+            private void burnRemaining(final long deadline) {
+                long remaining;
+                do {
+                    remaining = deadline - currentTimeMillis();
+                } while (remaining > 0);
             }
         };
     }
 
-    private void sleep() {
-        try {
-            timeUnit.sleep(delay);
-        } catch (InterruptedException e) {
-            throw new RuntimeException(e);
-        }
+    private static ThreadFactory maxPrioThreadFactory() {
+        return new ThreadFactory() {
+            @Override
+            public Thread newThread(Runnable r) {
+                Thread t = new Thread(r);
+                t.setDaemon(true);  // allows the JVM to exit when clients forget to call DelayedExecution.close()
+                t.setPriority(MAX_PRIORITY);
+                return t;
+            }
+        };
     }
 }
--- a/src/test/java/org/mockitousage/verification/VerificationWithTimeoutTest.java
+++ b/src/test/java/org/mockitousage/verification/VerificationWithTimeoutTest.java
@@ -45,25 +45,22 @@ public class VerificationWithTimeoutTest {
 
     private ScheduledExecutorService executor;
 
+    private DelayedExecution delayedExecution;
+    
     @Before
     public void setUp() {
-        executor = newSingleThreadScheduledExecutor();
+        delayedExecution = new DelayedExecution();
     }
 
     @After
     public void tearDown() throws InterruptedException {
-        executor.shutdownNow();
-        executor.awaitTermination(5, SECONDS);
+        delayedExecution.close();
     }
 
     @Test
     public void shouldVerifyWithTimeout() throws Exception {
-        // given
-        DelayedExecution delayedExecution = new DelayedExecution(executor, mock, 20, MILLISECONDS);
-        delayedExecution.recordAsyncCall('c');
-
         // when
-        delayedExecution.allAsyncCallsStarted();
+        delayedExecution.callAsync(30, MILLISECONDS, callMock('c') );
 
         // then
         verify(mock, timeout(100)).oneArg('c');
@@ -75,12 +72,8 @@ public class VerificationWithTimeoutTest {
 
     @Test
     public void shouldFailVerificationWithTimeout() throws Exception {
-        // given
-        DelayedExecution delayedExecution = new DelayedExecution(executor, mock, 30, MILLISECONDS);
-        delayedExecution.recordAsyncCall('c');
-
         // when
-        delayedExecution.allAsyncCallsStarted();
+        delayedExecution.callAsync(30, MILLISECONDS, callMock('c') );
 
         // then
         verify(mock, never()).oneArg('c');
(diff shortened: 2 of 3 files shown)


TimvdLippe:
We probably need to restart the builds a couple of times to get again a level of certainty.

src/test/java/org/mockitousage/verification/DelayedExecution.java
```diff
@@ -1,54 +1,78 @@
...
     }
 
-    void allAsyncCallsStarted() throws InterruptedException {
-        final CountDownLatch countDownLatch = new CountDownLatch(invocations.size());
-        for (final Character invocation : invocations) {
-            executor.execute(runnable(countDownLatch, invocation));
+    public void close() throws InterruptedException {
+        executor.shutdownNow();
+
+        if (!executor.awaitTermination(5, SECONDS)) {
```
bric3: I'd make that configurable

src/test/java/org/mockitousage/verification/DelayedExecution.java
```diff
@@ -1,54 +1,78 @@
...
+
+            /**
+             * Loop in tight cycles until we reach the dead line. We do this cause sleep or park is very not precise, 
+             * this can causes a Thread to under- or oversleep, sometimes by +50ms.  
+             */
+            private void burnRemaining(final long deadline) {
+                long remaining;
+                do {
+                    remaining = deadline - currentTimeMillis();
+                } while (remaining > 0);
```
bric3: Interesting idea. It looks like a spin loop, I have been [playing](https://gist.github.com/bric3/314c3d01a80e5e3c158965dcd459a8a5) with precise sleep duration. On my laptop this works well though you burn a CPU for the time being, however I'm not sure this behaves that well on virtualised environment.

```java
    public static void sleepNanos(long nanoDuration) throws InterruptedException {
        final long end = System.nanoTime() + nanoDuration;
        long timeLeft = nanoDuration;
        do {
            if (timeLeft > SLEEP_PRECISION) {
                Thread.sleep(1);
            } else {
                if (timeLeft > SPIN_YIELD_PRECISION)
                    Thread.yield();
            }
            timeLeft = end - System.nanoTime();

            if (Thread.interrupted())
                throw new InterruptedException();
        } while (timeLeft > 0);
    }
```

src/test/java/org/mockitousage/verification/DelayedExecution.java
```diff
@@ -1,54 +1,78 @@
...
-    private Runnable runnable(final CountDownLatch countDownLatch, final Character invocation) {
+    private static Runnable delayedExcution(final Runnable r, final long deadline) {
         return new Runnable() {
             @Override
             public void run() {
-                countDownLatch.countDown();
-                sleep();
-                mock.oneArg(invocation.charValue());
+                //we park the current Thread till 50ms before we want to execute the runnable 
+                parkUntil(deadline - MAX_EXPECTED_OVERSLEEP_MILLIS);
```
bric3: For uncommon static method I'd prefer to see the object owner, i.e. `LockSupport.parkUntil(...)`

src/test/java/org/mockitousage/verification/DelayedExecution.java
```diff
@@ -1,54 +1,78 @@
...
-        } catch (InterruptedException e) {
-            throw new RuntimeException(e);
-        }
+    private static ThreadFactory maxPrioThreadFactory() {
+        return new ThreadFactory() {
+
+            @Override
+            public Thread newThread(Runnable r) {
+                Thread t = new Thread(r);
+                t.setDaemon(true);  // allows the JVM to exit when clients forget to call DelayedExecution.close()
```
bric3: Nice catch

src/test/java/org/mockitousage/verification/VerificationWithTimeoutTest.java
```diff
@@ -45,25 +45,25 @@
...
 
+    private DelayedExecution delayedExecution;
+
+
+    
+    
     @Before
     public void setUp() {
-        executor = newSingleThreadScheduledExecutor();
+        delayedExecution = new DelayedExecution();
```
bric3: The before method seems overkill now, compared to a simple field initialisation.

src/test/java/org/mockitousage/verification/DelayedExecution.java
```diff
@@ -1,54 +1,78 @@
...
-        this.timeUnit = timeUnit;
+
+    public DelayedExecution() {
+        this.executor = newScheduledThreadPool(CORE_POOL_SIZE, maxPrioThreadFactory());
     }
 
-    void recordAsyncCall(char c) {
-        invocations.add(c);
+    public void callAsync(long delay, TimeUnit timeUnit, Runnable r) {
+        long deadline = timeUnit.toMillis(delay) + currentTimeMillis();
```
bric3: I think the code should check the deadline is not too small, especially since the code will decrease this amount [here](https://github.com/mockito/mockito/pull/704/files#diff-305ace128d183213cab9933b42729f20R44). The best option imho is to perform the deadline calculation in `delayedExecution`

src/test/java/org/mockitousage/verification/DelayedExecution.java
```diff
@@ -1,54 +1,78 @@
...
+        executor.shutdownNow();
+
+        if (!executor.awaitTermination(5, SECONDS)) {
+            throw new IllegalStateException("This delayed excution did not terminated after 5 seconds");
         }
-        countDownLatch.await();
     }
 
-    private Runnable runnable(final CountDownLatch countDownLatch, final Character invocation) {
+    private static Runnable delayedExcution(final Runnable r, final long deadline) {
```
bric3: typo ;)

bric3:
Interesting approach. I had in mind something more constrained using the concurrency utils like `CountDownLatch`. Let's see how it works.

src/test/java/org/mockitousage/verification/DelayedExecution.java
ChristianSchwarz: * `parkUntil(long) `does nothing if the deadline is in the past, so i don't see a problem here.
* The deadline is calculated as soon as possible, when it is calculated later e.g. in `delayedExecution` the deadline may not reflect the delay passed to `callAsync(..)` 
. The time between the call and the calculation is not zero, objects are created and code must be executed in the meantime or the current Thread is halted because of GC's and so on.

src/test/java/org/mockitousage/verification/DelayedExecution.java
ChristianSchwarz: It would make the usage uneccessary complex for the caller. I don't see a benefit now.

src/test/java/org/mockitousage/verification/DelayedExecution.java
ChristianSchwarz: I experiment also with a `sleep(1)`, ` sleep(0,500)`,  `sleep(0,100)` and `sleep(0,100)`. The precision of the sleep implementation in the JVM or OS seems not to be very accurate. In average the sleep calls oversleeps by 8ms and up to ~50ms, which also depends on the Thread priority. 

Anyway, I think the real problem is that we don't have a realtime JVM. The timer resolution on Win is pretty low (15,6ms), GC's may occure and thread scheduling is not deterministic. All of this lead to flickering tests. Maybe it is better to throw an exception in case a value <30ms is passed to timeout().

ChristianSchwarz:
The main problem I see with the concurrent utils is fact that they all depend on the timer resolution which affects sleep and park, whereas the cpu-burner strategy depends on the clock resolution which seems to be more precise (but still allows to oversleep).

src/test/java/org/mockitousage/verification/DelayedExecution.java
bric3: at least with two constructor

bric3:
I don't know which is better, I'm not enough skilled in that matter. Concurrency utility introduce synchronisation between threads, so in a way this is more deterministic in regard of behavior. While clock resolution is more precise regarding the time, yet other threads may be slowed down for some reason.
Although I think the spin loop is worth a try.

bric3:
This job failed https://travis-ci.org/mockito/mockito/jobs/169349729

src/test/java/org/mockitousage/verification/VerificationWithTimeoutTest.java
ChristianSchwarz: Since we need to call close() after every test to avoid delayed calls on mocks we need need a new instance before we start every test. Am I right or did I overlook something?

ChristianSchwarz:
I triggered some build to ensure the changes have an effect.

* [Build 1640.*](https://travis-ci.org/mockito/mockito/builds/169698150) succeed
* [Build 1641.3](https://travis-ci.org/mockito/mockito/builds/169702658) failed
 * [VerificationWithTimeoutTest > canIgnoreInvocationsWithJunit FAILED](https://travis-ci.org/mockito/mockito/jobs/169702661#L270)

bric3:
You can change the [file](https://github.com/mockito/mockito/blob/release/2.x/dummy-commit.txt) to trigger new builds

ChristianSchwarz:
@bric3 
Thanks, I have the problem that I am currently behind a firewall and can't commit via git. Therefore I must use the GitHub UI to commit. I squash the commits later.

bric3:
@ChristianSchwarz ah ok, sorry, firewall sucks :'(

bric3:
By the way shall we enable the midi interface to get high resolution timers? 😅

https://github.com/real-logic/Agrona/blob/master/src/main/java/org/agrona/concurrent/HighResolutionTimer.java

ChristianSchwarz:
> By the way shall we enable the midi interface to get high resolution timers? :sweat_smile:
>
>https://github.com/real-logic/Agrona/blob/master/src/main/java/org/agrona/concurrent/HighResolutionTimer.jar

Thats something between brilliant or insane, I am not sure.

bric3:
@ChristianSchwarz This looks good, could you cleanup commits and code, then we'll merge.

Thanks for your patience by the way, much appreciated !

bric3:
For reference this aims to fix #548, and chose a different approach than #619

bric3:
@ChristianSchwarz Damn it, it failed. Granted it is rare.

https://travis-ci.org/mockito/mockito/jobs/171369547#L283

```
org.mockitousage.verification.VerificationWithTimeoutTest > canIgnoreInvocationsWithJunit FAILED
    Wanted but not invoked:
    mock.oneArg('1');
    -> at org.mockitousage.verification.VerificationWithTimeoutTest.canIgnoreInvocationsWithJunit(VerificationWithTimeoutTest.java:142)
    Actually, there were zero interactions with this mock.
        at org.mockito.internal.exceptions.Reporter.wantedButNotInvoked(Reporter.java:329)
        at org.mockito.internal.verification.checkers.MissingInvocationChecker.checkMissingInvocation(MissingInvocationChecker.java:37)
        at org.mockito.internal.verification.Times.verify(Times.java:37)
        at org.mockito.internal.verification.VerificationOverTimeImpl.verify(VerificationOverTimeImpl.java:78)
        at org.mockito.internal.verification.VerificationWrapper.verify(VerificationWrapper.java:15)
        at org.mockito.internal.verification.MockAwareVerificationMode.verify(MockAwareVerificationMode.java:21)
        at org.mockito.internal.handler.MockHandlerImpl.handle(MockHandlerImpl.java:73)
        at org.mockito.internal.handler.NullResultGuardian.handle(NullResultGuardian.java:32)
        at org.mockito.internal.handler.InvocationNotifierHandler.handle(InvocationNotifierHandler.java:36)
        at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.doIntercept(MockMethodInterceptor.java:37)
        at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor$DispatcherDefaultingToRealMethod.interceptAbstract(MockMethodInterceptor.java:116)
        at org.mockitousage.IMethods$MockitoMock$100420155.oneArg(Unknown Source)
        at org.mockitousage.verification.VerificationWithTimeoutTest.canIgnoreInvocationsWithJunit(VerificationWithTimeoutTest.java:142)
```

ChristianSchwarz:
Ready to merge?

src/test/java/org/mockitousage/verification/DelayedExecution.java
bric3: ok

bric3:
OK let's try that approach :)
Maybe another approach is to have a _plugable ticker_; and then plugin a test ticker that could allow determinist order of execution ;  current approach rely on an [active loop](https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/internal/verification/VerificationOverTimeImpl.java#L75-L91) that rely on a `org.mockito.internal.util.Timer`.

ChristianSchwarz:
@bric3 

> Maybe another approach is to have a plugable ticker;

Yup! We use `timeout() `and `after()` to test concurrent legacy code that relies on System.currentTime() / nanoTime().  For new code we are using guava's [Ticker](https://google.github.io/guava/releases/19.0/api/docs/com/google/common/base/Ticker.html) instead. For our test we simply mock the Ticker and use derivations of jmock's [DeterministicScheduler](http://www.jmock.org/javadoc/jmock-2.6.0/doc/org/jmock/lib/concurrent/DeterministicScheduler.html) or [DeterministicExecutor](http://www.jmock.org/javadoc/jmock-2.6.0/doc/org/jmock/lib/concurrent/DeterministicExecutor.html) instead of real Executor implementaitions of the production code. This approach is relieable and must faster.

bric3:
I think that's the real deal here, those tests should not really care much about time, they should care about wether or not the behaviour according to their semantic :

* timeout : happens before deadline
* after : happens after offset

That's what we should care about there regardless of the actual time elapsed.

-------------------------------------

Additionally the issues / pull request serie indicates there's an issue when tests are asserting something according to time. One cannot expect such tests to be determinist without proper infrastructure and good enough understanding of how time and thread are related, how external factors like hardware and / or OS scheduler.

Typically : 

* The test thread where mockito verification is executed can run faster or slower than the thread where the method to verify is executed
* Due to some external thread both thread can run slower, while they maintain their relation (_happens-before_)

PR #711, opened 2016-10-24
Fixes #707, adds throwable to arity answers, also improved javadoc

### LINKED ISSUE #707: Improve usability of arity Answers with regards to checked exceptions
Recently introduced Answer1 style answers don't declare Throwable. They should! Otherwise the user is forced to add boilerplate code to AnswerX implementations, whenever they use methods that declare checked exceptions. Plus, AnswerX should behave exactly as good old Answer interface!

Reported via the [mailing-list](https://groups.google.com/d/msgid/mockito/82d08ba9-5536-495e-90db-78c1062b6400%40googlegroups.com)

bric3:
All in the title. This should fix #707

### DIFF
--- a/src/main/java/org/mockito/stubbing/Answer1.java
+++ b/src/main/java/org/mockito/stubbing/Answer1.java
@@ -3,13 +3,36 @@ package org.mockito.stubbing;
 import org.mockito.Incubating;
 
 /**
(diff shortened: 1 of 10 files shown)


TimvdLippe:
LGTM. Restarted the build that was failing on an unrelated test

PR #719, opened 2016-10-28
Verification listeners

LiamClark:
I had the need to be notified when ever a verification happens, this pr implements this.

To do anything useful with this notification the following data is important:
- The mock that we are verifying on.
- The verification mode we are using.
- And what method we wish to verify.

The one place where this data is collectable is in MockAwareVerificationMode.

What remains is to get the listeners there so we are able to notify them.

The tests I added try to follow those of the invocation listeners, however there might be more cases that are important that i'm unaware of.


szczepiq:
Thank you for submitting - very interesting PR! This is exactly the direction of development of MockitoListener. Can you:

1. Elaborate your use case. Why do you need this data / API?
2. Include an example of use in the description of the PR

In the meantime, we will review the PR.

TimvdLippe:
Chiming in, @LiamClark is a student colleague of mine. We are both teaching assistants for the course "Software Quality and Testing". In this course, we use Mockito (that's how I learned Mockito and started contributing :smile: ). Since high student growth is a real issue on our university, we have to resort to automatic ways of testing student tests. One of the requirements is to verify that the user has verified a certain method call. E.g. did the user do `verify(mock).method(anyString())`.

For this we can use the new `MockitoFramework` and attach some listeners before we build and grade the student code. This is fully automatic and solves the scaling issues we are coping with at the moment.

---

Regarding the implementation: Still not sure what the cleanest solution is to pass the listeners to the specific `VerificationMode`. Since we don't have the concrete type `MockAwareVerificationMode` in `MockingProgressImpl::verificationStarted`, we can't just set the listeners. Should we change the type for this? Would be a breaking change, so for Mockito 3, but the solution might be a lot cleaner. Still have to tinker about it.

src/main/java/org/mockito/listeners/VerificationListener.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+public interface VerificationListener extends MockitoListener {
+    /**
+     * Called when a verification happened.
+     *
+     * @param mock the mock that verify was called with.
+     * @param mode the mode that the mock will be verified with.
+     * @param verificationData the data of this verification.
+     */
+    void onVerification(Object mock, VerificationMode mode, VerificationData verificationData);
```
ChristianSchwarz: @LiamClark 
The method `listener.onVerification(..)` will only be called if the verification was successful but not when an Exception is thrown by `VerificationMode.verify(..)`. I think the JavaDocs and method name should reflect that.  
Here are my proposals:
 * JavaDoc: Called when a verification ~~happened~~ **succeeded**.
 * Method name:  onVerification**Succeeded**(..)

WDYT?

LiamClark: Makes a lot of sense, done.

My use cases is however not interested in the case that verify does throw. 
Would it however make any sense to add support for it?

src/main/java/org/mockito/verification/VerificationSucceededEvent.java
```diff
@@ -0,0 +1,51 @@
+package org.mockito.verification;
+
+import org.mockito.internal.verification.api.VerificationData;
```
szczepiq: Let's avoid leaking internal classes and expose VerificationData as a public type (e.g. let's move it to public package)

src/main/java/org/mockito/verification/VerificationSucceededEvent.java
```diff
@@ -0,0 +1,51 @@
...
+    public VerificationMode getMode() {
+        return mode;
+    }
+
+    public VerificationData getData() {
+        return data;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
szczepiq: Does this class need equals and hashcode? If not, let's keep things simple

src/main/java/org/mockito/listeners/VerificationListener.java
```diff
@@ -0,0 +1,22 @@
...
+public interface VerificationListener extends MockitoListener {
+    /**
+     * Called when a verification completes successfully without throwing an exception.
+     *
+     * @param verificationSucceededEvent contains all information about the event.
+     *                                   Consisting of the mock that was verified on,
+     *                                   the {@link org.mockito.verification.VerificationMode} used
+     *                                   and the {@link org.mockito.internal.verification.api.VerificationData}
+     */
+    void onVerificationSucceeded(VerificationSucceededEvent verificationSucceededEvent);
```
szczepiq: 1. More generic API that also covers verification failure would be better. 
2. 'onMockVerified' is more consistent in naming with existing 'onMockCreated'
3. Why don't we keep the API simple and ditch the 'event' object.

For example:

```java
//failure is nullable
onMockVerified(Object mock, VerificationMode mode, VerificationData data, Throwable failure)
```

LiamClark: @szczepiq 
wouldn't it be nicer to have two methods in the interface over having a null-able argument?

TimvdLippe: I think event is better, because we might want to add more information later. With an event this still compiles. With an extra argument it breaks which makes it very hard to extend.

LiamClark:
So, I added some support to also deal with exceptions thrown during verifcation.

As i see it now we have 3 distinct cases:

1. everything goes fine and we have a mock,a verification mode and some verification data.

2. verify was called incorrectly and it fails fast with a MockitoException.
For example it was called with  a null pointer, a non-mock object  or it has unfinished stubbing.  In this case we can only notify the listener about the exception thrown the mock and the verification mode. But we fail before the verification data is even there.

3. The call to verify was correct, but the test should fail because the verification mode doesn't match the actual results. I believe this always throws an MockitoAssertionError Now we can notify the listener about the exception, the mode, the mock and the verification data.

(Please tell me if i'm wrong about the types of exceptions and if there are any others that can be thrown.)

Now we could still have one method in the interface with some optional arguments, or we could make them 3 distinct cases. where we can put the things that are in common, namely the mock and verification mode, in an event object, so that if information is added we can do so. We can also have the two different exception cases have a more concrete type about what type of exception actually occurred.

Other possibilities would be dropping the event object completely, or introducing multiples to handle the multiple cases.

Lastly this introduces a lot of nasty exception catching and re-throwing, but i just don't see a way around it.

TimvdLippe:
I am very much in favor of two subtypes of a VerificationExceptionEvent. This makes it extensible and easy to reason about + less boilerplate when you use the interface.

ChristianSchwarz:
@LiamClark Nice write up! 

I think case 2 "wrong specified verification" should not be considered for now. It isn't part of the verification but more a specification error. For this case I propose to create a new issue like "introduce specification error listener" in the future. This could also address specification errors in `when`  or matcher misuse.

src/main/java/org/mockito/internal/MockitoCore.java
```diff
@@ -78,16 +80,25 @@ public boolean isTypeMockable(Class<?> typeToMock) {
...
+            }
+            if (!isMock(mock)) {
+                throw notAMockPassedToVerify(mock.getClass());
+            }
+            mockingProgress.verificationStarted(new MockAwareVerificationMode(mock, actualMode, listeners));
+            return mock;
+        } catch (MockitoException e) {
+            for (VerificationListener listener : listeners) {
+                listener.onVerificationException(mock, actualMode, e);
+            }
```
bric3: instead of looping here and elsewhere ([1](https://github.com/mockito/mockito/pull/719/files#diff-df53f45103af61fe77db9216591b7ad5R64), [2](https://github.com/mockito/mockito/pull/719/files#diff-60afe350d3acfd74fc0b6a42026e9dc1R34)), I propose to have an internal listener that can do the for loop, this would be easier to test too. Thoughts ?

src/main/java/org/mockito/verification/VerificationData.java
```diff
@@ -2,7 +2,7 @@
  * Copyright (c) 2007 Mockito contributors
  * This program is made available under the terms of the MIT License.
  */
-package org.mockito.internal.verification.api;
+package org.mockito.verification;
```
bric3: Making this public is a bit more complicated as it leaks `org.mockito.internal.invocation.InvocationMatcher` class, a simple thing would be to extract an interface from the `InvocationMatcher` class. But I don't want to make public all the API of this class.

src/main/java/org/mockito/listeners/VerificationListener.java
```diff
@@ -0,0 +1,24 @@
...
+     * Called when a verification completes successfully without throwing an exception.
+     *
+     * @param verificationSucceededEvent contains all information about the event.
+     *                                   Consisting of the mock that was verified on,
+     *                                   the {@link org.mockito.verification.VerificationMode} used
+     *                                   and the {@link VerificationData}
+     */
+    void onVerificationSucceeded(VerificationSucceededEvent verificationSucceededEvent);
+
+    void onVerificationException(Object mock, VerificationMode actualMode, Throwable failure);
```
ChristianSchwarz: I would prefer listener interfaces with only one method (FunctionalInterface), this way lambda-expressions can be used as listeners too.

bric3: Agreed with others, both should be consistent methods with an event class, also I'd rename the method

```
onVerificationSuccess(VerificationSuccess success);
onVerificationFailure(VerificationFailure failure);
```

src/main/java/org/mockito/verification/VerificationSucceededEvent.java
```diff
@@ -0,0 +1,26 @@
+package org.mockito.verification;
+
+public class VerificationSucceededEvent {
```
bric3: to be renamed, see other comment on `VerificationListener`

src/main/java/org/mockito/internal/verification/MockAwareVerificationMode.java
```diff
@@ -4,21 +4,39 @@
...
-        mode.verify(data);
+        try {
+            mode.verify(data);
+
+            for (VerificationListener listener : listeners) {
+                listener.onVerificationSucceeded(new VerificationSucceededEvent(mock, mode, data));
+            }
+        } catch (MockitoAssertionError e) {
+            for (VerificationListener listener : listeners) {
+                listener.onVerificationException(mock, mode, e);
```
szczepiq: I suggest to keep it simple and just catch Throwable. This will also cover cases where matcher breaks on other exceptions, etc.

src/main/java/org/mockito/listeners/VerificationListener.java
szczepiq: +1 to nullable Throwable and single interface + single method

src/main/java/org/mockito/verification/VerificationSucceededEvent.java
szczepiq: This type needs to be exposed as an interface, and not as a concrete class. It's very important that we expose all API with interfaces (with some exceptions for very good reasons, like Mockito, Answers, etc).

src/main/java/org/mockito/verification/VerificationData.java
szczepiq: Very nice catch, thank you! I'll take a stab at exposing this type and report back.

szczepiq:
See #728 for exposing new new public API for verification.

src/main/java/org/mockito/verification/VerificationSucceededEvent.java
ChristianSchwarz: @szczepiq Can you explain why you would like to expose the events as interfaces?  The events are pure value objects without any behaviour. So there no need to mock them. Also there will be only one implementation so the event classes can be made final to avoid subclassing.

LiamClark:
@szczepiq 
Regarding the leaking of the internal api, what is your preferred method of me undoing my move to the public package?

And i'm curious why you prefer one interface with one method with a nullable throwable argument over the two single method interfaces? doesn't this force the user to deal with both cases, when instead we could seperate them?

szczepiq:
Thanks for checking in and for working on the feature!!!

>Regarding the leaking of the internal api, what is your preferred method of me undoing my move to the public package?

No preferred method. Up to you. git revert, or just move class.

>And i'm curious why you prefer one interface with one method with a nullable throwable argument over the two single method interfaces? doesn't this force the user to deal with both cases, when instead we could seperate them?

For following reasons:
 - simple solution (I don't think there will more arguments in the future)
 - nullable throwable idiom is pretty standard (that's how listeners work in Gradle for example)
 - I don't see nullable throwable as a problem to the user - he can ignore it if he does not care

Above is recommendation - I'm ok with team's decision on that matter.

Should you choose to introduce new event object, I do insist on putting it behind a public interface (while the concrete class goes to 'internal' package). This ensures flexibility for future changes to the implementation class and reduces public API footprint (e.g. we don't expose constructors, any unexpected public methods of the concrete class, etc.).

Thanks for being persistent. Let's get this change in :)

src/main/java/org/mockito/verification/VerificationSucceededEvent.java
szczepiq: Good question. See my last comment to this PR. Thanks for checking in!

src/main/java/org/mockito/internal/verification/MockAwareVerificationMode.java
LiamClark: @szczepiq catching throwable doesn't allow me to rethrow the exception, so instead i'll use RunTimeException. However we can keep the type in the interface throwable.

src/main/java/org/mockito/internal/MockitoCore.java
LiamClark: So this particular loop is now gone, so that leaves us with three loops left.
one for finding the verification listeners, one for verification succes and a final one for verification failure.

If we do the api the way szczepiq  suggets that would leave us with 2. 
@bric3 do you still think it's worth it in that case? and how do you suggest we do this?
have the internal listener be some sort of composite listener that dispatches to all the other verification listeners? but then how would you add the listeners to this composite listener?

src/main/java/org/mockito/internal/verification/MockAwareVerificationMode.java
szczepiq: @LiamClark, nice catch!

Let's catch both, RuntimeException and Error types. This way, you can rethrow while keeping the implementation robust enough to support both cases.

LiamClark:
I've implemented some of the changes requested.

However since there was still a lot of different opinions about the api we should expose, I have implemented two different versions, not sure how to show that here (should i make two pr's to this pr?)

The first is the @szczepiq version with one SAM interface with an event object hidden behind an interface, and the event object has a nullable throwable cause.

The second version is the @bric3 version. It has the two SAM interfaces one for success and one for failures, both have their own event objects hidden behind the interfaces. Right now the failure extends the success interface to just get the throwable (non-null) cause. This seems nice but i'm not sure if we want the interface inheritance in there since it does introduce some dependency between them.

While I personally like this interface a lot more, it does introduce quite a few types and more instance of checking since we now need to distinguish between the two listener types.

So how do i share this properly so you guys can have a look at it? 

disclaimer: since half of it will probably be thrown away I did not spend any time on proper java doc for it.

src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
```diff
@@ -0,0 +1,123 @@
...
+    @Test
+    public void should_call_single_listener_on_verify() {
+        //given
+        RememberingListener listener = new RememberingListener();
+        ThreadSafeMockingProgress.mockingProgress().addListener(listener);
+        Foo foo = mock(Foo.class);
+
+        //when
+        VerificationMode never = never();
+        verify(foo, never).doSomething("");
```
kobbikobb: I then this verify should belong with the //then

src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+        //then
+        assertThatHasBeenNotified(listener, foo, never, invocationWanted);
```
kobbikobb: This assertion method is obscure. You shouldn't have to read the methods implementation to see what is being asserted.

src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
```diff
@@ -0,0 +1,123 @@
...
+        mockingProgress.addListener(listener2);
+
+        Foo foo = mock(Foo.class);
+
+        //when
+        VerificationMode never = never();
+        verify(foo, never).doSomething("");
+
+        //then
+        assertThatHasBeenNotified(listener1, foo, never, invocationWanted);
```
kobbikobb: It's not good to have two methods failing for the same reasons. Could we assert fewer things here?

src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
```diff
@@ -0,0 +1,123 @@
...
+        } catch (Throwable e) {
+            //then
+            assertThat(listener.cause).isInstanceOf(MockitoAssertionError.class);
+        }
+    }
+
+    private void assertThatHasBeenNotified(RememberingListener listener, Object mock, VerificationMode mode, String invocationWanted) {
+        assertThat(listener.mock).isEqualTo(mock);
+        assertThat(listener.mode).isEqualTo(mode);
+        assertThat(listener.data.getWanted().getMethod().getName()).isEqualTo(invocationWanted);
```
kobbikobb: This seems like a very complex and fragile way of verifying that a listener has been notified.
Law of Demeter smell.
Could we possibly improve this?

src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
```diff
@@ -0,0 +1,123 @@
...
+        mockingProgress.addListener(listener);
+
+        //when
+        Foo foo = null;
+        try {
+            verify(foo).doSomething("");
+            fail("Exception expected.");
+        } catch (Exception e) {
+            //then
+            assertNull(listener.cause);
```
kobbikobb: This way of verifying that a listener was not called seems fragile. Could we improve this?

src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+    @Test
+    public void wrong_verify_type() {
```
kobbikobb: Please provide a better test name. This is vague.

TimvdLippe:
@kobbikobb The tests are similar to https://github.com/mockito/mockito/blob/release/2.x/src/test/java/org/mockitousage/debugging/InvocationListenerCallbackTest.java and I disagree that they are overly complicated. We can take a look to improve, but let's do that in a separate PR and fix both tests at the same time.

kobbikobb:
@TimvdLippe Yes they are not complicated but a bit obscure. I don't think we should let former examples of "not so good tests" guide us into the future. However it is a good idea to do a little clean up for our tests. Fixing the name of this test: wrong_verify_type would be super easy then we could address other issue in a separate PR?

TimvdLippe:
@kobbikobb once this PR has landed you can go ahead and tidy up the tests if you want to :smile:

src/main/java/org/mockito/internal/verification/MockAwareVerificationMode.java
```diff
@@ -4,21 +4,46 @@
...
+        try {
+            mode.verify(data);
+
+            for (VerificationListener listener : listeners) {
+                listener.onVerificationSucceeded(new VerificationSucceededEvent(mock, mode, data));
+            }
+        } catch (RuntimeException e) {
+            notifyExceptionListeners(e);
+            throw e;
+        } catch (Error e) {
```
szczepiq: Great!

src/main/java/org/mockito/listeners/VerificationListener.java
```diff
@@ -0,0 +1,24 @@
...
+     * Called when a verification completes successfully without throwing an exception.
+     *
+     * @param verificationSucceededEvent contains all information about the event.
+     *                                   Consisting of the mock that was verified on,
+     *                                   the {@link org.mockito.verification.VerificationMode} used
+     *                                   and the {@link org.mockito.internal.verification.api.VerificationData}
+     */
+    void onVerificationSucceeded(VerificationSucceededEvent verificationSucceededEvent);
+
+    void onVerificationException(Object mock, VerificationMode actualMode, Throwable failure);
```
szczepiq: onVerificationSucceeded expects an event object but this method just take args. Let's make it consistent: either both methods take parameters or both methods take an event object.

src/main/java/org/mockito/verification/VerificationSucceededEvent.java
```diff
@@ -0,0 +1,28 @@
+package org.mockito.verification;
+
+import org.mockito.internal.verification.api.VerificationData;
+
+public class VerificationSucceededEvent {
```
szczepiq: Let's have an interface

szczepiq:
@LiamClark,

>The first is the @szczepiq version with one SAM interface with an event object hidden behind an interface, and the event object has a nullable throwable cause.

Let's go with above. It's simple and clean. It uses an event object as most core devs prefer.

>since half of it will probably be thrown away I did not spend any time on proper java doc for it.

Please write fantastic javadoc!!! Consider adding a paragraph in main Mockito class. This is how you promote the feature you wrote. It is really nice feature, it shows the flexibility, robustness of Mockito. Also it pushes us forward in exposing verification API as decent public API.

THANKS!!!

Let's get this feature in.

src/main/java/org/mockito/internal/verification/MockAwareVerificationMode.java
szczepiq: Do you have unit tests for both cases, errors and exception emitted during verification? If no, please add, if yes, ignore my request :)

src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
```diff
@@ -0,0 +1,123 @@
...
+        RememberingListener listener2 = new RememberingListener();
+
+        MockingProgress mockingProgress = ThreadSafeMockingProgress.mockingProgress();
+        mockingProgress.addListener(listener1);
+        mockingProgress.addListener(listener2);
+
+        Foo foo = mock(Foo.class);
+
+        //when
+        VerificationMode never = never();
```
LiamClark: It's being used in the assertion below however, why would you prefer a call to never twice?

TimvdLippe: To keep it consistent with all other tests we write :smile:

ChristianSchwarz:
@TimvdLippe 
> @kobbikobb once this PR has landed you can go ahead and tidy up the tests if you want to :smile:

Later means never! IMHO a feature should be complete that means it has the best quality possible and that includes tests javadoc and overall design.  What about an extra branch based on this PR? This way @LiamClark and @kobbikobb can work in parallel if they like.

TimvdLippe:
@ChristianSchwarz While I generally agree, we should prevent this PR from a `merge-it-all`-PR. If we can separate the effort, let's do that

src/main/java/org/mockito/internal/verification/MockAwareVerificationMode.java
```diff
@@ -4,21 +4,44 @@
...
         this.mock = mock;
         this.mode = mode;
+        this.listeners = listeners;
     }
 
     public void verify(VerificationData data) {
-        mode.verify(data);
+        try {
+            mode.verify(data);
+            notifyExceptionListeners(new VerificationEventImpl(mock, mode, data, null));
```
TimvdLippe: Sad that we can't use `Optional` for this. Let's leave a TODO for Mockito 3 to refactor this to `Optional`

src/main/java/org/mockito/internal/verification/VerificationEventImpl.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public VerificationMode getMode() {
+        return mode;
+    }
+
+    public VerificationData getData() {
+        return data;
+    }
+
+    public Throwable getCause() {
```
ChristianSchwarz: @LiamClark 
We should rename this to `getVerificationError()`.

src/main/java/org/mockito/internal/verification/MockAwareVerificationMode.java
```diff
@@ -4,21 +4,44 @@
...
+        } catch (RuntimeException e) {
+            notifyExceptionListeners(new VerificationEventImpl(mock, mode, data, e));
+            throw e;
+        } catch (Error e) {
+            notifyExceptionListeners(new VerificationEventImpl(mock, mode, data, e));
+            throw e;
+        }
```
bric3: This catch pattern feels wrong. I'm ok for `RuntimeException`, but not `Error` those are normally created by the JVM. At the very least I'd reduce the scope to `AssertionError`.

bric3: Or more questionable using the sneaky throw technic to catch and rethrow Throwable.

src/main/java/org/mockito/internal/verification/MockAwareVerificationMode.java
bric3: Optional in arguments is discouraged

src/main/java/org/mockito/listeners/VerificationListener.java
```diff
@@ -0,0 +1,18 @@
...
+ * For this to happen, it must be registered using {@link org.mockito.internal.progress.MockingProgress#addListener(MockitoListener)}.
+ */
+public interface VerificationListener extends MockitoListener {
+    /**
+     * Called after a verification happened.
+     * This includes verifications that fail due to the verification mode failing.
+     *
+     * @param verificationEvent the event that occurred.
+     */
+    void onVerification(VerificationEvent verificationEvent);
```
bric3: I'm picky but I don't want the term `Event` there, I'd prefer something like `InvocationVerification`

src/main/java/org/mockito/internal/verification/MockAwareVerificationMode.java
```diff
@@ -4,21 +4,44 @@
...
+    }
+
+
+    private void notifyExceptionListeners(VerificationEvent event) {
```
bric3: Why `Exception` ?

src/main/java/org/mockito/internal/verification/MockAwareVerificationMode.java
bric3: Also I'd suggest two static factory that will create the concrete types : `VerificationSuccess` and `VerificationFailure`. That would look like : 

```
try {
    mode.verify(data);
    onVerification(VerificationFactory.success(mock, mode, data)); // build a VerificationSuccess class
} catch (... runtime) {
    onVerification(VerificationFactory.failure(mock, mode, data, runtime)); // create a VerificationFailure class
    throw runtime;
}
```

src/main/java/org/mockito/internal/verification/MockAwareVerificationMode.java
TimvdLippe: Could you explain? Optional gives type safety, since `null` now implicitly means: this verification was successful. Personally I dislike using `null` to convey this information.

ChristianSchwarz:
1.  Why do we have now only one Event type for two different kind of events (success , failure)? We should consider the perspectiv of the event consumer, now it is not possible to register for a specific event only. Implementors have make additional checks if they are interested in only one case. That means boilerplate null checks.
1.  What about dropping listener interfaces and use an mechanism like guava event bus?
1. The whole thing should be annotated with `@Beta`  
1. I don't like to expose `VerificationData` there is too much other stuff that will be exposed too that should be refactored in the future (argument capturing, in order verification, ....)

src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
LiamClark: @bric3  would:

assertThat(listener.mode).isEqualToComparingFieldByField(never());

work for you?

bric3:
> Why do we have now only one Event type for two different kind of events (success , failure)? We should consider the perspectiv of the event consumer, now it is not possible to register for a specific event only. Implementors have make additional checks if they are interested in only one case. That means boilerplate null checks.

Because of the `VerificationListener` is now a SAM functional interface. That's a nice idea there to subscribe to events, currently these parts of mockito are designed after the infamous observer pattern, moving to an event bus may prove useful indeed, however it requires significant work without substantial gain (in the present situation).
That doesn't mean that it should't or that it won't happen, but that the observer pattern is currently enough to achieve the functionality.

Back to this PR. That is why in the first place I'm against the word _event_ yet as it shifts the mind to think this way. Also about the two events in this case they can be seen as the same event with two state : failure or success.

> What about dropping listener interfaces and use an mechanism like guava event bus?

Maybe in mockito 3 ?
With backward compatibility for existing listeners.

> The whole thing should be annotated with `@Beta`

👍 

> I don't like to expose VerificationData there is too much other stuff that will be exposed too that should be refactored in the future (argument capturing, in order verification, ....)

See #730 

> API changes should be discussed before implementatin starts! GitHub Reviews are an awful and chaotic tool for this. I propose to close this PR and send a new one with the same content to get rid comment chaos.

Yes but then you'll loose all the intelligence, and we'll have to link it here, and in the end the new comments will be : _"you forgot that and that in the that comment link and that comment <link>"_

However I agree that some point can be discussed outside. Like #730

src/main/java/org/mockito/internal/verification/MockAwareVerificationMode.java
bric3: > Personally I dislike using null to convey this information.

That's why I propose to remove it in the first place.

> Could you explain?

* Using `Optional` instead of reference has no real value, as the logic will be mostly the same, especially for a constructor.
* Having an optional argument does not prevent it form being null.
* Having an optional arguments may require to wrap in an `Optional`.

src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
bric3: I was commenting the comments ;)
Actually using reference equality is ok as long as it's explicit (like naming variable `neverModeReference` and using `assertThat(...).isSame`). But `isEqualToComparingFieldByField` is ok as well.

ChristianSchwarz:
@bric3 
>> Why do we have now only one Event type for two different kind of events (success , failure)? 

> Because of the VerificationListener is now a SAM functional interface. 

What I wanted to say is why don't we use 2 SAM interfaces and 2 Event types? This would make the life of event consumer easier. Using one event for 2 different cases violates seperation of concerns.

bric3:
@ChristianSchwarz 
> Using one event for 2 different cases violates seperation of concerns.

Not necessarily, it might be an _event_ with different state : _success_ or _failure_.

> What I wanted to say is why don't we use 2 SAM interfaces and 2 Event types? This would make the life of event consumer easier, see my comment above.

Yes it's not out of question (my suggestion on the factory methods indeed lean in that direction). But at this moment, I fear we are over-engineering this. @TimvdLippe and @LiamClark are the primary users of this, could you show us some concrete code snippets outside (not tests) how you want the listener to be registered or how the code you're working with can/should process the _verifications_.

Basically the design question is wether we want this :

```java
Mockito.framework().addListener((VerificationFailure) System.out::println);
Mockito.framework().addListener((VerificationSuccess) System.out::println);

interface Verification { verificationMode, wantedMethod, verificationData }
interface VerificationFailure extends Verification { failureCause }
interface VerificationFailureListener extends MockitoListener {
        void on(VerificationFailure arg);
}
interface VerificationSuccess extends Verification { }
interface VerificationSuccessListener extends MockitoListener {
        void on(VerificationSuccess arg);
}
```

Or 

```java
Mockito.framework().addListener((Verification) System.out::println);

// could be with sub-interfaces or public classes VerificationFailure and VerificationSuccess
interface Verification { verificationMode, wantedMethod, verificationData, failureCause }
interface VerificationListener extends MockitoListener {
        void on(Verification arg);
}
```

TimvdLippe:
Tomorrow Liam and I will work on the userside and see what option is most
appealing.

Op di 8 nov. 2016 om 11:52 schreef Brice Dutheil <notifications@github.com>:

ChristianSchwarz:
@bric3 
For now we have  the usecase of @LiamClark and @TimvdLippe that needes an notification for a "verification success". During the review a new usecase "verification failed" was introduced.  We should find out how possible clients would use the new API and what are possible uses cases. The more uses cases we find the better the API and client experience can be. 

Here is a usecase that requires to log only verification failure events:

specialized API solution: 
```java
Mockito.framework().addListener((VerificationFailure) System.out::println); `
```

vs. 

generic API solution:
```java
Mockito.framework().addListener((VerificationListener) v -> {
   Throwable error =  v.getVerificationError();
    if (e!=null){
          System.out.println(e);
    }
}); 
```

I think the main question is which kind of uses case is more likely: consuming both event types or consuming only one type.

src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
kobbikobb: @bric3 Yes I agree, we should consider changing that.

src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
kobbikobb: @bric3 Yes I agree that LOD is more target to production...

I like your ideas for improving the assertions. I think this would improve the tests a lot.

kobbikobb:
@bric3 @ChristianSchwarz @TimvdLippe @LiamClark 
I created a pull request to this branch to improve the unit tests as we have talked about.

Changes I made:
1. Use assertThat and custom assertion to verify notification
2. Match method rather then method name
3. Inline setup because it was small and this adds readability
4. Rename failure tests to be more descriptive

The pull request is here:
https://github.com/LiamClark/mockito/pull/1

Anything you would like to improve?

TimvdLippe:
@kobbikobb Please update https://github.com/mockito/mockito/blob/release/2.x/src/test/java/org/mockitousage/debugging/InvocationListenerCallbackTest.java accordingly, to keep them in sync. Still I would prefer to separate concerns in separate PRs

kobbikobb:
@TimvdLippe Ok no problem. Should I create a pull request to LiamClark:verifaction-listeners or mockito:release/2.x?

TimvdLippe:
@kobbikobb Well, they should be both fixed at the same time. I would create the PR on `release/2.x` after we merge this one.

kobbikobb:
@TimvdLippe Ok, I will do that after we merge this one.

bric3:
The tests change on https://github.com/LiamClark/mockito/pull/1 seems good to me. I agree with @TimvdLippe though

kobbikobb:
@bric3 it's fine by me to. You could check in the PR without the fixes to the test and I can create a new PR fixing tests for both listeners at the same time. I've implemented both so I could do the PR the same day. Just let me know, either way.

TimvdLippe:
@kobbikobb publish your PR after this one is merged

bric3:
@kobbikobb I'm waiting on usage feedback from @TimvdLippe and @LiamClark regarding verification events. I'd like simplicity first on this but @ChristianSchwarz arguments are valid as well. Whatever comes out of @TimvdLippe and @LiamClark work / experience will drive how to integrate with mockito verifications.

LiamClark:
So far, the listener I have been using looks something like:

public void onVerification(VerificationEvent verificationEvent) {
   final String methodName = verificationEvent.getData().getWanted().getMethod().getName();
   scopeAwareLogging("verification", new Object[]{verificationEvent.getMock(), verificationEvent.getMode(), methodName});
}

In which the failure is simply ignored, because the precondition of a passing suite reasonably removes any failures.
However if I were to drop this precondition in the future it would require the dreaded if failure != null.

This as discussed puts some boilerplate code onto the consumer of the api.
Now the arguments about simplicity versus better client experience, don't indicate a clear winner.

However in the invocation listener's MethodInvocationReport we have:

/**
* @return The throwable raised by the method invocation, maybe <code>null</code>
*/
Throwable getThrowable();

the exact same api, I would rather keep them consistent with each other.

TimvdLippe:
Ping per above comment. This is our usecase, so please let us know what you think :smile:

szczepiq:
Since it isn't merged yet, I reviewed the PR once again and some things stood out:

1. There is no high level unit test that shows how to use this feature. The test that only uses public API and demonstrates usage.
2. We don't have to expose the private VerificationData. I realize this has been discussed before and this type remains internal and leaky. However, in the public API we add we can start using public VerificationData. I will make a slight change in the branch to demonstrate what I mean.
3. Finally, I really like the 'event' object that is passed in as an arg to the listener. It is great idea and it allows the API to grow and evolve. Why didn't I see it before? :)

Besides by comments, what is missing for getting this PR merged? Is the only remaining question about single interface VS separate interfaces for success and failure? Let's make a decision and move on. Either way, the users will have their use cases covered. If that helps, my +1 vote goes for single interface.

Thanks guys for working on this!!! Really nice feature!

TimvdLippe:
High level unit test can be copypasted from the local repo of Liam.

Glad you also like the event!

I am for multiple interfaces, to allow users to use lambdas.

Excited we can almost land this PR!

LiamClark:
@szczepiq 
Could you eleborate a bit what's wrong with the verification data, I thought that was supposed to be moved in a seperate pr?

szczepiq:
>Could you eleborate a bit what's wrong with the verification data, I thought that was supposed to be moved in a seperate pr?

Let's just move on. If @TimvdLippe says it's good then ship it :)

szczepiq:
Restarted the build to ensure this is mergable. @bric3, your review still shows 'requested changes'. Is it up-to-date or we can merge it?

bric3:
Sorry I lag behind, there's some time consuming issues with my employer software ;)

I'll re-review it now.

src/main/java/org/mockito/internal/verification/MockAwareVerificationMode.java
LiamClark: On the one hand i like the readability of this, on the other is it really worth introducing a factory for 3 constructor calls? with the possibility of one going away due to multi catch.

Maybe adding them as static factory functions to the type itself?

src/main/java/org/mockito/verification/VerificationEvent.java
```diff
@@ -0,0 +1,31 @@
+package org.mockito.verification;
+
+import org.mockito.internal.verification.api.VerificationData;
+
+/**
+ * Contains all information about a verification that has happened.
+ */
+public interface VerificationEvent {
```
bric3: Add an `@Incubating` there as well. Since it may opportunistically change with `Optional` once the team switches to mockito 3 which will be based on JDK8.

src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
bric3: Yes ! Use mockito (i.e use `mock(VerificationListener.class)`)

src/main/java/org/mockito/internal/MockitoCore.java
bric3: Nope I don't think that's worth it now ;)

src/main/java/org/mockito/internal/verification/MockAwareVerificationMode.java
bric3: Don't bother I was nitpicky, not sure it's actually worth it.

TimvdLippe:
@LiamClark could you incorporate @bric3 comments and rebase this PR? Then we can merge it 🎉

TimvdLippe:
After rebase I will merge this PR

TimvdLippe:
I have restarted the JDK6 builds as they were again affected by the SSH issue. I think we can safely merge this one though

TimvdLippe:
I am merging this PR, even though the build has failed (per above comment). Thanks a lot @LiamClark for your time and effort into this PR!

PR #728, opened 2016-11-02
Exposed new public APIs for verification

szczepiq:
The intention of this PR is to streamline #719: merge with verification-data branch, add javadoc and prosper :) We can also check-in this change separately provided javadocs are fixed. Separate PR can also help us discuss this change separately from the new verification listeners.

Exposed following public APIs to allow verification listeners:
 - VerificationData
 - MatchableInvocation (public interface on top of InvocationMatcher)

I am super excited about those changes - we are making progress in exposing verification API and enabling users to write custom verification modes!


TimvdLippe:
I dont really understand the motivation for the renaming. Why would we do this?

szczepiq:
What renaming do you refer to?

szczepiq:
Let's ignore this PR, see the description in #730

PR #734, opened 2016-11-03
New nullable(Class<T>) matcher for convenient matching of nullable arguments

bric3:
After using mockito, I have found that the new behaviors of matcher are good and indeed show problems in the test code or in the production code. However I have found that I miss an easy way to express a matcher for `nullable` argument when the production code may produce either `null` or value types, typically `Stream.of(<mixed values>).map(o -> mapper.map(c))...`, the current options are :

* `<T> T any()` is nice but don't check the type.
* there's always `AdditionalMatchers.or()` but this feel more heavy to use.

I'm proposing the introduction of the argument matcher `<T> nullable(Class<T>)` which is a combination of `isA` and `isNull`. I'm not adding `<T> T nullable()` because this would have the same meaning as `<T> T any()`

### DIFF
--- a/src/main/java/org/mockito/ArgumentMatchers.java
+++ b/src/main/java/org/mockito/ArgumentMatchers.java
@@ -1,5 +1,14 @@
 package org.mockito;
 
+import static org.mockito.internal.progress.ThreadSafeMockingProgress.mockingProgress;
+import static org.mockito.internal.util.Primitives.defaultValue;
+import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


szczepiq:
I really like the description of this PR - it nicely describes the motivation of the changes and I don't need to read the code to grok it. Thanks!

Also, it's really cool that you use Mockito and keep improving it as you see missing features.

bric3:
Thanks @szczepiq

Issue #730, opened 2016-11-03
Expose public verification API


szczepiq:
### Problem

Public API under org.mockito.verification.* leaks internal APIs:
 - VerificationMode leaks internal VerificationData interface
 - VerificationData leaks InvocationMatcher concrete class

### Options

1. Move VerificationData to public package, hide behind an interface: #728
2. Deprecate internal types, create new public interfaces, delegate to deprecated APIs to avoid breaking API compatibility and behavior.

### Recommendation

- Option 1 is easier to implement. It introduces breaking change on an internal type. This is legal from the standpoint of semantic versioning.
- Option 2 adds some implementation overhead and leaves some deprecated code. It is fairly easy to implement, too. Although it is legal to change private API at will, this particular API was leaking from public api (for many releases) so users may have started depending on it.

+1 for option 2.

### Plan

- org.mockito.internal.verification.api.VerificationData deprecated by new org.mockito.verification.VerificationData
- org.mockito.internal.verification.api.VerificationData.getWanted() deprecated by new org.mockito.verification.VerificationData.getTarget()

szczepiq:
This issue has been fixed, although the original plan needed to be updated.

PR #735, opened 2016-11-04
Exposed existing verification API so that it no longer leaks internal API

szczepiq:
See motivation and plan in #730


szczepiq:
Friendly ping

bric3:
why not reopening #728 , what is the difference here, it seems to be the same branch.

src/main/java/org/mockito/invocation/MatchableInvocation.java
```diff
@@ -0,0 +1,70 @@
...
+ * It is used during verification process:
+ *
+ * <pre class="code"><code class="java">
+ *   mock.foo();   // <- invocation
+ *   verify(mock).bar();  // <- matchable invocation
+ * </code></pre>
+ *
+ * @since @since@
+ */
+public interface MatchableInvocation extends DescribedInvocation {
```
bric3: What do yon think of `VerifiableInvocation` ?

src/main/java/org/mockito/invocation/MatchableInvocation.java
```diff
@@ -0,0 +1,70 @@
...
+ * <code>MatchableInvocation</code> wraps {@link Invocation} instance
+ * and holds argument matchers associated with that invocation.
+ * It is used during verification process:
+ *
+ * <pre class="code"><code class="java">
+ *   mock.foo();   // <- invocation
+ *   verify(mock).bar();  // <- matchable invocation
+ * </code></pre>
+ *
+ * @since @since@
```
bric3: I may have missed something but is this placeholder replaced by the build script ?

src/main/java/org/mockito/verification/VerificationData.java
```diff
@@ -0,0 +1,43 @@
...
+import org.mockito.invocation.MatchableInvocation;
+
+import java.util.List;
+
+/**
+ * Data needed to perform verification of interactions.
+ *
+ * @since @since@
+ */
+public interface VerificationData extends org.mockito.internal.verification.api.VerificationData {
```
bric3: should it be the opposite, internal extending public interface ?

src/main/java/org/mockito/invocation/MatchableInvocation.java
szczepiq: I prefer Matchable because technically speaking it's also used for matching invocations for stubbing.

src/main/java/org/mockito/internal/reporting/PrintSettings.java
```diff
@@ -58,7 +59,7 @@ public String print(Invocation invocation) {
         return print(ArgumentsProcessor.argumentsToMatchers(invocation.getArguments()), invocation);
     }
 
-    public String print(InvocationMatcher invocationMatcher) {
+    public String print(MatchableInvocation invocationMatcher) {
```
szczepiq: Done

szczepiq: Nice catch

src/main/java/org/mockito/verification/VerificationData.java
szczepiq: I will write a test for this. Nice catch again.

szczepiq: You will like my last commit :)

src/main/java/org/mockito/verification/VerificationMode.java
```diff
@@ -30,8 +29,11 @@
  * See examples in javadoc for {@link Mockito#verify(Object, VerificationMode)}
  */
 public interface VerificationMode {
-    
-    void verify(VerificationData data);
+
+    /**
+     * Performs the verification
+     */
+    void verify(org.mockito.internal.verification.api.VerificationData data);
```
TimvdLippe: This type should be imported

src/test/java/org/mockito/internal/verification/OnlyTest.java
```diff
@@ -33,6 +34,9 @@ public VerificationDataStub(InvocationMatcher wanted, Invocation invocation) {
             return Arrays.asList(invocation);
         }
 
+        @Override
+        public MatchableInvocation getTarget() { return wanted; }
```
TimvdLippe: For consistency please add new lines

src/main/java/org/mockito/verification/VerificationMode.java
bric3: Also this is an internal type

src/test/java/org/mockito/internal/verification/OnlyTest.java
bric3: intellij will collapse the block code just good enough

src/main/java/org/mockito/internal/verification/api/VerificationData.java
```diff
@@ -6,26 +6,43 @@
...
 import java.util.List;
 
 /**
- * @deprecated - This internal class leaks from the public API and therefore was deprecated.
- * Please use {@link org.mockito.verification.VerificationData} instead.
- *
- * Deprecated since 2.2.12
+ * Data needed to perform verification of interactions.
+ * This interface is considered public even though it lives in private package.
+ * In the next major version of Mockito, this class will be moved to public space.
```
bric3: Actually this still does leak `org.mockito.internal.invocation.InvocationMatcher`

szczepiq: Yes, I cannot remove this for now. Next major version #744 :)

src/main/java/org/mockito/verification/VerificationMode.java
szczepiq: Right, this type will be made public (breaking change) in the next major version of Mockito, see  #744. It's not really feasible to stop leaking private API without introducing incompatible changes.

src/main/java/org/mockito/invocation/MatchableInvocation.java
```diff
@@ -0,0 +1,70 @@
...
+ * It is used during verification process:
+ *
+ * <pre class="code"><code class="java">
+ *   mock.foo();   // <- invocation
+ *   verify(mock).bar();  // <- matchable invocation
+ * </code></pre>
+ *
+ * @since 2.2.12
+ */
+public interface MatchableInvocation extends DescribedInvocation {
```
ChristianSchwarz: @szczepiq can you add `@Beta` or `@Incubation`, there should be some room for modifications of this API. Especially the methods `hasSimilarMethod()` and `hasSameMethod()` should be questioned.

szczepiq: BTW. Thanks for checking in!

PR #740, opened 2016-11-04
fixes #732 allow user to override configuration class name

### LINKED ISSUE #732: The mechanism to provide MockitoConfiguration is not compatible with java 9 jigsaw
Error on mockito 2.2.9

If your module is a test module it cannot export a package that is already exported by another package,

it is not possible to create the `org.mockito.configuration.MockitoConfiguration` inside a jigsaw module

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.6.0:testCompile (default-testCompile) on project sfm-datastax: Compilation failure: Compilation failure:
[ERROR] /home/aroger/dev/github/SimpleFlatMapper/sfm-datastax/src/test/java/org/mockito/configuration/MockitoConfiguration.java:[1] error: package exists in another module: mockito.core
[ERROR] /home/aroger/dev/github/SimpleFlatMapper/sfm-datastax/src/test/java/org/mockito/configuration/MockitoConfiguration.java:[3,42] error: cannot find symbol
[ERROR] symbol: class DefaultMockitoConfiguration
[ERROR] /home/aroger/dev/github/SimpleFlatMapper/sfm-datastax/src/test/java/org/mockito/configuration/MockitoConfiguration.java:[4,4] error: method does not override or implement a method from a supertype
```

arnaudroger:
The changes allow one to add a System property to change the configuration class from 
org.mockito.configuration.MockitoConfiguration to your own named class

-Dorg.mockito.configuration.className=my.package.MyMockitoClass

The issue with java9 module is that you can export a package that already contained in another module, the unamed mockito module here. With that change you can have the config in your own package that will not conflict.

check list

 - [ ] Read the [contributing guide](https://github.com/mockito/mockito/blob/master/.github/CONTRIBUTING.md)

### DIFF
--- a/src/main/java/org/mockito/configuration/IMockitoConfiguration.java
+++ b/src/main/java/org/mockito/configuration/IMockitoConfiguration.java
@@ -14,7 +14,8 @@ import org.mockito.stubbing.Answer;
  * when you might want to have different 'mocking style' this interface might be helpful. 
  * A reason of configuring Mockito might be if you disagree with the {@link ReturnsEmptyValues} unstubbed mocks return.
  * <p>
- * To configure Mockito create exactly <b>org.mockito.configuration.MockitoConfiguration</b> class that implements this interface.
+ * To configure Mockito create exactly <b>org.mockito.configuration.MockitoConfiguration</b> class that implements this interface. You can also specify another target class by setting the
+ * <b>org.mockito.configuration.className</b> system property.
  * <p>
  * Configuring Mockito is completely <b>optional</b> - nothing happens if there isn't any <b>org.mockito.configuration.MockitoConfiguration</b> on the classpath. 
  * <p>
--- a/src/main/java/org/mockito/internal/configuration/ClassPathLoader.java
+++ b/src/main/java/org/mockito/internal/configuration/ClassPathLoader.java
@@ -29,6 +29,8 @@ import org.mockito.plugins.MockMaker;
  *     // ...
  * }
  *     </code></pre>
+ *         The user can change the name of the class by setting the system property org.mockito.configuration.className
+ *         to the name of the IMockitoConfiguration implementation.
  *     </li>
  *     <li>
  *         Can load available mockito extensions. Currently Mockito only have one extension point the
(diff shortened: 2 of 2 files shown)


bric3:
Could you describe the change in the PR, the default template does not help ;)
The PR misses some unit tests.

How does that solve Java 9 ?

arnaudroger:
I change the description, I'll see if I can add unit test for that.

szczepiq:
Thanks for the PR! On the high level, the feature makes sense. However, I'd like the team to think through about different ways of configuring Mockito and come up with good cohesive strategy. 

Currently we have:
 - MockitoConfiguration
 - Plugins in "org.mockito.plugins" package
 - "inline-mock" inside plugin implementation

We can also include:
 - Mockito.framework().addListener()
 - JUnitRunner / JUnitRule

bric3:
Currently what I have to say

> * MockitoConfiguration

Old maybe even deprecated, but without replacement.

> * Plugins in "org.mockito.plugins" package

You mean the mockito-extensions system that is a service-loader clone.

> * "inline-mock" inside plugin implementation

It's the same as the mockito plugin, it just offer a short hand alias.

> * Mockito.framework().addListener()

Usage intended for framework developers 

> * JUnitRunner / JUnitRule

JUnit 5 don't have anymore _rules_ or _runners_, they have now _extensions_. However they are not done with the development. Also they are a good fit for configuring a test scope (the test class) not all the tests in a whole module.

szczepiq:
@arnaudroger, can you describe your use case - what do you use MockitoConfiguration for? We didn't hear a lot of feedback about this feature from our users. We don't know how useful it is.

arnaudroger:
@szczepiq I used it to disable the objenesis class cache. I had some test that used a different class loader and with the class cache to avoid ClassCastException because the mock was cached on a different class loader. 
Now I realized that I don't need it anymore but that was the original reason.

szczepiq:
Hey, given that you no longer need to disable objenesis cache, do you still need this feature? If not, I'll close the PR without merging. We can always reopen it if we need the feature.

arnaudroger:
@szczepiq yes that's fine with me.

szczepiq:
Thanks for feedback. Closing!

Issue #643, opened 2016-09-19
Support lambda when syntax


TimvdLippe:
It would be awesome to support a syntax like this:
```java
Mockito.when(mock::method).isInvokedWith(foo,bar).thenAnswer((foo, bar) -> 5);
```

The benefits: true type safety. Currently we have an interim solution which helps a lot with Java 8, but it still [relies on casting](https://github.com/mockito/mockito/blob/master/src/main/java/org/mockito/internal/stubbing/answers/AnswerFunctionalInterfaces.java#L39).
Also the `when` syntax can be confusing compared to the `verify` syntax, since verify actually has the method call on the return value of `verify()` whereas `when` wants the method call inside its function argument.

This reads a lot more fluent imo and given the compiler warnings and similarity to `verify` could greatly benefit new users.

The biggest issue with this syntax is probably: how do we track the lambda reference? I have tried to come up with a proof of concept, but since `mock::method` just returns a `Function` without any points to what mock it was actually referring too, I could not extract the actual mock we should stub for. Maybe we can do some Bytecode magic there?

szczepiq:
The benefits come at the cost of clarity and simplicity :)

However, it is definitely worth exploring. I'm wondering if Java8 has sufficient feature set that we can rethink the entire Mockito API :) After all, Mockito was built for java 6 and with the features available in java 6 :)

bric3:
I would even say the feature of Java 5 as there wasn't so much in Java 6 ;)

ChristianSchwarz:
@TimvdLippe I wonder how the signature of `isInvokedWith(foo,bar)` would look like, is it an Object-Vararg? If yes it is not possible to ensure a dev pass always the correct number of arguments and also type safty can't be guaranteed.

TimvdLippe:
@szczepiq I would actually argue that this syntax is a lot more clearer than the current `when` syntax.

@ChristianSchwarz So since we pass a `Function` into `when`, we can match on the generics of the function. E.g. if we pass a function we 2 arguments, the definition of `when` should be:

```java
public <F, S, R> LambdaWhenMatcher3<F, S, R> when(BiFunction<F, S, R> function) {
  // Do stuff with the function and use generics to ensure LambdaWhenMatcher has the correct type
}
```

Then in `LambdaWhenMatcher3` we define `isInvokedWith`:

```java
public LambdaWhenReturn<F, S, R> isInvokedWith(F first, S second) {
  // Again check for the matchers etc...
}
```

[...]

PR #742, opened 2016-11-04
[Heavy WIP] Lambda when

TimvdLippe:
This is a prototyping PR to investigate the feasibility of stubbing with lambda's.

### Motivation

There are several reasons for using lambda's as stubs, including but not limited to:
1. Full type-safety when stubbing. This includes
    1. Type-safety when using answers. You can now pass a normal lambda as answer instead of (internally) relying on `getArgumentAt`.
    2. Type-safety for using `ArgumentMatcher` vs literal values. When you are mixing `ArgumentMatcher` with literal values, an exception on run-time is thrown. This now brings it to compile-time
2. Unifying way to handle consumers and functions. In Mockito 2 void methods are handled with `doReturn` and `doAnswer`. In my experience teaching developers starting with Mockito this brings quite some confusion. Instead, Mockito should internally handle the way to process consumers.
    1. In essence we can remove all previous ways to stub a method. That would include `doReturn()`, `doThrow()`, `doAnswer()`, `doNothing()`, `doCallRealMethod()`. These APIs are in my experience very confusing and that was also feedback I got from Mockito starters.
3. Fix issues with unwanted side-effects when re-stubbing already stubbed methods. In the current engine it is hard to fix these issues, since the invocation is performed before the actual stubbing is done. Since the "invocation" is now delayed, these issue can actually be dealt with.
    1. This probably effects Spies the most where calling of methods is expected, but not when stubbing.

In a very similar fashion to `MockitoLambda.when`, `MockitoLambda.given` can also be implemented. The naming of the methods will be different, but the implementation is equal.

I am convinced that we can fix more use-cases and bugs using this syntax as we have full control over the actual method and corresponding invocation. Moreover we can upgrade the backing engine for full type-safety instead of relying on Reflection to trace back the corresponding stubbed answer.

### Syntax
```java
public class LambdaStubbingTest {

    @Test
    public void stubs_method_as_lambda() {
        IMethods mock = Mockito.mock(IMethods.class);
        MockitoLambda.when(mock::threeArgumentMethodWithStrings).isInvokedWith(anyInt(), anyString(), anyString()).thenReturn("foo");
        assertThat(mock.threeArgumentMethodWithStrings(5, "foo", "bar")).isEqualTo("foo");
    }

    @Test
    public void stubs_void_method_as_lambda() {
        IMethods mock = Mockito.mock(IMethods.class);
        AtomicBoolean atomicBoolean = new AtomicBoolean(false);
        MockitoLambda.when(mock::twoArgumentMethod).isInvokedWith(anyInt(), anyInt()).thenAnswer((a, b) -> {
            atomicBoolean.set(true);
        });
        mock.twoArgumentMethod(0, 1);
        assertThat(atomicBoolean.get()).isEqualTo(true);
    }
}
```

The difference boils down to:
```java
when(mock::threeArgumentMethodWithStrings).isInvokedWith(anyInt(), anyString(), anyString()).thenReturn("foo");
```
vs
```java
when(mock.threeArgumentMethodWithStrings(anyInt(), anyString(), anyString())).thenReturn("foo");
```
Note that `isInvokedWith` is a bit wordy, but happy to bike-shed on the naming. Could also just be `with`:

```java
when(mock::threeArgumentMethodWithStrings).with(anyInt(), anyString(), anyString()).thenReturn("foo");
```

The difference between the two is a total of 6 characters, which I don't think is too bad.

There were some initial concerns regarding the syntax. The syntax is of course different from the "old" syntax, but I do not think the difference harms users. I would actually argue that the lambda syntax is easier to grasp for starters and in my opinion it makes more sense.

It might be a good idea to, if the prototype appears to be working, ask users whether they like the new syntax.

### DIFF
--- a/src/main/java/org/mockito/MockitoLambda.java
+++ b/src/main/java/org/mockito/MockitoLambda.java
@@ -1,0 +1,30 @@
+package org.mockito;
+
+import org.mockito.internal.MockitoCore;
+import org.mockito.stubbing.OngoingStubbing;
+
+public class MockitoLambda {
+
+    static final MockitoCore MOCKITO_CORE = new MockitoCore();
+
+    public static <A, B, C, R> OngoingLambdaStubbing<A, B, C, R> when(TriFunction<A, B, C, R> method) {
+        return new OngoingLambdaStubbing<>(method);
+    }
+
+    @FunctionalInterface
+    public interface TriFunction<A, B, C, R> {
+        R apply(A var1, B var2, C var3);
+    }
+
+    public static class OngoingLambdaStubbing<A, B, C, R> {
+        private final TriFunction<A, B, C, R> method;
+
+        public OngoingLambdaStubbing(TriFunction<A, B, C, R> method) {
+            this.method = method;
+        }
+
+        public OngoingStubbing<R> isInvokedWith(ArgumentMatcher<A> var1, ArgumentMatcher<B> var2, ArgumentMatcher<C> var3) {
+            return MOCKITO_CORE.when(method.apply(var1, var2, var3));
+        }
+    }
+}
--- a/src/main/java/org/mockito/MockitoLambda.java
+++ b/src/main/java/org/mockito/MockitoLambda.java
@@ -1,14 +1,21 @@
 package org.mockito;
 
 import org.mockito.internal.MockitoCore;
+import org.mockito.stubbing.Answer2;
+import org.mockito.stubbing.Answer3;
 import org.mockito.stubbing.OngoingStubbing;
+import org.mockito.stubbing.VoidAnswer2;
 
 public class MockitoLambda {
 
     static final MockitoCore MOCKITO_CORE = new MockitoCore();
 
-    public static <A, B, C, R> OngoingLambdaStubbing<A, B, C, R> when(TriFunction<A, B, C, R> method) {
-        return new OngoingLambdaStubbing<>(method);
+    public static <A, B, C, R> OngoingLambdaStubbingReturn3<A, B, C, R> when(TriFunction<A, B, C, R> method) {
+        return new OngoingLambdaStubbingReturn3<>(method);
+    }
+
+    public static <A, B> OngoingLambdaStubbingVoid2<A, B> when(DuoConsumer<A, B> method) {
+        return new OngoingLambdaStubbingVoid2<>(method);
     }
 
     @FunctionalInterface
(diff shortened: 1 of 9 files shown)


szczepiq:
Cool. Keep prototyping!

ChristianSchwarz:
relates to #643 643

ChristianSchwarz:
Overloaded methods won't work with this kind of API cause the instance reference doesn't satifiy the functional interface contract?

```java
IMethods mock; //String:oneArg(boolean), String:oneArg(String)

when(mock::oneArg).isInvokedWith(anyString()).thenReturn("foo");
```

TimvdLippe:
Updated the PR to show the intent of how to use void and only literal values.

To solve the ambiguous issue, we could add an overload which takes a lambda instead. You lose the typesafety there. Then the user can opt-in of type-safety if the method can be determined unambiguously.

TimvdLippe:
The last commit now has some more implementations, inspired by #828 of @ChristianSchwarz . Essentially it calls the method itself, but with all typesafety around it of `ArgumentMatcher`s and concrete values.

At this point I am touching a lot of core code, namely handling of invocations. Mainly to make it typesafe or handle the new interfaces, which are different than the previous 1 `Answer` interface.

In essence handling of `ArgumentMatcher` should still be the same, since matchers have to be created and reported to the internal core, but instead of directly returning a value, they return the matcher. The actual value is retrieved during the `isInvokedWith` to ensure it typechecks.

Consumer <-> Function are handled differently in the way that Consumer only have `thenAnswer`, but Function also has `thenReturn`.

In the `MockingProgress`, instead of an `OnGoingStubbing`, the `invocationContainer` itself is stored and later passed on. This container can then be used for the new `FinishableXXX` interfaces to correctly pass on the value, see `pullFinishableStubbing`.

At this point everything compiles type-safely, but the tests fail as the invocation handling has not been written.

src/main/java/org/mockito/MockitoLambda.java
```diff
@@ -0,0 +1,118 @@
...
+
+    public static class OngoingLambdaStubbingVoid2<A, B> {
+        private final DuoConsumer<A, B> method;
+
+        public OngoingLambdaStubbingVoid2(DuoConsumer<A, B> method) {
+            this.method = method;
+        }
+
+        public FinishableAnswer<Void, VoidAnswer2<A, B>> isInvokedWith(ArgumentMatcher<A> var1, ArgumentMatcher<B> var2) {
+            method.apply(var1.getValue(), var2.getValue());
```
LiamClark: Per discussion with @TimvdLippe : we can reset any previous stubs here, so that when a user wrote a thenAnswer for this stub, it is not invoked.

szczepiq:
@TimvdLippe, did the API proposal change after your recent changes? Can you update the PR description with how it looks like now (example). Also, can you write in the description more information like: whether you propose to replace existing api VS an alias, and what do you suggest for given() api. One more request - can you include in the description the motivation behind the new API (is it for consistency with void methods? any other reasons?).

I feel relatively strongly against 'isInvokedWith'. It is very not mockito-like. Mockito aims for maximum readability where how you see invocations in tests is very similar how you see them in code under test.

I'm really happy that we're considering different kinds of APIs. Let's innovate!!! :D

TimvdLippe:
@szczepiq Planned to do that yesterday but ran out of time.

@mockito/core I have updated the description of this PR with a simple motivation. I have discussed this PR with several others and every discussion we had yet another usecase that could be fixed with this PR. I am very excited about this and the corresponding issues it can fix, but have to make it work first :smile:

szczepiq:
Thank you for updating the description - it's fantastic! The arguments are really strong. When you have a chance, can you also add few code examples to the description?

It would be great if we could test this new syntax out somehow and evaluate the impact on developer experience. The newer syntax seems to be slightly more wordy / less readable for simple cases (we want all unit tests to be simple!). However, it really solves some of the pains we have. It would also help with opt-in strict mocking because strict mocks would require doReturn syntax.

I'm still not quite sold on the idea but the arguments you put are very strong and perhaps we can figure out a good solution.

TimvdLippe:
Updated the description with example usages 😄

szczepiq:
@TimvdLippe, excellent examples, thank you. Is the prototype working?

TimvdLippe:
@szczepiq It compiles, but the actual stubbing is not working just yet. Still working on this PR, haven't had the time lately to continue working on it.

TimvdLippe:
After my exams I am going to continue working on this :smile:

bric3:
@TimvdLippe FOCUS on your exams :D

TimvdLippe:
I am going to take another stab at this PR after I am home again. I have a new plan to maybe make this work. #prototyping :smile:

TimvdLippe:
This PR has been superseded by #1264

PR #745, opened 2016-11-07
Removes deprecated way of documenting package by package-info.java

### LINKED ISSUE #665: Use package-info.java instead of the old html file
`package-info.java` can be annotated, and reflection can be used, this may be useful at some point, or not.

bric3:
Should fix #665 

Nothing to say really, but it has to be done at some point. The most important selling point is that it may allow to annotate a package.


bric3:
Thanks @TimvdLippe I believe this is OK javadoc wise

TimvdLippe:
I trust you then 😄

bric3:
Around midnight? 😛

szczepiq:
Nice change, thanks for keeping the codebase clean!!!

PR #746, opened 2016-11-07
Add missing copyright headers

### LINKED ISSUE #727: Copyright notice missing on files
Let's add automation that will make consistent copyright notice on all source files.

DmitriyZaitsev:
- [x] Fixes #727 
- [x] Add missing copyrights using [license-gradle-plugin](https://github.com/hierynomus/license-gradle-plugin).
- [x] Updated only source files. Test files were skipped. Existing copyright headers were not modified.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,6 +5,7 @@ buildscript {
 
     dependencies {
         classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.2' //publishing to bintray
+        classpath'nl.javadude.gradle.plugins:license-gradle-plugin:0.11.0' // later versions don't work on JDK6
     }
 }
 
@@ -25,6 +26,7 @@ apply from: "gradle/ide.gradle"
 apply from: 'gradle/coverage.gradle'
 apply from: 'gradle/osgi.gradle'
 apply from: 'gradle/gradle-fix.gradle'
+apply from: 'gradle/license.gradle'
 
 allprojects {
     repositories {
--- a/gradle/license.gradle
+++ b/gradle/license.gradle
@@ -1,0 +1,14 @@
+apply plugin: 'com.github.hierynomus.license'
+
+license {
+    header = rootProject.file('doc/licenses/HEADER.txt')
+    strictCheck = true
+    ignoreFailures = true
+    skipExistingHeaders = true
+    mapping {
+        java = 'SLASHSTAR_STYLE'
+        groovy = 'SLASHSTAR_STYLE'
+    }
+    ext.year = Calendar.getInstance().get(Calendar.YEAR)
+    ext.name = 'Mockito contributors'
+}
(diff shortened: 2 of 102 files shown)


bric3:
Unfortunately the plug-in is not compiled for Java 6.

I really would like to avoid as much as possible maintaining 2.x too long.

bric3:
However I suppose we can change the JDK that gradle use if this is a separate task in travis.

Why a separate task, because if gradle runs the under the code under another JDK we may miss specific JDK 6 runtime behavior.

TimvdLippe:
Per https://github.com/hierynomus/license-gradle-plugin/issues/35#issuecomment-238446002 we can use version 0.11.0 and make it work on JDK6

bric3:
Good point, let's do that instead.

bric3:
@DmitriyZaitsev By I'd like to say you have done a proper pull request, that's very nice!

DmitriyZaitsev:
oops. smth went wrong. need to re-push changes

DmitriyZaitsev:
@bric3 I've switched plugin version to `0.11.0`, rebased my branch and resubmitted.

HEADER.txt
```diff
@@ -0,0 +1,2 @@
+Copyright (c) ${year} ${name}
```
szczepiq: Can we put this file under doc/licenses dir?

gradle/license.gradle
```diff
@@ -0,0 +1,14 @@
...
+license {
+    header = rootProject.file('HEADER.txt')
+    strictCheck = true
+    ignoreFailures = true
+    skipExistingHeaders = true
+    mapping {
+        java = 'SLASHSTAR_STYLE'
+        groovy = 'SLASHSTAR_STYLE'
+    }
+    ext.year = Calendar.getInstance().get(Calendar.YEAR)
```
szczepiq: [not blocker] I'm curious, should the header contain both dates, e.g. 2007-2016?

build.gradle
```diff
@@ -160,3 +161,5 @@ buildScan {
     licenseAgreementUrl = 'https://gradle.com/terms-of-service'
     licenseAgree = 'yes'
 }
+
+apply from: 'gradle/license.gradle'
```
szczepiq: For clarity, is it possible to move this statement to the line where we apply other Gradle script plugins?

DmitriyZaitsev: Other gradle scripts aren't declared together. You can find `apply from` declarations all over the file. But yes, it's of course possible.

gradle/license.gradle
DmitriyZaitsev: Actually, there should be the year of the last change.
I've skipped existing headers because every time the plugin will update all source files. It's very unnecessary.

build.gradle
DmitriyZaitsev: Done

HEADER.txt
DmitriyZaitsev: Done

DmitriyZaitsev:
Added fixes after @szczepiq review

TimvdLippe:
@DmitriyZaitsev Thanks!

bric3:
thx @DmitriyZaitsev

PR #747, opened 2016-11-08
Allow fluent usage of Mockito.framework() listeners methods

bric3:
Currently `addListener` and `removeListener` returns `void` which makes it repetitive to add or remove these :

```java
Mockito.framework().addListener(...);
Mockito.framework().addListener(...);
Mockito.framework().addListener(...);
```

The main change of this PR is to return `MockitoFramework` instead, thus allowing to chain invocations :

```java
Mockito.framework().addListener(...)
                   .addListener(...)
                   .addListener(...);
```

This improvements is a breaking change, however the API is still **incubating** and should be expected to break, source wise there's no problem since nothing could be done before due to the `void` return.

### DIFF
--- a/src/main/java/org/mockito/MockitoFramework.java
+++ b/src/main/java/org/mockito/MockitoFramework.java
@@ -37,7 +37,7 @@ public interface MockitoFramework {
      * @param listener to add
      */
     @Incubating
-    void addListener(MockitoListener listener);
+    MockitoFramework addListener(MockitoListener listener);
 
     /**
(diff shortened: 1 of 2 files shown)


szczepiq:
Very nice change, thank you! This is exactly the direction we should take - making the api easy to use, step by step :)

@TimvdLippe, you are exactly right that this change is technically backwards compatible. Since MockitoFramework is not intended for extending, we can go ahead release it as is. +1

PR #751, opened 2016-11-12
Pushed release automation code into separate GitHub repository

szczepiq:
Moved release automation code into separate GitHub repository 'mockito-release-tools'. This way it can be tested and shared easily.

 - Removed most code from buildSrc, instead we're using binary dependency
 - Added necessary dependencies

Next steps:

 - Move most of 'release.gradle' and 'version.gradle' into opinionated Gradle plugins under 'mockito-release-tools'
 - Enable autopublication of 'mockito-release-tools'

Tested by running: "./gradlew release -PdryRun"


gradle/release.gradle
```diff
@@ -1,3 +1,16 @@
+buildscript {
+    repositories {
+        maven { url "https://plugins.gradle.org/m2/" }
+    }
+    dependencies {
+        //so that PublicationsComparatorTask can be imported
+        //this entire 'buildscript' section will be removed when the entire release automation is moved to new repo 'mockito-release-tools'
```
szpak: 2 years ago it didn't work to use classes from external plugins in a separate gradle file without adding `buildscript` section in that file, even if a separate plugin was used. However, maybe it has been fixed in the meantime.

buildSrc/build.gradle
```diff
@@ -1,30 +1,5 @@
...
-repositories { jcenter() }
-
 dependencies {
-    compile gradleApi()
-    //TODO SF use jcabi to edit issues after the release so that they have the milestone attached
-    //compile "com.jcabi:jcabi-github:0.17"
-    compile "com.googlecode.json-simple:json-simple:1.1.1@jar"
+    //needed by JavadocExclude
+    //TODO find out if we really need JavadocExclude. Why don't we just exclude internal packages using Gradle API (Javadoc.exclude)?
     compile files("${System.properties['java.home']}/../lib/tools.jar")
```
bric3: I don't think that's possible at this time, or that can be brittle, the reason I believe is that gradle invoke javadoc file by file. This mode disable some features of `javadoc` tool, and may even make the `javadoc` tool fail if there's a reference to some excluded file within the javadoc itself.

PR #759, opened 2016-11-16
Updated the verification documentation to correctly use argThat with Java8 lambdas

LiamClark:
The example for custom argument matchers using a lambda in the main mockito documentation is missing a call to argThat.

This pr fixes the docs to correctly use argThat.


TimvdLippe:
SGTM? Can't test it right now, but almost certain it is the correct fix

szczepiq:
Very nice catch, thank you! Merging... @TimvdLippe, you'll have a chance to test the Javadoc changes in real.

PR #760, opened 2016-11-17
Updated internal documentation regarding Javadoc doclet

szczepiq:
Updated internal documentation regarding Javadoc doclet.


TimvdLippe:
Right, I ran the javadoc locally and I remember again what the issue was. If you remove the files **before** the JavaDoc runs (via the exclude option), the tool will generate a warning for **every** missing import. E.g. all of the internal classes imports will generate a warning. It cuts of the amount of warnings after 100, but in total there are a lot more.

Therefore the doclet is required to remove the files **after** the JavaDoc runs. This mitigates these warnings, will generate correct JavaDoc nonetheless and is therefore the better solution.

szczepiq:
Great code review, thanks @TimvdLippe for catching this. We don't want to increase the number of warnings because we will miss out important warnings.

I've updated this PR to only fix the documentation. Reopening, please review / merge.

TimvdLippe:
I think this is superseded by #763 then, since @bric3 did almost the same changes

szczepiq:
Thank you for raising this! I'll wait for @bric3 changes to get merged, rebase this PR if there isn't anything useful left in this PR, I'll drop it.

bric3:
@szczepiq I merged #763, I think it covers what you have done here.

szczepiq:
Let's drop this rb. Thanks for the improvements!

PR #763, opened 2016-11-17
Fixes #762 Removes ReturnsEmptyValue javadoc references

### LINKED ISSUE #762: Public javadoc offers links to internal classes
In few places in public javadoc we offer links to internal classes. Instead, we should document the public API without links to private classes. How to find offenders? Generate javadoc and look for:

- \<code\>ReturnsEmptyValues
- \<code\>org.mockito.internal

bric3:
- Removes `ReturnsEmptyValue` (internal) generated links.
- Didn't remove `<code>org.mockito.internal.creation.bytebuddy.InlineByteBuddyMockMaker</code>`, the only result for `<code>org.mockito.internal` as it is a documentation not an actual link.
- Removes `ReturnsMoreEmptyValue` references
- Make use of gradle javadoc task exclude
- Fixes package view rendering : see [2.2.1](https://javadoc.io/doc/org.mockito/mockito-core/2.2.21)

See #762


bric3:
@TimvdLippe I tweaked the HTTPS switch in `8f968f8` as it was messing with the all package frame

TimvdLippe:
@bric3 Ah yes, was already wondering: why is there no option for this. Must have missed it. :+1:

bric3:
@TimvdLippe Well no, I don't know why the gradle guys [didn't include this option](https://docs.gradle.org/3.2/javadoc/org/gradle/external/javadoc/StandardJavadocDocletOptions.html). It's not the only thing that is weird with the javadoc task.

bric3:
By the way I don't know why there's less coverage. Could be due to the [variable renaming](https://codecov.io/gh/mockito/mockito/src/8f968f8a4bb5b59a766913d9c722fcc5ca1e21c2/src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsMoreEmptyValues.java)...

gradle/javadoc.gradle
```diff
@@ -14,8 +14,11 @@ task mockitoJavadoc(type: Javadoc) {
     title = "Mockito ${project.version} API"
     def javadocJdk6Hack = "jdk6-project-version-insert.min.js"
 
-    options.doclet "org.mockito.javadoc.JavadocExclude"
-    options.docletpath = [rootProject.file('./buildSrc/build/classes/main/')]
+    // The doclet is more flexible but still allow to generate internal links, while exclude is more of a hammer.
+    exclude 'org/mockito/internal'
```
TimvdLippe: This change does not work, see https://github.com/mockito/mockito/pull/760#issuecomment-261229080

bric3:
@TimvdLippe Also I'm not sure about using the exclude instead of the doclet, locally it works but I'm using jdk 7

bric3:
I played a bit more with the doclet code, it's a mess. It is really hard to filter internal code everywhere.

TimvdLippe:
Yes it is a mess. JDK9 should fix this with the new javadoc tooling api.

bric3:
After a lot of experimentation with the doclet and gradle. I may have found a nice compromise regarding the use of the `exclude` statement of the gradle task. It's not perfect, but since extending the doclet to filter everything internal is an immense task, that can be thrown away with JDK9, LI chose to go the gradle way. The reason is explained in the commit and in the gradle `javadoc` task.

But the thing is these errors still allow to generate the javadoc, just without the internal symbols. To **hide** these errors I found a trick to lower the log output level.

Here's what the javadoc look like without the internal symbol :

* The internal parent doesn't appear, it's weird yet I find that acceptable
    Timeout

* Here links are not generated, instead it's just text (to compare with [code](https://github.com/mockito/mockito/pull/763/files#diff-507cbf8e34e6d7f8bb99123acae1cc29R17))
    StackTraceCleaner

The nice thing is that developers can still navigate those in their IDE.

TimvdLippe:
The one question I still have: why do we need to modify the doclet now? If it is working right now, we can keep it as is.

bric3:
It has been discussed over slack, but for reference the doclet still lets internal links slip through in the javadoc for exemple : 

* in the [`After`](https://static.javadoc.io/org.mockito/mockito-core/2.2.21/org/mockito/verification/After.html) javadoc, click the extends class `VerificationWrapper`
* in the [`StackTraceCleaner`](https://static.javadoc.io/org.mockito/mockito-core/2.2.21/org/mockito/exceptions/stacktrace/StackTraceCleaner.html) click either link `DefaultStackTraceCleaner` or `StackTraceFilter` in the javadoc.

And still get an error.

TimvdLippe:
Oh I misunderstood on Slack, but that is clear now. Sad to see it go, but no free lunch in computer science :stuck_out_tongue:

bric3:
Yeah I kinda liked the doclet approach too, and would have preferred it if it was possible to fix in a reasonable way and time. Could be another github project ;)

PR #768, opened 2016-11-19
Experimentation with MockitoJUnit.rule().strict() rule

szczepiq:
Started experimenting with strict rules. Take a look, not intended for merge, yet :)

The target is following new API:

```java
public @Rule MockitoRule rule = MockitoJUnit.rule().strict();
```


src/main/java/org/mockito/internal/junit/JUnitRule.java
```diff
@@ -8,38 +8,49 @@
...
+    public JUnitRule(MockitoLogger logger, Strictness strictness) {
 		this.logger = logger;
-        this.silent = silent;
+        this.strictness = strictness;
     }
 
 	@Override
 	public Statement apply(final Statement base, FrameworkMethod method, final Object target) {
-        if (silent) {
+        if (strictness == Strictness.SILENT) {
```
TimvdLippe: Lets do `strictness.applyStatement(...)` here. Makes it easier to add a new enum value for later.

src/main/java/org/mockito/internal/junit/JUnitRule.java
```diff
@@ -87,8 +100,17 @@ private void performEvaluation(RuleStubbingHintsReporter reporter) throws Throwa
...
             }
-            reporter.getUnusedStubbings().format(testName, logger);
             Mockito.validateMockitoUsage();
+            if (strictness == Strictness.ERROR) {
+                SimpleMockitoLogger log = new SimpleMockitoLogger();
+                reporter.getUnusedStubbings().format(testName, log);
+                if (!log.isEmpty()) {
+                    throw new MockitoAssertionError(log.getLoggedInfo());
+                }
+            } else if (strictness == Strictness.WARN) {
```
TimvdLippe: Same as above, but now `strictness.evaluate(...)`

src/main/java/org/mockito/junit/MockitoRule.java
```diff
@@ -76,4 +76,6 @@
      * @since 2.1.0
      */
     MockitoRule silent();
+
+    MockitoRule strict();
```
TimvdLippe: TODO add javadoc

szczepiq:
Scratching this PR. I'll get there with smaller increments.

Issue #769, opened 2016-11-19
Strictness in Mockito


szczepiq:
This issue explores the addition of "strict" behavior and APIs to Mockito. Traditionally, Mockito is a lenient mocking framework.

Why strictness in Mockito:
  - productivity - improved debugging and writing tests
  - clean code - avoid unnecessary stubbings

Strictness in Mockito can have 2 forms:
  - Strict stubbing that requires that all stubs are actually used
  - Strict mocks, that require all non-void methods

Strictness behavior can be implemented as:
  - warning on the console
  - exception thrown

Strictness API can be implemented in:
  - JUnit runner / rule
  - TestNG runner
  - explicit method to call by the user

Future direction:
  - strict stubbing on by default, opt-out available
  - strict mocking off by default, opt-in available

Implementation as of Mockito 2.2.x:
  - JUnitRunner - fails at the end of test class when unused stubbings are present. Opt out available via JUnitRunner.Silent
  - JUnitRule - warn at the end of test method about a) unused stubs b) stubs invoked with different arguments. Opt-out available via rule().silent().

Mockito 2.3 proposal:
  - new rule().strictStubs() API that:
    - fail when mock method called with different arguments
    - fail when mocking is not used
  - can declare “optional” stubbings
  - new verifyStubs() method to use when rule/runner not in used
  - new Answer.StrictMock that fails immediately when non-void method invoked on without prior stubbing
  - verifyNoMoreInvocations() no longer require to verify “used” stubs _only_ when the strictStubs() are used

Mockito 3 proposal:
  - strict stubbing is the default
  - change behavior of verifyNoMoreInvocations()
  - runner and rule behave the same

imod:
I think from a code maintenance point of view, "behavior: warning on the console" is not really useful - in such a case a CI system would not inform anyone about unused stubs. An exception would be a lot better.

TimvdLippe:
@szczepiq could you add list items (`- [ ] ` in markdown) to signify what the remaining action points in this issue are? That gives a clearer overview, especially as multiple open issues are related to this single epic.

szczepiq:
>@szczepiq could you add list items (- [ ] in markdown) to signify what the remaining action points in this issue are? That gives a clearer overview, especially as multiple open issues are related to this single epic.

I will do that once this epic turns into an execution plan. Currently it is intended to scope out the details and I don't know if all of them will be executed (and in what form). I annotated one item as merged by using strikethrough. I hope it is clearer now.

[...]

PR #770, opened 2016-11-19
Opt-in stubbing strictness implemented in JUnit rules

szczepiq:
New API (for rationale, see #769):

```java
//Existing API:
@Rule public MockitoRule mockito = MockitoJUnit.rule()

//Proposed improvements:
rule().strictness(Strictness.LENIENT)
rule().strictness(Strictness.WARN)       // <-- is the default
rule().strictness(Strictness.STRICT_STUBS)  // <-- Mockito 3.0 default
```

What "strict stubbing" means?
- Improved debuggability: the test fails early when code under test invokes stubbed method with different arguments.
- Cleaner tests without unnecessary stubbings: the test fails when there are any unused stubs declared.
- Cleaner, more DRY tests ("Don't Repeat Yourself"): If you use Mockito.verifyNoMoreInteractions() you no longer need to explicitly verify stubbed invocations. They are automatically verified.

For background and motivation, check out [Szczepan's article on LinkedIn](https://www.linkedin.com/pulse/curious-how-get-even-cleaner-tests-new-mockito-features-faber).


src/main/java/org/mockito/internal/junit/WarningTestListener.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public void afterTest(Collection<Object> mocks, Throwable testFailure) {
+        if (testFailure != null) {
+            //print stubbing mismatches only when there is a test failure
+            //to avoid false negatives. Give hint only when test fails.
+            new ArgMismatchFinder().getStubbingArgMismatches(mocks).format(testName, logger);
+        } else {
+            //Validate only when there is no test failure to avoid reporting multiple problems
+            Mockito.validateMockitoUsage();
```
bric3: I wonder, if there's a test failure. But if some arguments are misplaced, users won't see the [`throw misplacedArgumentMatcher(lastMatchers)`](https://github.com/mockito/mockito/blob/v2.2.21/src/main/java/org/mockito/internal/progress/ArgumentMatcherStorageImpl.java#L102) ?

szczepiq:
Thanks for feedback!!! This RB is in progress. I'm so itching to work on this!

szczepiq:
Added this API in recent update:

```java
public interface StubbingLookUpListener {

    /**
     * Called by the framework when Mockito looked up an answer for invocation on a mock.
     *
     * @param invocation - the invocation on the mock
     * @param stubbingFound -  can be null - it indicates that the invocation was not stubbed. If not null, it means that the stubbing was found on the mock.
     */
    void onStubbingLookUp(Invocation invocation, MatchableInvocation stubbingFound);
```

Feedback welcome.

src/main/java/org/mockito/internal/junit/StrictStubsTestListener.java
```diff
@@ -0,0 +1,71 @@
...
+
+    private final Map<Object, MockCreationSettings> mocks = new HashMap<Object, MockCreationSettings>();
+
+    public void beforeTest(Object testClassInstance, String testMethodName) {
+        //TODO init mocks and validate mockito usage is duplicated in the listeners, refactor or make sure all is tested
+        MockitoAnnotations.initMocks(testClassInstance);
+    }
+
+    public void afterTest(Throwable testFailure) {
+        if (testFailure == null) {
```
ChristianSchwarz: @szczepiq what about returning asap? This avoids indentation and improves readability.
```java
if (testFailure!=null){
   return;
}
```

src/test/java/org/mockitousage/junitrule/StrictJUnitRuleTest.java
```diff
@@ -0,0 +1,182 @@
...
+            public void evaluate(IMethods mock, IMethods mock2) throws Throwable {
+                given(mock.simpleMethod(10)).willReturn("foo");
+                mock.simpleMethod(10);
+                mock.simpleMethod(15);
+            }
+        });
+    }
+
+    @Test public void fails_when_unused_stubbings() throws Throwable {
+        try {
```
ChristianSchwarz: @szczepiq 
Would you mind using the ExpectedException rule, this would improve the readability and avoid try/fail/catch blocks which distracts the reader from the test case IMO.

Stephan202:
@szczepiq: I'm not a native speaker, but I think `StubbingLookUpListener` should be `StubbingLookupListener` (lower case `u`)?

src/main/java/org/mockito/internal/creation/settings/CreationSettings.java
```diff
@@ -104,6 +106,10 @@ public SerializableMode getSerializableMode() {
         return invocationListeners;
     }
 
+    public List<StubbingLookUpListener> getStubbingLookUpListeners() {
```
ChristianSchwarz: @szczepiq 
`Mockito.framework()` has `add`/`removeListener` to handle all types of listeners. What do you think about a unification of API listener concepts here?  I mean `CreationSettings.add/removeListener` instead of `CreationSettings.getStubbingLookUpListeners().add()` /` remove()` / `CreationSettings.getInvocationListeners().add()` /` remove()`.

This would make the API as a whole more clear and avoids different patterns for the same concept.

src/main/java/org/mockito/internal/junit/JUnitRule.java
```diff
@@ -7,88 +7,85 @@
...
-            MockitoAnnotations.initMocks(target);
-            base.evaluate();
-            Mockito.validateMockitoUsage();
+    /**
+     * @param logger target for the stubbing warnings
+     * @param strictness how strict mocking / stubbing is concerned
+     */
+    public JUnitRule(MockitoLogger logger, Strictness strictness) {
+        this.logger = logger;
+        if (strictness == Strictness.SILENT) {
```
ChristianSchwarz: @szczepiq 
please replace this by an swich case e.g like this to avoid code duplication:

```java
 listeners.add(newTestListener(strictness));
```
```java
MockitoTestListener newTestListener(Strictness strictness){
  switch(strictness){
    case SILENT:
       return new SilentTestListener();
   case WARN:
      return new WarningTestListener(logger);
  case STRICT_STUBS:
     return new StrictStubsTestListener();
  }
  throw new IAE("Unsuported Strictness "+strictness);
}
```

src/main/java/org/mockito/internal/junit/JUnitRule.java
```diff
@@ -7,88 +7,97 @@
...
+            listeners.add(new SilentTestListener());
+        } else if (strictness == Strictness.WARN) {
+            listeners.add(new WarningTestListener(logger));
+        } else if (strictness == Strictness.STRICT_STUBS) {
+            listeners.add(new StrictStubsTestListener());
```
TimvdLippe: Yes agreed. That was also what I proposed in #768 before that PR got closed

src/test/java/org/mockitousage/junitrule/StrictJUnitRuleTest.java
```diff
@@ -0,0 +1,182 @@
...
+                    //invocation in the code under test uses different argument and should fail immediately
+                    //this helps with debugging and is essential for Mockito strictness
+                    mock.simpleMethod(15);
+                }
+            });
+
+            //then
+            fail();
+        } catch (MockitoAssertionError e) {
+            e.printStackTrace();
```
TimvdLippe: Left-over statement. You can prevent this using `ExpectedException` as pointed out by @ChristianSchwarz

src/main/java/org/mockito/internal/creation/settings/CreationSettings.java
szczepiq: Nice catch!

In short: Framework listener != Mock object listener.

Currently framework listeners are kept in thread-local state and hence they cannot receive events from different threads. It is legal to have the same mock shared in different threads for normal interactions with mock.

Mock object listeners are kept with the mock object, hence they can receive events from multiple threads.

I don't claim the current distinction is good :) It is best I could think of given current state of mockito architecture. I didn't want to have 'global' static listeners.

Instead of unifying the listeners, I tentatively plan to make the distinction explicit in the API, e.g. MockitoListener as top marker interface, and new MockObjectListener as another top marker interface for mock object listeners.

szczepiq:
Thanks guys for great code review feedback!!! All suggestions make perfect sense.

While I still work on this feature / PR, the best feedback you can give me is at higher level (e.g. public API changes, the change to public behavior). I'm worried that smaller code suggestions will get lost as I shuffle the code ;)

szczepiq:
The feature is shaping up. If you review, keep an eye on the public API changes and the new behavior (the tests explain the new behavior) - I'm really keen on feedback on those. Rule strictness will be enabled by following API and I tentatively plan to make it a default in Mockito 3:

```java
MockitoJUnit.rule().strictStubs()
```

Thanks for keeping an eye on this and the review!!!

szczepiq:
I'm close to ready with this PR. Missing:
 - documentation
 - exposing of strictStubs() method

TimvdLippe:
Could you please give me some time to review this big PR? Dont have time
right now, but soon will!

src/main/java/org/mockito/internal/framework/DefaultMockitoFramework.java
```diff
@@ -23,4 +24,24 @@ public MockitoFramework removeListener(MockitoListener listener) {
         mockingProgress().removeListener(listener);
         return this;
     }
+
+    /**
+     * Adds listener only if it is an instance of MockitoListener.
+     * Useful to avoid casting in internal code.
+     */
+    public static void tryAddListner(Object listener) {
```
ChristianSchwarz: @szczepiq This smells, it looks like a design problem. Why don't you let MockitoTestListener  extend MockitoListener?

src/main/java/org/mockito/internal/junit/DefaultTestFinishedEvent.java
```diff
@@ -0,0 +1,28 @@
+package org.mockito.internal.junit;
+
+/**
+ * Created by sfaber on 11/24/16.
+ */
+public class DefaultTestFinishedEvent implements TestFinishedEvent {
```
ChristianSchwarz: @szczepiq Other implementations follow the *Impl pattern...

src/main/java/org/mockito/internal/junit/JUnitRule.java
```diff
@@ -8,87 +8,76 @@
...
-            try {
-                base.evaluate();
-            } catch(Throwable t) {
-                reporter.getStubbingArgMismatches().format(testName, logger);
-                throw t;
-            }
-            reporter.getUnusedStubbings().format(testName, logger);
-            Mockito.validateMockitoUsage();
-        }
+    public JUnitRule strictStubs() {
```
ChristianSchwarz: @szczepiq The specialisations should follow the same naming pattern. Either silentStubs()/strictStubs() or sielent()/strict()

src/main/java/org/mockito/internal/framework/DefaultMockitoFramework.java
szczepiq: Here're my reasons:

- Multiple interfaces with single methods are easier to evolve than APIs with "big" interfaces. It makes it easier to grow / add new APIs by addition of new interfaces without breaking current compatibility.
- MockitoTestListener and MockitoListener are different. The former is used by JUnit rule to add behavior after each test that uses the rule, we cannot add this listener using Mockito.framework(). The latter is a listener for events triggered during "mocking" and is intended to be added by Mockito.framework()

Hope that helps! I'm happy to discuss better implementation. Please suggest!

src/main/java/org/mockito/internal/junit/JUnitRule.java
szczepiq: Great questions and feedback.

silent() and strictStubs() are different: the first silences all warnings whatever they are (currently , the only warnings are related to stubbing but it will change in the future). The latter specifically makes the stubbing stricter. It does not do Mockito stricter, for example, it does not make the mock objects "strict".

Thoughts/ideas/suggestions for improvements?

src/main/java/org/mockito/internal/junit/DefaultTestFinishedEvent.java
szczepiq: Nice catch! Actually, we use both patters quite eagerly :)

szczepiq:
@TimvdLippe, no worries, I can wait. Plus, I haven't finished yet.

szczepiq:
Pushed in few refactorings, mostly using the new SafeJUnitRule test utility (btw. I love it :D).

I plan to work on the docs and public api exposure in the next round. The code is reviewable now.

szczepiq:
Ok, documentation ready and public api exposed :D

Thank you guys for following this PR and guiding the implementation via code review feedback!!!

src/test/java/org/mockitousage/junitrule/StrictJUnitRuleTest.java
bric3: Actually I find try/catch has a better readability than `ExpectedException`, AssertJ and Java 8 will help in this.

src/main/java/org/mockito/junit/MockitoRule.java
```diff
@@ -76,4 +78,24 @@
...
+     * </ol>
+     * We are very eager to hear feedback about "strict stubbing" feature, let us know by commenting on GitHub
+     * <a href="https://github.com/mockito/mockito/issues/769">issue 769</a>.
+     * Strict stubbing is an attempt to improve testability and productivity with Mocktio. Tell us what you think!
+     * <p>
+     * This feature is tentatively planned to be turned on by default in Mockito v3.
+     *
+     * @since 2.3.0
+     */
+    MockitoRule strictStubs();
```
bric3: It came to me earlier, but I have doubts on the word _strict_. When I first opened this PR, I thought this was about easymock style stubs. Then I realised it was not.

_Strict stubs_ may refer / apply to different things regarding stubs, I think a different name should be used. I'm halfway re-reviewing it (it's long) and my understanding of this change is that it is much more about potential stubbing problem detection than about strict mocks.

Also I'd add `@Incubating` there. 

Finally I'm thinking about rule builders, and I'd like to avoid a proliferation of this kind of API right now. Typically what I have in mind is rule builder where I can

* alter the mockito configuration (i.e. plugins override)
* alter the behaviour of mock (e.g. the easymock style)
* alter the logger
* etc.

src/test/java/org/mockitousage/junitrule/StubbingWarningsJUnitRuleTest.java
```diff
@@ -1,242 +1,191 @@
...
+        });
+
+        //when
+        declareStubbingWithArg(mock, "a"); // <-- used
+        declareStubbingWithArg(mock, "b"); // <-- unused
+
+        useStubbingWithArg(mock, "a");
+        useStubbingWithArg(mock, "d"); // <-- arg mismatch
+
+        throw new AssertionError("x");
```
bric3: Could be a method in `SafeJUnitRule` ? like `triggerFailure()`, this could avoid to test ourselves `assertEquals("x", t.getMessage());`, `SafeJUnitRule` would do that before running the custom `FailureAssert`.

And this would make explicit that the assertion error is related to the rule.

src/main/java/org/mockito/internal/junit/DefaultTestFinishedEvent.java
bric3: On that I don't mind much as it is internal, not a fan of `Impl`, but I can live with it.

src/main/java/org/mockito/junit/MockitoRule.java
szczepiq: Thanks for feedback! Please suggest a different name and/or API for rule.

I named it 'strictStubs' because 1) it is about strictness 2) it is about stubbing. The behavior is very similar to classic strict mocks like in EasyMock/JMock where unconfigured method on a mock fails immediately and unsatisfied expectations (in Mockito those are stubbings) fail at the end of test.

Using incubating tag is a good idea.

src/test/java/org/mockitousage/junitrule/StubbingWarningsJUnitRuleTest.java
szczepiq: Could be a good idea. I like the current assertion for simplicity. I admit it is not completely DRY.

src/main/java/org/mockito/internal/junit/DefaultTestFinishedEvent.java
```diff
@@ -0,0 +1,28 @@
+package org.mockito.internal.junit;
+
+/**
+ * Created by sfaber on 11/24/16.
+ */
```
szczepiq: Ah, will remove this boilerplate! Good catch :)

src/main/java/org/mockito/internal/framework/DefaultMockitoFramework.java
szczepiq: Since this raises concern from multiple contributors, I'll change/update the code. Thanks guys!

src/main/java/org/mockito/internal/exceptions/Reporter.java
```diff
@@ -835,11 +835,45 @@ public static UnnecessaryStubbingException formatUnncessaryStubbingException(Cla
...
+                "  " + actualInvocation.getLocation(),
+                "Has following stubbing(s) with different arguments:",
+                stubbings,
+                "Typically, stubbing argument mismatch indicates user mistake when writing tests.",
+                "In order to streamline debugging tests Mockito fails early in this scenario.",
+                "However, there are legit scenarios when this exception generates false negative signal:",
+                "  - stubbing the same method multiple times using 'given' or 'when' syntax",
+                "    Please use willReturn/doReturn API for stubbing",
+                "  - stubbed method is intentionally invoked with different arguments by code under test",
+                "    Please use 'default' or 'silent' JUnit Rule.",
```
szczepiq: I agree.

1. Why would they remove the rule instead of silencing the behavior?
2. I plan to add to the API a way to declare "optional" stubbings (not part of this RB)

src/main/java/org/mockito/internal/exceptions/Reporter.java
```diff
@@ -835,11 +835,45 @@ public static UnnecessaryStubbingException formatUnncessaryStubbingException(Cla
...
+                "This invocation of '" + actualInvocation.getMethod().getName() + "' method:",
+                "  " + actualInvocation,
+                "  " + actualInvocation.getLocation(),
+                "Has following stubbing(s) with different arguments:",
+                stubbings,
+                "Typically, stubbing argument mismatch indicates user mistake when writing tests.",
+                "In order to streamline debugging tests Mockito fails early in this scenario.",
+                "However, there are legit scenarios when this exception generates false negative signal:",
+                "  - stubbing the same method multiple times using 'given' or 'when' syntax",
+                "    Please use willReturn/doReturn API for stubbing",
```
szczepiq: ok, nice feedback!

src/main/java/org/mockito/internal/creation/settings/CreationSettings.java
szczepiq: Thanks for feedback Brice! I like it. Created #793

src/main/java/org/mockito/internal/framework/DefaultMockitoFramework.java
szczepiq: I removed tryAddListener for clarity.

src/main/java/org/mockito/internal/exceptions/Reporter.java
szczepiq: Done. Nice feedback!

src/main/java/org/mockito/internal/exceptions/Reporter.java
szczepiq: Interesting feedback. Can you suggest a solution?

Why do you think users will not use the rule at all? You think they will prefer to remove the rule from the test class and replace it by hand-crafted setup method with initMocks, etc.?

src/main/java/org/mockito/internal/junit/DefaultTestFinishedEvent.java
szczepiq: Done :)

src/main/java/org/mockito/internal/junit/StrictStubsTestListener.java
```diff
@@ -0,0 +1,57 @@
...
+        //MockCreationSettings is intended to be an immutable view of the creation settings
+        //In future, we should start passing MockSettings object to the creation listener
+        settings.getStubbingLookUpListeners().add(new StubbingLookUpListener() {
+            public void onStubbingLookUp(Invocation invocation, MatchableInvocation stubbingFound) {
+                if (stubbingFound == null) {
+                    List<Invocation> matchingStubbings = new LinkedList<Invocation>();
+                    Collection<Stubbing> stubbings = mockingDetails(invocation.getMock()).getStubbings();
+                    for (Stubbing s : stubbings) {
+                        if (!s.wasUsed() && s.getInvocation().getMethod().getName().equals(invocation.getMethod().getName())) {
+                            matchingStubbings.add(s.getInvocation());
```
szczepiq: I agree with your concern. It is a subjective decision to make the match only based on method name (and same mock object - it's implied here). It is consistent with how we currently report stubbing are mismatches as warnings.

We could change it per user feedback.

src/main/java/org/mockito/internal/junit/StrictStubsTestListener.java
```diff
@@ -0,0 +1,57 @@
...
+                        }
+                    }
+                    if (!matchingStubbings.isEmpty()) {
+                        Reporter.potentialStubbingProblemByJUnitRule(invocation, matchingStubbings);
+                    }
+                } else {
+                    //when strict stubs are in use, every time a stub is realized in the code it is implicitly marked as verified
+                    //this way, the users don't have to repeat themselves to verify stubbed invocations (DRY)
+                    invocation.markVerified();
+                }
```
szczepiq: Good feedback. Updated the code for readability. Thanks!

szczepiq:
Updated the code based on latest code review feedback. TODOs:
- review public API changes in the mockito PR
- add blurp to the main mockito class

szczepiq:
Nice feedback @ChristianSchwarz , thank you!

>silent() is renamed to silentStubs(). What do you think?

I'd still stay with silent() and strictStubs(). However, I'm open to discussion about the best API. Other ideas:

1. MockitoJUnit.rule(Strictness.STRICT_STUBS) // <-- easy to discover
2. Annotation:

```java
@Mockito(strictStubs=true) // <-- would work with both runner and the rule
public class SomeTest {
  @Rule public MockitoRule rule = MockitoJUnit.rule(); 
}
```

szczepiq:
The feature is complete. I don't plan to make changes except of some public API changes that are currently discussed. Please review!

src/main/java/org/mockito/internal/exceptions/Reporter.java
bric3: e.g. with pseudo code

```
class TheTest {

MockitoRule.strictStubs() // current naming

  test1 {
    // works ok with strict stubs
  }
  test2 {
    // works ok with strict stubs
  }
  test3 {
    // incompatible with strict stubs 
  }
}
```

as test 3 is incompatible the solution is either to move it to another class, or to _silence_ the rule. In the later case, once silenced maintainance or evolution tasks will not benefit from the strict stubs, because the rule will be silent (or something else).

While the interesting part of quality is in maintaining the codebase. My idea would be to have the ability to deactivate by test, e.g.

```
  test3 {
    rule.deactivate()
    // incompatible with strict stubs 
  }
```

src/main/java/org/mockito/internal/junit/StrictStubsTestListener.java
bric3: Yeah no problem really, I just wanted to check if we're ok with algorithm, maybe we should just name it, like `PassbleMismatchMethodComparator`.

src/main/java/org/mockito/junit/MockitoRule.java
bric3: > Please suggest a different name and/or API for rule.

A proper name still eludes me.

> The behavior is very similar to classic strict mocks like in EasyMock/JMock where unconfigured method on a mock fails immediately and unsatisfied expectations (in Mockito those are stubbings) fail at the end of test.

OK after a few days, that sinks in ;)

bric3:
I really prefer the `Strictness` enum to enable strict mocks.

1. > MockitoJUnit.rule(Strictness.STRICT_STUBS) // <-- easy to discover

    Although I'd prefer something like :

    ```java
    MockitoJUnit.rule().strictness(Strictness.STRICT_STUBS)
    ```

    ```java
    MockitoJUnit.rule().strictness(Strictness.STRICT_STUBS)
                       .logger(my_logger)
    ```

    This configuration design is discoverable and evolutive.

2. > Annotation `@Mockito`

    Annotation is nice, though I'd favor a name with matching the feature / option like `@Srictness`. From experience I'd like to avoid annotation overload, so I'd prefer a programatic way of changing some option like strictness

src/main/java/org/mockito/Mockito.java
```diff
@@ -1279,6 +1280,31 @@
...
+ * This is also different from the classic mocking frameworks of the past, they were "strict" by default.
+ * <p>
+ * Being "loose" by default makes Mockito tests harder to debug at times.
+ * There are scenarios where misconfigured stubbing (like using a wrong argument) forces the user to run the test with a debugger.
+ * Ideally, tests failures are immediately obvious and don't require debugger to identify the root cause.
+ * Starting with version 2.1 Mockito has been getting new features that nudge the framework towards "strictness".
+ * We want Mockito to offer fantastic debuggability while not losing its core mocking style, optimized for
+ * intuitiveness, explicitness and clean test code.
+ * <p>
+ * Help Mockito! Try the new features, give us feedback, join the discussion about Mockito strictness at GitHub <a href="">issue XXX</a>.
```
bric3: empty link

src/main/java/org/mockito/internal/exceptions/Reporter.java
```diff
@@ -851,10 +851,11 @@ public static void unncessaryStubbingException(List<Invocation> unused) {
         throw formatUnncessaryStubbingException(null, unused);
     }
 
-    public static PotentialStubbingProblem potentialStubbingProblemByJUnitRule(Invocation actualInvocation, Collection<Invocation> matchingStubbings) {
+    public static PotentialStubbingProblem potentialStubbingProblemByJUnitRule(
+            Invocation actualInvocation, Collection<Invocation> argMismatchStubbings) {
         StringBuilder stubbings = new StringBuilder();
         int count = 1;
-        for (Invocation s : matchingStubbings) {
+        for (Invocation s : argMismatchStubbings) {
```
bric3: 👍

src/main/java/org/mockito/internal/junit/StrictStubsTestListener.java
```diff
@@ -54,4 +50,15 @@ public void onStubbingLookUp(Invocation invocation, MatchableInvocation stubbing
...
+    private static List<Invocation> potentialArgMismatches(Invocation invocation) {
+        List<Invocation> matchingStubbings = new LinkedList<Invocation>();
+        Collection<Stubbing> stubbings = mockingDetails(invocation.getMock()).getStubbings();
+        for (Stubbing s : stubbings) {
+            if (!s.wasUsed() && s.getInvocation().getMethod().getName().equals(invocation.getMethod().getName())) {
+                matchingStubbings.add(s.getInvocation());
+            }
+        }
+        return matchingStubbings;
+    }
```
bric3: 👍

src/main/java/org/mockito/Mockito.java
szczepiq: Ha! Thanks :)

szczepiq:
Very nice feedback @bric3!

>Since the API is incubating, it can be removed safely (users have been warned) even in minor version to follow ideas expressed in above comments.

Absolutely! Despite that, I want us to make well thought out API decision and bring to our users an opinionated feature. Hence, let's keep the discussion going for a little bit, before we merge.

>I really prefer the Strictness enum to enable strict mocks

I really like your suggestion with an enum!

>From experience I'd like to avoid annotation overload, so I'd prefer a programatic way of changing some option like strictness

I really like this idea!

I'm with you about the annotations. One value I see in annotations is that we would have _exactly_ the same API for runner and the rule. They both can use the annotation.

I'll design the API based on @bric3 feedback and present it here for more feedback. In the meantime, tell us what you think!

szczepiq:
Based on feedback, I came up with this:

```java
//Existing API:
@Rule public MockitoRule mockito = MockitoJUnit.rule()

//Proposed improvements:
rule().strictness(Strictness.LENIENT)
rule().strictness(Strictness.HINTING)       // <-- is the default
rule().strictness(Strictness.STRICT_STUBS)  // <-- Mockito 3.0 default
```
Potential future changes (based on feedback / necessity):
```java
//Further work on "strictness":
rule().strictness(Strictness.STRICT_MOCKS)  // <-- different than strict stubs
rule().strictness(Strictness.STRICT_MOCKS_HINTING)

//Exposing the listener API used by JUnit rules:
rule().listener(MockitoTestListener)

//Annotation API that also integrates seamlessly with MockitoJUnitRunner
@Mockito(strictness = Strictness.STRICT_STUBS)
public class SomeTest {

	@Test public void shouldWork() {
		//...	
	}

	@Mockito(strictness = Strictness.LENIENT)
	@Test public void shouldWork() {
		//...	
	}
}
```

szczepiq:
This feature is ready to merge. Please make a final pass!

szczepiq:
@TimvdLippe, I'm about to merge it. Do you still want to review? Any feedback about the new public API.

@ChristianSchwarz, feedback?

src/main/java/org/mockito/internal/listeners/StubbingLookupListener.java
```diff
@@ -0,0 +1,31 @@
+package org.mockito.internal.listeners;
+
+import org.mockito.invocation.Invocation;
+import org.mockito.invocation.MatchableInvocation;
+
+/**
+ * Listens to attempts to look up stubbing answer for given mocks. This class is private for now.
```
TimvdLippe: Nit: `private` -> `internal`

src/main/java/org/mockito/internal/listeners/StubbingLookupListener.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+public interface StubbingLookupListener {
+
+    /**
+     * Called by the framework when Mockito looked up an answer for invocation on a mock.
+     *
+     * @param invocation the invocation on the mock
+     * @param stubbingFound - can be null - it indicates that the invocation was not stubbed.
+     */
+    void onStubbingLookup(Invocation invocation, MatchableInvocation stubbingFound);
```
TimvdLippe: Agreed an event object makes this more maintainable/extendible for the future

src/test/java/org/mockitoutil/SafeJUnitRule.java
```diff
@@ -0,0 +1,111 @@
...
+import org.assertj.core.api.Assertions;
+import org.junit.rules.MethodRule;
+import org.junit.runners.model.FrameworkMethod;
+import org.junit.runners.model.Statement;
+
+/**
+ * Junit rule for testing exception handling other JUnit rules, like Mockito JUnit rules.
+ * Makes it easy to assert on expected exceptions triggered by the rule under test.
+ *
+ * TODO update other tests that validate rule error handling so that they use this class.
```
TimvdLippe: Might want to fix this TODO before merging?

src/test/java/org/mockitoutil/SafeJUnitRule.java
```diff
@@ -0,0 +1,111 @@
...
+        this.testedRule = testedRule;
+    }
+
+    public Statement apply(final Statement base, final FrameworkMethod method, final Object target) {
+        return new Statement() {
+            public void evaluate() throws Throwable {
+                try {
+                    testedRule.apply(base, method, target).evaluate();
+                    successAssert.run();
+                } catch (Throwable t) {
```
TimvdLippe: I think `ExpectedException` can make this code a lot more readable for other developers. You can probably also take advantage of AssertJ with `expectException`

src/main/java/org/mockito/internal/junit/JUnitRule.java
```diff
@@ -8,87 +8,75 @@
...
      * @param logger target for the stubbing warnings
-     * @param silent whether the rule
+     * @param strictness how strict mocking / stubbing is concerned
      */
-    public JUnitRule(MockitoLogger logger, boolean silent) {
-		this.logger = logger;
-        this.silent = silent;
+    public JUnitRule(MockitoLogger logger, Strictness strictness) {
+        this.logger = logger;
+        switch (strictness) {
```
TimvdLippe: Implementation detail: I would prefer to remove the `switch` and instead do `this.listener = strictness.obtainListener()`. Therefore if we add a new strictness value, the compiler will complain :smile:

szczepiq: Interesting idea. I like compilers helping out :)

This however would make private API leak out to the clients (Strictness is public).

src/test/java/org/mockitoutil/SafeJUnitRule.java
szczepiq: Interesting feedback! You mean that the SafeJUnitRule would be more readable? SafeJUnitRule is not a test.

src/test/java/org/mockitoutil/SafeJUnitRule.java
szczepiq: This TODO is cleared, I'll remove it. Nice catch! :)

szczepiq:
Thanks @TimvdLippe !!! I assume that you're ok with public API changes, e.g. new rule().strict() method. I made subjective tweaks based on your feedback and will merge the change today!

src/main/java/org/mockito/internal/junit/JUnitRule.java
bric3: Since strictness is now public I may not be a good idea anymore, though I liked it.

src/main/java/org/mockito/internal/listeners/StubbingLookupListener.java
szczepiq: This is internal class, we can create an event at any time.

szczepiq: I added a TODO for it to execute should we choose to make this public.

bric3: I agree with @ChristianSchwarz here, though it's an internal listener so that's les of a problem.

TimvdLippe:
@szczepiq Yes I am okay with the API. I do think the Event Object should be fixed, other than that LGTM

src/main/java/org/mockito/internal/listeners/StubbingLookupListener.java
szczepiq: FYI - I absolutely like the idea of event objects. They are tremendously helpful when we need to grow/change the listeners. Thanks for pushing for this!!!

szczepiq:
>@szczepiq Yes I am okay with the API. I do think the Event Object should be fixed, other than that LGTM

Thank you! FYI - Since the listener is internal, I don't plan to introduce an event at this time.

szczepiq:
Was about to merge it but did a final pass over the changed API and found out that I'm leaking private API via MockCreationSettings :) Will fix this shortly.

version.properties
```diff
@@ -1,2 +1,2 @@
-version=2.2.28
-mockito.testng.version=1.0
+version=2.2.3
```
TimvdLippe: Should be `2.3.0`

szczepiq:
I made a bad rebase and polluted commit history of this PR. Opened a separate PR with the same code: #807

version.properties
szczepiq: Thank you! Updated.

szczepiq:
I cannot reopen PR after it has been force-pushed. Learning Git/GitHub every day :)

PR #771, opened 2016-11-20
Fixes #731 Implements retry rule for flaky tests

### LINKED ISSUE #731: Consider rerunning failed tests
Consider rerunning failed tests to avoid https://travis-ci.org/mockito/mockito/jobs/172673452

Some options:
 - rerun test Gradle task
 - rerun test rule http://stackoverflow.com/questions/8295100/how-to-re-run-failed-junit-tests-immediately

bric3:
Previous attempts at fixing the time related tests failed. Indeed the previous attempts are not bullet proof when the CI can run schedule some threads slower than _expected_.

A proper harness should be implemented, but in the meantime, I implemented the retry rule as proposed in #731 to gain back stability on the build.

### DIFF
--- a/src/test/java/org/mockitoutil/RetryRule.java
+++ b/src/test/java/org/mockitoutil/RetryRule.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2007 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+
+package org.mockitoutil;
+
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+public class RetryRule implements TestRule {
+    private final TestRule innerRule;
+
+    public static RetryRule attempts(final int attempts) {
+        return new RetryRule(new NumberedAttempts(attempts));
+    }
+
+    private RetryRule(TestRule innerRule) {
+        this.innerRule = innerRule;
+    }
+
+    public Statement apply(final Statement base, final Description description) {
+        return innerRule.apply(base, description);
+    }
+
+    private static class NumberedAttempts implements TestRule {
+        private final int attempts;
+
+        NumberedAttempts(int attempts) {
+            assert attempts > 1;
+            this.attempts = attempts;
+        }
+
+        @Override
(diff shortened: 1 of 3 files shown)


TimvdLippe:
The amount of import reordering and stylistic changes is quite high sadly. Would be nice to prevent these for new PRs

src/test/java/org/mockitoutil/RetryRule.java
```diff
@@ -0,0 +1,52 @@
...
+                public void evaluate() throws Throwable {
+                    for (int remainingAttempts = attempts; remainingAttempts > 0 ; remainingAttempts--) {
+                        try {
+                            base.evaluate();
+                        } catch (Throwable throwable) {
+                            if (remainingAttempts > 0) {
+                                throw throwable;
+                            }
+                        }
+                    }
```
szpak: For negative attempts defined by developer the code in the for loop will never be executed - no evaluation of a base statement will be performed which in turn may hide potential exceptions in a test.

bric3:
> The amount of import reordering and stylistic changes is quite high sadly. Would be nice to prevent these for new PRs

Regarding the imports I followed the new style guide. To lower this noise, we can apply the style guide on the whole project. Also I think it's possible to export/import the IntelliJ/Eclipse style config.

src/test/java/org/mockitoutil/RetryRule.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+package org.mockitoutil;
+
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+public class RetryRule implements TestRule {
+    private TestRule innerRule;
```
bric3: ok

src/test/java/org/mockitoutil/RetryRule.java
bric3: Added an `assert` check

TimvdLippe:
@bric3 Yes it is good you enforce it, it's that others don't :P Seems like a good idea to fix it once and then block PRs if they change it.

PR #775, opened 2016-11-25
Removed private isMockitoMock(..) in MockUtil

ChristianSchwarz:
Small internal refactoring.


TimvdLippe:
While it is internal, I know some other projects are using this class and the `boolean` -> `<T>` could potentially break them. Let's investigate quickly to make sure we are not creating unwanted work for these users.

ChristianSchwarz:
@TimvdLippe good catch!

ChristianSchwarz:
The signature <T> boolean isMock(T) doesn't make sense, cause <T> will be erased to Object anyway. I restored the original signature.

PR #776, opened 2016-11-25
InvocationMatcher internal improvements

ChristianSchwarz:
InvocationMatcher internal improvements
 * Added missing @Override annotations to reduce warning count.
 * Simplified parameter comparison in hasSameMethod(..)
 * Fixed compile error due to char conversion in MatchersTest


src/test/java/org/mockitousage/matchers/MatchersTest.java
```diff
@@ -596,7 +596,7 @@ public void same_matcher_and_nulls() {
     public void nullable_matcher() throws Exception {
         // imagine a Stream.of(...).map(c -> mock.oneArg(c))...
         mock.oneArg((Character) null);
-        mock.oneArg(Character.valueOf('€'));
```
bric3: Let's remove that line as it's not part of the improvements

PR #777, opened 2016-11-25
Replaced ObjectBox with AtomicReference

ChristianSchwarz:
Removed ObjectBox in favor of AtomicReference.


bric3:
Please add more description, even though the change is easy and better ;)

ChristianSchwarz:
@bric3 Done.

PR #778, opened 2016-11-25
ValuePrinter small improvements

ChristianSchwarz:
* added private construtor
 * removed unneccessary 'else' statements
 * added missing type arguments
 * replaced `new ArrayIterator(new String[]{""}) `with `emptyIterator()` in
_printValues(..)_


ChristianSchwarz:
:angry:  build break for java 1.6
  emptyIterator() -> since 1.7

szpak:
Btw, @ChristianSchwarz You can make PR with `emptyIterator()` to master.

ChristianSchwarz:
@szpak  I figured out that it is misleading touse an empty iterator it may hide a bug if a  null Iterator is passed, therefore I decided to remove the call. Now a NPE is thrown in this case.

PR #779, opened 2016-11-25
Removed argument from MockingProgress.stubbingCompleted(..)

ChristianSchwarz:
Removed unneccessary argument from internal MockingProgress.stubbingCompleted(..)


ChristianSchwarz:
bump

PR #782, opened 2016-11-25
Introduce default answers for primitive Optionals/Streams

Stephan202:
I introduced `#invokeNullaryFactoryMethod` to remove some of the existing code duplication, but am unhappy about the _new_ duplication, especially in the test code. What's your view on introducing deduplicating logic in test code? Also, the singleton fields in `JavaEightUtil` could be replaced with a `ConcurrentMap<String, Object>` (i.e. a map from FQCNs to singletons). If you agree, I'd love to reduce the size of this PR.

### DIFF
--- a/src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValues.java
+++ b/src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValues.java
@@ -37,7 +37,10 @@ import java.util.*;
  * Returns zero if references are equals otherwise non-zero for Comparable#compareTo(T other) method (see issue 184)
  * </li>
  * <li>
- * Returns an {@code java.util.Optional#empty() empty Optional} for Optional (see issue 191).
+ * Returns an {@code java.util.Optional#empty() empty Optional} for Optional. Similarly for primitive optional variants.
+ * </li>
+ * <li>
+ * Returns an {@code java.util.stream.Stream#empty() empty Stream} for Stream. Similarly for primitive stream variants.
  * </li>
  * <li>
  * Returns null for everything else
@@ -109,8 +112,20 @@ public class ReturnsEmptyValues implements Answer<Object>, Serializable {
             return new LinkedHashMap<Object, Object>();
         } else if ("java.util.Optional".equals(type.getName())) {
             return JavaEightUtil.emptyOptional();
+        } else if ("java.util.OptionalDouble".equals(type.getName())) {
+            return JavaEightUtil.emptyOptionalDouble();
+        } else if ("java.util.OptionalInt".equals(type.getName())) {
+            return JavaEightUtil.emptyOptionalInt();
+        } else if ("java.util.OptionalLong".equals(type.getName())) {
+            return JavaEightUtil.emptyOptionalLong();
         } else if ("java.util.stream.Stream".equals(type.getName())) {
             return JavaEightUtil.emptyStream();
+        } else if ("java.util.stream.DoubleStream".equals(type.getName())) {
+            return JavaEightUtil.emptyDoubleStream();
+        } else if ("java.util.stream.IntStream".equals(type.getName())) {
+            return JavaEightUtil.emptyIntStream();
+        } else if ("java.util.stream.LongStream".equals(type.getName())) {
+            return JavaEightUtil.emptyLongStream();
         }
 
         //Let's not care about the rest of collections.
--- a/src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesTest.java
+++ b/src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesTest.java
@@ -89,7 +89,7 @@ public class ReturnsEmptyValuesTest extends TestBase {
     }
 
     @Test
-    public void should_return_empty_optional() throws Exception {
+    public void should_return_empty_Optional() throws Exception {
         Class<?> streamType = getClassOrSkipTest("java.util.stream.Stream");
 
         //given
@@ -108,7 +108,64 @@ public class ReturnsEmptyValuesTest extends TestBase {
     }
 
     @Test
-    public void should_return_empty_stream() throws Exception {
+    public void should_return_empty_OptionalDouble() throws Exception {
+        Class<?> streamType = getClassOrSkipTest("java.util.stream.DoubleStream");
+
+        //given
+        Object stream = mock(streamType);
+        Object optional = streamType.getMethod("findAny").invoke(stream);
+        assertNotNull(optional);
+        assertFalse((Boolean) Class.forName("java.util.OptionalDouble").getMethod("isPresent").invoke(optional));
+
+        Invocation findAny = this.getLastInvocation();
+
+        //when
+        Object result = values.answer(findAny);
+
+        //then
+        assertEquals(optional, result);
+    }
+
+    @Test
+    public void should_return_empty_OptionalInt() throws Exception {
+        Class<?> streamType = getClassOrSkipTest("java.util.stream.IntStream");
+
+        //given
+        Object stream = mock(streamType);
+        Object optional = streamType.getMethod("findAny").invoke(stream);
+        assertNotNull(optional);
+        assertFalse((Boolean) Class.forName("java.util.OptionalInt").getMethod("isPresent").invoke(optional));
+
+        Invocation findAny = this.getLastInvocation();
+
+        //when
+        Object result = values.answer(findAny);
+
+        //then
+        assertEquals(optional, result);
+    }
+
+    @Test
+    public void should_return_empty_OptionalLong() throws Exception {
+        Class<?> streamType = getClassOrSkipTest("java.util.stream.LongStream");
+
+        //given
+        Object stream = mock(streamType);
+        Object optional = streamType.getMethod("findAny").invoke(stream);
+        assertNotNull(optional);
+        assertFalse((Boolean) Class.forName("java.util.OptionalLong").getMethod("isPresent").invoke(optional));
+
+        Invocation findAny = this.getLastInvocation();
+
+        //when
+        Object result = values.answer(findAny);
+
+        //then
+        assertEquals(optional, result);
+    }
+
+    @Test
+    public void should_return_empty_Stream() throws Exception {
         // given
         Class<?> streamType = getClassOrSkipTest("java.util.stream.Stream");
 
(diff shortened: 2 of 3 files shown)


bric3:
Thanks it looks good to me

bric3:
Regarding the test code it's ok, it'll be simplified in master as the code will target Java 8

Stephan202:
Alright, that works for me as well. Thanks for merging! :)

src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesTest.java
```diff
@@ -108,7 +108,64 @@ public void should_return_empty_optional() throws Exception {
...
+
+        //when
+        Object result = values.answer(findAny);
+
+        //then
+        assertEquals(optional, result);
+    }
+
+    @Test
+    public void should_return_empty_OptionalInt() throws Exception {
```
szpak: It just asks to simplify it with parameterized tests, e.g. with https://github.com/Pragmatists/JUnitParams/wiki/Quickstart . @Stephan202, would you dare to do it? ;)

szpak:
It just asks to simplify it with parameterized tests, e.g. with https://github.com/Pragmatists/JUnitParams/wiki/Quickstart . @Stephan202, would you dare to do it? ;)

src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesTest.java
Stephan202: @szpak: I'm fine opening an extra PR, but I think the Mockito team should first decide on which parameterized test framework to use. The one you linked is an option, but JUnit also comes with built-in support for [`@RunWith(Theories.class)`](https://github.com/junit-team/junit4/wiki/Theories) and [`@RunWith(Parameterized.class)`](https://github.com/junit-team/junit4/wiki/parameterized-tests). There may be others, and it seems right now none of these approaches are used.

Alternatively, the test duplication can be avoided as follows:
```diff
diff --git a/src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesTest.java b/src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesTest.java
index 480a9a1..ab9aad9 100644
--- a/src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesTest.java
+++ b/src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesTest.java
@@ -90,70 +90,32 @@ public class ReturnsEmptyValuesTest extends TestBase {
 
     @Test
     public void should_return_empty_Optional() throws Exception {
-        Class<?> streamType = getClassOrSkipTest("java.util.stream.Stream");
-
-        //given
-        Object stream = mock(streamType);
-        Object optional = streamType.getMethod("findAny").invoke(stream);
-        assertNotNull(optional);
-        assertFalse((Boolean) Class.forName("java.util.Optional").getMethod("isPresent").invoke(optional));
-
-        Invocation findAny = this.getLastInvocation();
-
-        //when
-        Object result = values.answer(findAny);
-
-        //then
-        assertEquals(optional, result);
+        verify_empty_Optional_is_returned("java.util.stream.Stream", "java.util.Optional");
     }
 
     @Test
     public void should_return_empty_OptionalDouble() throws Exception {
-        Class<?> streamType = getClassOrSkipTest("java.util.stream.DoubleStream");
-
-        //given
-        Object stream = mock(streamType);
-        Object optional = streamType.getMethod("findAny").invoke(stream);
-        assertNotNull(optional);
-        assertFalse((Boolean) Class.forName("java.util.OptionalDouble").getMethod("isPresent").invoke(optional));
-
-        Invocation findAny = this.getLastInvocation();
-
-        //when
-        Object result = values.answer(findAny);
-
-        //then
-        assertEquals(optional, result);
+        verify_empty_Optional_is_returned("java.util.stream.DoubleStream", "java.util.OptionalDouble");
     }
 
     @Test
     public void should_return_empty_OptionalInt() throws Exception {
-        Class<?> streamType = getClassOrSkipTest("java.util.stream.IntStream");
-
-        //given
-        Object stream = mock(streamType);
-        Object optional = streamType.getMethod("findAny").invoke(stream);
-        assertNotNull(optional);
-        assertFalse((Boolean) Class.forName("java.util.OptionalInt").getMethod("isPresent").invoke(optional));
-
-        Invocation findAny = this.getLastInvocation();
-
-        //when
-        Object result = values.answer(findAny);
-
-        //then
-        assertEquals(optional, result);
+        verify_empty_Optional_is_returned("java.util.stream.IntStream", "java.util.OptionalInt");
     }
 
     @Test
     public void should_return_empty_OptionalLong() throws Exception {
-        Class<?> streamType = getClassOrSkipTest("java.util.stream.LongStream");
+        verify_empty_Optional_is_returned("java.util.stream.LongStream", "java.util.OptionalLong");
+    }
+
+    private void verify_empty_Optional_is_returned(String streamFqcn, String optionalFqcn) throws Exception {
+        Class<?> streamType = getClassOrSkipTest(streamFqcn);
 
         //given
         Object stream = mock(streamType);
         Object optional = streamType.getMethod("findAny").invoke(stream);
         assertNotNull(optional);
-        assertFalse((Boolean) Class.forName("java.util.OptionalLong").getMethod("isPresent").invoke(optional));
+        assertFalse((Boolean) Class.forName(optionalFqcn).getMethod("isPresent").invoke(optional));
 
         Invocation findAny = this.getLastInvocation();
 
@@ -166,54 +128,34 @@ public class ReturnsEmptyValuesTest extends TestBase {
 
     @Test
     public void should_return_empty_Stream() throws Exception {
-        // given
-        Class<?> streamType = getClassOrSkipTest("java.util.stream.Stream");
-
-        // when
-        Object stream = values.returnValueFor(streamType);
-        long count = (Long) streamType.getMethod("count").invoke(stream);
-
-        // then
-        assertEquals("count of empty Stream", 0L, count);
+        verify_empty_Stream_is_returned("java.util.stream.Stream");
     }
 
     @Test
     public void should_return_empty_DoubleStream() throws Exception {
-        // given
-        Class<?> streamType = getClassOrSkipTest("java.util.stream.DoubleStream");
-
-        // when
-        Object stream = values.returnValueFor(streamType);
-        long count = (Long) streamType.getMethod("count").invoke(stream);
-
-        // then
-        assertEquals("count of empty DoubleStream", 0L, count);
+        verify_empty_Stream_is_returned("java.util.stream.DoubleStream");
     }
 
     @Test
     public void should_return_empty_IntStream() throws Exception {
-        // given
-        Class<?> streamType = getClassOrSkipTest("java.util.stream.IntStream");
-
-        // when
-        Object stream = values.returnValueFor(streamType);
-        long count = (Long) streamType.getMethod("count").invoke(stream);
-
-        // then
-        assertEquals("count of empty IntStream", 0L, count);
+        verify_empty_Stream_is_returned("java.util.stream.IntStream");
     }
 
     @Test
     public void should_return_empty_LongStream() throws Exception {
+        verify_empty_Stream_is_returned("java.util.stream.LongStream");
+    }
+
+    private void verify_empty_Stream_is_returned(String streamFqcn) throws Exception {
         // given
-        Class<?> streamType = getClassOrSkipTest("java.util.stream.LongStream");
+        Class<?> streamType = getClassOrSkipTest(streamFqcn);
 
         // when
         Object stream = values.returnValueFor(streamType);
         long count = (Long) streamType.getMethod("count").invoke(stream);
 
         // then
-        assertEquals("count of empty LongStream", 0L, count);
+        assertEquals("count of empty " + streamFqcn, 0L, count);
     }
 
     /**
```

TimvdLippe: I think that your proposed refactor is sufficient enough.

szpak: It's a poor man's parameterized tests, but just for that case it is probably enough.

Btw, I haven't used `Theories`, but `Parameterized` from JUnit is very verbose and limited. To define one parameterized test you need to sacrifice the whole class (you need to use fields and a constructor). There's no way to create more than one parameterized test in the same test class (which very often is very handy). JUnitParams is much more flexible (no to mentioned Spock brings it to the another level :) ).

src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValues.java
```diff
@@ -109,8 +112,20 @@ Object returnValueFor(Class<?> type) {
             return new LinkedHashMap<Object, Object>();
         } else if ("java.util.Optional".equals(type.getName())) {
             return JavaEightUtil.emptyOptional();
+        } else if ("java.util.OptionalDouble".equals(type.getName())) {
```
Stephan202: Well it's a behavioural change, so by definition one can write a test that passed prior but not afterwards. For this particular change that'd look something like the following:

1. One has a class `C` with a method `m` returning `OptionalDouble`.
2. One mocks `C`, producing an instance `c`.
3. One fails to explicitly mock the behaviour of `c.m`.
4. One uses `c` in some code that invokes `c.m` and relies on the result being `null`.

I think the "breakage" in this case is acceptable for the following reasons:

1. Since values of type `Optional*` should never be `null`, the failure points out a misuse of `Optional*`, and as such does the user a favour.
2. If one insists on a "nullable `Optional*`", adding a simple mock statement in step 3 above clearly documents that contract, consistent with the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).
3. More generally, if a test crucially relies on a certain method's return value, then it's good practise to document this precondition explicitly.

I feel the above arguments apply to `Stream*` as much as they do to `Optional*`, because a stream is a (proto) collection in some sense, and nullable collections are also decidedly an anti-pattern. WDYT?

szczepiq: Your change is really good, it improves Mockito behavior for Java8!

Regardless of the final decision whether to merge it or not, I find it disturbing that it wasn't discussed as a breaking change in this PR. @bric3, @TimvdLippe thoughts?

TimvdLippe: To be honest, I had not realised this. Especially because I dont think someone relies on the previous null behavior.

Nonetheless we should have discussed this, good point.

src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesTest.java
bric3: JUnitParams is fine, I have nothing against, let's go for it, test readability and maintainability is just as important as production code.

src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValues.java
bric3: Well my reasoning was fast : 

1. We forgot that in mockito 2, as a user I usually see that as a bug.
2. The change is consistent with the goal of mockito 2 aka support java 8 best way.
3. The change is breaking for users that uses primitive variant and that expect null for these stubs, which is not what those type are made for, so pretty rare.

All in a all this a reasonable change, that imo does not mean mockito should bump any version or postpone the change.

PR #783, opened 2016-11-26
Simplify `ReturnsEmptyValuesTest`

Stephan202:
As [discussed](https://github.com/mockito/mockito/pull/782#discussion_r89658266) in the comments under #782. Longer term this approach could be replaced with a proper test parametrization framework, but then at least the foundation has been laid.


szpak:
Thanks @Stephan202!

PR #786, opened 2016-11-27
Add showing stacktrace in Travis build

szpak:
To see more details in a case of build failure.


szczepiq:
Nice change, thank you.

Also, Gradle's build scan should take a care of it.

Issue #788, opened 2016-11-27
Failing InlineByteBuddyMockMakerTest with Java 9


szpak:
InlineByteBuddyMockMakerTest seems to fail with Java 9.

```
org.mockito.internal.creation.bytebuddy.InlineByteBuddyMockMakerTest > test_parameters_retention FAILED
    org.junit.ComparisonFailure: expected:<"[bar]"> but was:<"[arg0]">
        at jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(java.base@9-ea/Native Method)
        at jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(java.base@9-ea/NativeConstructorAccessorImpl.java:62)
        at jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(java.base@9-ea/DelegatingConstructorAccessorImpl.java:45)
        at org.mockito.internal.creation.bytebuddy.InlineByteBuddyMockMakerTest.test_parameters_retention(InlineByteBuddyMockMakerTest.java:246)

1671 tests completed, 1 failed, 43 skipped

:test FAILED
```

Detected during backporting 2.x changes to master. Sample [Travis build](https://travis-ci.org/mockito/mockito/jobs/179234063#L309-L314).

Reproducible locally with Java 8 129 and 149 and BB 1.5.5 and 1.5.7 in a `tech/backport-2.x-20161127` branch (`release/2.x` fails with Java 9 for other reasons).

TimvdLippe:
Hm it does not fail on my machine when I was working on JDK9 compatibility. See also https://github.com/mockito/mockito/commit/818aa4c1189a812ad61bda91f5ab90ca6065e107 and the green build

szpak:
I don't see this commit in master. Have you run it from my branch `tech/backport-2.x-20161127` or some other?

TimvdLippe:
https://github.com/mockito/mockito/commit/c2a670cf8c8c22c7fb54e7ffcce0fdc5e7264410 was the merge commit of the PR that was merged into master

raphw:
ASM 6 is not yet final so its not yet used by Byte Buddy. Once Java 9 is finalized, asm 6 gets merged what is required for parsing Java 9 class files.

TimvdLippe:
Let's move the assume from the one specific test to the `@Before` of the test class. We can fix them once ASM 6 is released.

TimvdLippe:
Since #791 has been merged, this can be closed.

PR #791, opened 2016-11-28
Test improvements

bric3:
Simple tests improvements

Basically disable some test for Java 9, and report tried attempts. And actually make the retry rule retry ;)

### DIFF
--- a/src/test/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMakerTest.java
+++ b/src/test/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMakerTest.java
@@ -1,5 +1,9 @@
 package org.mockito.internal.creation.bytebuddy;
 
+import java.util.List;
+import java.util.Observable;
+import java.util.Observer;
+import java.util.regex.Pattern;
 import net.bytebuddy.ByteBuddy;
 import net.bytebuddy.ClassFileVersion;
 import net.bytebuddy.description.modifier.Visibility;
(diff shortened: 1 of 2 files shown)


bric3:
Relates to #771

src/test/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMakerTest.java
```diff
@@ -47,6 +45,8 @@ public void should_create_mock_from_final_class() throws Exception {
 
     @Test
     public void should_create_mock_from_final_class_in_the_JDK() throws Exception {
+        assumeTrue(ClassFileVersion.ofThisVm().isLessThan(ClassFileVersion.JAVA_V9)); // Change when ByteBuddy has ASM6
```
TimvdLippe: This will be a conflict with current master: https://github.com/mockito/mockito/blob/master/src/test/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMakerTest.java#L42 Let's choose one way or another. I am fine with either way, no specific preference.

bric3: Yes I know, but the parameter name retention test already use the BB API. I think BB is nicer.

TimvdLippe: Yes agreed. Was just a comment that when backporting we need to fix the conflict.

src/test/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMakerTest.java
```diff
@@ -225,6 +225,7 @@ public void is_type_mockable_give_allow_final_mockable_from_JDK() {
     @Test
     public void test_parameters_retention() throws Exception {
         assumeTrue(ClassFileVersion.ofThisVm().isAtLeast(ClassFileVersion.JAVA_V8));
+        assumeTrue(ClassFileVersion.ofThisVm().isLessThan(ClassFileVersion.JAVA_V9)); // Change when ByteBuddy has ASM6
```
szpak: I see an additional change in the logic here. At least Java 8 vs less than Java 9 (so also Java 6 & 7). Is it an intentional change?

bric3: Yes intentional, alternate way could be to write : 

```java
assumeTrue(ClassFileVersion.ofThisVm().isAtLeast(ClassFileVersion.JAVA_V8) 
        && ClassFileVersion.ofThisVm().isLessThan(ClassFileVersion.JAVA_V9));
```

> Could you also by the way mention #788 in a comment (twice).

Yes I can, but I'm not sure what do you mean ? (in the commit comment or gh comment)

bric3: ok will do

bric3:
ping @szpak

PR #797, opened 2016-12-01
Introduce EventBus

ChristianSchwarz:
**DO NOT MERGE YET!**

This is a prototype of an EventBus implementation that can be used by Mockito in the future to get rid of listener interfaces. The implementation follows the well known Guava EventBus.

Listener methods must have a method with a sole argument and must be marked with `@Subscribe`. It can be used like this

``` java
EventBus b = new EventBus();

b.register(new Object(){
    @Subscribe 
    public void log(MockitoEvent e){
        System.out.println(e);
    };
});

b.post(mockitoEvent);
```

There is much work to do, but please feel free to comment.

Todo's:
 - [ ] a healty API discussion
 - [ ] JavaDocs
 - [ ] provide transparent compability-mode for the currently known Mockito listener types
 - [ ] use this EventBus where ever possible
 - [ ] check Lambda compability
 - [ ] increase coverage
 - [ ] make EventBus Thread safe

### DIFF
--- a/src/main/java/org/mockito/MockitoFramework.java
+++ b/src/main/java/org/mockito/MockitoFramework.java
@@ -42,7 +42,7 @@ public interface MockitoFramework {
      * @since 2.1.0
      */
     @Incubating
-    MockitoFramework addListener(MockitoListener listener);
+    MockitoFramework addListener(Object listener);
 
     /**
      * When you add listener using {@link #addListener(MockitoListener)} make sure to remove it.
@@ -58,5 +58,5 @@ public interface MockitoFramework {
      * @since 2.1.0
      */
     @Incubating
-    MockitoFramework removeListener(MockitoListener listener);
+    MockitoFramework removeListener(Object listener);
 }
--- a/src/main/java/org/mockito/internal/progress/MockingProgress.java
+++ b/src/main/java/org/mockito/internal/progress/MockingProgress.java
@@ -5,7 +5,10 @@
 
 package org.mockito.internal.progress;
 
+import org.mockito.listeners.MockCreatedReport;
+import org.mockito.listeners.MockCreationListener;
 import org.mockito.listeners.MockitoListener;
+import org.mockito.listeners.Subscribe;
 import org.mockito.mock.MockCreationSettings;
 import org.mockito.stubbing.OngoingStubbing;
 import org.mockito.verification.VerificationMode;
@@ -39,9 +42,30 @@ public interface MockingProgress {
     
     void mockingStarted(Object mock, MockCreationSettings settings);
 
-    void addListener(MockitoListener listener);
+    /**
+     * Registers the given Listener that will be notified when a mock object was created, see {@link #mockingStarted(Object, MockCreationSettings)}. 
+     * The listener must provide a void method annotated with {@link Subscribe} and a sole 
+     * {@link MockCreatedReport} parameter or implement the interface {@link MockCreationListener} . 
+     * Here is an example:
+     *  <pre>
+     *  Object listener =  new Object(){
+     *      public void logCreatedMocks(MockCreatedReport report){
+     *         logger.log(report);
+     *      }
+     *  };
+     *  </pre>
+     *  <p>
+     *  Note: Listeners implementing the interface {@link MockCreationListener} will be notified via {@link MockCreationListener#onMockCreated(Object, MockCreationSettings)}.  
+     * 
+     * @param listener must not be <code>null</code>
+     */
+    void addListener(Object listener);
 
-    void removeListener(MockitoListener listener);
+    /**
+     * Unregisters the given listener. It will not be notified about mock creations anymore.
+     * @param listener
+     */
+    void removeListener(Object listener);
 
     void setVerificationStrategy(VerificationStrategy strategy);
 
--- a/src/main/java/org/mockito/internal/util/eventbus/EventBus.java
+++ b/src/main/java/org/mockito/internal/util/eventbus/EventBus.java
@@ -1,4 +1,4 @@
-package org.mockito.internal.eventbus;
+package org.mockito.internal.util.eventbus;
 
 import static org.mockito.internal.util.Checks.checkNotNull;
 
--- a/src/main/java/org/mockito/internal/util/eventbus/EventBus.java
+++ b/src/main/java/org/mockito/internal/util/eventbus/EventBus.java
@@ -1,10 +1,14 @@
 package org.mockito.internal.util.eventbus;
 
 import static org.mockito.internal.util.Checks.checkNotNull;
+import static org.mockito.internal.util.eventbus.compability.ListenerSupport.legacyListenerOf;
 
 import java.lang.reflect.Method;
+import java.util.Collection;
 import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.Set;
+import org.mockito.listeners.Subscribe;
 
 public class EventBus {
 
@@ -14,7 +18,10 @@ public class EventBus {
         checkNotNull(listener, "The listener must not be null!");
 
         eventHandlers.addAll(eventHandlerOf(listener));
+        eventHandlers.addAll( legacyListenerOf(listener));
     }
+
+    
 
     public void post(Object event) {
         checkNotNull(event, "The event must not be null!");
@@ -22,6 +29,17 @@ public class EventBus {
         for (Subscriber eventHandler : eventHandlers) {
             eventHandler.tryHandleEvent(event);
         }
+    }
+    
+    public void unregister(Object listener) {
+        Collection<Subscriber> toBeRemoved= new LinkedList<Subscriber>();
+        for (Subscriber subscriber : eventHandlers) {
+            if (subscriber.getListener()==listener){
+                toBeRemoved.add(subscriber);
+            }
+        }
+        
+        eventHandlers.removeAll(toBeRemoved);
     }
 
     private static Set<Subscriber> eventHandlerOf(Object listener) {
@@ -34,6 +52,7 @@ public class EventBus {
             }
         }
 
+        
         return result;
     }
 
@@ -55,4 +74,9 @@ public class EventBus {
 
         return true;
     }
+
+    
+
+
+    
 }
--- a/src/main/java/org/mockito/internal/util/eventbus/Subscriber.java
+++ b/src/main/java/org/mockito/internal/util/eventbus/Subscriber.java
@@ -6,30 +6,31 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import org.mockito.exceptions.base.MockitoException;
 
-final class Subscriber {
+public class Subscriber {
 
     private final Object listenerInstance;
     private final Method listenerMethod;
     private final Class<?> eventType;
 
-    Subscriber(Object listenerInstance, Method listenerMethod) {
+    public Subscriber(Object listenerInstance, Method listenerMethod) {
         this.listenerInstance = listenerInstance;
         this.listenerMethod = listenerMethod;
         this.eventType = listenerMethod.getParameterTypes()[0];
     }
 
-    void tryHandleEvent(Object event) {
-        if (!canHandle(event)) {
-            return;
+    public final void tryHandleEvent(Object event) {
+        if (canHandle(event)) {
+            invokeListenerMethod(event);
         }
+    }
 
-        boolean isAccessible = listenerMethod.isAccessible();
+    protected void invokeListenerMethod(Object event) {
+        Boolean initialState = listenerMethod.isAccessible();
         try {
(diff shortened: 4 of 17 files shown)


szczepiq:
I am very excited about this idea!

Thanks for adding the code sample in the description - it really helps understanding the idea.

Without looking at the code, my immediate feedback is that annotation-driven APIs are harder to work with (discoverability, compile-time error reporting, implementation complexity).

TimvdLippe:
> The implementation follows the well known Guava EventBus.

Instead of reimplementing, could we take use/provide compatibility for their EventBus? Reduces maintenance cost for us :smile:

ChristianSchwarz:
> Instead of reimplementing, could we take use/provide compatibility for their EventBus? 
I inverstigated to "copy" the guava implementation. The guava solution is quite heavy (10 types) and has also dependencies to other guava classes that must be copied to or reimplemented too(MultiMap, Precondition, MoreExecutors, Objects, ...). Also the tests and the test tools have to be copied. An other point is the license, guava is released under Apache 2 and I have no clue if code can be transferred to an MIT project.

We can make the mockto event bus compatible with guava's @Subscribe annotation, but I don't know if it make much sense.

ChristianSchwarz:
Added backward compatible support to register for mock creation events:
 * in MockitoFramework I changed `addListener(MockitoListener)` to  `addListener(Object)`  it supports to register `MockCreationListener` as before or listeners that provide one or more subscriber methods. 
 * added a new event type `MockCreationReport` that must be used as sole parameter type in listener methods using the `@Subscribe` annotation

The _"old"_ way still works:

```java
Mockito.framework().addListener(new MockCreationListener(){
    public void onMockCreated(Object mock, MockCreationSettings settings){
        //jibbi it still works
    };
});
```

and the annotation based implementation (the listener method can have any visibility incl. private):

```java
Mockito.framework().addListener(new MockCreationListener(){
    @Subscribe void nameTheMethodAsYouLike(MockCreationReport report){
        //MockCreationReport includes the mock and the creation settings
    };
});
```

src/main/java/org/mockito/internal/progress/MockingProgress.java
```diff
@@ -39,9 +42,30 @@
...
 
-    void addListener(MockitoListener listener);
+    /**
+     * Registers the given Listener that will be notified when a mock object was created, see {@link #mockingStarted(Object, MockCreationSettings)}. 
+     * The listener must provide a void method annotated with {@link Subscribe} and a sole 
+     * {@link MockCreatedReport} parameter or implement the interface {@link MockCreationListener} . 
+     * Here is an example:
+     *  <pre>
+     *  Object listener =  new Object(){
+     *      public void logCreatedMocks(MockCreatedReport report){
```
TimvdLippe: The example is not annotated with `@Subscribe` :wink:

src/main/java/org/mockito/internal/progress/MockingProgress.java
```diff
@@ -39,9 +42,30 @@
...
+     *  <p>
+     *  Note: Listeners implementing the interface {@link MockCreationListener} will be notified via {@link MockCreationListener#onMockCreated(Object, MockCreationSettings)}.  
+     * 
+     * @param listener must not be <code>null</code>
+     */
+    void addListener(Object listener);
 
-    void removeListener(MockitoListener listener);
+    /**
+     * Unregisters the given listener. It will not be notified about mock creations anymore.
```
TimvdLippe: Mock creations is too specific, we plan to add more right?

src/main/java/org/mockito/internal/util/eventbus/EventBus.java
```diff
@@ -0,0 +1,82 @@
...
+    
+    public void unregister(Object listener) {
+        Collection<Subscriber> toBeRemoved= new LinkedList<Subscriber>();
+        for (Subscriber subscriber : eventHandlers) {
+            if (subscriber.getListener()==listener){
+                toBeRemoved.add(subscriber);
+            }
+        }
+        
+        eventHandlers.removeAll(toBeRemoved);
```
TimvdLippe: Since `eventHandlers` is a `Set`, there can only be one?

src/main/java/org/mockito/internal/util/eventbus/EventBus.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public void post(Object event) {
+        checkNotNull(event, "The event must not be null!");
+
+        for (Subscriber eventHandler : eventHandlers) {
+            eventHandler.tryHandleEvent(event);
+        }
+    }
+    
+    public void unregister(Object listener) {
```
TimvdLippe: I think this whole method can be replaced with `this.eventHandlers.remove(listener)`

src/main/java/org/mockito/internal/util/eventbus/EventBus.java
```diff
@@ -0,0 +1,82 @@
...
+
+        Class<?> eventType = parameterTypes[0];
+
+        if (eventType.isPrimitive()) {
+            return false;
+        }
+
+        return true;
+    }
+
```
TimvdLippe: Unnecessary whitespace

src/main/java/org/mockito/internal/util/eventbus/Subscriber.java
```diff
@@ -0,0 +1,78 @@
...
+        this.eventType = listenerMethod.getParameterTypes()[0];
+    }
+
+    public final void tryHandleEvent(Object event) {
+        if (canHandle(event)) {
+            invokeListenerMethod(event);
+        }
+    }
+
+    protected void invokeListenerMethod(Object event) {
```
TimvdLippe: This method feels very hacky. Especially since you are using reflection modify the accessibility. Would prefer a cleaner and nicer solution for this.

src/main/java/org/mockito/internal/progress/MockingProgress.java
ChristianSchwarz: Yes, currently only MockCreationListener extends MockitoLister, other listener interfaces can be added in the same way.

TimvdLippe:
To not lose type safety, maybe it is better to change `Object` to `MockitoListener` again and let `MockitoListener` implement default methods (requires Java 8 obv). Then whenever we want to add a new hook, we add a default empty method in `MockitoListener`. This is type-safe and lets users listen to multiple events at the same time.

src/main/java/org/mockito/internal/util/eventbus/EventBus.java
ChristianSchwarz: Actually no, `eventHandlers` is a set of Subscriber instances, every Subscriber  represents a listener method of a listener instance. So every subscriber of the listener instance must be removed. The name `eventHandlers` is confusing.

If the mockito team decides to introduce an EventBus I will start more refactoring and polishing.

src/main/java/org/mockito/internal/util/eventbus/EventBus.java
ChristianSchwarz: see my comment above

src/main/java/org/mockito/internal/util/eventbus/Subscriber.java
ChristianSchwarz: Yes it is hacky, the error handling is also not final. In the current state the whole is a prototyp to see if it works at all. 

The idea is to allow methods and the declaring class to be non public. If the accessability is not changed we can't call public methods in private inner classes.

ChristianSchwarz:
The question for me is, why do we need type safety here? What are the risks of the user. The addListener method can be implment more strict and throw is an instance is passed that implements no listener or has no subscriber methods. WDYT?

szczepiq:
Thank you for pushing the the idea forward!

At the moment, I don't want this implementation to be merged.

I'd like to step back and
a) identify what is the goal of the API change. What is the value of "getting rid of listener interfaces"?
b) evaluate alternative APIs, that don't use annotations.

Looking at the API in its current form, I don't want this to be merged because the API is:
 - harder to discover
 - fail in runtime instead of failing at compile time (type safety)
 - require writing a lot more code to handle
 - harder to work with and maintain in practice (in my experience of using them in Gradle)
 - annotations/reflection are typically slow (perhaps negligible but

Getting rid of listener interfaces (if that's the goal) has negative value. Mockito has and will have different families of listeners / subscriptions. Encoding this information in type hierarchy makes it easy to work with (cheap documentation, compiler checks).

src/main/java/org/mockito/MockitoFramework.java
```diff
@@ -42,7 +42,7 @@
      * @since 2.1.0
      */
     @Incubating
-    MockitoFramework addListener(MockitoListener listener);
```
szczepiq: I reviewed the implementation and it seems like an overkill. What's the goal? See my comment to the entire PR.

ChristianSchwarz:
Thanks Szczepan and thanks Tim for your feedback!

About the general motivation: 
 * Annotation based event distribution allows the clients to name the listener method as they like. In traditional listener interface pattern clients are forced follow the method name of the listener interface. It is a good idea to name a method after its purpose. E.g:
```java
  @Subscribe
  private void sendVerificationFailureToBuildProcess(VerificationFailed event){
     ...
  }
``` 
* Annotation based event distribution allows to implement more than one event handling method per class. In an traditional approach you need to setup the chaining manually. 

```java
  @Subscribe
  private void sendToBuildProcess(VerificationFailed event){
     ...
  }

  @Subscribe
  private void logToConsole(VericationFailed event){
   ...
  }
``` 

 * The event bus allows to send event to private methods. The advantage here is that you don't have to write a private listener class and deleagte to a private method in order keep the visibility as low as possible. Example:

```java
  @Subscribe
  private void sendToBuildProcess(VerificationFailedEvent event){ ... }
``` 

vs:

```java
   //private or anonymous listeners are often used to implement  a listener iterface in favour of letting the top level classes implement it, cause this makes the public listener method a part of API of the top level class which is considered an anti pattern.
  private final VerificationFailedListener listener = new VerificationFailedListener(){
     public onVerificationFailed(VerificationFailedEvent event){
        sendToBuildProcess(event);
    }
  }

  private void sendToBuildProcess(VerificationFailedEvent event){ ...  }
``` 

* Events of the same family can be processed without the need to implement different listener methods and interfaces.

```java
  /* Assume  VerificationFailedEvent and VerificationSuccessEvent extend the base class VerificationEvent 
      both event types will be passed. 
  */
  @Subscribe
  private void sendToBuildProcess(VerificationEvent event){ ... }
```

ChristianSchwarz:
Benefits for mockio:
 * no need to to write sometimes redundant JavaDoc do for ListenerInterfaces and Events
 * same listener registration everywhere, e.g. currently some classes hold listeneres in a list where other hold the in a set
 * same error handling
 * Easy to discover API , we would have all events in one package and one annotation `@Subscribe` 
 * no listener interface to maintain 
 * no code duplication of listener iteration/notification/error handling

szczepiq:
@ChristianSchwarz, thanks a lot for taking the time and laying out the arguments!!!

I really like this idea and the pattern - but not for Mockito ;) It just seems like an overkill for our purposes. Also, you completely discount the cons of the proposed API (I guess following my strategy of discounting the benefits ;).

I don't feel too happy to curb your enthusiasm. I'd like to move on and stay with the current API for listeners. For now. Keep the ideas coming!!!

ChristianSchwarz:
Okay no problem! What should I do with this PR?
 * close
 * close and send again agains mockito 3
 * keep it open

bric3:
I have the same opinion as @szczepiq on this. We already struggle to maintain existing code base. Having another one without substantial benefit (or benefit for a few) seems overkill. Nevertheless I'm not entirely closed to the idea.

Imho you should target master anyway. No need to open a new PR on this, as it's possible to change the target to master (maybe reserved to admins). Ping me if you can't.

TimvdLippe:
@ChristianSchwarz If you want to, I would be very interested if the default-method interface would work out. That would be type-safe and solves the duplication of listeners. Since you now target master, we can take advantage of Java 8 :smile:

szczepiq:
Let's close. We can always reopen :)

ChristianSchwarz:
@all Does it make sense to work on this idea for mockito 3?
@bric3 Currently  I can't change the target to master, maybe because this ticket is closed.

bric3:
> @all Does it make sense to work on this idea for mockito 3?

That's not a priority, I think upgrading the codebase to Java 8 has more value.

@ChristianSchwarz I just changed it. Let's see if it's updated. 
@szczepiq Reopening temporarily

TimvdLippe:
> Does it make sense to work on this idea for mockito 3?

Agreed with @bric3 , we should reimplement core code for Java 8 first, then we can add new features on top of that.

Keep in mind a strong requirement is type-safety and up to this moment I am unconvinced you can develop an EventBus in a type-safe way

Issue #801, opened 2016-12-06
Mockito 2 fails when running on IBM J9 (SR1 FP10) VM


grimsa:
Mocking an interface fails when running on IBM's JDK with the error attached below. 
Works well on Oracle JDK 1.8.0_91.

Attaching a small project that can be used to reproduce the issue: [mockito-ibm-jdk-test.zip](https://github.com/mockito/mockito/files/633253/mockito-ibm-jdk-test.zip)

Original Mailing list post: https://groups.google.com/forum/#!topic/mockito/IWPQj2bg280

Output from `java -version` of the failing version:
```
java version "1.8.0"
Java(TM) SE Runtime Environment (build pwa6480sr1fp10-20150711_01(SR1 FP10))
IBM J9 VM (build 2.8, JRE 1.8.0 Windows 7 amd64-64 Compressed References 20150630_255633 (JIT enabled, AOT enabled)
J9VM - R28_jvm.28_20150630_1742_B255633
JIT  - tr.r14.java_20150625_95081.01
GC   - R28_jvm.28_20150630_1742_B255633_CMPRSS
J9CL - 20150630_255633)
JCL - 20150711_01 based on Oracle jdk8u51-b15
```
Error message and stack trace:
```
org.mockito.exceptions.base.MockitoException: 
Mockito cannot mock this class: interface service.Service.

Mockito can only mock non-private & non-final classes.
If you're not sure why you're getting this error, please report to the mailing list.

Java               : 1.8
JVM vendor name    : IBM Corporation
JVM vendor version : 2.8
JVM name           : IBM J9 VM
JVM version        : pwa6480sr1fp10-20150711_01 (SR1 FP10)
JVM info           : JRE 1.8.0 Windows 7 amd64-64 Compressed References 20150630_255633 (JIT enabled, AOT enabled)
J9VM - R28_jvm.28_20150630_1742_B255633
JIT  - tr.r14.java_20150625_95081.01
GC   - R28_jvm.28_20150630_1742_B255633_CMPRSS
J9CL - 20150630_255633
OS name            : Windows 7
OS version         : 6.1


Underlying exception : java.lang.IllegalArgumentException: object is not an instance of declaring class
	at org.mockito.internal.runners.SilentJUnitRunner$1.withBefores(SilentJUnitRunner.java:29)
	at org.junit.runners.BlockJUnit4ClassRunner.methodBlock(BlockJUnit4ClassRunner.java:276)
[...]

PR #803, opened 2016-12-07
Adds a warning for IBM J9 VMs if mock generation fails

bric3:
Related to #801, this PR simply adds a warning for IBM J9 VMs if the mock maker fails.


bric3:
Damn build not passing due to some certificate issue.

TimvdLippe:
Yes I have seen that error multiple times now. It seems to be related to the release tool of @szczepiq

bric3:
I believe it's the JDK6 installed on travis that is not updated with new certificates of plugins.gradle, bu I don't understand why atm, didn't investigated either.

TimvdLippe:
The sooner we purge JDK6, the better :stuck_out_tongue:

bric3:
Agreed, there's a few things to do on mockito 2 still, but I can't agree more.

PR #805, opened 2016-12-09
Never throw mockito exceptions from a plugin type's constructor

raphw:
Throwing a `MockitoException` from a mock maker constructor will cause the stack trace cleaning to be applied which itself relies on the plugin registry which is not yet initialized during the mock maker plugin's creation.

### DIFF
--- a/src/main/java/org/mockito/exceptions/base/MockitoInitializationException.java
+++ b/src/main/java/org/mockito/exceptions/base/MockitoInitializationException.java
@@ -1,0 +1,19 @@
+/*
+ * Copyright (c) 2007 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+
+package org.mockito.exceptions.base;
+
+public class MockitoInitializationException extends RuntimeException {
+
+    private static final long serialVersionUID = 1L;
+
+    public MockitoInitializationException(String message) {
+        super(message);
+    }
+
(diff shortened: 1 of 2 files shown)


TimvdLippe:
Is it possible to write a test for this too?

raphw:
Not in a way that justifies the effort. The constructors need to be default constructors, otherwise, the plugin mechanism cannot invoke them. Its only error validation and all we could test would be the avoided circularity. I do not think that there is much to gain in testing in this particular case.

src/main/java/org/mockito/exceptions/base/MockitoInitializationException.java
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright (c) 2007 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+
+package org.mockito.exceptions.base;
+
+public class MockitoInitializationException extends RuntimeException {
```
bric3: I suggest `MockitoMockMakerException` and to move it in the same mockmaker package since the exception is not that public and is specialized for constructors. Thought ?

bric3: I'm mentioning this, because as a public exception, `MockitoInitializationException` can be used by other framework and pissibly be misused.

Also this exception misses javadoc.

PR #808, opened 2016-12-11
JUnit runner reports argument stubbing mismatches

szczepiq:
Mockito v2 introduced a new feature for reporting unused stubs and reporting argument mismatches (when stubbing uses different arguments than invocation in code under test). We received feedback from our users that the behavior should be more consistent for JUnit rules and JUnit runner. This change introduces that consistency, making JUnit runner also report stubbing argument mismatches to the console. This way, the users should be able to debug tests faster and gain more productivity when writing and maintaining tests!

Also checkout new feature added in Mockito 2.3.x - [opt-in strictness](https://static.javadoc.io/org.mockito/mockito-core/2.3.10/org/mockito/junit/MockitoRule.html#strictness(org.mockito.quality.Strictness)) for Mockito JUnit rules, making the tests even cleaner and easier to debug!


szczepiq:
Friendly ping

szczepiq:
Friendly ping no. 2 ;) If no other feedback I'll merge the PR.

src/main/java/org/mockito/internal/runners/SilentJUnitRunner.java
```diff
@@ -7,35 +7,69 @@
...
-            protected Statement withBefores(FrameworkMethod method, Object target,
-                    Statement statement) {
+
+            public Object target;
+            private MockitoTestListener mockitoTestListener;
+
+            protected Statement withBefores(FrameworkMethod method, Object target, Statement statement) {
+                // get new test listener and add add it to the framework
+                mockitoTestListener = listenerSupplier.get();
+                Mockito.framework().addListener(mockitoTestListener);
```
bric3: comment paraphrase ;)

src/main/java/org/mockito/internal/runners/util/RunnerProvider.java
```diff
@@ -4,24 +4,24 @@
...
-
 public class RunnerProvider {
 
-    public RunnerImpl newInstance(String runnerClassName, Class<?> constructorParam) throws Exception {
+    public RunnerImpl newInstance(String runnerClassName, Object ... constructorArgs) throws Exception {
         Constructor<?> constructor;
         try {
             Class<?> runnerClass = Class.forName(runnerClassName);
-            constructor = runnerClass.getConstructor(Class.class.getClass());
+            constructor = runnerClass.getConstructors()[0];
```
bric3: It's not safe to rely on order of constructor, I don't believe the JVM garantees it. If there's a single constructor that's ok, but any implementor can add another (e.g. for tests), etc..

src/main/java/org/mockito/internal/runners/RunnerFactory.java
```diff
@@ -16,9 +21,34 @@
...
+        });
+    }
+
+    /**
+     * Creates runner implementation with provided listener supplier
+     */
+    public RunnerImpl create(Class<?> klass, Supplier<MockitoTestListener> listenerSupplier) throws InvocationTargetException {
         try {
-            return new RunnerProvider().newInstance("org.mockito.internal.runners.SilentJUnitRunner", klass);
+            return new RunnerProvider().newInstance("org.mockito.internal.runners.SilentJUnitRunner", klass, listenerSupplier);
```
bric3: Why not `SilentJUnitRunner.class.getName()` ?

src/main/java/org/mockito/internal/runners/RunnerFactory.java
```diff
@@ -16,9 +21,34 @@
...
+    /**
+     * Creates strict runner implementation
+     */
+    public RunnerImpl createStrict(Class<?> klass) throws InvocationTargetException {
+        return create(klass, new Supplier<MockitoTestListener>() {
+            public MockitoTestListener get() {
+                return new MismatchReportingTestListener(new ConsoleMockitoLogger());
+            }
+        });
+    }
```
bric3: Why not using the same construction with strictness enum and a switch statement ?

src/main/java/org/mockito/internal/runners/RunnerFactory.java
szczepiq: To avoid static linking that leads to uncaught class errors. Good question!

src/main/java/org/mockito/internal/runners/util/RunnerProvider.java
szczepiq: Good point! For simplicity I'll just add a check that we expect exactly one constructor.

src/main/java/org/mockito/internal/runners/SilentJUnitRunner.java
szczepiq: Actually, I kinda like it in the code :). Leaving it for now.

src/main/java/org/mockito/internal/runners/SilentJUnitRunner.java
```diff
@@ -7,35 +7,69 @@
...
 import org.junit.runners.model.FrameworkMethod;
 import org.junit.runners.model.InitializationError;
 import org.junit.runners.model.Statement;
+import org.mockito.Mockito;
 import org.mockito.MockitoAnnotations;
-import org.mockito.internal.runners.util.FrameworkUsageValidator;
+import org.mockito.internal.junit.MockitoTestListener;
+import org.mockito.internal.util.Supplier;
 
+//TODO rename to DefaultInternalRunner and the parent to InternalRunner
```
szczepiq: Given that it has more features that just being "silent" (initializing mocks, validating framework) would still prefer the Silent in the name?

src/main/java/org/mockito/internal/runners/RunnerImpl.java
```diff
@@ -11,8 +11,10 @@
 /**
  * I'm using this surrogate interface to hide internal Runner implementations.
  * Surrogate cannot be used with &#064;RunWith therefore it is less likely clients will use interal runners.
+ *
+ * TODO rename to InternalRunner
  */
-public interface RunnerImpl extends Filterable{
+public interface RunnerImpl extends Filterable {
```
szczepiq: It will make the diffs in GitHub unusable. Are you sure? :)

src/main/java/org/mockito/internal/runners/RunnerFactory.java
szczepiq: Good point. It feels that the current implementation is simpler then when I add conditional logic.

szczepiq:
Merging today. Thanks guys for really nice feedback!!!

src/main/java/org/mockito/internal/junit/NoOpTestListener.java
```diff
@@ -2,7 +2,7 @@
 
 import org.mockito.mock.MockCreationSettings;
 
-class NoOpTestListener implements MockitoTestListener {
+public class NoOpTestListener implements MockitoTestListener {
```
szczepiq: yes. We still have 2 internal packages: internal.runners and internal.junit. We should merge them but perhaps after we stop maintaining 2.x?

bric3: That makes sense.

src/main/java/org/mockito/internal/runners/RunnerImpl.java
bric3: Nope git (and so is github) can track renames too

src/main/java/org/mockito/internal/runners/RunnerFactory.java
bric3: Yes I supposed this was that, but since the class is in the same package it feels not needed. And may complicate a tad more renaming.

PR #774, opened 2016-11-24
Added subproject for supporting class loading on Android.

bric3:
This is a work from @raphw to bring mockito to android devices without dexmaker, using ByteBuddy Android instead.

PR #809, opened 2016-12-11
Automated publication of mockito submodule like mockito-android

szczepiq:
The goal of this PR is:
 - create a sample subproject "android" that is automatically released to Bintray / Maven central
 - refactor the build code so that it is easy to publish more artifacts in the future. TestNG? :)

What happens when this PR is merged?
 - with every mockito-core release, mockito-android jar will be automatically uploaded to Bintray, however, it will be "invisible" until it somebody "publishes" it from Bintray UI
 - when we are comfortable with the android release, we can toggle the "bintrayAutoPublish" and "mavenCentralSync" property in "android.gradle" files and have fully automated publications of mockito-android

Gotchas
- Mockito release has a tweak that when mockito-core source code hasn't changed since previous version, the release will be skipped and no jars will be published to artifactory. We need to fix this logic so that it is more flexible for publishing changes to other submodules like mockito-android. In the meantime, to work around this limitation, just make a whitespace change in Mockito javadoc to force release.

How did I test this change?
- ran build and compare the distributions created, the source, javadoc, pom files were good.
- ran publications from my local to Bintray test repo, inspected Bintray content
- dry / test runs of release from local

Enjoy :)


gradle/release.gradle
```diff
@@ -147,6 +158,14 @@ task assertCleanWorkingCopy {
     }
 }
 
+task assertEnvVariables {
+    doLast {
+        ['TRAVIS_BRANCH', 'GH_TOKEN', 'MOCKITO_BINTRAY_API_KEY'].each {
+            assert System.env[it]: "Please export '$it' envinronment variable first"
```
TimvdLippe: environment

gradle/publishable-java-library.gradle
```diff
@@ -0,0 +1,46 @@
...
+        }
+    }
+}
+
+//Gradle requires to configure pom after the publication above is configured
+apply from: "$rootDir/gradle/pom.gradle"
+apply from: "$rootDir/gradle/publish.gradle"
+
+//so that we flesh out problems with maven publication during the build process
+build.dependsOn publishToMavenLocal
```
TimvdLippe: :+1: :+1:

gradle/publishable-java-library.gradle
```diff
@@ -0,0 +1,46 @@
+def licenseFiles = copySpec {
+    //mockito license
+    from(".") { include 'LICENSE' }
+}
+
+//TODO why do we still ship the main jar with sources?
```
TimvdLippe: So users have javadoc in the mockito jar?

szpak: Yup, it's an old issue: #158. What was the original reason to do that? Is it still present?

szczepiq: The use case was for Eclipse users that manually dropped mockito jar on the classpath. They would have automatic source for debugging / javadoc out of the box. It didn't help IDEA out of the box because IDEA does not look for sources inside jar. This was the time before declarative dependency resolution where people kept jars checked-in.

We should move on and stop putting sources in main jar. Nobody does it this way in the modern world. Makes the jar bigger and has some other issues.

szczepiq:
I'm ready see the description.

TimvdLippe:
The CI is failing though, so that has to be fixed

build.gradle
```diff
@@ -17,20 +17,28 @@ plugins {
...
+apply from: "gradle/publishable-java-library.gradle"
+
+apply from: 'gradle/root/ide.gradle'
+apply from: 'gradle/root/coverage.gradle'
+apply from: 'gradle/root/gradle-fix.gradle'
+
+apply from: 'gradle/mockito-core/osgi.gradle'
+apply from: 'gradle/mockito-core/javadoc.gradle'
+apply from: 'gradle/mockito-core/license.gradle'
+apply from: 'gradle/release.gradle'
```
bric3: So every `apply from:` after `archivesBaseName = 'mockito-core'` are applied for the `mockito-core` ?

build.gradle
```diff
@@ -17,20 +17,28 @@ plugins {
...
-apply from: "gradle/ide.gradle"
-apply from: 'gradle/coverage.gradle'
-apply from: 'gradle/osgi.gradle'
-apply from: 'gradle/gradle-fix.gradle'
-apply from: 'gradle/license.gradle'
+ext {
+    artifactName = 'mockito-core'
+    bintrayAutoPublish = true
+    mavenCentralSync = true
+}
```
bric3: What does mean `ext` ?

bric3:
@TimvdLippe this is because bintray has already had the artifact, so this upload task failed. I just removed it from bintray and restarted the builds but it fails for another reason.

TimvdLippe:
Hm probably we need to disable the actual upload, but do trigger the release logic.

TimvdLippe:
@szczepiq The last commit triggered a release again and published once again 2.3.1 to bintray: https://bintray.com/mockito/maven/mockito/2.3.1

bric3:
> @szczepiq The last commit triggered a release again and published once again 2.3.1 to bintray: https://bintray.com/mockito/maven/mockito/2.3.1

I'm probably the culprit here as I tried to get the pull request green.

szczepiq:
Still working on this. The publication should be fixed but the pom file is still busted. Thanks for patience!

szczepiq:
I fixed the pom.xml problem and I have added an integration test for it.

build.gradle
```diff
@@ -81,52 +59,19 @@ dependencies {
...
-}
-
-task sourcesJar(type: Jar) {
-    jar {
-        baseName = 'mockito-core'
-        from(sourceSets.main.allSource)
-        with licenseFiles
+//need to declare dependencies before applying publishable java library
+//otherwise the pom file is incorrect
+apply from: "gradle/publishable-java-library.gradle"
```
bric3: Isn't it possible to add a check in the `publishable-java-library.gradle`

subprojects/android/android.gradle
```diff
@@ -0,0 +1,8 @@
+ext {
+    artifactName = 'mockito-android'
+    bintrayAutoPublish = false
+    mavenCentralSync = false
+}
+
+apply from: "$rootDir/gradle/java-library.gradle"
+apply from: "$rootDir/gradle/publishable-java-library.gradle"
```
bric3: Android subproject is not there yet, but if we want correct pom we'd need to add that, right ?

```
dependency {
    runtime "org.mockito:mockito-core:$version"
    runtime "net.bytebuddy:byte-buddy-android:1.5.5"
}
```

TimvdLippe:
This conflicts with #774. @szczepiq could you incorporate the changes in #774 into this PR?

subprojects/android/android.gradle
szczepiq: The aspiration of this PR is to make new module publishable, not to implement android submodule :) When we implement the module, it probably only needs single dependency on root project, e.g:

```groovy
dependencies {
  compile project(":")
}
```

build.gradle
szczepiq: I really like this idea. I can emit a warning. I'll also check if we can work around using afterEvaluate for this use case.

szczepiq:
>This conflicts with #774. @szczepiq could you incorporate the changes in #774 into this PR?

The aspiration of this PR is to make new module publishable, not to implement android submodule :) So, I intend to merge it without android code yet. If we're ready to ship android, just flip the flags in the build.gradle file.

build.gradle
szczepiq: I've tried but it's not working very well. Let's move on.

eygraber:
Would you guys be interested in using [this](https://github.com/eygraber/bintray-gradle-scripts) to make deploying to bintray much more simple? I can create a PR if you do.

szczepiq:
@eygraber, thanks for offering! I looked at your site but I'm not sure what value would it bring. Also, I have a feeling that we need finer control - for example releasing submodules to different repos. We'd rather avoid more build dependencies, unless they provide decent value.

eygraber:
@szczepiq It would provide an abstraction around all the bintray and maven stuff, and move configuration to "property" files instead of the main gradle build files. All you would have to do to get artifacts to bintray would be to add a property file to the root project and/or each module that specify things like the artifact name, package, repo, version, dry run, etc...

Then there is one task to call that uploads the artifacts to bintray (`deployToBintray`) which can deploy all the artifacts for the project, or you can pick which modules to upload artifacts for. There is also a task for uploading snapshot artifacts (but I believe you guys don't use that).

szczepiq:
Thank you for information! I suggest to put the goals / benefits of the plugin on its GitHub site / readme.

At the moment, we don't need an abstraction around bintray / maven. Or rather, we already have an abstraction, e.g. as separate gradle files that configure maven / Bintray for all modules we produce. Thanks for offering the plugin, though!

eygraber:
Great suggestion; I'll do that.

PR #811, opened 2016-12-12
Enable annotation engine as plugin

bric3:
Currently `AnnotationEngine` can be overrriden with the _deprecated_ `MockitoConfiguration`, in order to be more consistent with the new way, I propose to get the `AnnotationEngine` from the `PluginRegistry`.

Regarding backward compatible behavior, if the `MockitoConfiguration` class exists and can be seen then Mockito will select the engine of this configuration instead of the `Plugins` one. If `MockitoConfiguration` class don't exist then Mockito wil chose the `Plugins` one.

Bonus I extended the classloader util to support some tests.


bric3:
Still work in progress

bric3:
For reference last commit skips `static final` fields otherwise the dead simple field copier raises an `IllegalAccessException`

```
java.lang.IllegalAccessException: Can not set static final [Z field org.mockitoutil.ClassLoadersTest$1.$jacocoData to [Z
	at sun.reflect.UnsafeFieldAccessorImpl.throwFinalFieldIllegalAccessException(UnsafeFieldAccessorImpl.java:73)
	at sun.reflect.UnsafeFieldAccessorImpl.throwFinalFieldIllegalAccessException(UnsafeFieldAccessorImpl.java:77)
	at sun.reflect.UnsafeQualifiedStaticObjectFieldAccessorImpl.set(UnsafeQualifiedStaticObjectFieldAccessorImpl.java:77)
	at java.lang.reflect.Field.set(Field.java:741)
	at org.mockitoutil.ClassLoaders$ClassLoaderExecutor.reloadTaskInClassLoader(ClassLoaders.java:153)
	... 6 more
```

This part of the classloader test have some limitations, but it should be ok for simple tests.

bric3:
I suggest to review this PR commit by commit.

PR #812, opened 2016-12-12
Fixes #810 : inconsistency_with_spy_and_injectmocks

### LINKED ISSUE #810: Consistent behaviour for @InjectMocks with @Spy
Based on a lot of discussions I found:
https://code.google.com/archive/p/mockito/issues/489
https://groups.google.com/forum/#!msg/mockito-dev/YFqTPm8yJSc/Y-ty81kLKI4J
and a few threads on SO, it seems mockito doesn't have an intention to support multiple-layer injection. I would like to see it work, but as many justified the decision: "Mockito was never intended to be a fully-fledged injection framework, it serves a different purpose". It's a shame, but I understand and respect that.

Now, it doesn't look consistent in the situation I'm facing:
This test will run successfully:
```
public class PaymentHelperTest  extends WhatEverClassYouWant {

    @Mock
    private DateUtil dateUtil;

    @InjectMocks
    private PaymentHelper paymentHelper;
```

But the exact same test without inheritance will fail:
```
public class PaymentHelperTest {

    @Mock
    private DateUtil dateUtil;

    @InjectMocks
    private PaymentHelper paymentHelper;
```
It happens because of the tricky loops like:
`        while (clazz != Object.class) {
          .....
            clazz = clazz.getSuperclass();
        }`
And a check in InjectMocksScanner.scan():
            `if (null != field.getAnnotation(InjectMocks.class)) `

I am planning to add another check in there:

yuranos87:
Based on a lot of discussions I found:
https://code.google.com/archive/p/mockito/issues/489
https://groups.google.com/forum/#!msg/mockito-dev/YFqTPm8yJSc/Y-ty81kLKI4J
and a few threads on SO, it seems mockito doesn't have an intention to support multiple-layer injection. I would like to see it work, but as many justified the decision: "Mockito was never intended to be a fully-fledged injection framework, it serves a different purpose". It's a shame, but I understand and respect that.

Now, it doesn't look consistent in the situation I'm facing:
This test will run successfully:
```
public class PaymentHelperTest  extends WhatEverClassYouWant {

    @Mock
    private DateUtil dateUtil;

    @InjectMocks
    private PaymentHelper paymentHelper;
```

But the exact same test without inheritance will fail:
```
public class PaymentHelperTest {

    @Mock
    private DateUtil dateUtil;

    @InjectMocks
    private PaymentHelper paymentHelper;
```
It happens because of the tricky loops like:
`        while (clazz != Object.class) {
          .....
            clazz = clazz.getSuperclass();
        }`
And a check in InjectMocksScanner.scan():
            `if (null != field.getAnnotation(InjectMocks.class)) `

I am planning to add another check in there:

### DIFF
--- a/src/main/java/org/mockito/internal/configuration/injection/scanner/InjectMocksScanner.java
+++ b/src/main/java/org/mockito/internal/configuration/injection/scanner/InjectMocksScanner.java
@@ -7,13 +7,14 @@ package org.mockito.internal.configuration.injection.scanner;
 import org.mockito.Captor;
 import org.mockito.InjectMocks;
 import org.mockito.Mock;
-
-import static org.mockito.internal.exceptions.Reporter.unsupportedCombinationOfAnnotations;
+import org.mockito.Spy;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.util.HashSet;
 import java.util.Set;
+
+import static org.mockito.internal.exceptions.Reporter.unsupportedCombinationOfAnnotations;
 
 /**
  * Scan field for injection.
@@ -24,7 +25,7 @@ public class InjectMocksScanner {
     /**
      * Create a new InjectMocksScanner for the given clazz on the given instance
      *
-     * @param clazz    Current class in the hierarchy of the test
+     * @param clazz Current class in the hierarchy of the test
      */
     public InjectMocksScanner(Class<?> clazz) {
         this.clazz = clazz;
@@ -50,7 +51,9 @@ public class InjectMocksScanner {
         Set<Field> mockDependentFields = new HashSet<Field>();
         Field[] fields = clazz.getDeclaredFields();
         for (Field field : fields) {
-            if (null != field.getAnnotation(InjectMocks.class)) {
+            if (null != field.getAnnotation(InjectMocks.class)
+                    && null == field.getAnnotation(Spy.class)
+                    && null == field.getAnnotation(Mock.class)) {
                 assertNoAnnotations(field, Mock.class, Captor.class);
                 mockDependentFields.add(field);
             }


TimvdLippe:
For us to accept this pull request, please include a testcase that reproduces your original problem to verify it has been fixed

bric3:
Hi thanks for proposing a fix. But this report is incomplete as it doesn't show what's wrong with mockito, the proposed change breaks current test cases and as @TimvdLippe noticed there's no new testcases showcasing the original problem.
I'm closing for now, if valid data is added I'll reopen the ticket.

yuranos87:
Hi, @bric3!

I noticed things like:
@Spy @InjectMocks private MyObject myClass
are used a lot in tests.
I also found your comments in: https://github.com/mockito/mockito/issues/169

If multiple @InjectMocks are not to be supported in one test, adding @Mock/@Spy together with @InjectMocks should also not be supported, because otherwise nothing stops people from abusing it.
What's worse is that tests that have ANY inheritance will pass, while the ones that won't - will fail.
Inconsistency is evil.

bric3:
> multiple @InjectMocks are not to be supported in one test

This is supported, just not injecting one into the other.

> adding @Mock/@Spy together with @InjectMocks should also not be supported, because otherwise nothing stops people from abusing it.

I don't understand the argument. Also there's no reason to have a field with both `@Mock @InjectMocks`, this just makes no sense.

> What's worse is that tests that have ANY inheritance will pass, while the ones that won't - will fail.
> Inconsistency is evil.

Mockito works well with test inheritance or injectee object graph, if the unit test does not work, there must be something else in play.

yuranos87:
@bric3, I explained it with example. Create the same test class, one with inheritance, another one - without.
Mockito will initialise @Spy @InjectMocks for the former, but not for the latter.
I debugged it and it's clearly because the logic inside Mockito.

yuranos87:
https://code.google.com/archive/p/mockito/issues/489 - the same issue is explained here. The difference is that if your test inherits something, the issue is not reproducible, otherwise it is.

bric3:
The old 489 won't be fixed. The non-compilable examples in the description of the #810 issue or of this PR work as expected.

PR #819, opened 2016-12-14
Fixes bug were previously verified invocations could not capture argu…

bric3:
Since Mockito 2 / #380 captured arguments is not anymore done for already verified interactions.

```java
mock.oneArg("first");
ArgumentCaptor<String> argument = ArgumentCaptor.forClass(String.class);
verify(mock, times(1)).oneArg(argument.capture());
assertThat(argument.getAllValues()).isEqualTo(1);

// additional interactions
mock.oneArg("second");
argument = ArgumentCaptor.forClass(String.class);
verify(mock, times(2)).oneArg(argument.capture());
assertThat(argument.getAllValues()).isEqualTo(2); // fail with mockito 2
```

The current behavior (Mockito 2.1 - 2.3.3) of `NumberedInvocationChecker` clears already verified interactions, before verifying the remaining interactions so the new argument captor is not populated with previous values. Which is akward if the times value is different that the lit size.

### DIFF
--- a/src/main/java/org/mockito/internal/verification/checkers/AtLeastXNumberOfInvocationsChecker.java
+++ b/src/main/java/org/mockito/internal/verification/checkers/AtLeastXNumberOfInvocationsChecker.java
@@ -5,7 +5,12 @@
 
 package org.mockito.internal.verification.checkers;
 
-import java.util.Iterator;
+import java.util.List;
+import org.mockito.internal.verification.api.InOrderContext;
+import org.mockito.invocation.Invocation;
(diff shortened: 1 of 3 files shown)


TimvdLippe:
But that would regress #379 right?

bric3:
Actually no, the tests added in #380 (which fixes #379) are still green (look at `VerficationAfterDelayTest`). There have been other change in this area since 380, but I haven't tracked where it is fixed. Running the tests added in #380 on mockito 1.x show they fail, so they are still valuable and still fixed.

PR #821, opened 2016-12-14
Fixes #820 ReturnsArgAt to handle returning vararg as arrays

### LINKED ISSUE #820: [VarArgs] unexpected behaviour of returnsArgAt(int)
I stumble about a test that failed after refactoring an array parameter to an varargs.

The following test succeed:
```java
interface Foo{
        int[] getOdd(int[] input);
}

@Test
    when(foo.getOdd(any())).then(returnsArgAt(0));
        
    int[] odd = mock.getOdd(new int[]{1,3,5});
    assertThat(odd).contains(1,3,5);
}
```
Now lets change the parameter _input_ to an varargs.
```java
interface Foo{
        int[] getOdd(int... input);
}
```
The test fails now with:

```
org.mockito.exceptions.misusing.WrongTypeOfReturnValue: 
The argument of type 'int' cannot be returned because the following 
method should return the type 'int[]'
 -> foo.getOdd()
```

This is not expected cause the change is source and binary(?) compatible, an varargs-parameter is still an array. ReturnsArgumentAt interprets the vararg array items as real parameters. In other words it tries to return an element type where the array type is required.

Can the behaviour of interpreting vararg elements like parameters be changed so an vararg is treated as an array? E.g. in Mockito 3? There are several other cases were the current vararg handling leads to unexpected behaviour.

bric3:
Attempt at fixing #820

Basically it makes the ReturnsArgumentAt answer a bit more smart about varargs. But I'm not satisfied with the fix yet.

### DIFF
--- a/src/main/java/org/mockito/internal/stubbing/answers/ReturnsArgumentAt.java
+++ b/src/main/java/org/mockito/internal/stubbing/answers/ReturnsArgumentAt.java
@@ -5,17 +5,16 @@
 package org.mockito.internal.stubbing.answers;
 
 import java.io.Serializable;
+import java.lang.reflect.Method;
+import org.mockito.internal.exceptions.Reporter;
+import org.mockito.invocation.Invocation;
 import org.mockito.invocation.InvocationOnMock;
 import org.mockito.stubbing.Answer;
 import org.mockito.stubbing.ValidableAnswer;
 
-import static org.mockito.internal.exceptions.Reporter.invalidArgumentPositionRangeAtInvocationTime;
-import static org.mockito.internal.exceptions.Reporter.invalidArgumentRangeAtIdentityAnswerCreationTime;
-import static org.mockito.internal.exceptions.Reporter.wrongTypeOfArgumentToReturn;
-
 /**
  * Returns the passed parameter identity at specified index.
- *
+ * <p>
  * <p>The <code>argumentIndex</code> represents the index in the argument array of the invocation.</p>
  * <p>If this number equals -1 then the last argument is returned.</p>
  *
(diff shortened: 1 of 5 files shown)


src/main/java/org/mockito/internal/stubbing/answers/ReturnsArgumentAt.java
```diff
@@ -41,9 +42,18 @@ public ReturnsArgumentAt(int wantedArgumentPosition) {
 
     public Object answer(InvocationOnMock invocation) throws Throwable {
         validateIndexWithinInvocationRange(invocation);
-        return invocation.getArgument(actualArgumentPosition(invocation));
+        // si vararg && actualArgumentPosition = last parameter && parameter compatible avec return type
```
TimvdLippe: French comment? 😂

src/main/java/org/mockito/internal/stubbing/answers/ReturnsArgumentAt.java
```diff
@@ -97,16 +107,21 @@ private boolean argumentPositionInRange(InvocationOnMock invocation) {
...
         }
 
-        Class<?>[] parameterTypes = invocation.getMethod().getParameterTypes();
         int varargPosition = parameterTypes.length - 1;
 
         if(actualArgumentPosition < varargPosition) {
             return parameterTypes[actualArgumentPosition];
         } else {
+            // try the vararg array type
+            if(actualArgumentPosition == varargPosition
```
TimvdLippe: spacing

src/main/java/org/mockito/internal/stubbing/answers/ReturnsArgumentAt.java
bric3: haha yes, I'm sick, so my brain had to think in french and slowly

ChristianSchwarz:
Wouldn't it a bit confusing that `returnsArgAt(0)` returns different types for the same argument type?

```java
int[] foo(int ...input) //returnsArgAt(0) would return an array
int foo(int ..input) //returnsArgAt(0) would returns an int
```

What about handling varargs always as array? It would break the actual behaviour, but I wonder if this would be an option in mockito 3?

bric3:
@TimvdLippe Not sure what [_stfm_](https://www.google.fr/search?q=SFTM&oq=SFTM&aqs=chrome..69i57j0l5.334j0j7&sourceid=chrome&ie=UTF-8#q=define:SFTM) means ;P

@ChristianSchwarz Yeah I pondered on this ; the use case is pretty limited though, i.e. all following conditions have to be met for the type change to take effect : 

* vararg method
* wanted index have to be the index of the vararg
* type of the vararg have to be assignable to the return type

In the end I think the new behavior makes sense

> ```java
> int[] foo(int... input) // returnsArgAt(0) would return an array
> int foo(int... input) // returnsArgAt(0) would returns an int
> ```

> Wouldn't it a bit confusing that returnsArgAt(0) returns different types for the same argument type?

Actually the current behavior is confusing hence the #820 ;)

> What about handling varargs always as array? It would break the actual behaviour, but I wonder if this would be an option in mockito 3?

I'm not sure what you exactly mean, but what about arg index beyond the first position of the vararg.

```java
int foo(int... input)
foo(1,2,3,4)
when(...).then(returnsArgAt(3))
```

As a user I wouldn't be happy without and I wouldn't be happy this behaviour changed.

bric3:
One of the thing that displease me here is that `AnswersValidator` knows too much about answers, I'd like a validable answer.

ChristianSchwarz:
@bric3 

> I'm not sure what you exactly mean [...]

What I wanted to say is a bit out of scope for this PR, I try keep it short: Mocktio should not expand varargs and treat them as regular arguments, this causes also problems in various vararg+captor scenarios. Maybe it is better to handle varargs seperate:

example: `int` return type 
```java
int foo(int... input)
foo(1,2,3,4)
when(...).then(returnsArgAt(0) 
//^ throws an Exception: arg0 of type int[] is not compatible with return type int

//new API  for varargs
when(..).then(returnVarargAt(0))//returns 1
```
example: `int[]` return type 
```java
int[] bar(int... input)
bar(1,2,3,4)
when(...).then(returnsArgAt(0) //return the vararg int[]

//new API  for varargs
when(..).then(returnVarargAt(0))
//^ throws an Exception: arg0[0] of type int is not compatible with return type int[]
```

ChristianSchwarz:
> One of the thing that displease me here is that AnswersValidator knows too much about answers, I'd like a validable answer.

+100 Great idea! To keep `Answer` a SAM/FunctionalInterface an extra interface `ValidatableAnswer` would be required, right?

bric3:
Yes

TimvdLippe:
> @TimvdLippe Not sure what stfm means ;P

Was a typo, meant Seems Fine To Me.

bric3:
I still think the change is safe and predicatable from a user's perspective. However let's wait on #826

bric3:
Rebased on top of release/2.x, it's better now. I'll see if I can improve the code now.

szczepiq:
Nice work and really cool idea about ValidableAnswer.

>Mockito should not expand varargs and treat them as regular arguments, this causes also problems in various vararg+captor scenarios.

I have high level feedback, hopefully you find it useful. Mockito is optimized for simplicity, intuitiveness, readability. Var args look in the code as a sequence of args and Mockito models it similarly in the API (that's why we expand them). Can we solve captor+vararg edge cases while keeping the API simple?

bric3:
Yeah agreed

bric3:
Ok now the implementation seems more readable. It is ready for review.
I've updated the javadoc to document the behavior as well.

TimvdLippe:
On mobile right now, but implementation seems fine!

PR #825, opened 2016-12-15
#719 Test clean up according to discussions for RememberingListeners

kobbikobb:
According to issue @719 we wanted to clean up the tests for `RememberingListener`s.

The tests should follow the same style and conventions.
Set up should not contain variables used by all the tests.

### DIFF
--- a/src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
+++ b/src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
@@ -1,24 +1,28 @@
 package org.mockitousage.debugging;
 
+import java.lang.reflect.Method;
 import org.assertj.core.api.Condition;
+import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.JUnitCore;
 import org.mockito.Mockito;
 import org.mockito.MockitoFramework;
+import org.mockito.StateMaster;
 import org.mockito.exceptions.base.MockitoAssertionError;
 import org.mockito.internal.verification.api.VerificationData;
 import org.mockito.listeners.VerificationListener;
 import org.mockito.verification.VerificationEvent;
 import org.mockito.verification.VerificationMode;
 
-
-import java.lang.reflect.Method;
-
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.Assert.assertNull;
 import static org.junit.Assert.fail;
-import static org.mockito.Mockito.*;
+import static org.mockito.Mockito.any;
+import static org.mockito.Mockito.framework;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
 
 public class VerificationListenerCallBackTest {
     private Method invocationWanted;
@@ -31,6 +35,13 @@ public class VerificationListenerCallBackTest {
         mockitoFramework = Mockito.framework();
         mockitoFramework.addListener(listener);
         invocationWanted = Foo.class.getDeclaredMethod("doSomething", String.class);
+    }
+
+    @After
+    public void reset_mockito() {
+        StateMaster stateMaster = new StateMaster();
+        stateMaster.reset();
+        stateMaster.clearMockitoListeners();
     }
 
     @Test


src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
```diff
@@ -12,30 +11,22 @@
...
 import static org.junit.Assert.fail;
 import static org.mockito.Mockito.*;
 
 public class VerificationListenerCallBackTest {
-    private Method invocationWanted;
-    private RememberingListener listener;
-    private MockitoFramework mockitoFramework;
-
-    @Before
-    public void setUp() throws NoSuchMethodException {
```
TimvdLippe: I would say to keep the setUp. It is cleaner and more DRY.

src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
```diff
@@ -47,41 +38,50 @@ public void should_call_single_listener_on_verify() {
...
-        assertThat(listener).is(notifiedFor(foo, never, invocationWanted));
+        assertThat(listener1).is(notifiedFor(foo, never, invocationWanted));
         assertThat(listener2).is(notifiedFor(foo, never, invocationWanted));
     }
 
     @Test
     public void should_not_call_listener_when_verify_was_called_incorrectly() {
         //when
+        VerificationListener listener = mock(VerificationListener.class);
+        framework().addListener(listener);
```
TimvdLippe: Why the reordering?

src/test/java/org/mockitousage/debugging/InvocationListenerCallbackTest.java
```diff
@@ -23,39 +24,36 @@
  */
 public class InvocationListenerCallbackTest {
 
-    // Cannot use a mockito-mock here: during stubbing, the listener1 will be called
-    // and mockito will confuse the mocks.
-    private RememberingListener listener1 = new RememberingListener();
-    private RememberingListener listener2 = new RememberingListener();
```
TimvdLippe: I would say let's keep them here

src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
bric3: But interdependant. Since the _given_ part of each tests is about the framework listner I think this is a good practice to put it in the test method. If this was regular setup that had no impact on the _given_ part of the scenario then it's probably ok in a before method.

src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
```diff
@@ -12,30 +11,22 @@
...
 
     @Test
-    public void should_call_single_listener_on_verify() {
+    public void should_call_single_listener_on_verify() throws NoSuchMethodException {
         //given
+        RememberingListener listener = new RememberingListener();
+        MockitoFramework mockitoFramework = Mockito.framework();
+        mockitoFramework.addListener(listener);
+
+        Method invocationWanted = Foo.class.getDeclaredMethod("doSomething", String.class);
```
bric3: I suggest to inline variable `invocationWanted`, at the moment this variable is declared in the given section but has no impact on the scenario.

This apply for other tests.

src/test/java/org/mockitousage/debugging/InvocationListenerCallbackTest.java
```diff
@@ -95,4 +87,15 @@ public void reportInvocation(MethodInvocationReport mcr) {
...
+    static Condition<RememberingListener> notifiedFor(final Object returned, final String location) {
+        return new Condition<RememberingListener>() {
+            public boolean matches(RememberingListener toBeAsserted) {
+                assertThat(toBeAsserted.returnValue).isEqualTo(returned);
+                assertThat(toBeAsserted.invocation).isNotNull();
+                assertThat(toBeAsserted.locationOfStubbing).contains(location);
+                return true;
+            }
+        };
+    }
```
bric3: Since it's used in multiple tests, I suggest to move the method to `org.mockitoutil.Conditions`.

src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
kobbikobb: @TimvdLippe so some tests use 0, 1 and 2 concrete listeners. This gets confusing and in the mock test the name of the listener needs to be mockXXX because listener is taken. I find the tests more easy to read and maintain when only common logic is in the setup. The most popular unit testing books would agree on this.
This way yes there is some duplication but very minimal. I thought about moving some other pieces to the setup but I felt this was more clean.
Personally I have been burned by this...
Does this make some sense to you?

src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
kobbikobb: Yes I agree. Was thinking about it... I will change it.

src/test/java/org/mockitousage/debugging/VerificationListenerCallBackTest.java
TimvdLippe: Then it is just my personal preference to do it otherwise, but I am okay with this approach so lets leave it.

kobbikobb:
@bric3 I've inlinded `Foo.class.getDeclaredMethod("doSomething", String.class)`
Regarding putting the method `notifiedFor` into org.mockitoutil.Conditions.
The method is not same between `InvocationListenerCallbackTest` and `VerificationListenerCallBackTest` and they both use a different implementation of `RememberingListener`.
I would like to wait for more test classes wanting to do the same thing and then move it. Else we need to move also the class and find new suitable names for the classes.
Unless you have another idea to improve this?

bric3:
I just tried the webview to resolve conflicts, it's ok but limited, and it creates a commit named like `Merge branch 'release/2.x' into verification-listeners-cleanup`. I'd rather have more control on this.

bric3:
> Regarding putting the method notifiedFor into org.mockitoutil.Conditions.
The method is not same between InvocationListenerCallbackTest and VerificationListenerCallBackTest and they both use a different implementation of RememberingListener.
I would like to wait for more test classes wanting to do the same thing and then move it. Else we need to move also the class and find new suitable names for the classes.
Unless you have another idea to improve this?

OK fair enough :)

PR #826, opened 2016-12-15
Introduce validable answers

### LINKED ISSUE #820: [VarArgs] unexpected behaviour of returnsArgAt(int)

bric3:
While trying to fix #820 / #821, I noticed that `AnswersValidator` knew too much about other answers. To fix the design issue I propose the following design :

* Introduce `ValidableAnswer` in public API which is a single method interface that doesn't extends `Answer`.
* Each answer that wants to be validated hence implement this interface

Benefit, reduced visibility of some methods in these answers. Each new answer that want to be validated can just by implementing the new contract.

Validating returned value of a mock's default answer had to be handled differently.


src/main/java/org/mockito/internal/stubbing/answers/MethodInfo.java
```diff
@@ -18,7 +17,7 @@
 
     private final Method method;
 
-    public MethodInfo(Invocation theInvocation) {
```
ChristianSchwarz: Better pass a Method-Object

szczepiq:
Hey! From the description, I don't quite get why we need additional public API. What does this public API give to the users?

ChristianSchwarz:
> What does this public API give to the users?

Users that like to write there own Answers have now the opportunity to add a runtime usage validation. This was not possible before. The `AnswersValidator` had the knowledge about predefined mockito answers harded coded, it was not extensible from outside. Thats why this API is public.

src/main/java/org/mockito/internal/stubbing/InvocationContainerImpl.java
```diff
@@ -59,8 +58,9 @@ public void addConsecutiveAnswer(Answer answer) {
     public StubbedInvocationMatcher addAnswer(Answer answer, boolean isConsecutive) {
         Invocation invocation = invocationForStubbing.getInvocation();
         mockingProgress().stubbingCompleted();
-        AnswersValidator answersValidator = new AnswersValidator();
-        answersValidator.validate(answer, invocation);
+        if ((Answer<?>) answer instanceof ValidableAnswer) {
+            ((ValidableAnswer) (Answer<?>) answer).validateFor(invocation);
```
bric3: Oups missed that. I think that's the IDE that did that !

src/main/java/org/mockito/internal/stubbing/answers/DefaultValidableAnswer.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package org.mockito.internal.stubbing.answers;
+
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+
+import static org.mockito.internal.exceptions.Reporter.wrongTypeReturnedByDefaultAnswer;
+
+public class DefaultValidableAnswer {
```
bric3: Actually this almost the same code as [`AnswerValidator.validateDefaultAnswerReturnedValue`](https://github.com/mockito/mockito/pull/826/files#diff-200d4fd7223f1f11f9971c8a3148ec42L99). I just renamed and inlined the code here. Look at the commit c588737f0e11dea3741abbcfd7205bd4eb77f913

src/main/java/org/mockito/internal/stubbing/answers/MethodInfo.java
bric3: Yes at this moment it does, I just changed the arg type by the proper interface. It's the responsability of `MethodInfo` to find the proper element on the invocation, not the caller. Maybe this class should be renamed `InvocationInfo`. I'd rather have some flexibility on this internal class.

src/main/java/org/mockito/internal/stubbing/answers/DefaultValidableAnswer.java
bric3: Problem with the proposal is that I need the `invocation`, for type checking against the method signature and to raise an error if needed. Also it extract too much in the `MockHandlerImpl`

ChristianSchwarz: Ah I see! I think it is possible to seperate the concerns here, if you keep the parameters of the "old" validateDefaultAnswerReturnedValue method, something like that:

```java
static void validatedReturnValue(InvocationOnMock invocation, Object actualReturnValue) 
```

IMO this would read a bit nicer on the call side:

```java
Object returnValue = answer.answer(invocation);
validatedReturnValue(invocation, returnValue);
```

vs.

```java
Object ret = new DefaultValidableAnswer(mockSettings.getDefaultAnswer()).validatedAnswerFor(invocation);  
//     ^ for a reader it may not be clear that this is the return value of the answer
```

src/main/java/org/mockito/internal/stubbing/answers/MethodInfo.java
ChristianSchwarz: +1 for renaming it to InvocationInfo

src/main/java/org/mockito/internal/stubbing/answers/ThrowsException.java
```diff
@@ -30,7 +33,20 @@ public Object answer(InvocationOnMock invocation) throws Throwable {
         throw t;
     }
 
-    public Throwable getThrowable() {
-        return throwable;
+    @Override
+    public void validateFor(InvocationOnMock invocation) {
+        Throwable throwable = this.throwable;
```
ChristianSchwarz: this line can be removed

src/main/java/org/mockito/internal/stubbing/answers/DefaultValidableAnswer.java
bric3: ok that works for me

src/main/java/org/mockito/stubbing/ValidableAnswer.java
```diff
@@ -0,0 +1,68 @@
...
+package org.mockito.stubbing;
+
+import org.mockito.Incubating;
+import org.mockito.invocation.InvocationOnMock;
+
+/**
+ * Allow to validate this answer is correct for the invocation.
+ *
+ * <p>
+ * Suppose there's If the code one was to write an answer like
```
TimvdLippe: This sentence is unclear, probably some typo's?

src/test/java/org/mockito/internal/invocation/InvocationBuilder.java
```diff
@@ -58,17 +59,33 @@ public Invocation toInvocation() {
...
         }
-        
-        Invocation i = new InvocationImpl(mock, new SerializableMethod(method), args, sequenceNumber, null,
-                location == null? new LocationImpl() : location);
+
+        Invocation i = new InvocationImpl(mock,
+                                          new SerializableMethod(method),
+                                          args,
+                                          sequenceNumber,
+                                          new RealMethod() {
```
TimvdLippe: Break this out into a local variable for readibility?

src/main/java/org/mockito/internal/stubbing/answers/MethodInfo.java
bric3: Done

src/main/java/org/mockito/stubbing/ValidableAnswer.java
bric3: Oups big typo !

bric3:
Updated `@since` to `2.3.8`

src/main/java/org/mockito/stubbing/ValidableAnswer.java
```diff
@@ -0,0 +1,84 @@
...
+ * </code></pre>
+ * </p>
+ *
+ * @since 2.3.8
+ */
+@Incubating
+public interface ValidableAnswer {
+
+    /**
+     * Perform validation of the answer at <em>stub time</em>.
```
ChristianSchwarz: Currently the JavaDoc describes the method from a caller perspective. The method is not meant to be called by client code but to be implemented. The JavaDoc should reflect that and guide the user how to implement it. Since the method has no boolean return type, it is not clear how to do it right. 

Maybe a comment like this helps here:

> The implementation must throw an MockitoException to indicate that this answer is not valid for the given invocation.  If the validation succeed the implementation must simply return without throwing.

bric3: OK

bric3:
I think this could be used for better maintainability for framework developers. Like mockito sugar that could add specifics regarding scala method signatures, etc...

bric3:
Damn the correct version is 2.3.10

szczepiq:
Thanks for quick turnaround. Really nice change!

szczepiq:
Added a press release to the top of this PR description :)

bric3:
@szczepiq A press release ?

szczepiq:
>@szczepiq A press release ?

By "press release" I mean a paragraph that describes the benefits for the users, at a high level.

bric3:
Oh you meant on the Mockito class pointing to [this](https://github.com/mockito/mockito/pull/826/files#diff-b414b69679387431a0e74e77f5c8460cR12). I find the feature less interesting than strict stubbing for end users. I think the release notes should be enough.

Issue #815, opened 2016-12-13
Proposal for lambda and void friendly when+then syntax


ChristianSchwarz:
This is a API proposal for an "when+then" that can be used to stub void methods (and static calls if they will be suported in the future) with a lambda friendly "then" API.

This ticket relates to #643 

Uses case:

 * Stubbing using an instance method reference: `when(mock::voidMethod).then(...);`
 * Stubbing using an Lambda expression:` when(()-> mock.voidMethod(gt(4))).then(...);`
 * Possible future use case, stubbing a static method:` when(()->Util.voidMethod(4)).then(..);`
 * Stubbing and processing the first argument:` when(()-> mock.voidMethod(gt(4))).then(firstArg -> println(firstArg));`
  * Stubbing and processing the InvocationOnMock: `when(mock::voidMethod).thenAnswer(invocation -> println(invocation.getMethod()));`

The API:
```java
class Mockito{
    OngoingVoidStubbing when(VoidCall call);
}

interface VoidCall { 
    void run() throws Throwable;
}

interface OngoingVoidStubbing {
    OngoingVoidStubbing thenAnswer(VoidAnswer answer);
    OngoingVoidStubbing then(VoidAnswer0 answer);
    <A> OngoingVoidStubbing then(VoidAnswer1<A> answer);
    <A,B> OngoingVoidStubbing then(VoidAnswer2<A,B> answer);
    <A,B,C> OngoingVoidStubbing then(VoidAnswer3<A,B,C> answer);
    OngoingVoidStubbing thenDoNothing();
    //... all other thenXXX methods from OngoingStubbing except thenReturn(..)
}

interface VoidAnswer{
     void answer(InvocationOnMock invocation) throws Throwable;
}

interface VoidAnswer0{
     <A> void answer(A firstArg) throws Throwable;
}
```
     
Current state: I have a working prototype ([gist patch](https://gist.github.com/ChristianSchwarz/db9973bb4a7043d4db410d36b7cf87d1)), currently I can't send a PR cause I am behind a firewall.

TimvdLippe:
Isn't this duplicate of #643? Adding the interfaces for void methods seems sufficient to include in that implementation to achieve this goal.

Do note that `OngoingVoidStubbing` must have generic arguments to obtain typesafety between the `when` and the `then`

ChristianSchwarz:
This proposal can be seen as different approach to realize stubbing of void methods, it just a base for further discussions. 

[...]

PR #828, opened 2016-12-15
Prototype implementation of when syntax for void methods

ChristianSchwarz:
This PR relates to #815, it allows to stub void methods like returning methods by using a method reference or a lambda containing a call to a void method.

```java
when(mock::voidMethod).then(...);
when(()->mock.voidMethod(gt(4))).then(...);
```

The prototype supports: argument matchers, argument capturing, consecutive answers, overriding answers and checks for lambda expressions that call more than one mock-method or no mock method.

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -8,6 +8,7 @@ import org.mockito.internal.MockitoCore;
 import org.mockito.internal.creation.MockSettingsImpl;
 import org.mockito.internal.debugging.MockitoDebuggerImpl;
 import org.mockito.internal.framework.DefaultMockitoFramework;
+import org.mockito.internal.stubbing.OngoingVoidStubbingImpl;
 import org.mockito.internal.verification.VerificationModeFactory;
 import org.mockito.junit.MockitoJUnit;
 import org.mockito.junit.MockitoJUnitRunner;
@@ -16,6 +17,7 @@ import org.mockito.mock.SerializableMode;
 import org.mockito.stubbing.Answer;
 import org.mockito.stubbing.Answer1;
 import org.mockito.stubbing.OngoingStubbing;
+import org.mockito.stubbing.OngoingVoidStubbing;
 import org.mockito.stubbing.Stubber;
 import org.mockito.stubbing.VoidAnswer1;
 import org.mockito.verification.After;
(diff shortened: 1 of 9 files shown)


TimvdLippe:
Looks cool! Your solution to running the method lambda is nice, had not thought of it. I am going to try to incorporate it into my PR because I think that one is feasible now! Thanks :)

szczepiq:
Good direction. Potentially useful for stubbing static methods down the road.

szczepiq:
Nice. Good job ;)

TimvdLippe:
@ChristianSchwarz Thank you for this PR. I will incorporate this in #742 as a full API change for also non-void methods :smile:

ChristianSchwarz:
Was there any disscussion about the pros and cons of both proposals? Can you give me a reason why you closed it, it is not transparent to me.

TimvdLippe:
I really need to start working on #742 again, but your proposal in this PR is exactly equal to what I was planning in #742. No difference. Therefore as I was planning to not only fix void-stubbing but also non-void-stubbing, I closed this PR as it will be incorporated in the bigger PR.

ChristianSchwarz:
> ...but your proposal in this PR is exactly equal to what I was planning in #742. 

The two proposals have a different primary goal and different API. But yes some of the secondary goals are equal, like unification of the stubbing API for void and returning methods. I really miss a healthy discussion of the pros and cons of both proposals. What problem of the past should be fixed, what about readabily and complexity of the API, is it possible to misuse the API, how are corner cases are handled?
Please don't get me wrong, I only want to ensure that mockito users get the best API possible. 

----

The *when/isInvokedWith/then* proposal
Primary goal: Provide a type safe API for Argument-Matchers and Answers that can be used to stub void and returning methods.
API:  `when(mock::method).isInvokedWith(anyInt()).thenAnswer((a) -> { ... });`

----

The *when-void/then* proposal
Primary goal: Keep the API to stub of void methods as close as possible to the well known when/then syntax. This is not intended as an alias to the exsiting when/then for returning methods. 
API:  `when(()->mock.voidMethod(anyInt())).thenAnswer((a) -> { ... })`
Note: when(VoidCall) return a VoidOngoingStubbing that lacks a thenReturn, so clients can't misuse this API and use answers that dosn't make sense in the contxt of a void stubbing.

Issue #822, opened 2016-12-14
Bogus NPE if any() is used with primitive type arguments


axos88:
Mockito: 2.3.0, Java 8

When mocking a method invocation, if any() is used instead of anyInt() for exampe in the .when method call, a bogus NPE is thrown. There is quite a lot of explanations to the root cause of that, but it does not contain this hint.

I would expect either a compilation error or a more explanatory exception.
Code: https://gist.github.com/axos88/163061ed82fcfdd9756d60d12eba8809

```
java.lang.NullPointerException
	at xxx.WtfTest.testCaseNPE(WtfTest.java:34)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.mockito.internal.runners.SilentJUnitRunner.run(SilentJUnitRunner.java:39)
	at org.mockito.internal.runners.StrictRunner.run(StrictRunner.java:39)
[...]

PR #830, opened 2016-12-16
Improves exception message in case a matcher is used for a primitive parameter

bric3:
This should help for cases like #822 

NPE can be raised by the JVM on the callsite of a matcher returning object like `any()`.
The thing is that the JVM does several thing like to allows a type to match a signature like `<T> T any()`, type inference with Java 8, cast and unboxing. However `any()` can only return `null` so the JVM raises an NPE on the cast opcode. Of course the problem isn't new but more possible due to Java 8 flexibility. That's why Mockito had for a long time the primitive.

When mockito is used with the junit runner or the junit rule there's a `InvalidUseOfMatchersException` raised **afer** the NPE with a message about misplaced matchers. This exception is raised because the matchers were not consumed by the mock since the NPE is raised first. So this exception is correct and can help the neophyte to discover the primitive variant like `anyInt()`.

### DIFF
--- a/src/main/java/org/mockito/internal/exceptions/Reporter.java
+++ b/src/main/java/org/mockito/internal/exceptions/Reporter.java
@@ -5,10 +5,32 @@
 
 package org.mockito.internal.exceptions;
 
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
 import org.mockito.exceptions.base.MockitoAssertionError;
 import org.mockito.exceptions.base.MockitoException;
-import org.mockito.exceptions.misusing.*;
-import org.mockito.exceptions.verification.*;
+import org.mockito.exceptions.misusing.CannotStubVoidMethodWithReturnValue;
+import org.mockito.exceptions.misusing.CannotVerifyStubOnlyMock;
+import org.mockito.exceptions.misusing.FriendlyReminderException;
+import org.mockito.exceptions.misusing.InvalidUseOfMatchersException;
+import org.mockito.exceptions.misusing.MissingMethodInvocationException;
+import org.mockito.exceptions.misusing.NotAMockException;
+import org.mockito.exceptions.misusing.NullInsteadOfMockException;
+import org.mockito.exceptions.misusing.PotentialStubbingProblem;
+import org.mockito.exceptions.misusing.UnfinishedStubbingException;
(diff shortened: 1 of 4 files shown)


src/test/java/org/mockitousage/matchers/InvalidUseOfMatchersTest.java
```diff
@@ -93,4 +98,35 @@ public void should_scream_when_Matchers_count_dont_match_parameter_count() {
...
+    }
+
+    @RunWith(MockitoJUnitRunner.class)
+    public static class ObjectMatcherMisuseOnPrimitiveSite {
+        @Test
+        public void fails_with_NPE() {
+            IMethods mock = Mockito.mock(IMethods.class);
+            doNothing().when(mock)
+                       .twoArgumentMethod(eq(73),
+                                          (Integer) any()); // <= Raise NPE on this call site
```
TimvdLippe: Oof indentation looks weird on this one. Nit: move this to the previous line for readibility

src/main/java/org/mockito/internal/exceptions/Reporter.java
```diff
@@ -515,6 +531,12 @@ public static MockitoException misplacedArgumentMatcher(List<LocalizedMatcher> l
                 "    doThrow(new RuntimeException()).when(mock).someVoidMethod(anyObject());",
                 "    verify(mock).someMethod(contains(\"foo\"))",
                 "",
+                "This message may appear after an NullPointerException if the last matcher is returning an object ",
+                "like any() but the stubbed method signature expect a primitive argument, in this case,",
+                "use primitive alternatives.",
+                "    when(mock.get(any())); // bad use, will raise NPE",
+                "    when(mock.get(anyInt())); // correct usage use",
```
axos88: grammar: correct usage / correct use

src/test/java/org/mockitousage/matchers/InvalidUseOfMatchersTest.java
bric3: I added the line breaks because the NPE is throw on the second matcher, not the first.

src/main/java/org/mockito/internal/exceptions/Reporter.java
bric3: Ok thx

szczepiq:
I love this change! The description of the PR is really good, I immediately was able to understand the use case. Mockito is great because we never stop improving the UX - the exception messages and such.

Big THANKS!

PR #831, opened 2016-12-16
Test clean up according to discussions for RememberingListeners

kobbikobb:
According to issue @719 we wanted to clean up the tests for RememberingListeners.

The tests should follow the same style and conventions.
Set up should not contain variables used by all the tests.

We previously had another PR for this in #825. That build was failing for some reason not obvious and had 1 merge commit that we wanted to remove. In an attempt to rebase and reset the history github lost the connection to the remote fork. Instead of spending time investigating I just created a new PR.

### DIFF
--- a/src/test/java/org/mockitousage/debugging/InvocationListenerCallbackTest.java
+++ b/src/test/java/org/mockitousage/debugging/InvocationListenerCallbackTest.java
@@ -4,6 +4,7 @@
  */
 package org.mockitousage.debugging;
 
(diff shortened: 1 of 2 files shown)


kobbikobb:
@bric3 @TimvdLippe Sorry for this new PR with the same code. Previously we were discussing #825. In an attempt to fix the PR github lost the link and I couldn't reconnect it. This PR is building and has only one commit message and no merge...

bric3:
Looks good to me. Thanks for the improvement.

PR #837, opened 2016-12-20
Disabled maven central sync

szczepiq:
I need to investigate:
 - release of 2.3.7 to central
 - errors logging in to oss.sonatype and syncing from Bintray

I suspect that maven central sync does not work at the moment, anyway due to login / credentials problem.


szpak:
There were some internal issues with Maven Central today: https://issues.sonatype.org/browse/MVNCENTRAL-1433

PR #839, opened 2016-12-22
Fixes #838 : Adds a matches(Pattern)

### LINKED ISSUE #838: Feature request: ArgumentMatchers.matches(Pattern regex)
you ladies and gentlemen are geniuses by the way.

ArgumentMatchers.match( String regex ) exists... but wouldn't it be slightly better if it were match( Pattern regex )? (i.e. catering for multi-line Strings, etc.).

Also, devising a regex which distinctly does NOT match a given String is non-trivial.  And yet not atypical of a mocking use case.  Hence I suggest you clever people take the heavy lifting away from grunts like me by offering ArgumentMatchers.doesNotMatch( Pattern regex ).

I posed this question today on Stack Overflow [here](http://stackoverflow.com/questions/41270738/mockito-argumentmatchers-doesnotmatch?noredirect=1#comment69745269_41270738).  

------

**EDIT 1** by @bric3 : removed the issue guidelines
**EDIT 2** by @bric3 : Renamed the issue title. Previous title was `Feature request: ArgumentMatchers.doesNotMatch( Pattern regex )`

bric3:
Following discussion in #838 it is sensible to add a mockito matcher that can take a `Pattern` as an argument.

The usage is the same as `matches(String)`.

```java
when(mock.oneArg(matches(Pattern.compile("[a-z]+\\d\\d")))).thenReturn(...);
```

Although it allows to build more complex regex, especially with regular expression flags.

### DIFF
--- a/src/main/java/org/mockito/ArgumentMatchers.java
+++ b/src/main/java/org/mockito/ArgumentMatchers.java
@@ -4,8 +4,6 @@
  */
 package org.mockito;
 
-import static org.mockito.internal.progress.ThreadSafeMockingProgress.mockingProgress;
-import static org.mockito.internal.util.Primitives.defaultValue;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


bric3:
@mockito/core Since the PR is fairly in scope amd fairly straightforward limited I'll merge it right away.

szczepiq:
>@mockito/core Since the PR is fairly in scope amd fairly straightforward limited I'll merge it right away.

Absolutely.

Updating the public API needs review from other team mates but in this case it's best to move fast. It's an overloaded method, very natural complement to the existing one. Thanks @bric3!

bric3:
Yes exactly :)

PR #842, opened 2016-12-23
Added missing Javadoc and few other tidy-ups

szczepiq:
1. No API / no behavior changes.
2. Publicly visible changes:
  - Added missing Javadoc and updated an existing one for MockitoException and MockitoAssertionError classes
3. Remaining changes, refactoring:
 - Rename job for consistency
 - Removed some unnecessary code
 - Made the tests more stable

### DIFF
--- a/src/main/java/org/mockito/internal/runners/RunnerFactory.java
+++ b/src/main/java/org/mockito/internal/runners/RunnerFactory.java
@@ -48,7 +48,11 @@ public class RunnerFactory {
      */
     public InternalRunner create(Class<?> klass, Supplier<MockitoTestListener> listenerSupplier) throws InvocationTargetException {
         try {
-            return new RunnerProvider().newInstance("org.mockito.internal.runners.DefaultInternalRunner", klass, listenerSupplier);
+            String runnerClassName = "org.mockito.internal.runners.DefaultInternalRunner";
+            //Warning: I'm using String literal on purpose!
+            //When JUnit is not on classpath, we want the code to throw exception here so that we can catch it
+            //If we statically link the class, we will get Error when class is loaded
+            return new RunnerProvider().newInstance(runnerClassName, klass, listenerSupplier);
         } catch (InvocationTargetException e) {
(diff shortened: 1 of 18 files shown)


src/main/java/org/mockito/internal/runners/RunnerFactory.java
```diff
@@ -48,7 +48,7 @@ public MockitoTestListener get() {
      */
     public InternalRunner create(Class<?> klass, Supplier<MockitoTestListener> listenerSupplier) throws InvocationTargetException {
         try {
-            return new RunnerProvider().newInstance("org.mockito.internal.runners.SilentJUnitRunner", klass, listenerSupplier);
+            return new RunnerProvider().newInstance("org.mockito.internal.runners.DefaultInternalRunner", klass, listenerSupplier);
```
bric3: Since the runner is in the same package as the factory I don't see why it should be a String ?

src/test/java/org/mockito/internal/runners/util/RunnerProviderTest.java
```diff
@@ -17,7 +17,7 @@ public void shouldCreateRunnerInstance() throws Throwable {
         //given
         RunnerProvider provider = new RunnerProvider();
         //when
-        RunnerImpl runner = provider.newInstance("org.mockito.internal.runners.SilentJUnitRunner", this.getClass(), null);
+        InternalRunner runner = provider.newInstance("org.mockito.internal.runners.DefaultInternalRunner", this.getClass(), null);
```
bric3: Since the runner is in the same package as the factory I don't see why it should be a String ?

szczepiq: Good point! Code updated!

bric3: OK, about the JUnit on the classpath, but will the `RunnerFactory` be used at all then ? The only reasonable use is in the `Silent` runner class which itself extends the JUnit `Runner` class.

Issue #840, opened 2016-12-22
Tweak JUnit rule strictness at the method level


szczepiq:
Team, I would really appreciate feedback about this enhancement.

There is a feature request reported - a way to exempt from unnecessary stubbing at method level (#792). The feature request makes sense to me. Currently, if the user wants to opt-out from JUnit rule or JUnit runner strictness, he needs to do it at the scope of the entire class. Ideally, we can opt-out from strictness at finer granularity level (method or even at mock level).

Inspired by @bric3 [idea](https://github.com/mockito/mockito/issues/792#issuecomment-265762478) I'd like to get your feedback about this API enhancement for strict JUnit rules.

Example test class:

```java
public class MyTest {
    //Mockito stubbing is strict in the entire test class
    @Rule public MockitoRule mockito = MockitoJUnit.rule().strictness(Strictness.STRICT_STUBS);

    @Test public void lenient_mockito() throws Throwable {
        //making Mockito lenient only for this test method
        //below API already exist but it does not have any effect inside of the test method
        mockito.strictness(Strictness.LENIENT);

        //rest of the test
    }
}
```

Pros:
 - we already have 'strictness' method in the MockitoRule API so it's not adding any new API
 - least surprise principle - users might expect this behavior to work but currently 'strictness' method _creates_ new instance of the rule and _not_ change the state of the rule.

Cons:
 - it's slightly not intuitive that 'strictness' method actually changes state (it's not a setter).

TimvdLippe:
Instead of a call to the rule (which I would expect to cause side-effects when I see the code) I think it is clearer to use an annotation on the method. Something like `@MockitoStrictness(LENIENT)`

szczepiq:
Annotations are clearer and they are a good idea, we discussed it in #792, @bric3 was not in favor of annotations.

Call to the rule is something we would offer in the meantime, before we have annotations. Long term, I think annotations are a way forward.

I'm waiting for more feedback and I'm leaning to implement this new rule behavior regardless if we decide to do annotations or not :)

bric3:
Yes I'm not in favor of annotation, but I'm not totally against either. However annotation API, is to be carefully designed as there is way less flexibility, override, argument types, etc.

Pros

* I believe a programtic way offers more choice for users and framework developers, including mockito.

> * least surprise principle - users might expect this behavior to work but currently 'strictness' method creates new instance of the rule and not change the state of the rule.

[...]

PR #843, opened 2016-12-24
JUnit rule strictness can be tweaked per test method

szczepiq:
See the design and rationale at #840

Please review!


src/main/java/org/mockito/internal/junit/DefaultStubbingLookupListener.java
```diff
@@ -0,0 +1,57 @@
...
+class DefaultStubbingLookupListener implements StubbingLookupListener {
+
+    Strictness currentStrictness;
+
+    DefaultStubbingLookupListener(Strictness strictness) {
+        this.currentStrictness = strictness;
+    }
+
+    public void onStubbingLookup(Invocation invocation, MatchableInvocation stubbingFound) {
+        if (currentStrictness != Strictness.STRICT_STUBS) {
```
TimvdLippe: Instead of passing the Strictness to this class, I think it is better to remove it from here and when this class is used add the conditional. E.g. `if (currentStrictness == Strictness.STRICT_STUBS)` then add this listener to the CreationSettings

szczepiq: Hey Tim, thanks for reviewing the code!

I need to add this listener during mock creation. After that, the strictness can be changed during test execution. If I conditionally add listener to the mock, I won't be able to modify the behavior during test execution. Can you elaborate more on your suggestion? I don't think it will work here.

szczepiq:
Any other feedback? I'd love to merge this change!

src/main/java/org/mockito/internal/junit/DefaultStubbingLookupListener.java
TimvdLippe: Ah okay I see, yeah then my suggestion does not work. Will see later if we run into troubles here.

szczepiq:
Merging tomorrow morning unless there's feedback. Thanks @TimvdLippe for review!

PR #846, opened 2016-12-26
Fix #845

### LINKED ISSUE #845: Rename MockMethodDispatcher.class to MockMethodDispatcher.raw on build
The `org.mockito.internal.creation.bytebuddy.MockMethodDispatcher.class` file should be renamed to `org.mockito.internal.creation.bytebuddy.MockMethodDispatcher.raw` on building Mockito. This way, class loaders are no longer capable of loading the class directly what should never happen. Doing so, the bootstrap injection mechanism cannot be superseeded by child-first class loaders which break mockability due to the dispatcher class being loaded twice by two different class loaders.

This problem was observed with the Robolectrics class loader.

Additionally, the following change is required in the `InlineByteBuddyMockMaker` when creating the boot jar

```java
String source = "org/mockito/internal/creation/bytebuddy/MockMethodDispatcher";
outputStream.putNextEntry(new JarEntry(source + ".class"));
InputStream inputStream = InlineByteBuddyMockMaker.class.getClassLoader().getResourceAsStream(source + ".raw");
if (inputStream == null) {
  throw new MockitoInitializationException(join(
    "The MockMethodDispatcher class file could not be located " + source,
    "",
    "The class loader responsible: " + InlineByteBuddyMockMaker.class.getClassLoader()
  ));
}
try {
  int length;
  byte[] buffer = new byte[1024];
  while ((length = inputStream.read(buffer)) != -1) {
    outputStream.write(buffer, 0, length);
  }
} finally {
  inputStream.close();
}
outputStream.closeEntry();
```

cemelo:
This change renames `org/mockito/internal/creation/bytebuddy/MockMethodDispatcher.class` to `org/mockito/internal/creation/bytebuddy/MockMethodDispatcher.raw` after compilation of the root project, more precisely, after the `classes` task is done executing.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -74,6 +74,12 @@ apply from: 'gradle/mockito-core/javadoc.gradle'
 apply from: 'gradle/mockito-core/license.gradle'
 apply from: 'gradle/mockito-core/testing.gradle'
 
+classes.doLast {
+    java.nio.file.Files.move(
+        file("${buildDir}/classes/main/org/mockito/internal/creation/bytebuddy/MockMethodDispatcher.class").toPath(),
+        file("${buildDir}/classes/main/org/mockito/internal/creation/bytebuddy/MockMethodDispatcher.raw").toPath())
+}
+
 task wrapper(type: Wrapper) {
     gradleVersion = '2.14.1'
 }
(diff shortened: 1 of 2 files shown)


raphw:
Tracked by #847 (Your committer tag remains intact.)

PR #847, opened 2016-12-26
Rename compiled MockMethodDispatcher class so that it works with Robolectric

raphw:
Adresses #845.


raphw:
Outstanding issue: My IDE is no longer capable to process the project as it relies on identifying the class file for the `MockMethodDispatcher` class. Any idea how we work around this? Maybe we should duplicate the file during compilation and only delete the original class file during creating the jar?

Or we should only delete the class file when a specific build profile is run.

cemelo:
@raphw I think you could just copy it after compilation in order to test and tweak the jar task to filter the original resource.

raphw:
As long as this happens prior to testing, this could work. Could you make a code suggestion?

cemelo:
Sure. I'm away from the computer atm, but I can send something in an hour or so. Is that ok?

raphw:
Of course, thank you so much for helping out!

cemelo:
@raphw, no problem! I sent a new pull request where I created a different file so we don't polute the main `build.gradle`. I create one task to copy the resource and configured the jar task to exclude the original `.class` file from the final distribution.

gradle/mockito-core/inline-mock.gradle
```diff
@@ -0,0 +1,13 @@
+task copyMockMethodDispatcher(type: Copy) {
+    from "${buildDir}/classes/main/org/mockito/internal/creation/bytebuddy"
+    into "${buildDir}/classes/main/org/mockito/internal/creation/bytebuddy"
+    include "MockMethodDispatcher.class"
+    rename { String fileName ->
+        fileName.replace('.class', '.raw')
+    }
+}
+classes.dependsOn(copyMockMethodDispatcher)
```
szczepiq: In order for this to work reliably, we need to also set one more task dependency on compileJava. Let me try to update the code in the branch...

szczepiq: Above change makes the compiled classes directory contain both: the 'raw' and normal class. Is this what we want?

raphw: Yes, for the time until packaging the jar. This is required to make IDEs work with Mockito which do otherwise claim that there is a class missing (experienced with IntellIJ). It does not hurt us as we know that our unit tests are run by a parent-first class loader where class loading works as expected. This is only an issue when dealing with parent-first loaders where the `.class` file must not exist.

raphw:
This is a file that must go into the bootstrap class loader and to avoid that regular class loaders attempt to load it, we renamed its extension.

suztomo:
Thank you for quick response.

PR #849, opened 2016-12-26
Copies compiled MockMethodDispatcher class and excludes the original file from the jar distribution.

cemelo:
Copies the `MockMethodDispatcher.class` to `MockMethodDispatcher.raw` without deleting the former so IDEs can still recognize the compiled file, and excludes it from the final JAR distribution.


raphw:
Incorporated in #847. Thanks a lot!

PR #850, opened 2016-12-27
Stopped publishing java source inside main jar

szczepiq:
This feature does not make sense in modern world of declarative dependency management:
 - we didn't hear from the community that keeping the sources inside jar is useful
 - modern IDEs and dependency management automatically pulls sources
 - the feature was originally needed (2008) to help users that check in jars into VCS and do not use modern dependency management
 - keeping source code in main jar is very unusual and makes the jar bigger

Bumped minor version because the change is relatively visible

### DIFF
--- a/version.properties
+++ b/version.properties
@@ -1,2 +1,2 @@
-version=2.4.6
+version=2.5.0
 mockito.testng.version=1.0
--- a/gradle/publishable-java-library.gradle
+++ b/gradle/publishable-java-library.gradle
@@ -13,11 +13,8 @@ def licenseFiles = copySpec {
     from(".") { include 'LICENSE' }
 }
 
-//TODO why do we still ship the main jar with sources?
 jar {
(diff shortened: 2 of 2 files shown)


version.properties
```diff
@@ -1,2 +1,2 @@
-version=2.4.4
+version=2.5.0
```
szpak: I assume this is put in the PR intentionally.

szpak: Nevermind, it is mentioned in the PR desc.

raphw:
LGTM!

TimvdLippe:
@szczepiq Instead of merging, please rebase this PR. That gives a cleaner git history and is easier to review for us.

szczepiq:
>@szczepiq Instead of merging, please rebase this PR. That gives a cleaner git history and is easier to review for us.

Will do. I merged via GitHub web UI, rebase on the way...

PR #853, opened 2016-12-27
Update master with release/2.x

TimvdLippe:
Quite a large number of commits again, but most of them are the CD-drone and @szczepiq with refactoring commits. We have to take a good look at the changes to the build file and make sure no changes on master were accidentally reverted.


TimvdLippe:
CI failed with:

```
org.mockitoutil.ClassLoadersTest > can_run_in_given_classloader FAILED
    java.lang.IllegalStateException: Given task could not be loaded properly in the given classloader 'org.mockitoutil.ClassLoadersTest$1@7b08f2f6', error 'java.lang.IllegalAccessException: Class org.mockitoutil.ClassLoaders$ClassLoaderExecutor can not access a member of class org.mockitoutil.ClassLoadersTest$1 with modifiers "private static transient"
```

szpak:
I would suspect byte-buddy upgrade, however it fails also with Java 8 what works fine in the 2.x branch...

szpak:
I've found the reason. Jacoco 0.7.8 breaks tests for the new plugin mechanism (it works fine with 0.7.6 and 0.7.7). Reported #860.

Closing this PR as it is superseded by #861.

TimvdLippe:
@szpak Jacoco 0.7.8 is required for Java 9, so we cant just downgrade.

TimvdLippe:
Oh nvm spoke too soon. Per above comment I thought you downgraded, but I see you used a different fix :)

Issue #851, opened 2016-12-27
Performance degradation when upgrading from 1.10 to 2.2 


theigl:
Hi,

I just ran into an issue when I upgraded our codebase from Mockito 1.10 to 2.2. Test execution time nearly doubled for unit tests using Mockito Mocks and Spies. We have around 4000 unit tests, so this increase was very noticeable on my development box as well as our CI server.

I couldn't believe it at first, but I did extensive profiling with Yourkit and significantly more time is spent in Mockito (and ByteBuddy) classes compared to 1.10.

Before I try to create simple sample projects to reproduce the issue, I wanted to make sure nobody else has run into this and has filed a ticket already.

- My development setup is: MacOs, Java 1.8.0_66
- Our CI setup is: Ubuntu 14.04.4, 1.8.0_101

Tests were run with Maven Surefire and with IntelliJ.

Any feedback would be greatly appreciated.

Best regards,

Thomas

TimvdLippe:
Hello Thomas,

This is the very first report we received. We are looking forward to an example project which reproduces the issue!

theigl:
Hi @TimvdLippe,

I created two extremely simple Maven projects. One with Mockito 1.10.19 and one wich 2.3.11. One empty POJO, and one test class that simply mocks the POJO.

When I run the projects with `mvn test` repeatedly on my box, the execution times converge around:

Mockito 1.10 | Mockito 2.3
------------ | -------------
180ms | 360ms

E.g.
`Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.181 sec` vs:
`Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.383 sec`


[mockito-performance-tests.zip](https://github.com/mockito/mockito/files/674105/mockito-performance-tests.zip)

szczepiq:
Thank you very much for reporting the issue!

TimvdLippe:
Opened #856 to start a benchmarking suite so we can investigate this. The results on my laptop:

```
:jmh
# JMH 1.17.3 (released 14 days ago)
# VM version: JDK 1.8.0_111, VM 25.111-b14
# VM invoker: /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
[...]

PR #856, opened 2016-12-29
Add gradle benchmark task to test performance of Mockito

TimvdLippe:
You can run this benchmark with `./gradlew jmh`. This adds a very simple benchmark suite which we can extend later too if we want to.

Related to #851


TimvdLippe:
@raphw I have updated the benchmarks. The first benchmark is correct I think (with the custom classloader), however the second one isn't. The reason being that I can't seem to cast to the correct class after invoking `loader.loadClass()`. It seems that the cast has a different `Class` than what the loader returns, even though they are both `IMethods.class`

raphw:
A class is only if its the same type name and loaded by the same class loader. You have to use reflection for such interaction or define a common super interface; maybe create a subclass of `IMethods` (using ByteBuddy), load this class in a different class loader (`ClassLoadingStrategy.Default.WRAPPER`) and create a mock of this class.

TimvdLippe:
@raphw I tried to turn your comment into code, but sadly it is still happening. It complains that the type of the class returned by `Mockito.mock` is not a subclass of `IMethods`.

TimvdLippe:
I am unable to progress on this PR as I clearly lack the experience and knowledge. Could someone maybe help out? :heart:

TimvdLippe:
Can't seem to get this to work. Closing the PR for now, but we can leave the branch alive.

PR #858, opened 2016-12-30
Safeguard for MockitoListener API

szczepiq:
Reminds user if he forgets about "removeListener()" when working with [MockitoListener API](https://static.javadoc.io/org.mockito/mockito-core/2.5.0/org/mockito/MockitoFramework.html#addListener(org.mockito.listeners.MockitoListener)).

- Added a safeguard that reminds users to clean up MockitoListeners via removeListener() method. The implementation is very simple, checking if the type of the listener is already added. I found out about this problem when working on strict stubbing support without JUnit runner.
- This change makes MockitoListener API safer. It's easy to leak state which we have encountered in our own test suite.
- If we encounter legit use cases for adding the same listener type, we can remove the safeguard or updated it

The PR also contains a couple of refactorings and tidy-ups. I hope it's ok.


TimvdLippe:
What if two libraries are used which add the same listener for their respective usecase? I dont think we can make this change, eventually someone will hit this.

szczepiq:
@TimvdLippe, thanks for feedback!

>What if two libraries are used which add the same listener for their respective usecase?

How is it possible? Can you elaborate this scenario?

TimvdLippe:
@szczepiq Let's say a user has both `Spring` and `LibraryX` as dependency. They both add a `MockCreationListener` to the framework. Whichever dependency was loaded last will have an exception thrown. I do not think this is a change we should pursue.

szczepiq:
This use case should work ok. Unless they add exactly the same implementation class which I dont think is realistic or useful.

TimvdLippe:
@szczepiq Ah okay, I see now. Then this seems okay for now, still wondering if this has any negative impact in a certain situation.

szczepiq:
@TimvdLippe, thanks for reviewing!

I'd say we check this in.

1. We can always change/remove this safeguard. It will be a compatible change.
2. There is an easy workaround for an unanticipated use case: create a subclass / delegating implementation to bypass the check

How does it sound?

TimvdLippe:
Yes LGTM

src/test/java/org/mockitoutil/ThrowableAssert.java
```diff
@@ -0,0 +1,43 @@
+package org.mockitoutil;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Clean asserts for exception handling
+ */
+public class ThrowableAssert {
```
bric3: This can be removed when we upgrade to AssertJ 3 (but it requires Java 8)

szczepiq: Great idea!

Issue #297, opened 2015-10-04
Mockito java8 distribution


mockitoguy:
There's an open source java8 feature set for mockito developed by Marcin Zajączkowski, see: https://solidsoft.wordpress.com/2015/06/30/more-compact-mockito-with-java-8-lambda-expressions-and-mockito-java8-add-ons

Let's figure out a way to include this in main distribution of Mockito

Discussions on this subject on the dev list: 

1) https://groups.google.com/forum/#!searchin/mockito-dev/java8/mockito-dev/pBb0Pv6x_7E/Sz8WIew6OK8J
2) https://groups.google.com/forum/#!searchin/mockito-dev/bytebuddy$20plan/mockito-dev/dG2KltOz-gQ/VG4B5wcRL0AJ

This story needs to be scoped out. Main challenge: how do we offer java8 features with Mockito distribution and yet provide compatibility to users on java version < 8

JeremybellEU:
Not sure if it is possible, but could Java 8 specific features be directed to a seperate module. Therefore the 'old' module supports Java <8 and the new + old module support Java >8. I am not even sure if it is able to be compiled.

szpak:
The current features can be (and in fact are) simply integrated as a separate mockito-java8 module.

Although I cannot provide a particular feature in Mockito a case study from AssertJ. Long time ago I started implementing AssertJ module for Java 8. It turned up quickly that due to complex class hierarchy of assertions it was not possible to add a new class Assertions8 which depends on a standard Java 7 Assertions class and also allows common subclasses from Java 7 to rely on it (Assertions8). Everything ended up as a separate branch with different version number (assertj-core 2.x and 3.x).

Mockito has quite different internal structure, but in my opinion the key question is of there are known features which woudn't be able to be implemented as an add-on module? What API limitation would a separate module approach bring?

szczepiq:
@szpak, can we have mockito-java8 __1.0__? There seems to be no big changes going on, you already are backwards compatible as you know well enough the power of server, lets do 1.0. This way, we can deemphasize java8 support.

I guess the the goal of this ticket is to pull in @szpak module into mockito codebase (and bring new fantastic team member along ;p). Since mockito-java8 is can be neatly used right now, there is no urgency on this.

bric3:
I definately would like to have java 8 only distribution. If you look at java 9 big changes this will make sense due to the modular system where mockito will probably need to be able to work with the module system [JEP 261](http://openjdk.java.net/jeps/261).

[...]

PR #859, opened 2016-12-30
[#297] Add AssertionMatcher and LambdaMatcher

szpak:
The matchers leverages Java 8 to provide inlined ArgumentCaptor with a lambda expression and inlined argument matcher with a predicate lambda expression. More information below and in JavaDoc.

### ~Lambda matcher~

~Allows to define matcher logic within a lambda expression. Useful when working with complex classes pass as an argument.~

Lambda matcher has been removed from the PR as it can be implemented with the enhancements introduced in the master branch in the meanwhile.

### Argument Captor - Java 8 edition

Allows to use ArgumentCaptor in one line:

```
@Test
public void shouldAllowToUseAssertionInLambda() {
    //when
    ts.findNumberOfShipsInRangeByCriteria(searchCriteria);
    //then
    verify(ts).findNumberOfShipsInRangeByCriteria(
        assertArg(sc -> assertThat(sc.getMinimumRange()).isLessThan(2000)));
}
```

In comparison to 3 lines in the classic way:

```
@Test
public void shouldAllowToUseArgumentCaptorInClassicWay() {  //old way
    //when
    ts.findNumberOfShipsInRangeByCriteria(searchCriteria);
    //then
    ArgumentCaptor<ShipSearchCriteria> captor = 
        ArgumentCaptor.forClass(ShipSearchCriteria.class);
    verify(ts).findNumberOfShipsInRangeByCriteria(captor.capture());
    assertThat(captor.getValue().getMinimumRange()).isLessThan(2000);
}
```

Backported from https://github.com/szpak/mockito-java8/
Related to #297.

Btw, I'm not sure if `Primitives` is the best name for a class with `defaultValue()` working also with non primitives classes. I probably liked `HandyReturnValues` more.


szpak:
typetools fails with Java 9. Reported an [issue](https://github.com/jhalterman/typetools/issues/34 in ) for that.

ChristianSchwarz:
Can you add some sample code to the description? Can you explain the benefit of the new matcher types, why not use a plain ArgumentMatcher?

szpak:
There are examples in JavaDoc and on the project webpage: https://github.com/szpak/mockito-java8#provided-add-ons . If you would I can paste them inline.

In short, the syntax is usually much shorter.

src/main/java/org/mockito/internal/matchers/AssertionMatcher.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public AssertionMatcher(Consumer<T> assertingLambda) {
+        this.assertingLambda = assertingLambda;
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public boolean matches(Object argument) {
+        try {
+            assertingLambda.accept((T) argument);
```
TimvdLippe: Casting here feels wrong. Can we get rid of this?

build.gradle
```diff
@@ -75,6 +75,7 @@ dependencies {
 
     provided "junit:junit:4.12", "org.hamcrest:hamcrest-core:1.3"
     compile "org.objenesis:objenesis:2.4"
+    compile 'net.jodah:typetools:0.4.8'
```
TimvdLippe: Is this library really required, would like to not add a library for 1 method invocation.

src/main/java/org/mockito/internal/matchers/LambdaMatcher.java
```diff
@@ -0,0 +1,103 @@
...
+
+    public LambdaMatcher(Predicate<T> predicate, String description) {
+        this.predicate = predicate;
+        this.description = description;
+    }
+
+    @Override
+    public boolean matches(T item) {
+        try {
+            return predicate.test((T)item);
```
TimvdLippe: Also this cast, this should work just fine right?

build.gradle
szczepiq: Absolutely. No extra libraries please, especially pre 1.0 (no compatibility guarantee).

szczepiq:
Thanks a lot @szpak for pushing this forward!

ArgumentMatcher interface is already a predicate and it can be used as lambda. Why do we need an explicit argLambda method?

Is it possible / good idea to overload argThat() method instead of creating new API methods? (discoverability, clean API).

build.gradle
szpak: Looking at the [code](https://github.com/jhalterman/typetools/blob/master/src/main/java/net/jodah/typetools/TypeResolver.java) there seem to be some corner cases in resolving return types.

In this code it is needed to determine a return type of lambda implementing `Consumer` interface. @raphw Maybe you have some code in ByteBuddy which could be used to do that?

szpak:
I simplified the code and removed `argLambda` and `LambdaMatcher`. I created a new wrapper matcher to be able to provide custom message/description in a case of failure. However, I'm not very satisfied with the name `DescriptableWrapperMatcher`. Do you have any better proposals?

src/main/java/org/mockito/internal/matchers/DescriptableWrapperMatcher.java
```diff
@@ -0,0 +1,55 @@
...
+ * }
+ * </code></pre>
+ *
+ * @param <T> type of argument
+ *
+ * @author Marcin Zajączkowski
+ * @since 3.0.0
+ */
+@Incubating
+public class DescriptableWrapperMatcher<T> implements ArgumentMatcher<T> {
```
ChristianSchwarz: @szpak Do we need this class? An other option is to provide an addtional constructor in AssertionMatcher with description parameter.

ChristianSchwarz:
@szpak A more general question about this PR: Why should one favor  `assertArg` over an `argThat`. Is the main purpose better readabilty of the error message in case a lamba ArgumentMatcher is used?

```java
//assertArg
verify(ts).findNumberOfShips( assertArg(sc -> assertThat(sc.getMinimumRange()).isLessThan(2000)));

//argThat 
verify(ts).findNumberOfShips( argThat(sc -> sc.getMinimumRange()<2000));
//                                        ^ easier to read condition, cause it is short
```

szpak:
@ChristianSchwarz Because of the readability in a case of error (assertion failure). AssertJ generates very meaningful error message (the same as in the ArgumentCaptor used in the old 3-line-long style). With Lambda you only know that "condition was not fulfilled".

TimvdLippe:
Instead of relying on run-time type-resolution I would rather solve this problem with a solution that is type-safe. I can probably implement this feature in #742 which should be type-safe?

szpak:
@TimvdLippe That's great and it could be another option (or even may deprecate this one in the future), however, it's type safe. It's just a default "empty" value to be returned (to not get runtime error) which is not use externally.

src/main/java/org/mockito/internal/matchers/DescriptableWrapperMatcher.java
szpak: I don't see a way to use `AssertionMatcher` here as it accepts a `Cunsumer` while in removed `LambdaMatcher` a `Predicate` argument was required. I simplified it to use just `ArgumentMatcher` interface, however, it doesn't provide a way to store a description (it is just a function argument with one `matches` method (which is equivalent to a `Predicate`).

szczepiq:
>@ChristianSchwarz Because of the readability in a case of error (assertion failure). AssertJ generates very meaningful error message (the same as in the ArgumentCaptor used in the old 3-line-long style). With Lambda you only know that "condition was not fulfilled".

Good use case. +1

@szpak, why new assertArg() method and not new overload of argThat()?

szpak:
@szczepiq As I already replied to Christian:

> ChristianSchwarz Because of the readability in a case of error (assertion failure). AssertJ generates > very meaningful error message (the same as in the ArgumentCaptor used in the old 3-line-long style). > With Lambda you only know that "condition was not fulfilled".

ChristianSchwarz:
@szczepiq  
overloading with argThat(Consumer<T>) would be nice, but it doesn't work well with lambdas cause the compiler don't like ambiguity.

szpak:
I limited the compatibility of AssertionMatcher to Java 8.

Regarding typetools as a dependency I was considering a repackage of it, however it uses `sun.reflect.ConstantPool` which would lead to compilation errors with Java 9. However, it is just one class with no transitive dependencies. Would it be really be such a problem to use it?

TimvdLippe:
I am not in favor of merging this particular implementation of lambda matchers. First of all it requires an extra dependency for this one method, secondly we have no guarantee that it actually works all of the time (no type-safety). I would prefer to create a solution that does not have these two problems, then we can support lambda matchers safely :smile:

szpak:
> secondly we have no guarantee that it actually works all of the time (no type-safety)

What do you exactly mean?

TimvdLippe:
@szpak [This implementation](https://github.com/mockito/mockito/pull/859/files#diff-ded7a7ed5931f11027ce4f6f43d648f6R80) is not type-safe and relies on run-time type retrieval. As there are bugs with the dependency and therefore type resolution, we have no guarantee we won't run into problems with users.

szpak:
I reworked my PR to remove external dependency (one class has been repackaged) and to be Java 9 compatible.

@TimvdLippe The return type of `Consumer<T>` AFAIK cannot be determined in a simpler way. It seems to work (even with Java 9) and even in a case of problems in the future only that one feature would be affected (which can be rewritten on the user side to the classic 3 lines long ArgumentCapture (instead of just one with my solution)). For me it is good to give users a try. This is still an incubating feature.

ChristianSchwarz:
@all 
What about renaming `assertArg` to a more generic name?  The interface `Consume<T>` allows to do even more than assertions. 

Some examples:

* Argument manipulation: `when(mock).method(assertArg(queue ->queue.poll ));`
* Argument logging: `verify(mock).method(assertArg(a -> System.out.println(a)));`
* Argument capuring without captor: `verify(mock).method(assertArg(a -> list.add(a)));`

src/test/java/org/mockito/internal/matchers/AssertionMatcherTest.java
```diff
@@ -0,0 +1,89 @@
...
+    @Rule
+    public MockitoRule mockitoRule = MockitoJUnit.rule();
+
+    @Mock
+    private TacticalStation ts;
+
+    private ShipSearchCriteria searchCriteria = new ShipSearchCriteria(1000, 4);
+
+    @Test
+    public void shouldAllowToUseArgumentCaptorInClassicWay() {
```
ChristianSchwarz: This test can be removed ArgumentCaptors are tested in an other test class.

src/test/java/org/mockito/internal/matchers/AssertionMatcherTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+public class AssertionMatcherTest implements WithAssertions {
+
+    @Rule
+    public MockitoRule mockitoRule = MockitoJUnit.rule();
+
+    @Mock
+    private TacticalStation ts;
+
+    private ShipSearchCriteria searchCriteria = new ShipSearchCriteria(1000, 4);
```
ChristianSchwarz: The class `TacticalStation` and  `ShipSearchCriteria` add complexity that isn't needed to test the feature. It should be possible to test assertArg() with a well known interface like `Consumer` or `Function` and a plain `String` or `int` parameter. WDYT?

ChristianSchwarz:
@szpak 

> The return type of Consumer<T> AFAIK cannot be determined in a simpler way.

Maybe this existing solution works:  [GenericTypeExtractor.genericTypeOf(..)](https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/internal/util/reflection/GenericTypeExtractor.java)

src/main/java/org/mockito/internal/util/Primitives.java
```diff
@@ -60,6 +63,17 @@ public static boolean isAssignableFromWrapper(Class<?> valueClass, Class<?> refe
...
 
+    /**
+     * Returns the boxed default value for the lambda return type (handling primitive and primitive wrapper types).
+     *
+     * @param consumer the consumer function
+     * @param <T> the function's return type
+     * @return the boxed default value for the lambda return type (handling primitive and primitive wrapper types)
+     */
+    public static <T> T defaultValueForConsumerLambda(Consumer<T> consumer) {
+        Class<?>[] typeArgs = TypeResolver.resolveRawArguments(Consumer.class, consumer.getClass());
```
ChristianSchwarz: > The return type of Consumer<T> AFAIK cannot be determined in a simpler way.

Maybe the existing [GenericTypeExtractor](https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/internal/util/reflection/GenericTypeExtractor.java) works.

```java
Class<?>[] typeArgs =GenericTypeExtractor.genericTypeOf(consumer.getClass(), Object.class, Consumer.class)
```

build.gradle
bric3: Can't we augment our `GenericMetadataSupport` tool ?

src/main/java/org/mockito/internal/util/Primitives.java
szpak: Thanks for your suggestion. I checked it and after some modifications (interfaces are considered after classes) I was able to make it work for an anonymous class. Unfortunately it doesn't work with a lambda (Consumer is "an instance of" `ParameterizedType` which is the only supported type).

I've checked also `GenericMetadataSupport` suggested by @bric3, however, it works also only for anonymous classes.

There is probably a possibility to enhance Mockito's internal mechanisms to support both lambdas and Java 9 - see https://github.com/jhalterman/typetools/pull/35/files .

src/test/java/org/mockito/internal/matchers/AssertionMatcherTest.java
szpak: In my project I use those two classes also for some other use cases and in fact it can be too complex for just `AssertionMatcherTest`.

szczepiq:
>@szczepiq overloading with argThat(Consumer<T>) would be nice, but it doesn't work well with lambdas cause the compiler don't like ambiguity.

@ChristianSchwarz, thank you for explaining it. Makes prefect sense now.

>@all
>What about renaming assertArg to a more generic name? The interface Consumer<T> allows to do >even more than assertions.

Absolutely! assertArg is nicely expressive but would look awkward if used not for assertions (and could be even confusing). How about one of the following new methods:

```java
//with
when(mock).method(with(queue ->queue.poll));

//arg
when(mock).method(arg(queue ->queue.poll));

//withArg
when(mock).method(withArg(queue ->queue.poll));
```

src/main/java/org/mockito/internal/util/Primitives.java
ChristianSchwarz: Maybe @raphw can help here, probably byte-buddy has similar utils.

raphw: Byte Buddy has the `TypeDescription` abstraction which binds type variables down the hierarchy. You can simply navigate down the hierarchy to a given type and then query generic values which are resolved on demand. If this works, I can probably help you to go further from there. Let me know.

ChristianSchwarz: @raphw How to get an `TypeDescription` for a Class-Object? I looked into the byte-buddy code but  don't get it.

ChristianSchwarz:
@szczepiq 
> How about one of the following new methods:

- `arg(..)`  feels too close to `argThat(..)` maybe it is confusing for users
- `with(..)` and `withArg(..)` looks both good to me, I may favor the shorter name

src/main/java/org/mockito/internal/util/Primitives.java
raphw: By `TypeDescription.ForLoadedType`.

TimvdLippe:
After reconsideration I am still convinced that the current solution is not good enough. Especially no guaranteed type-safety is a blocker for me. I would therefore like to hold off until we can find a solution that is guaranteed type-safe. I will check if we can have a type-safe solution with #742 where we have the opportunity to redesign these features. WDYT?

szpak:
I'm not convinced it is a major issue, but on the other hand I don't want to thrust it forward at all cost. Closing the PR.

PR #864, opened 2017-01-02
Updated to Byte Buddy 1.6.0

raphw:
Updated Byte Buddy, several performance and API improvements. Also with much better Android support.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,8 +45,8 @@ apply from: 'gradle/root/version.gradle'
 apply from: "gradle/java-library.gradle"
 
 dependencies {
-    compile 'net.bytebuddy:byte-buddy:1.5.12'
-    compile 'net.bytebuddy:byte-buddy-agent:1.5.12'
+    compile 'net.bytebuddy:byte-buddy:1.6.0'
+    compile 'net.bytebuddy:byte-buddy-agent:1.6.0'
 
     provided "junit:junit:4.12", "org.hamcrest:hamcrest-core:1.3"
     compile "org.objenesis:objenesis:2.4"
--- a/gradle/mockito-core/osgi.gradle
+++ b/gradle/mockito-core/osgi.gradle
@@ -16,7 +16,7 @@ afterEvaluate {
                     "org.mockito.*;version=${version}"
 
             instruction 'Import-Package',
-                    'net.bytebuddy.*;version="[1.5.5,2.0)"',
+                    'net.bytebuddy.*;version="[1.6.0,2.0)"',
                     'junit.*;resolution:=optional',
                     'org.junit.*;resolution:=optional',
                     'org.hamcrest;resolution:=optional',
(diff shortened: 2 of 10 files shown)


bric3:
Can you update `osgi.gradle` too ;)

bric3:
By the way here's [release notes of BB 1.6.0](https://github.com/raphw/byte-buddy/blob/master/release-notes.md#2-january-2017-version-160)

### 2. January 2017: version 1.6.0

* Added `InjectingClassLoader` with class loading strategy that allows for reflection-free loading.
* Added proper class loader locking to injection strategy.
* Fixed method lookup to not use declared accessors unless necessary to avoid security manager check.
* Added `@SuperMethod` and `@DefaultMethod` annotations for `MethodDelegation`.
* Refactored `AsmVisitorWrapper` to accept a list of fields and methods that are intercepted. This allows to use the wrapper also for methods that are overridden.
* Added a `MethodGraph.Compiler.ForDeclaredMethods` to avoid processing full type hierarchy if only type enhancement should be done without declaring new methods on a type. This should be used in combination with `Advice` instead of `MethodGraph.Empty` as those methods are supplied to the ASM visitor wrappers.
* Refactored `MethodDelegation` to precomile records for all candidates to avoid duplicate annotation processing.

### 29. December 2016: version 1.5.13

* Updates to ASM 5.2
* Updates Android DX-maker.
* Adds API to `MultipleParentClassLoader` to use other loader than bootstrap loader as a parent which is not always legal, e.g. on Android.
* Make `ClassInjector` use official class loading lock if one is available.
* Make `ClassInjector` use `getDefinedPackage` instead of `getPackage` if available.
* Declare UNIX socket library as provided in Byte Buddy agent to only add the dependency on demand.

raphw:
Good catch, a good fraction of the changes target Java 9 class loading; cglib and Javassist do not currently work on Jigsaw for the class loading. Byte Buddy also needed to add an unfortunate fallback on sun.misc.Unsafe for the moment.

bric3:
That's a nice improvement anyway !

raphw:
Mockito now also uses the built-in Byte Buddy type cache which is a bit more efficient.

TimvdLippe:
Nvm, tests do not pass, I looked at the wrong status indicator :sob:

bric3:
@raphw What was the bug with the serialization _feature_ in the type cache ?

raphw:
The first created type would be returned as the cache result, independently of the required serialization feaure.

raphw:
Tests pass again, had to unwrap an exception.

TimvdLippe:
This is fine by me. If @bric3 agrees too he/you can merge :smile:

bric3:
I don't think that was the case, because the cache made use of a MockKey that tested equality in class and interfaces, and if the mock was serializable, [the code added the `Serializable` interface](https://github.com/bric3/mockito/blob/release/2.x/src/main/java/org/mockito/internal/creation/MockSettingsImpl.java#L185-L191). I believe there's tests for that.

However I agree this wasn't really straightforward and could break in some way after refactoring.

Anyway the new type cache is better !

Issue #857, opened 2016-12-30
Strictness without JUnit - New MockitoSession API


szczepiq:
### Background

- Mockito strictness (improved debuggability, cleaner tests) is described in #769
- Strict stubbing is already supported with JUnit rules - #807
- Open PR for supporting strict stubbing with JUnit Runners - #854
- This ticket intends to discuss the goals and implementation of strict stubbing without JUnit Rules / Runner

### Initial design

```java
public class SampleTest {   
    @Mock IMethods mock;
    
    //Initialize mocks similar to MockitoAnnotations.initMocks(this), also configures strictness
    MockitoSession mockito = Mockito.mockitoSession().initMocks(this).strictness(Strictness.STRICT_STUBS).startMocking();

    @After public void after() {
        //After the test has completed, we detect unused stubs and validate mockito usage
        mockito.finishMocking();
    }

    //start and finish mocking invocations can be hidden in a common test base class
    //or abstracted out in unit test framework specific way (custom rule, etc.)

    @Test public void some_test() {
        //...
    }
}    
```

### Alternative new public API methods

🚀 Feedback needed! 🚀

```java
//1
MockitoMocking mockito = Mockito.startMocking(this, Strictness.STRICT_STUBS);
mockito.finishMocking();
//2
MockitoSession mockito = Mockito.mockitoSession().initMocks(this).strictness(Strictness.STRICT_STUBS).startMocking();
mockito.finishMocking();
//3
MockitoSession mockito = MockitoSessionBuilder.mockitoSession().testClassInstance(this).strictness(Strictness.STRICT_STUBS).start();
mockito.stop();
```

### Discussion

- other ideas to meet the goal (better debuggability / cleaner tests without JUnit)
- good naming / API for the feature
- consistency with existing support for strictness with Runner and the Rule

We would love to hear your feedback!

bric3:
Here's my opinion :

* Regarding alternative 1/2/3, I prefer the API with an actual variable. Using static methods requires a hidden thread local state, while it's ok for stubbing, I think this could be bad design choice for the mockiting session. Also without `MockitoSession` variable it's hard to interact with, and always requires some static method.
* Not in favor of option 1, as it kinda dismiss the builder style API. And it's probably harder to evolve such API in a non breaking way.
* Not in favor of option 3, regarding the `testClassInstance` as it does not explain what it does with the test class instance. While `initMocks` suggests the behavior behind this.
[...]

PR #865, opened 2017-01-03
New strict stubbing API - MockitoSession

szczepiq:
See proposed design at #857

The code is reviewable, especially new public API. Please give feedback! More work is pending.

Current status:

- [x]  address @bric3 feedback
  - [x] getter
  - [x] javadoc
- [x] review documentation for consistency
  - [x] Mockito.session()
  - [x] MockitoSessionBuilder
  - [x] Strictness
  - [x] MockitoHint
  - [x] PotentialStubbingProblem
  - [x] UnnecessaryStubbingException
  - [x] MockitoJUnitRunner and subclasses
  - [x] MockitoSession
  - [x] MockitoRule
- [x] Add / update mentions in the main Mockito class
- [x] add validate mockito usage (should be very easy)
- [x] Try making rules/runner use the session. They already reuse the code, they just don't use the API directly. Created #898 to track this work.
- [x] Address @TimvdLippe feedback
- [x] Create concurrent test (multiple sessions active in different threads)
- [x] (non-code change, not blocking merge) GitHub tickets linked from source code should be better documented #769, #384. Alternatively, create ticket for tracking.


bric3:
Bit busy in this couple of weeks I'l give my review asap, probably next week.

TimvdLippe:
I think this API makes more sense if it is actually used in the JUnit rule and runner. E.g. this API exposes the methods and implementation and the rule/runner call this API accordingly

szczepiq:
Thanks guys! Waiting for feedback.

@TimvdLippe, I completely agree this API should be used by the Rule / Runner. What are your thoughts about exposing this API to the users?

TimvdLippe:
@szczepiq Not sure what the benefit is for users. Therefore I would like this PR to implement it as internal API used in the rule and runner. Later we can see if we want to expose it or not.

szczepiq:
@timvdlippe thanks for feedback! You are not sure about benefits of a) strict stubbing or b) strict stubbing availibility/need outside of runner/rule?

TimvdLippe:
@szczepiq b. Strict stubbing is very valuable, but I am not sure what a user could do with the API outside of a runner/rule. I think keeping it internal is just fine for now.

szczepiq:
Ok. What about users that cannot use our JUnit Rule/runner?

TimvdLippe:
At this moment I am not aware of users requesting this. Still, we can initially keep it internal and expose it later if requested?

szczepiq:
I'm with you on being very cautious about introducing features without user feedback or request. I think that this exactly the approach the core team should have to build a great library.

Thing is that in this situation there are users needing it. For example, I'm frustrated that myself and other engineers at LinkedIn cannot use strict stubbing (most of the tests use TestNG, there are also usages of custom runners).

Another use case are existing integrations like [Jukito](https://github.com/ArcBees/Jukito), [Springockito](http://www.springockito.org), etc. Users of those frameworks won't be able to take advantage of strict stubbing.

In general, if the team chooses to implement a feature such us strict stubbing, then we should make the usage of the feature easy and flexible. We should not lock in users with JUnit. For example, Mockito has "@Mock" annotation and offers flexible way of using that annotation: with JUnit Rule, Runner and without them (initMocks() method).

Also keep in mind that the new API is "@Incubating" so we could potentially take it away if we don't find it valuable.

Thoughts?

TimvdLippe:
> myself and other engineers at LinkedIn cannot use strict stubbing

I was not aware of this usecase, as no issue was created. Given that this is requested by user, making it public is okay with me!

src/main/java/org/mockito/Mockito.java
```diff
@@ -2702,4 +2696,17 @@ static MockitoDebugger debug() {
...
+     * <p>
+     * TODO javadoc
+     *
+     * @param testClassInstance
+     * @param strictness
+     */
+    @Incubating
+    public static MockitoMocking startMocking(Object testClassInstance, Strictness strictness) {
+        return new DefaultMockitoMocking(testClassInstance, strictness, new ConsoleMockitoLogger());
+    }
```
bric3: * While public API is ok for me, I wonder if this API should be exposed in the Mockito class at first. Why not a class like `MockitoMocking.startMocking()...`, and rename the current `MockitoMocking` interface to `OngoingMockitoMocking`.
* This method maybe more interesting if started some kind of builder instead of taking arguments. One could want to modify the logger. And future options could be passed much more easily without breaking binary compatibility.

bric3: Could become something like :

```java
MockitoMocking.mocking().testInstance(this)
                        .logger(devnull)
                        .strictness(...)
                        // possible configuration options to add at later time
                        .initAnnotations(...)
                        .overridePlugin(...)
                        .validate(...)
                        // ...
                        .start(); // returns OngoingMockitoMocking
```

src/main/java/org/mockito/internal/framework/DefaultMockitoMocking.java
```diff
@@ -0,0 +1,50 @@
...
+            }
+
+            public String getTestMethodName() {
+                return null;
+            }
+        });
+
+        //TODO
+        // 1. validate mockito usage
+        // 2. can we have JUnit rule and runner use MockitoMocking API directly?
```
bric3:
```java
// 3. also JUnit 5 / Jupiter extensions
// 4. There's also the unreleased TestNG code.
```

src/main/java/org/mockito/internal/junit/DefaultStubbingLookupListener.java
```diff
@@ -20,6 +20,7 @@
 class DefaultStubbingLookupListener implements StubbingLookupListener {
 
     Strictness currentStrictness;
+    boolean mismatchesReported;
```
bric3: Since it not `final` but muted once, maybe introduce a getter ?

src/main/java/org/mockito/internal/junit/UniversalTestListener.java
```diff
@@ -44,8 +47,9 @@ public void testFinished(TestFinishedEvent event) {
         }
     }
 
-    private static void reportUnusedStubs(TestFinishedEvent event, Collection<Object> mocks) {
-        if (event.getFailure() == null) {
+    private void reportUnusedStubs(TestFinishedEvent event, Collection<Object> mocks) {
+        if (event.getFailure() == null && !stubbingLookupListener.mismatchesReported) {
+            //If there is some other failure (or mismatches were detected) don't report another exception to avoid confusion
```
bric3: this comment should be above (I understand because I did follow on this code and did the review, but in 2 years form now if someone stumble on this code)

src/main/java/org/mockito/internal/junit/UniversalTestListener.java
```diff
@@ -34,6 +36,7 @@ public void testFinished(TestFinishedEvent event) {
         Collection<Object> createdMocks = mocks.keySet();
         //At this point, we don't need the mocks any more and we can mark all collected mocks for gc
         //TODO make it better, it's easy to forget to clean up mocks and we still create new instance of list that nobody will read, it's also duplicated
+        //TODO clean up all other state, null out stubbingLookupListener
```
bric3: Should we promote the test util class `StateMaster` to track mockito state ?

src/test/java/org/mockitousage/stubbing/StrictStubbingE2ETest.java
```diff
@@ -0,0 +1,141 @@
...
+import org.mockito.exceptions.misusing.PotentialStubbingProblem;
+import org.mockito.exceptions.misusing.UnfinishedMockingException;
+import org.mockito.exceptions.misusing.UnnecessaryStubbingException;
+import org.mockito.quality.Strictness;
+import org.mockitousage.IMethods;
+import org.mockitoutil.JUnitResultAssert;
+
+import static org.mockito.BDDMockito.given;
+
+public class StrictStubbingE2ETest {
```
bric3: nit-picky I'd rename `E2E` to `EndToEnd`

src/test/java/org/mockitoutil/JUnitResultAssert.java
```diff
@@ -47,6 +50,17 @@ public JUnitResultAssert fails(int expectedFailureCount, Class expectedException
     }
 
     /**
+     * Expects single failure with specific exception and exception message.
+     * Automatically filters line numbers from exception messages.
+     */
+    public JUnitResultAssert fails(Class expectedException, String exceptionMessage) {
+        fails(1, expectedException);
+        Failure f = result.getFailures().iterator().next();
```
bric3: I remember that at some point some `Iterator` implementations failed if `hasNext` was not invoked. I'd be careful on that one. (Better make a utility method in `org.mockito.internal.util.collections.Iterables`)

src/main/java/org/mockito/internal/framework/DefaultMockitoMocking.java
szczepiq: Nice ideas!

src/main/java/org/mockito/internal/junit/DefaultStubbingLookupListener.java
szczepiq: I'd rather keep it simple, without a getter. Is it really important for you?

src/main/java/org/mockito/internal/junit/UniversalTestListener.java
szczepiq: Interesting idea. StateMaster manages thread local state. In this case, it's the state of test listener that we want to keep clean. Hence, StateMaster does not really apply here.

src/main/java/org/mockito/internal/junit/UniversalTestListener.java
szczepiq: Good idea. Thanks!

src/test/java/org/mockitousage/stubbing/StrictStubbingE2ETest.java
szczepiq: ok :)

src/test/java/org/mockitoutil/JUnitResultAssert.java
szczepiq: Interesting. iterator().next() is a fairly common idiom to get first element, for example from a set. It's a good idea to put it in Iterables. There is no harm to call hasNext so I'll do it. Thanks for the suggestion!

src/main/java/org/mockito/internal/junit/DefaultStubbingLookupListener.java
bric3: Actually having the field public (package visible) made me wonder if anyone could change the state, if the field is private and there's a getter I know instantly this field is supposed to be read elsewhere.

src/main/java/org/mockito/internal/junit/UniversalTestListener.java
bric3: I meant a tool like `StateMaster`, or refactor it to work for Mockito in a more general way.

src/test/java/org/mockitoutil/JUnitResultAssert.java
bric3: Also this will be easier to replace by Stream with Mockito 3

src/main/java/org/mockito/Mockito.java
szczepiq: I like a lot the idea of a builder. "MockitoMocking" name is somewhat meaningless. How about we call it "MockitoSession" or "MockitoRun" to indicate a lifecycle of some kind with clear starting and ending point:

```java
MockitoSession mockito = Mockito.mockitoSession().initMocks(this).strictness(Strictness.STRICT_STUBS).startMocking();
mockito.finishMocking();
```

I'm leaning towards new method on Mockito class instead for new static class endpoint. This way, the API is easier to discover / easier to use.

bric3: OK for either `MockitoRun` or `MockitoSession`.

szczepiq: Mockito.mockitoSession() VS MockitoSession.mockitoSession() ?

TimvdLippe: Let's do `Mockito.mockitoSession()` for now, but we should take another look at the method creep in this class. It is getting out of hand slowly but steadily :P

szczepiq: Absolutely we need to be cautious about adding more public method to Mockito class because it is so big. On the other hand, our users expect to find new API on Mockito class, it is a main entry to our entire API. MockitoAnnotations is an odd exception that we can live with because everybody is so used to it ;)

An alternative approach is to new create classes like MockitoSession, and place new public static methods there. However, this would fragment the API and would make it harder to use. Having lots of methods on Mockito is not convenient. However, having several static classes with API entry points is even more inconvenient. It makes the API less discoverable.

Given decent use case and no other sensible host, I'm ok with new method on Mockito class. 

Given @TimvdLippe vote on this direction, I'm going ahead with Mockito.mockitoSession() in this PR.

Thanks everybody for great feedback!

src/main/java/org/mockito/internal/junit/DefaultStubbingLookupListener.java
szczepiq: >Actually having the field public (package visible) made me wonder if anyone could change the state

This is the intent of this code. This state is mutable. The alternative is to provide getter and setter, which is more code for a simple collaboration between classes in the same package. If you feel that getter and setter yield code that is easier to understand, I'm happy to change it. LMK!

src/main/java/org/mockito/internal/junit/UniversalTestListener.java
szczepiq: Your suggestion is not super clear to me. Given that this refactoring is not critical for the feature nor code quality, let's move on for now. We can get back to this if needed!

szczepiq:
Pushed changes, new API is ready. Some TODOs, javadoc updates in code are pending.

src/main/java/org/mockito/internal/junit/DefaultStubbingLookupListener.java
bric3: I'm an anti setter getter person. But on the other hand I don't like leaking mutable variables.

src/main/java/org/mockito/internal/junit/UniversalTestListener.java
bric3: Yes it's not really related to this PR. Just an idea to discuss. ;)

bric3:
I will try to review the pushed change tonight. But I'm a tad overwhelmed at this moment.

src/main/java/org/mockito/MockitoSession.java
```diff
@@ -0,0 +1,79 @@
...
+
+import org.mockito.junit.MockitoJUnitRunner;
+import org.mockito.junit.MockitoRule;
+import org.mockito.quality.Strictness;
+
+/**
+ * {@code MockitoSession} helps driving cleaner tests but eliminating boilerplate code and adding extra validation.
+ * <p>
+ * {@code MockitoSession} is a session of mocking, during which the user creates and uses Mockito mocks.
+ * Typically the session is an execution of a single test method.
```
szczepiq: Thank you! I'll address your feedback.

src/main/java/org/mockito/MockitoSession.java
```diff
@@ -0,0 +1,79 @@
...
+ * {@code MockitoSession} initializes mocks, validates usage and detects incorrect stubbing.
+ * It helps keeping tests clean, {@link MockitoJUnitRunner} and {@link MockitoRule} use it behind the hood.
+ * <p>
+ * {@code MockitoSession} is useful when you cannot use {@link MockitoJUnitRunner} or {@link MockitoRule}.
+ * Example use cases include TestNG users or tests that cannot use Mockito's JUnit support
+ * because they already use different JUnit support (Jukito, Springockito).
```
szczepiq: Nice, thanks for reading the new docs!

szczepiq:
I added game plan to the PR description. Thanks so much for review! The code will not be changed much (individual commits can be reviewed anyway).

src/main/java/org/mockito/internal/session/DefaultMockitoSessionBuilder.java
```diff
@@ -0,0 +1,33 @@
...
+import org.mockito.quality.Strictness;
+import org.mockito.session.MockitoSessionBuilder;
+
+public class DefaultMockitoSessionBuilder implements MockitoSessionBuilder {
+
+    private Object testClassInstance;
+    private Strictness strictness;
+
+    @Override
+    public MockitoSessionBuilder initMocks(Object testClassInstance) {
```
bric3: Something just poped in my head, before merging, I thought that `initMocks` should be renamed to `initAnnotations`, `initMocks` is the old name when there was only `@Mock` annotation, but since a long time Mockito handles several types of annotations and has even a plugable annotation engine.

szczepiq: Yes, initMocks is not fully accurate.

OTOH, users are used to "initMocks" method from MockitoAnnotations, so the API is consistent. Also, technically we are not "initializing annotations", we are initializing annotated fields, and most of the time we annotate mocks ;)

I'd say we keep it as "initMocks".

bric3: Yes technically it is indeed fields ;)
But `initMocks` is still inacurate. Ideas : `initAnnotated` (thanks to @raphw), `processAnnotations`, `initFields`. What I'd like is to give these users the hint that it's not only about mocks only, but about mockito annotations in general. And then in the javadoc one can mention the `AnnotationEngine`.

src/main/java/org/mockito/internal/junit/DefaultStubbingLookupListener.java
szczepiq: Fixed it. Indeed it looks better now because after all I needed only 2 out of 4 setter/getter methods. The public/package local footprint of the class is smaller now (will push soon).

Thank you for useful code review feedback. My code is easy for me to understand (most of the time :), but I want it to be clear for other engs. Keep it coming.

src/main/java/org/mockito/internal/session/DefaultMockitoSessionBuilder.java
szczepiq: What about introduced inconsistency. Do you want to deprecate and rename MockitoAnnotations method, too?

src/main/java/org/mockito/internal/junit/UniversalTestListener.java
```diff
@@ -49,7 +49,7 @@ public void testFinished(TestFinishedEvent event) {
 
     private void reportUnusedStubs(TestFinishedEvent event, Collection<Object> mocks) {
         //If there is some other failure (or mismatches were detected) don't report another exception to avoid confusion
-        if (event.getFailure() == null && !stubbingLookupListener.mismatchesReported) {
+        if (event.getFailure() == null && !stubbingLookupListener.isMismatchesReported()) {
```
bric3: +1000 ;)

src/main/java/org/mockito/internal/session/DefaultMockitoSessionBuilder.java
bric3: I think the new name will speak more than the old name.
On MockitoAnnotations, this `initMocks` tickled me for quite some time, but since I use moregularly the JUnit runner or the JUnit rule, I didn't acted on it.

I think we could deprecate the `MockitoAnnotations.initMocks` and add the alternative method, it could in another PR though.

szczepiq: Not sure about "initAnnotated(this)". Looks kind of awkward, half-finished and not speaking to me on what is actually going on. Verb + noun would be my preference. Other options: "initFields" or "initAnnotatedFields". The more I think about it, the more I like "initMocks" :) I acknowledge that it is not 100% accurate but it is slick and simple. At the end of the day, we're working with mocks!

There is a separate question about ROI. Does changing this method name actually make any tangible difference in the quality of the API? There is some churn caused by deprecation of 9-year old method in a very mature library that has lots of users (all code, sample and documentation all over the world).

@bric3, I'd like to move on with your original idea, e.g. session().initMocks(this). If we want to change initMocks in both, session and MockitoAnnotations, let's move it to a separate ticket.

szczepiq:
I just rebased and bumped minor version to 2.7.0. This code is ready for final review and merge. I don't plan to work on the code unless there is more feedback.

The remaining unfinished action item in the description is a non-code change and does not block the review / merge.

src/main/java/org/mockito/internal/framework/DefaultMockitoSession.java
```diff
@@ -0,0 +1,44 @@
...
+public class DefaultMockitoSession implements MockitoSession {
+
+    private final Object testClassInstance;
+    private final UniversalTestListener listener;
+
+    public DefaultMockitoSession(Object testClassInstance, Strictness strictness, MockitoLogger logger) {
+        this.testClassInstance = testClassInstance;
+        listener = new UniversalTestListener(strictness, logger);
+        try {
+            Mockito.framework().addListener(listener);
```
TimvdLippe: Why add a listener here? It seems that the code is trying to detect invalid usage of mocking, but in `finishMocking` this listener is automatically removed and manually invoked with `testFinished`. This seems like a misusage of `Mockito.framework()`

src/main/java/org/mockito/internal/session/DefaultMockitoSessionBuilder.java
```diff
@@ -0,0 +1,33 @@
...
+    @Override
+    public MockitoSessionBuilder strictness(Strictness strictness) {
+        this.strictness = strictness;
+        return this;
+    }
+
+    @Override
+    public MockitoSession startMocking() {
+        //Configure default values
+        Object effectiveTest = this.testClassInstance == null ? new Object() : this.testClassInstance;
```
TimvdLippe: I would rather move this argument to the constructor, in my eyes it does not make sense to not track the testclassInstance when using this builder. Moreover, it does not notify framework authors when they are using the listener.

src/test/java/org/mockito/internal/session/DefaultMockitoSessionBuilderTest.java
```diff
@@ -0,0 +1,22 @@
+package org.mockito.internal.session;
+
+import org.junit.Test;
+import org.mockito.quality.Strictness;
+
+public class DefaultMockitoSessionBuilderTest {
+
+    @Test public void creates_sessions() throws Exception {
```
TimvdLippe: Why does this test not have any asserts?

src/test/java/org/mockito/internal/session/DefaultMockitoSessionBuilderTest.java
```diff
@@ -0,0 +1,22 @@
...
+
+        //passing null to configuration is legal, default value will be used
+        new DefaultMockitoSessionBuilder().initMocks(null).startMocking().finishMocking();
+        new DefaultMockitoSessionBuilder().initMocks(null).strictness(null).startMocking().finishMocking();
+        new DefaultMockitoSessionBuilder().strictness(null).startMocking().finishMocking();
+
+        //happy path
+        new DefaultMockitoSessionBuilder().initMocks(this).startMocking().finishMocking();
+        new DefaultMockitoSessionBuilder().initMocks(new Object()).startMocking().finishMocking();
+        new DefaultMockitoSessionBuilder().strictness(Strictness.LENIENT).startMocking().finishMocking();
```
TimvdLippe: What happens if you invoke for example `startMocking` twice? Would that be legal?

src/test/java/org/mockitousage/stubbing/StubbingWarningsTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Test public void stubbing_argument_mismatch() throws Throwable {
+        //when
+        given(mock.simpleMethod(100)).willReturn("100");
+        mock.simpleMethod(200);
+
+        mockito.finishMocking();
+
+        //TODO - currently we warn about "Unused" instead of "Arg mismatch" below
+        //because it was simpler to implement. This can be improved given we put priority to improve the warnings.
```
TimvdLippe: Is it too big of a hassle to fix now instead of delaying?

szczepiq: Yes. Given that warnings are not that much useful, I would probably never prioritize it (unless users ask).

src/test/java/org/mockito/internal/session/DefaultMockitoSessionBuilderTest.java
szczepiq: We have coverage for this scenario at org.mockitousage.stubbing.StrictStubbingEndToEndTest#detects_unfinished_mocking

However, it does not have correct documentation. Thank you for raising this. Nice catch. I will improve this scenario.

src/test/java/org/mockito/internal/session/DefaultMockitoSessionBuilderTest.java
szczepiq: It is a test for the builder, if it creates objects that we can call startMocking() and finishMocking() on, it is a sufficient verification. 

I'll improve it so that it is better and does not raise question like yours. Good feedback!

src/main/java/org/mockito/internal/session/DefaultMockitoSessionBuilder.java
szczepiq: >I would rather move this argument to the constructor

To the constructor of DefaultMockitoSessionBuilder? Are you advocating for API modification:

```java
//current
Mockito.session().initMocks(this).strictness(Strictness.WARN).startMocking();

//suggested by Tim
Mockito.session().strictness(Strictness.WARN).startMocking(this);
//or
Mockito.session(this).strictness(Strictness.WARN).startMocking();
```

>Moreover, it does not notify framework authors when they are using the listener

This feedback is unclear to me.

src/main/java/org/mockito/internal/framework/DefaultMockitoSession.java
szczepiq: > Why add a listener here?

When the session is started, the listener is added so that it can track creation of mocks and stubbing misuse.

In finish mocking we need to clean up state hence we remove the listener and report and findings.

Does it explain the workflow sufficiently? Do you suggest changes to the code / adding code comments?

src/main/java/org/mockito/internal/session/DefaultMockitoSessionBuilder.java
TimvdLippe: `Mockito.session(this).strictness(Strictness.WARN).startMocking();` yes that's what I meant.

> This feedback is unclear to me.

To me the usage of the listener is confusing and thus far I do not understand how it is used and why. Probably I am missing an implementation detail.

src/main/java/org/mockito/internal/session/DefaultMockitoSessionBuilder.java
bric3: OK for `initAnnotated`, but what about `processAnnotations`.

> @bric3, I'd like to move on with your original idea, e.g. session().initMocks(this). If we want to change initMocks in both, session and MockitoAnnotations, let's move it to a separate ticket.

That's an option bu in the meantime we introduce an API that could be deprecated. And in the next PR, it's probably gonna be discarded because the API has just been introduced in mockitosession and we won't deprecate something that just entered the scene. We rarely change incubating APIs. At least we could mention it in the javadoc.

bric3: About ROI : `mockitoSession` will probably be more especially useful to framework developers. `initMocks` is for them the way to initialize mockito annotations. Or other annotations if they plug another `AnnotationEngine`. If we were to introduce other annotations that are not about mocks, it will feel awkward.

If we ever implement a way to inject non mock in the `@InjectMocks` field then `initMocks` will be even more awkward. If we add other annotations that non mock related, but for Mockito lifecycle it'll be even more _off_.

src/main/java/org/mockito/internal/session/DefaultMockitoSessionBuilder.java
bric3: This is confusing to me `Mockito.session(this)`, especially since the instance could be `null`, if one wants it that way. And this is a builder.

I like this part of the API the way it is. The listener could be extracted in a private method ?

src/main/java/org/mockito/internal/framework/DefaultMockitoSession.java
bric3: Eventually wrap this try block in a private method that self explain itself.

src/main/java/org/mockito/internal/session/DefaultMockitoSessionBuilder.java
szczepiq: >To me the usage of the listener is confusing and thus far I do not understand how it is used and why. Probably I am missing an implementation detail.

Good feedback. I'll try to rework the code / add code comments to make this part clearer.

szczepiq: "initMocks(this)" is optional, the session will work without it. It also communicates the intent best when reading the code. @TimvdLippe, are you worried that users will forget to call this method?

"Mockito.session(this)" does not communicate the intent well and looks somewhat awkward. "startMocking(this)" is slightly better but suffers from similar problem. I'd say we stick to the current API in the PR.

bric3: I agree with @szczepiq.

TimvdLippe: Okay fair points, let's leave it as is.

szczepiq:
If there is no more feedback, I'd like to merge. Thanks guys for comprehensive review and pointing out improvements!!!

PR #867, opened 2017-01-04
Fixes #866 should not throw NPE, when custom Exception fillInStackTrace returns null

### LINKED ISSUE #866: When custom exception fillInstackTrace() returns null, Exception mock cannot work properly
```java
public class CustomException extends Exception {

    public CustomException() {
        super();
    }

    @Override
    public Exception fillInStackTrace() {
        return null;
    }
}
```

TestCase like this
```java
 when(demo.process()).thenThrow(new CustomException())

//... test code
```

```
	at org.mockito.internal.exceptions.stacktrace.ConditionalStackTraceFilter.filter(ConditionalStackTraceFilter.java:23)
	at org.mockito.internal.stubbing.answers.ThrowsException.answer(ThrowsException.java:32)
	at org.mockito.internal.stubbing.StubbedInvocationMatcher.answer(StubbedInvocationMatcher.java:35)
	at org.mockito.internal.handler.MockHandlerImpl.handle(MockHandlerImpl.java:95)
```

wuwen5:
fixes #866.

### DIFF
--- a/src/main/java/org/mockito/internal/stubbing/answers/ThrowsException.java
+++ b/src/main/java/org/mockito/internal/stubbing/answers/ThrowsException.java
@@ -29,6 +29,10 @@ public class ThrowsException implements Answer<Object>, ValidableAnswer, Seriali
             throw throwable;
         }
         Throwable t = throwable.fillInStackTrace();
+
+        if (t == null) {
+            throw throwable;
+        }
         filter.filter(t);
         throw t;
     }
--- a/src/test/java/org/mockitousage/bugs/NPEWhenCustomExceptionStackTraceReturnNullTest.java
+++ b/src/test/java/org/mockitousage/bugs/NPEWhenCustomExceptionStackTraceReturnNullTest.java
@@ -1,0 +1,32 @@
+package org.mockitousage.bugs;
+
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockitousage.IMethods;
+import org.mockitoutil.TestBase;
+
+import static junit.framework.TestCase.fail;
+import static org.mockito.Mockito.when;
+
+public class NPEWhenCustomExceptionStackTraceReturnNullTest extends TestBase {
+
+    @Mock
+    IMethods mock;
+
+    class NullStackTraceException extends RuntimeException {
+        @Override
+        public Exception fillInStackTrace() {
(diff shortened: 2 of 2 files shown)


src/main/java/org/mockito/internal/stubbing/answers/ThrowsException.java
```diff
@@ -29,6 +29,10 @@ public Object answer(InvocationOnMock invocation) throws Throwable {
             throw throwable;
         }
         Throwable t = throwable.fillInStackTrace();
+
+        if (t == null) {
+            throw throwable;
+        }
```
bric3: Actually I'd rather make the check in `ConditionalStackTraceFilter.filter()`. And if throwable is null, just return.

wuwen5: 👌 already changed,please review.

bric3:
The build is failing.

src/main/java/org/mockito/internal/stubbing/answers/ThrowsException.java
```diff
@@ -30,7 +30,7 @@ public Object answer(InvocationOnMock invocation) throws Throwable {
         }
         Throwable t = throwable.fillInStackTrace();
         filter.filter(t);
-        throw t;
+        throw throwable;
```
bric3: Sorry it seems I was wrong. It seems that indeed your previous attempt was correct.

```java
        if (t == null) {
            return throwable;
        }
```

wuwen5: Ok, But I don't know why the build is failing?

> Execution failed for task ':assertEnvVariables'.
> Please export 'GH_TOKEN' env variable first. Expression: java.lang.System.env[it]. Values: it = GH_TOKEN

bric3: Yes I'm sure your change is correct, it's something else.

src/main/java/org/mockito/internal/stubbing/answers/ThrowsException.java
bric3: Discarding this comment as it requires to rewrite `filter` a bit more.

PR #875, opened 2017-01-11
Update Byte Buddy and enable injecting loading strategy for Android

raphw:
This enables Byte Buddy's new injection strategy for Android which loads types within a target class loader instead of creating a new one. Also, this updates Byte Buddy with a few bug fixes.


bric3:
Didn't even have time to press the approve button ;)

PR #885, opened 2017-01-15
Spy annotation reports better error message if instance creation is impossible

### LINKED ISSUE #878: @Spy dosn't report the correct error when it can't instantiate abstract class
The annotation engine doesn't inform the user correctly when the spying an inner `private static abstract class`.

```
org.mockito.exceptions.base.MockitoException: Unable to initialize @Spy annotated field 'a'.
null

	at org.mockito.internal.configuration.SpyAnnotationEngine.process(SpyAnnotationEngine.java:70)
	at org.mockito.internal.configuration.InjectingAnnotationEngine.processIndependentAnnotations(InjectingAnnotationEngine.java:59)
	at org.mockito.internal.configuration.InjectingAnnotationEngine.process(InjectingAnnotationEngine.java:41)
```

Indeed when the class is private, javac still generates a synthetic 1-arg constructor taking the enclosing class.

That should be easy to improve, as the `SpyAnnotationEngine` just needs to report an error message for this case.

bric3:
This simple piece of code prevents the `SpyAnnotationEngine` to try create a spy instance for a some inner private class with different set of modifiers.

While preventing execution it reports slightly better error messages, before, some time the cause was `null` . This should fix #878

### DIFF
--- a/src/main/java/org/mockito/internal/configuration/SpyAnnotationEngine.java
+++ b/src/main/java/org/mockito/internal/configuration/SpyAnnotationEngine.java
@@ -61,12 +61,9 @@ public class SpyAnnotationEngine implements AnnotationEngine, org.mockito.config
                         // for example happens when MockitoAnnotations.initMocks is called two times.
                         Mockito.reset(instance);
                     } else if (instance != null) {
-                        field.set(testInstance, Mockito.mock(instance.getClass(),
-                                                             withSettings().spiedInstance(instance)
-                                                                           .defaultAnswer(CALLS_REAL_METHODS)
-                                                                           .name(field.getName())));
+                        field.set(testInstance, spyInstance(field, instance));
                     } else {
-                        field.set(testInstance, newSpyInstance(testInstance, field));
+                        field.set(testInstance, spyNewInstance(testInstance, field));
                     }
                 } catch (Exception e) {
                     throw new MockitoException("Unable to initialize @Spy annotated field '" + field.getName() + "'.\n" + e.getMessage(), e);
@@ -82,7 +79,14 @@ public class SpyAnnotationEngine implements AnnotationEngine, org.mockito.config
         }
     }
 
-    private static Object newSpyInstance(Object testInstance, Field field)
+    private Object spyInstance(Field field, Object instance) {
+        return Mockito.mock(instance.getClass(),
+                            withSettings().spiedInstance(instance)
+                                                           .defaultAnswer(CALLS_REAL_METHODS)
+                                                           .name(field.getName()));
+    }
+
+    private static Object spyNewInstance(Object testInstance, Field field)
             throws InstantiationException, IllegalAccessException, InvocationTargetException {
         MockSettings settings = withSettings().defaultAnswer(CALLS_REAL_METHODS)
                                               .name(field.getName());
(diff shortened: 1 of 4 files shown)


src/test/java/org/mockitousage/annotation/SpyAnnotationTest.java
```diff
@@ -173,18 +192,86 @@ public void should_report_when_encosing_instance_is_needed() throws Exception {
...
+    @Test
+    public void should_report_private_inner_not_supported() throws Exception {
+        try {
+            MockitoAnnotations.initMocks(new WithInnerPrivate());
+            fail();
+        } catch (MockitoException e) {
+            // Currently fails at instantiation time, because the mock subclass don't have the
+            // 1-arg constructor expected for the outerclass.
+            // org.mockito.internal.creation.instance.ConstructorInstantiator.withParams()
+            assertThat(e).hasMessageContaining("Unable to initialize @Spy annotated field 'spy_field'")
```
bric3: @raphw I found a slight edge issue here, when working on inner private. In this case, the created mock class loses it's 1-arg constructor, the error is raised in the above method `ConstructorInstantiator.withParams`.

I feel this is either a bug or a misconfiguration. But I haven't looked too much at it yet. Could it be the combination of modifiers that causes this problem ?

The class to mock is [`InnerPrivate`](https://github.com/mockito/mockito/pull/885/files#diff-3404c0c83dc4764960b54a85a537418fR252).

TimvdLippe:
The diff is quite cluttered with unrelated changes. https://github.com/mockito/mockito/pull/885/files?w=1 makes it a bit easier, but still a lot of other changes going on.

If I can digest the changes correctly, I think they are fine.

src/test/java/org/mockitousage/annotation/SpyAnnotationTest.java
raphw: The problem is that `InnerPrivate` only declares a `private` constructor. Even if private classes are compiled to be package-private (there are no private classes in Java) what allows subclassing them, you cannot call any constructor of this class from a subclass. Therefore, Byte Buddy does not add a constructor, it is not possible to legally do so.

bric3: Ah right of course. However though, one can write : `private class InnerPrivateSub extends InnerPrivate {}` in the same enclosing class. Then this `InnerPrivateSub` will be instantiable, and will have a single private 1-arg (of the enclosing class type) constructor.

Here's bytecode outline

```
// class version 50.0 (50)
// access flags 0x20
class org/mockitousage/annotation/SpyAnnotationTest$WithInnerPrivate$InnerPrivateSub extends org/mockitousage/annotation/SpyAnnotationTest$WithInnerPrivate$InnerPrivate  {

  // compiled from: SpyAnnotationTest.java
  // access flags 0x8
  static INNERCLASS org/mockitousage/annotation/SpyAnnotationTest$WithInnerPrivate org/mockitousage/annotation/SpyAnnotationTest WithInnerPrivate
  // access flags 0x2
  private INNERCLASS org/mockitousage/annotation/SpyAnnotationTest$WithInnerPrivate$InnerPrivateSub org/mockitousage/annotation/SpyAnnotationTest$WithInnerPrivate InnerPrivateSub
  // access flags 0x2
  private INNERCLASS org/mockitousage/annotation/SpyAnnotationTest$WithInnerPrivate$InnerPrivate org/mockitousage/annotation/SpyAnnotationTest$WithInnerPrivate InnerPrivate
  // access flags 0x1008
  static synthetic INNERCLASS org/mockitousage/annotation/SpyAnnotationTest$1 org/mockitousage/annotation/SpyAnnotationTest null

  // access flags 0x1010
  final synthetic Lorg/mockitousage/annotation/SpyAnnotationTest$WithInnerPrivate; this$0

  // access flags 0x2
  private <init>(Lorg/mockitousage/annotation/SpyAnnotationTest$WithInnerPrivate;)V
   L0
    LINENUMBER 263 L0
    ALOAD 0
    ALOAD 1
    PUTFIELD org/mockitousage/annotation/SpyAnnotationTest$WithInnerPrivate$InnerPrivateSub.this$0 : Lorg/mockitousage/annotation/SpyAnnotationTest$WithInnerPrivate;
    ALOAD 0
    ALOAD 1
    ACONST_NULL
    INVOKESPECIAL org/mockitousage/annotation/SpyAnnotationTest$WithInnerPrivate$InnerPrivate.<init> (Lorg/mockitousage/annotation/SpyAnnotationTest$WithInnerPrivate;Lorg/mockitousage/annotation/SpyAnnotationTest$1;)V
    RETURN
   L1
    LOCALVARIABLE this Lorg/mockitousage/annotation/SpyAnnotationTest$WithInnerPrivate$InnerPrivateSub; L0 L1 0
    MAXSTACK = 3
    MAXLOCALS = 2
}
```

I think the `INNERCLASS` represents the `InnerClasses` attributes as specified in [§4.7.5](https://docs.oracle.com/javase/specs/jvms/se6/html/ClassFile.doc.html#79996).

bric3:
There's some reformat and cleanup commit, the most interesting is the latest commit.

szczepiq:
>@TimvdLippe There's some reformat and cleanup commit, the most interesting is the latest commit.

This is useful information. I can focus on the last commit.

src/main/java/org/mockito/internal/configuration/SpyAnnotationEngine.java
```diff
@@ -68,6 +68,7 @@ public void process(Class<?> context, Object testInstance) {
                         field.set(testInstance, newSpyInstance(testInstance, field));
                     }
                 } catch (Exception e) {
+                    e.printStackTrace();
```
szczepiq: I assume you want to get rid of this line?

src/main/java/org/mockito/internal/configuration/SpyAnnotationEngine.java
```diff
@@ -89,7 +90,13 @@ private static Object newSpyInstance(Object testInstance, Field field)
         if (type.isInterface()) {
             return Mockito.mock(type, settings.useConstructor());
         }
-        if (!Modifier.isStatic(type.getModifiers())) {
+        int modifiers = type.getModifiers();
+        if (Modifier.isPrivate(modifiers) && Modifier.isAbstract(modifiers) && type.getEnclosingClass() != null) {
```
szczepiq: Is it worth to extract it to small utility method to drive clarity?

src/main/java/org/mockito/internal/configuration/SpyAnnotationEngine.java
```diff
@@ -89,7 +90,13 @@ private static Object newSpyInstance(Object testInstance, Field field)
...
+            throw new MockitoException("@Spy annotation can't initialize private abstract inner classes. "
```
szczepiq: Is it possible to offer a suggestion to the user? For example, make the class public, etc.

src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
```diff
@@ -36,7 +37,7 @@ public ConstructorInstantiator(Object outerClassInstance) {
         } catch (Exception e) {
             throw paramsException(cls, e);
         }
-        throw paramsException(cls, null);
+        throw noMatchingConstructor(cls);
```
szczepiq: I really like continuously improving the exception messages. Thank you!!!

src/main/java/org/mockito/internal/configuration/SpyAnnotationEngine.java
bric3: Ah yes forgot it, that's why it's labeled on hold :D

src/main/java/org/mockito/internal/configuration/SpyAnnotationEngine.java
bric3: Yes indeed, I'm still experimenting, but I think this can be safely extracted.

src/main/java/org/mockito/internal/configuration/SpyAnnotationEngine.java
bric3: Good idea !

bric3:
Comments addressed.

PR #886, opened 2017-01-15
Fixed build for external contributors

### LINKED ISSUE #871: Travis builds from non-core developers fail on assertEnvVariables
Example build: https://travis-ci.org/mockito/mockito/jobs/188911380

Seems that this does not affect core developers, as my build has been succesful: https://travis-ci.org/mockito/mockito/builds/188321204 as well as those in #865 and #859

szczepiq:
Moved the validation of env variables required for release to further release step. We don't need to assert the variables if the release is skipped.

Fixes #871

Merge at will. Reordering those release steps is safe. Those steps do not impact the release process in any way. Removing those steps completely would not break the release. It would  just make it fail later VS fast, and would make the failure less readable. Hence, the validation steps are useful but reordering them does not impact the release.

I tested it by running "./gradlew release -PdryRun". If you export the variables, you can actually use above to dry run the release.

### DIFF
--- a/gradle/root/release.gradle
+++ b/gradle/root/release.gradle
@@ -254,10 +254,10 @@ task pushChanges(type: Exec) {
 }
 
 releaseWorkflow {
-    step assertEnvVariables
-    step assertCleanWorkingCopy
     step releaseNeeded
     onlyIf { releaseNeeded.needed }
+    step assertEnvVariables
+    step assertCleanWorkingCopy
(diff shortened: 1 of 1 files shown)


TimvdLippe:
Oh nice! Now how do we test this? :joy:

szczepiq:
Update the ticket with steps to test and info how I have tested it. I should have included that earlier :)

bric3:
Thanks @szczepiq

PR #872, opened 2017-01-07
Prepare Android library for publication and bump version to 2.6.0

TimvdLippe:
References first step of #848

PR #893, opened 2017-01-20
Add detection for Android environments and give warnings if used incorrectly. Corrected documentation.

raphw:
The documentation gave an incorrect configuration example. Also, warnings were added for using Android on a non-Android environment.

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -95,15 +95,23 @@ import org.mockito.verification.VerificationWithTimeout;
  * <h3 id="0.1">0.1. <a class="meaningful_link" href="#mockito-android">Mockito Android support</a></h3>
  *
  * With Mockito version 2.6.1 we ship "native" Android support. To enable Android support, add the `mockito-android` library as dependency
- * to your project. This artifact is published to the same mockito organization:
+ * to your project. This artifact is published to the same Mockito organization and can be imported for Android as follows:
  *
  * <pre class="code"><code>
- * repositories { jcenter() }
- * dependencies { testCompile "org.mockito:mockito-android:2.+" }
+ * repositories {
+ *   jcenter()
+ * }
+ * dependencies {
+ *   testCompile "org.mockito:mockito-core:+"
+ *   androidTestCompile "org.mockito:mockito-android:+"
+ * }
  * </code></pre>
  *
- * All features of Mockito are still in the `mockito-core` artifact. The `mockito-android` artifact implements a compatibility layer
- * with Android. If you encounter issues with mocking on Android, please open an issue <a href="https://github.com/mockito/mockito/issues/new">on the official issue tracker</a>.
+ * You can continue to run the same unit tests on a regular VM by using the `mockito-core` artifact in your "testCompile" scope as shown
+ * above. Be aware that you cannot use the <a href="#39">inline mock maker</a> on Android due to limitations in the Android VM.
+ *
+ * If you encounter issues with mocking on Android, please open an issue
+ * <a href="https://github.com/mockito/mockito/issues/new">on the official issue tracker</a>.
  * Do provide the version of Android you are working on and dependencies of your project.
  *
  * <h3 id="1">1. <a class="meaningful_link" href="#verification">Let's verify some behaviour!</a></h3>
--- a/src/main/java/org/mockito/internal/util/Platform.java
+++ b/src/main/java/org/mockito/internal/util/Platform.java
@@ -4,8 +4,11 @@
  */
 package org.mockito.internal.util;
 
+import java.util.Locale;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+
+import static org.mockito.internal.util.StringJoiner.join;
 
 public abstract class Platform {
 
@@ -23,8 +26,16 @@ public abstract class Platform {
     private Platform() {
     }
 
+    public static boolean isAndroid() {
+        return System.getProperty("java.vendor", "").toLowerCase(Locale.US).contains("android");
+    }
+
+    public static boolean isAndroidMockMakerRequired() {
+        return Boolean.getBoolean("org.mockito.mock.android");
+    }
+
     public static String describe() {
-        return String.format("Java               : %s\n" +
+        String description = String.format("Java               : %s\n" +
                              "JVM vendor name    : %s\n" +
                              "JVM vendor version : %s\n" +
                              "JVM name           : %s\n" +
@@ -40,6 +51,18 @@ public abstract class Platform {
                              JVM_INFO,
                              OS_NAME,
                              OS_VERSION);
+        if (isAndroid()) {
+            description = join(
+                    "IMPORTANT INFORMATION FOR ANDROID USERS:",
+                    "",
+                    "The regular Byte Buddy mock makers cannot generate code on an Android VM!",
+                    "To resolve this, please use the 'mockito-android' dependency for your application:",
+                    "http://search.maven.org/#search%7Cga%7C1%7Ca%3A%22mockito-android%22%20g%3A%22org.mockito%22",
+                    "",
+                    description
+            );
+        }
+        return description;
     }
 
     public static boolean isJava8BelowUpdate45() {
--- a/subprojects/android/src/main/java/org/mockito/android/internal/creation/AndroidByteBuddyMockMaker.java
+++ b/subprojects/android/src/main/java/org/mockito/android/internal/creation/AndroidByteBuddyMockMaker.java
@@ -1,13 +1,34 @@
 package org.mockito.android.internal.creation;
 
 import org.mockito.internal.creation.bytebuddy.SubclassByteBuddyMockMaker;
+import org.mockito.internal.util.ConsoleMockitoLogger;
+import org.mockito.internal.util.Platform;
 import org.mockito.invocation.MockHandler;
 import org.mockito.mock.MockCreationSettings;
 import org.mockito.plugins.MockMaker;
 
+import static org.mockito.internal.util.StringJoiner.join;
+
 public class AndroidByteBuddyMockMaker implements MockMaker {
 
-    private final MockMaker delegate = new SubclassByteBuddyMockMaker(new AndroidLoadingStrategy());
+    private final MockMaker delegate;
+
+    public AndroidByteBuddyMockMaker() {
+        if (Platform.isAndroid() || Platform.isAndroidMockMakerRequired()) {
+            delegate = new SubclassByteBuddyMockMaker(new AndroidLoadingStrategy());
+        } else {
+            new ConsoleMockitoLogger().log(join(
+                    "IMPORTANT NOTE FROM MOCKITO:",
+                    "",
+                    "You included the 'mockito-android' dependency in a non-Android environment.",
+                    "The Android mock maker was disabled. You should only include the latter in your 'androidTestCompile' configuration",
+                    "If disabling was a mistake, you can set the 'org.mockito.mock.android' property to 'true' to override this detection.",
+                    "",
+                    "Visit https://javadoc.io/page/org.mockito/mockito-core/latest/org/mockito/Mockito.html#0.1 for more information"
+            ));
+            delegate = new SubclassByteBuddyMockMaker();
+        }
+    }
 
     @Override
     public <T> T createMock(MockCreationSettings<T> settings, MockHandler handler) {


src/main/java/org/mockito/Mockito.java
```diff
@@ -95,15 +95,18 @@
  * <h3 id="0.1">0.1. <a class="meaningful_link" href="#mockito-android">Mockito Android support</a></h3>
  *
  * With Mockito version 2.6.1 we ship "native" Android support. To enable Android support, add the `mockito-android` library as dependency
- * to your project. This artifact is published to the same mockito organization:
+ * to your project. This artifact is published to the same Mockito organization and can be imported for Android as follows:
  *
  * <pre class="code"><code>
  * repositories { jcenter() }
- * dependencies { testCompile "org.mockito:mockito-android:2.+" }
+ * dependencies { androidTestCompile "org.mockito:mockito-android:2.+" }
```
bric3: I think we should add the version 2.6 for android artefact, and add the mockit-core dependency otherwise, it is not imported by mockito-android.

```
androidTestCompile "org.mockito:mockito-core:2.6.+"
androidTestCompile "org.mockito:mockito-android:2.6.+"
```

subprojects/android/src/main/java/org/mockito/android/internal/creation/AndroidByteBuddyMockMaker.java
```diff
@@ -1,13 +1,31 @@
...
 public class AndroidByteBuddyMockMaker implements MockMaker {
 
-    private final MockMaker delegate = new SubclassByteBuddyMockMaker(new AndroidLoadingStrategy());
+    private final MockMaker delegate;
+
+    public AndroidByteBuddyMockMaker() {
+        if (Platform.isAndroid() || Boolean.getBoolean("org.mockito.mock.android")) {
+            delegate = new SubclassByteBuddyMockMaker(new AndroidLoadingStrategy());
+        } else {
+            System.err.println(join(
```
bric3: Maybe use and extends `ConsoleMockitoLogger` to log on `System.err` ?

bric3: OTH maybe `err` is a bit too much, since this code enable the mockito-android to fallback on the core mockmaker ? info level maybe sufficient ?

src/main/java/org/mockito/Mockito.java
TimvdLippe: Shouldn't it be `testCompile "org.mockito:mockito-core:2.6.+"` then?

The original format was approved in #872

bric3: Actually both could be necessary.

They are different classpath, `testCompile` to run on the jvm, `androidTestCompile` to run on the emulator / device. 
Only JUnit appears in both even without the explicit dependency (I believe this is due to the default Espresso dependency that transitively import JUnit).

bric3: Here's what a gradle file should look like.

```gradle
apply plugin: 'com.android.application'
apply plugin: 'com.neenbedankt.android-apt'

android {
    compileSdkVersion 25
    buildToolsVersion "25.0.2"
    defaultConfig {
        applicationId "com.github.bric3.mymockitoapplication"
        minSdkVersion 25
        targetSdkVersion 25
        versionCode 1
        versionName "1.0"
        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
    }
    buildTypes {
        release {
            minifyEnabled false
            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
        }
    }
}

dependencies {
    compile fileTree(dir: 'libs', include: ['*.jar'])
    androidTestCompile('com.android.support.test.espresso:espresso-core:2.2.2', {
        exclude group: 'com.android.support', module: 'support-annotations'
    })
    compile 'com.android.support:appcompat-v7:25.1.0'
    compile 'com.google.dagger:dagger:2.8'
    testCompile 'junit:junit:4.12'
    testCompile 'org.mockito:mockito-core:2.+'
    androidTestCompile 'org.mockito:mockito-core:2.+'
//    androidTestCompile 'com.linkedin.dexmaker:dexmaker-mockito:2.2.0'
    androidTestCompile 'org.mockito:mockito-android:2.6.+'
    apt 'com.google.dagger:dagger-compiler:2.8'
}
```

raphw: Changed the doc. I think it is best if we do not refer to a version alltogether to keep the documentation "time-less" in case there are old versions somewhere in the future. Most people do know how to specify a version.

subprojects/android/src/main/java/org/mockito/android/internal/creation/AndroidByteBuddyMockMaker.java
raphw: I changed this!

src/main/java/org/mockito/Mockito.java
bric3: That makes sense, especially since the artifact is known to have been released with version 2.6

subprojects/android/src/main/java/org/mockito/android/internal/creation/AndroidByteBuddyMockMaker.java
```diff
@@ -1,13 +1,32 @@
...
+import org.mockito.internal.util.Platform;
 import org.mockito.invocation.MockHandler;
 import org.mockito.mock.MockCreationSettings;
 import org.mockito.plugins.MockMaker;
 
-public class AndroidByteBuddyMockMaker implements MockMaker {
+import static org.mockito.internal.util.StringJoiner.join;
 
-    private final MockMaker delegate = new SubclassByteBuddyMockMaker(new AndroidLoadingStrategy());
+public class AndroidByteBuddyMockMaker extends ConsoleMockitoLogger implements MockMaker {
```
bric3: I'd rather have it as a field. Actually I'm thinking about having a plugin for the logger.

subprojects/android/src/main/java/org/mockito/android/internal/creation/AndroidByteBuddyMockMaker.java
```diff
@@ -1,13 +1,32 @@
...
+
+    private final MockMaker delegate;
+
+    public AndroidByteBuddyMockMaker() {
+        if (Platform.isAndroid() || Boolean.getBoolean("org.mockito.mock.android")) {
```
ChristianSchwarz: @raphw Maybe it is a good idea to move `Boolean.getBoolean("org.mockito.mock.android")` to `Platform.hasDependencyMockitoAndroid()`. Currently it feels a bit cryptic.

subprojects/android/src/main/java/org/mockito/android/internal/creation/AndroidByteBuddyMockMaker.java
```diff
@@ -1,13 +1,32 @@
...
+            delegate = new SubclassByteBuddyMockMaker(new AndroidLoadingStrategy());
+        } else {
+            log(join(
```
ChristianSchwarz: @raphw What about moving the log creation to a new method like `logUnexpectedAndroidDependency()`. The log is just a corn case but in this constructor it consumes the most lines.

src/main/java/org/mockito/Mockito.java
```diff
@@ -95,15 +95,23 @@
...
  *
  * <pre class="code"><code>
- * repositories { jcenter() }
- * dependencies { testCompile "org.mockito:mockito-android:2.+" }
+ * repositories {
+ *   jcenter()
+ * }
+ * dependencies {
+ *   androidTestCompile "org.mockito:mockito-core:+"
+ *   androidTestCompile "org.mockito:mockito-android:+"
```
szczepiq: I like that the example uses dynamic versions. It's easier for us to maintain and also potentially gets us users that are always on the latest version.

Why does the user need to declare both dependencies? mockito-android depends on mockito-core and pulls it as transitive dependency. I would recommend to just depend on mockito-android.

src/main/java/org/mockito/Mockito.java
```diff
@@ -95,15 +95,23 @@
...
+ * }
  * </code></pre>
  *
- * All features of Mockito are still in the `mockito-core` artifact. The `mockito-android` artifact implements a compatibility layer
- * with Android. If you encounter issues with mocking on Android, please open an issue <a href="https://github.com/mockito/mockito/issues/new">on the official issue tracker</a>.
+ * You can continue to run the same unit tests on a regular VM by using the `mockito-core` artifact for your "testCompile" scope.
+ * Be aware that you cannot use the <a href="#39">inline mock maker</a> on Android due to limitations in the Android VM.
+ *
+ * If you encounter issues with mocking on Android, please open an issue
+ * <a href="https://github.com/mockito/mockito/issues/new">on the official issue tracker</a>.
```
szczepiq: Nice, quick link to our tracker.

src/main/java/org/mockito/Mockito.java
```diff
@@ -95,15 +95,23 @@
...
+ * }
+ * dependencies {
+ *   androidTestCompile "org.mockito:mockito-core:+"
+ *   androidTestCompile "org.mockito:mockito-android:+"
+ * }
  * </code></pre>
  *
- * All features of Mockito are still in the `mockito-core` artifact. The `mockito-android` artifact implements a compatibility layer
- * with Android. If you encounter issues with mocking on Android, please open an issue <a href="https://github.com/mockito/mockito/issues/new">on the official issue tracker</a>.
+ * You can continue to run the same unit tests on a regular VM by using the `mockito-core` artifact for your "testCompile" scope.
```
szczepiq: Consider a code sample for 'testCompile'

src/main/java/org/mockito/internal/util/Platform.java
```diff
@@ -40,6 +47,18 @@ public static String describe() {
                              JVM_INFO,
                              OS_NAME,
                              OS_VERSION);
+        if (isAndroid()) {
+            description = join(
+                    "IMPORTANT INFORMATION FOR ANDROID USERS:",
```
szczepiq: Fantastic message!

It would be great to link to official documentation, for example: https://javadoc.io/page/org.mockito/mockito-core/latest/org/mockito/Mockito.html#0.1

subprojects/android/src/main/java/org/mockito/android/internal/creation/AndroidByteBuddyMockMaker.java
raphw: Fixed.

subprojects/android/src/main/java/org/mockito/android/internal/creation/AndroidByteBuddyMockMaker.java
raphw: Good idea, changed this.

subprojects/android/src/main/java/org/mockito/android/internal/creation/AndroidByteBuddyMockMaker.java
raphw: I would not want to tear the two apart. As it is the only class and the constructor is only a few lines, I do not think that this would improve readability as this would only mean that the same logic was moved to a method.

src/main/java/org/mockito/internal/util/Platform.java
raphw: Good idea, added it.

src/main/java/org/mockito/Mockito.java
raphw: You are right!

src/main/java/org/mockito/Mockito.java
raphw: I merged the samples to explain how 99% of the people are going to use it.

bric3:
@szczepiq the additional `mockito-core` dependency is necessary because `mockito-android` `pom.xml` doesn't depends on `mockito-core`.

Also with Android the scopes `testCompile` and `androidTestCompile` are different scopes and one doesn't inherit former the other. Hence the additional `mockito-core` declaration.

raphw:
@bric3 That does not seem to be true. According to the Maven Central artifact https://repo1.maven.org/maven2/org/mockito/mockito-android/2.7.0/mockito-android-2.7.0.pom - Mockito Android does depend on Mockito Core.

src/main/java/org/mockito/Mockito.java
```diff
@@ -95,15 +95,23 @@
...
  *
  * <pre class="code"><code>
- * repositories { jcenter() }
- * dependencies { testCompile "org.mockito:mockito-android:2.+" }
+ * repositories {
+ *   jcenter()
+ * }
+ * dependencies {
+ *   testCompile "org.mockito:mockito-core:+"
+ *   androidTestCompile "org.mockito:mockito-android:+"
```
bric3: As long as mockito-core is not listed as a dependency of mockito-android the tests that are supposed to run on the emulator or the device won't compile.

raphw: It is: https://repo1.maven.org/maven2/org/mockito/mockito-android/2.7.0/mockito-android-2.7.0.pom

bric3: Oh you're right, I must have been in another world when I thought this.

TimvdLippe:
@szczepiq Do you approve the changes? If so you can merge this PR :smile:

bric3:
I think we can merge this PR. 

We can still improve error message in another PR if that feel  absolutely necessary.

PR #896, opened 2017-01-20
Clarify documentation for consecutive stubbing

### LINKED ISSUE #895: Documentation missing clarification for consecutive stubbing
The [documentation](https://static.javadoc.io/org.mockito/mockito-core/2.6.3/org/mockito/Mockito.html#10) for consecutive stubbing is missing a clarification.

When multiple when/thenReturn statements are being used instead of chaining .thenReturn() statements, each stubbing overrides the existing one.

dimosr:
Fixes #895

As of now, the documentation does not clarify the difference of behaviour between chaining multiple .thenReturn() statements and using multiple when/thenReturn statements, when someone attempts to perform consecutive stubbing.

I added a warning/clarification in the documentation with a corresponding example, so that it's more clear (especially for new users of Mockito).

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -452,8 +452,8 @@ import org.mockito.verification.VerificationWithTimeout;
  *   .thenReturn("one", "two", "three");
  * </code></pre>
  *
- * <b>Warning</b>, if instead of chaining .thenReturn() calls, multiple when/thenReturn commands are used,
- * then each stubbing will override the previous one:
+ * <strong>Warning</strong> : if instead of chaining {@code .thenReturn()} calls, multiple stubbing with the same matchers or arguments
+ * is used, then each stubbing will override the previous one:
  *
  * <pre class="code"><code class="java">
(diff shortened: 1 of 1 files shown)


src/main/java/org/mockito/Mockito.java
```diff
@@ -452,6 +452,16 @@
  *   .thenReturn("one", "two", "three");
  * </code></pre>
  *
+ * <b>Warning</b>, if instead of chaining .thenReturn() calls, multiple when/thenReturn commands are used,
+ * then each stubbing will override the previous one:
```
bric3: I'd rephrase to : 

```
<strong>Warning</strong> : if instead of chaining {@code .thenReturn()} calls, multiple stubbing with same matchers or arguments are used then each stubbing will override the previous one :
```

bric3:
Hi @dimosr thank you for the javadoc update !
Could you rework the javadoc a bit ?

src/main/java/org/mockito/Mockito.java
dimosr: Sure, sounds much better. I'm going to update the PR to this.

PR #897, opened 2017-01-21
CleansEmpty value returns empty arrays

### LINKED ISSUE #876: Mockito should return an empty array for default values
Mockito returns by default interesting default values for some types : 

* Long ⇒ 0
* Optional ⇒ Optional.empty
* List => new ArrayList(0)

But it don't return an empty array currently.

bric3:
Fixes #876 and do a bit more cleanup

* Moves returning empty array from `ReturnsMoreEmtpyValues` to `ReturnsEmptyValues`
* Cleans up javadoc
* Removes reflection for Java8 types (implementation and tests)

Best to review commit by commit.


src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsMoreEmptyValues.java
```diff
@@ -61,13 +36,10 @@ public Object answer(InvocationOnMock invocation) throws Throwable {
         Class<?> returnType = invocation.getMethod().getReturnType();
         return returnValueFor(returnType);
     }
-    
+
     Object returnValueFor(Class<?> type) {
         if (type == String.class) {
             return "";
```
TimvdLippe: If this class now only returns an emptry String, why don't we remove this class and move the empty string to `ReturnsEmptyValues`?

bric3: Actually I don't think that's part of this PR. And I find that returning empty string instead of `null` more debattable (hence a separate PR).

bric3:
That's weird, there's a failure in this test org.mockito.internal.creation.bytebuddy.TypeCachingMockBytecodeGeneratorTest

Which has nothing to do with my change. Other JDK 8 builds are fine. It looks like a race condition.

src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValues.java
```diff
@@ -130,22 +129,22 @@ Object returnValueFor(Class<?> type) {
...
-            return JavaEightUtil.emptyOptionalLong();
-        } else if ("java.util.stream.Stream".equals(type.getName())) {
-            return JavaEightUtil.emptyStream();
-        } else if ("java.util.stream.DoubleStream".equals(type.getName())) {
-            return JavaEightUtil.emptyDoubleStream();
-        } else if ("java.util.stream.IntStream".equals(type.getName())) {
-            return JavaEightUtil.emptyIntStream();
-        } else if ("java.util.stream.LongStream".equals(type.getName())) {
-            return JavaEightUtil.emptyLongStream();
+        } else if (type == Optional.class) {
```
szczepiq: Oh nice! It's good to take advantage of Java8!

src/main/java/org/mockito/internal/util/JavaEightUtil.java
```diff
@@ -1,144 +0,0 @@
...
-
-import org.mockito.internal.creation.instance.InstantiationException;
-
-import java.lang.reflect.Method;
-
-/**
- * Helper class to work with features that were introduced in Java versions after 1.5.
- * This class uses reflection in most places to avoid coupling with a newer JDK.
- */
-public final class JavaEightUtil {
```
szczepiq: Fantastic! Less code to maintain.

src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesTest.java
```diff
@@ -15,21 +15,28 @@
...
 
 import static junit.framework.TestCase.assertEquals;
 import static junit.framework.TestCase.assertFalse;
-import static junit.framework.TestCase.assertNotNull;
 import static junit.framework.TestCase.assertNull;
 import static junit.framework.TestCase.assertTrue;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.mock;
 
 public class ReturnsEmptyValuesTest extends TestBase {
```
szczepiq: I like the idea to condense the tests into 2 methods.

src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsMoreEmptyValues.java
szczepiq: As general feedback, not particularly to this PR, I'm +1 to returning empty String. We can poll more feedback from the community. Empty String seems like a better default, avoiding potential NPE. The idea behind the defaults is to return the most sensible value if the user does not care about the mock.

bric3: @szczepiq Let's create another PR for default Strings, it'll be easier to track for end users as it will appear in the changelog.

Issue #892, opened 2017-01-20
Mockito 2.6.4 hangs on JDK 1.8.0_31


Stephan202:
We just upgraded to Mockito 2.6.4. When we run our tests on Travis CI without the `oracle-java8-installer`, JDK 1.8.0_31 is used. We then have two threads hanging in the following state:
```
"pool-1-thread-2" #24 prio=5 os_prio=0 tid=0x00007fcd208ca800 nid=0x27b3 in Object.wait() [0x00007fcc8c1a1000]
   java.lang.Thread.State: RUNNABLE
        at net.bytebuddy.description.type.TypeDescription$Generic.<clinit>(TypeDescription.java:319)
        at net.bytebuddy.ByteBuddy.subclass(ByteBuddy.java:363)
        at net.bytebuddy.ByteBuddy.subclass(ByteBuddy.java:334)
        at net.bytebuddy.ByteBuddy.subclass(ByteBuddy.java:231)
        at org.mockito.internal.creation.bytebuddy.SubclassBytecodeGenerator.mockClass(SubclassBytecodeGenerator.java:66)
        at org.mockito.internal.creation.bytebuddy.TypeCachingBytecodeGenerator$1.call(TypeCachingBytecodeGenerator.java:35)
        at org.mockito.internal.creation.bytebuddy.TypeCachingBytecodeGenerator$1.call(TypeCachingBytecodeGenerator.java:32)
        at net.bytebuddy.TypeCache.findOrInsert(TypeCache.java:138)
        at net.bytebuddy.TypeCache$WithInlineExpunction.findOrInsert(TypeCache.java:346)
        at net.bytebuddy.TypeCache.findOrInsert(TypeCache.java:161)
        - locked <0x0000000689673668> (a java.lang.Class for com.microsoft.windowsazure.services.servicebus.ServiceBusContract)
        at net.bytebuddy.TypeCache$WithInlineExpunction.findOrInsert(TypeCache.java:355)
        at org.mockito.internal.creation.bytebuddy.TypeCachingBytecodeGenerator.mockClass(TypeCachingBytecodeGenerator.java:30)
        at org.mockito.internal.creation.bytebuddy.SubclassByteBuddyMockMaker.createMockType(SubclassByteBuddyMockMaker.java:71)
        at org.mockito.internal.creation.bytebuddy.SubclassByteBuddyMockMaker.createMock(SubclassByteBuddyMockMaker.java:42)
[...]

PR #902, opened 2017-01-23
Fix TypeCache dead lock

raphw:
I could trace the problem to more eager resolution of types in Java u31 upon loading where concurrent mock creation with locking on the class level causes a dead lock. This can happen in other VM implementations but can be solved with a more granular lock on our type cache.

This fixes https://github.com/mockito/mockito/issues/892.

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/TypeCachingBytecodeGenerator.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/TypeCachingBytecodeGenerator.java
@@ -13,6 +13,8 @@ import java.util.concurrent.Callable;
 
 class TypeCachingBytecodeGenerator extends ReferenceQueue<ClassLoader> implements BytecodeGenerator {
 
+    private final Object BOOTSTRAP_LOCK = new Object();
+
     private final BytecodeGenerator bytecodeGenerator;
 
     private final TypeCache<SerializationFeatureKey> typeCache;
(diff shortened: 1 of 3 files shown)


bric3:
@raphw Could that be the issue that I saw in #897, see travis log : https://travis-ci.org/mockito/mockito/jobs/194254780#L360-L366

raphw:
No, that is a flaky test as we rely on triggering a garbage collection which is something we do not controll on a VM even though there is a good chance to. We should ideally rerun this test multiple times upon a failure.

bric3:
OK. Agreed.

src/main/java/org/mockito/internal/creation/bytebuddy/TypeCachingBytecodeGenerator.java
```diff
@@ -13,6 +13,8 @@
 
 class TypeCachingBytecodeGenerator extends ReferenceQueue<ClassLoader> implements BytecodeGenerator {
 
+    private final Object BOOTSTRAP_LOCK = new Object();
```
szczepiq: Is this field intended to be constant? If not, it should have lower case name.

raphw: It is true, this should be lower case but yes, it should be a constant.

bric3: However this lock should not be static.

PR #905, opened 2017-01-23
Fixes #814 : A elements were missing the name attribute

### LINKED ISSUE #814: Javadoc links in main Mockito doc are not linkable
The links to line items in the main [Mockito documentation](https://static.javadoc.io/org.mockito/mockito-core/2.3.0/org/mockito/Mockito.html) are not linkable, e.g. I cannot create a link to item no. 40 in the list.

bric3:
Meaningful links probably never worked as expected.

The `A` elements were missing the `name` attribute, so the browser couldn't find the matching part of the document.

For reference : [5.9.8 Navigating to a fragment identifier](http://www.w3.org/html/wg/drafts/html/master/browsers.html#scroll-to-fragid)

> For HTML documents (and the text/html MIME type), the following processing model must be followed to determine what the indicated part of the document is.
> 
> 1. Parse the URL, and let fragid be the <fragment> component of the URL.
> 2. If fragid is the empty string, then the indicated part of the document is the top of the document.
> 3. If there is an element in the DOM that has an ID exactly equal to fragid, then the first such element in tree order is the indicated part of the document; stop the algorithm here.
> 4. If there is an a element in the DOM that has a name attribute whose value is exactly equal to fragid, then the first such element in tree order is the indicated part of the document; stop the algorithm here.
> 5. Otherwise, there is no indicated part of the document.

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -82,7 +82,7 @@ import org.mockito.verification.VerificationWithTimeout;
  *      <a href="#40">40. (**new**) Improved productivity and cleaner tests with "stricter" Mockito (Since 2.+)</a><br/>
  * </b>
  *
- * <h3 id="0">0. <a class="meaningful_link" href="#mockito2">Migrating to Mockito 2</a></h3>
+ * <h3 id="0">0. <a class="meaningful_link" href="#mockito2" name="mockito2">Migrating to Mockito 2</a></h3>
  *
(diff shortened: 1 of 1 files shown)


bric3:
I am merging this one as this is very simple.

TimvdLippe:
Hm, don't want a quick review? I try to never self-merge as it got me numerous times already

bric3:
the change is damn simple (it's just adding `name` attribute), fixed with a simple regexp

Issue #285, opened 2015-08-21
"Argument should be a mock, but is" error when combine @InjectMocks with @Spy


vbuell:
In mockito <=2.0.12 there was possible to use @InjectMocks on spy object, but seems like this was changed since 2.0.13-beta version:

```
Argument should be a mock, but is: class mockito.bug.snippet.ArgumentShouldBeAMockButIsTest$B
```

Test case:

```
package mockito.bug.snippet;

import org.junit.Before;
import org.junit.Test;
import org.mockito.InjectMocks;
import org.mockito.Mock;
import org.mockito.MockitoAnnotations;
import org.mockito.Spy;

public class ArgumentShouldBeAMockButIsTest {

    // Inject here
    static class A {
        private B b;
    }

    // Inject here also
    static class B {
        private C c;
    }

    // Some dependency
    static class C {
    }

    @InjectMocks
    private final A a = new A();

    @Spy
    @InjectMocks
    private final B b = new B();

    @Mock
    private C c;

    @Before
    public void setUp() throws Exception {
        MockitoAnnotations.initMocks(this);
    }

    @Test
    // Worked in 2.0.12-beta
    public void testArgumentShouldBeAMockButIsNonMock() {
        b.toString();
    }
}
```

JeremybellEU:
Using the real objects PR, I was able to create a fix for it.

The problem was an internal name-retrieval of mocks during injection to fix a previous issue. The commit that caused this bug was: https://github.com/mockito/mockito/commit/110ffa80070bf54ab8efabdaaa27b193a93d6128

bric3:
Looking at that again it seems the use case is a bit adge case, and worked by chance, Mockito injection was never designed to support complicated injection stuff, like building a graph.

harek:
I've just noticed that there is still this misbehavior (undocumented feature) when you provide spy object by hand. If you run test provided by @vbuell with object created by `Mockito#spy` method then test passes (of course it's not important if there is @Spy annotation on that field).

I've reproduced this issue on latest Mockito version: 2.7.14
```
import org.junit.Before;
import org.junit.Test;
import org.mockito.InjectMocks;
import org.mockito.Mock;
import org.mockito.Mockito;
import org.mockito.MockitoAnnotations;
import org.mockito.Spy;

public class ArgumentShouldBeAMockButIsTest {

	// Inject here
	static class A {
		private B b;
	}

	// Inject here also
	static class B {
		private C c;
	}

	// Some dependency
	static class C {
[...]

PR #906, opened 2017-01-24
Cleans up the injection engine code

bric3:
This is mostly a boyscout change, reviewing commit by commit should makes more sense.

Basically the refactoring removes some old identified code duplication. And few tweaks on reporting.
More refining can be done in this area.

I'm quite unsure at this moment if allowing edge cases like in #810 is a godd thing. Yet, with the current code base it is still possible to have multi level depedency working by incoking serveral time `MockitoAnnotations.initMocks(this)`.

### DIFF
--- a/src/main/java/org/mockito/internal/configuration/injection/SpyOnInjectedFieldsHandler.java
+++ b/src/main/java/org/mockito/internal/configuration/injection/SpyOnInjectedFieldsHandler.java
@@ -23,7 +23,8 @@ public class SpyOnInjectedFieldsHandler extends MockInjectionStrategy {
     @Override
     protected boolean processInjection(Field field, Object fieldOwner, Set<Object> ignored) {
         if (field.isAnnotationPresent(Spy.class)) {
-            SpyFieldInitializer.initializeSpy(fieldOwner, field);
+            Object spy = SpyFieldInitializer.initializeSpy(fieldOwner, field);
+            ignored.add(spy);
         }
 
         return false;
--- a/src/main/java/org/mockito/internal/configuration/injection/SpyOnInjectedFieldsHandler.java
+++ b/src/main/java/org/mockito/internal/configuration/injection/SpyOnInjectedFieldsHandler.java
@@ -7,15 +7,8 @@ package org.mockito.internal.configuration.injection;
 
 import java.lang.reflect.Field;
 import java.util.Set;
-import org.mockito.Mockito;
 import org.mockito.Spy;
-import org.mockito.exceptions.base.MockitoException;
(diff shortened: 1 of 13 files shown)


bric3:
This refactoring may allow to add new unit tests.

src/main/java/org/mockito/internal/util/Checks.java
```diff
@@ -24,4 +24,11 @@
         }
         return iterable;
     }
+
+    public static <T> T check(boolean assertion, T arg, String checkedValue) {
```
TimvdLippe: There is no coverage for this check. Is it actually used?

src/main/java/org/mockito/internal/configuration/injection/SpyOnInjectedFieldsHandler.java
```diff
@@ -28,27 +21,10 @@
...
-					    .name(field.getName()));
-					setField(fieldOwner, field, mock);
-                }
-            } catch (Exception e) {
-                throw new MockitoException("Problems initiating spied field " + field.getName(), e);
-            }
+    protected boolean processInjection(Field field, Object fieldOwner, Set<Object> ignored) {
+        if (field.isAnnotationPresent(Spy.class)) {
+            Object spy = SpyFieldInitializer.initializeSpy(fieldOwner, field);
+            ignored.add(spy);
```
TimvdLippe: Nit: can be one line.

bric3: Yes I hesitated, as I wanted to make the second statement a different line. That works for me.

src/main/java/org/mockito/internal/util/Checks.java
bric3: Ha good it was added then removed after refactoring, maybe it's the reason why the code coverage is less.

bric3:
@TimvdLippe I'll look at #285 .

Also I'll have a look at improvement discussed on #680 regarding spy creation.

TimvdLippe:
@bric3 is this still something you would like to land any time soon?

bric3:
@TimvdLippe I'm still on my break form mockito. There's just too much going on in the private life and the professional life. So I probably won't work on this soon.

TimvdLippe:
@bric3 That is absolutely fine!! Take your time "off". No obligations to do anything at all :smile:

TimvdLippe:
Closing this PR as it got out-of-date with the current branch for quite a bit. Feel free to open a new PR if you have some time again.

Issue #904, opened 2017-01-23
Remove unnecessary classes


TimvdLippe:
Looking at https://codecov.io/gh/mockito/mockito/list/e92b0be0c6362157e2c8ae349a2b9d1295065d2e/src and sorting it by coverage, it appears that there are some classes that are completely untested. Taking another closer look reveals some candidate classes that can be removed. For example `MockitoLimitations` is [only used in the Reporter](https://github.com/mockito/mockito/search?utf8=%E2%9C%93&q=MockitoLimitations), the field can therefore be placed in the Reporter directly.

There are also some debugging and util classes that seem unused. If we do decide to retain them, I think we should test them to actually make sure they are working properly.

Issue #765, opened 2016-11-17
Cannot build Mockito on Windows 64b with Java 8 


scais:
Hi guys,

 I am trying to build Mockito on my Windows 64 machine with jdk1.8.0_66 and the build is failing. I tried the `master `branch and also the `release/2.x` branch, both are having the same problem. 

When I run `gradlew clean build` I am getting:

```
mockito\src\test\java\org\mockitousage\matchers\MatchersTest.java:599: error: unclosed character literal
        mock.oneArg(Character.valueOf('Ôé¼'));
                                      ^
mockito\src\test\java\org\mockitousage\matchers\MatchersTest.java:599: error: illegal character: '\u201a'
        mock.oneArg(Character.valueOf('Ôé¼'));
                                        ^
mockito\src\test\java\org\mockitousage\matchers\MatchersTest.java:599: error: illegal character: '\u00ac'
        mock.oneArg(Character.valueOf('Ôé¼'));
                                         ^
mockito\src\test\java\org\mockitousage\matchers\MatchersTest.java:599: error: unclosed character literal
        mock.oneArg(Character.valueOf('Ôé¼'));
                                          ^
mockito\src\test\java\org\mockitousage\matchers\MatchersTest.java:599: error: ';' expected
        mock.oneArg(Character.valueOf('Ôé¼'));
                                            ^
5 errors
:compileTestJava FAILED
:removeTestResources

FAILURE: Build failed with an exception.
```

After removing the line containing `        mock.oneArg(Character.valueOf('€'));` inside **MatchersTest.java** on line 599, the build works OK.

Anyone else facing similar trouble?

TimvdLippe:
On Linux this is working fine. I thought that @bric3 (who made the test in https://github.com/mockito/mockito/commit/c9e8562c994d23286d326a8ee19e2598d36ebfda) is on a Mac. I suppose we can alter the value just fine for Windows. We do not build our tests on Windows ourselves, so thanks for trying it out!

bric3:
This looks like a character encoding issue.
The last time I checked on Windows there was nothing wrong. I'll recheck when I have access to a windows box.

scais:
Thanks guys! FYI: Just checked on different Windows 64 and build fails there as well.

raphw:
Just checked on a Windows VM and the problem does not happen here. Strange.

scais:
I managed to build it on my Windows 7 64B machine by using `gradlew build -Dfile.encoding=UTF-8`

[...]

PR #912, opened 2017-01-29
Small improvements

ChristianSchwarz:
This PR fixes some compile errors and removes dead code:
 - compile error: duplicate org.mockito.package-info.java in _src/main/java_ and _src/test/java_
 - ~~compile error: char enconding error in MatchersTest -> relates to #765~~
 - dead code: Constructors -> relates to #904


TimvdLippe:
It seems to be compiling just fine on CI :O

ChristianSchwarz:
Yes, the JDT compiler seems to be more strict...

TimvdLippe:
The unicode issue should be fixed with the option encoding in the compilation step. Don't think that is necessary to be fixed here.

src/test/java/org/mockitousage/matchers/MatchersTest.java
```diff
@@ -608,7 +608,7 @@ public void same_matcher_and_nulls() {
     public void nullable_matcher() throws Exception {
         // imagine a Stream.of(...).map(c -> mock.oneArg(c))...
         mock.oneArg((Character) null);
-        mock.oneArg(Character.valueOf('€'));
```
bric3: That statement should work, there's no need to modify the source here. If something is wrong maybe it's the file encoding during the git clone or other editor that changed the encoding. On my clone : 

```
$ file src/test/java/org/mockitousage/matchers/MatchersTest.java
src/test/java/org/mockitousage/matchers/MatchersTest.java: UTF-8 Unicode c program text
```

ChristianSchwarz:
Okay, I reverted the UTF-8 change.

PR #914, opened 2017-01-29
Refactored RemoveFirstLine to a static utility class.

ChristianSchwarz:
Relates to #426 

Refactored `RemoveFirstLine` to a static utility class.

I would like to unifiy `RemoveFirstLine`, `Decamelizer` and `StringJoiner` to a new type `StringUtil`. They all manipulate strings and contain only a few methods that are not worth to live in a dedicated classes, any thoughts?


TimvdLippe:
I think refactoring these classes to `StringUtil` makes the most sense. We can do that in 1 PR instead of 4 separate. Let's see what the others think.

ChristianSchwarz:
Okay lets see, what others say! I added a commit containing the unification + missing tests. If requested I will remove the commit.

PR #920, opened 2017-01-31
Added subproject for configuration-free inline mock making.

raphw:
Added a preconfiguration for using the inline mock maker as a convenience artifact that can be included instead of creating the plugin file. This is very helpful when creating multi-module projects where the mock maker file has to repeated many times.

Once we choose to change the mock maker to be default or enable programmatic access, we can decide to drop this artifact similar to `mockito-all` from version one.

### DIFF
--- a/subprojects/inline/inline.gradle
+++ b/subprojects/inline/inline.gradle
@@ -1,0 +1,18 @@
+description = "Mockito preconfigured inline mock maker (intermediate and to be superseeded by automatic usage in a future version)"
+
+ext {
+    artifactName = 'mockito-inline'
+    bintrayAutoPublish = true
+    bintrayRepo = 'maven'
+    mavenCentralSync = false
+}
+
+apply from: "$rootDir/gradle/java-library.gradle"
+
+dependencies {
+    compile project.rootProject
+}
+
+apply from: "$rootDir/gradle/publishable-java-library.gradle"
+
+tasks.javadoc.enabled = false
--- a/subprojects/inline/src/resources/mockito-extensions/org.mockito.plugins.MockMaker
+++ b/subprojects/inline/src/resources/mockito-extensions/org.mockito.plugins.MockMaker
@@ -1,0 +1,1 @@
+mock-maker-inline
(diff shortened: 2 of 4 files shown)


subprojects/inline/inline.gradle
```diff
@@ -0,0 +1,18 @@
+description = "Mockito preconfigured inline mock maker"
```
bric3: Maybe note this preconfigured mock maker may be discontinued at some point.

raphw: Are we sure we want this? For the moment, this is the best alternative. If we removed the artifact, we would probably rather announce it somewhere. It will probably not be before Mockito 3 anyways where people are naturally going through the release notes.

bric3: Yes until we find a better way to activate final mocking either explicitly or automatically.

raphw:
Ok, added a comment to the description.

subprojects/inline/src/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
szczepiq: This will be smallest but most useful jar ever published to central :)

subprojects/inline/inline.gradle
```diff
@@ -0,0 +1,18 @@
...
+    artifactName = 'mockito-inline'
+    bintrayAutoPublish = true
+    bintrayRepo = 'maven'
+    mavenCentralSync = false
+}
+
+apply from: "$rootDir/gradle/java-library.gradle"
+
+dependencies {
+    compile project.rootProject
```
szczepiq: When you run build, can you inspect the pom files generated somewhere under the build dir, and make sure the pom has dependencies? We hit this problem in the past twice so far. I will add validation for this caveat at some point.

raphw: I checked and the created POM does contain the `mockito-core` dependency. Adding verification seems like a good idea!

raphw:
All done, I added documentation and checked the POM generation!

TimvdLippe:
W00h00!

Issue #685, opened 2016-10-09
Support constructor parameters for spying on abstract classes


szczepiq:
Nice, self-contained enhancement that makes Mockito API more robust. The implementation does not require proficiency with Mockito codebase.

### Feature

We already support spying on abstract classes by allowing parameterless constructor. However, there is no support for constructor parameters. This has been asked about long time ago. Mockito API is not robust enough if it supports mocking with constructor **but** not when one has any constructor parameters.

``` java
//current api:
Foo spy = mock(Foo.class, withSettings() .useConstructor().defaultAnswer(CALLS_REAL_METHODS));
//existing method (will remain):
MockSettings useConstructor();

//new api (change existing method):
Foo spy = mock(Foo.class, withSettings() .useConstructor("someArg").defaultAnswer(CALLS_REAL_METHODS));
//changed method:
MockSettings useConstructorArgs(Object ... args);
```

### Open questions

- in case we find multiple matching constructors, do we just use the 1st matching (option 1) or throw an exception (option 2)?

I'd say we go for option 1 because it's simpler to implement and seems more convenient (works out of the box for certain use cases). If we go for option 2 we need to inform the user what to do to resolve the problem (for example document and suggest @fluentfuture idea of creating an inner implementation)

- do we add new method or add vararg to existing useConstructor() method?

We decided that using existing method is simpler, keeps the API small and is easy to discover.

### Implementation notes

- the main complexity is to identify and detect the right constructor to use based on types of parameters supplied by the user
- we already deal with detecting constructors for the @InjectMocks functionality - there should be code to reuse

### Test coverage

- see existing tests that cover "useConstructor" method for
- ensure decent, informative exception messages
  - if user supplies wrong constructor args (wrong types, we cannot find matching constructor)
  - if the constructor throws some exception (constructors of some types have code that can fail)
  - when one uses existing parameter-less "useConstructor" method but the mocked class requires constructor args, the exception message should tell the user about new "useConstructorArgs" method.
- what if arguments supplied by the user match more than 1 constructor - either we fail gracefully with decent message or we pick one of the constructors.
- update documentation to describe new feature. Update documentation for existing parameter-less "useConstructor" method. Update documentation in main Mockito class if it references “useConstructor”.
- other use cases?

fluentfuture:
I've thought about some variant of this feature for a while, so here's some feedback.

It'd be nice if we have some concrete use cases to study.

[...]

PR #928, opened 2017-02-04
Fix whitespace issues throughout the code

mureinik:
This patch applies de-facto standards the codebsae currently has regarding whitespaces on all the java files in the project:

- Newlines were added at the end of files that didn't have them.
- Whitespaces at the end of source lines were removed.

All these changes were done by running `sed` on the codebase. The exact details can be found in the individual commit messages.


TimvdLippe:
Cool that was easier tp fix than I thought. I think we should enforce this with a simple Checkstyle rule. @mockito/developers WDYT about adding checkstyle to only enforce these two rules?

mureinik:
@TimvdLippe Adding such a checkstyle is pretty easy to do. If @mockito/developers agree this is the right way to go, I'd happily contribute such a patch, either on this PR or on a separate one, whatever makes more sense to you.

mureinik:
For the reference, here's an example of a path adding these checkstyle rules (didn't want to post a patch and clatter up this PR before there's an agreement on the subject)

[0001-Checkstyle-conf.-to-enforce-whitespace-rules.patch.gz](https://github.com/mockito/mockito/files/752243/0001-Checkstyle-conf.-to-enforce-whitespace-rules.patch.gz)

szczepiq:
@mureinik, thank you for contribution and dedicating time to Mockito!

@TimvdLippe, @mureinik, it is a great idea to enforce checkstyle! Consistent code style makes codebase easier to work with.

What is the practical benefit of enforcing whitespace at the end of line / file? I vote for enforcing practical rules that help maintaining the project VS rules that fail the build occasionally, triggering extra effort from developers, without adding much value.

If you want to fix whitespace in the project or set a checkstyle rule for it, then go for it. If checkstyle rule becomes a hindrance we can always tweak it.

@mureinik, do you want to contribute an enhancement to Mockito? How about #685?

TimvdLippe:
Discussion regarding usefulness was resolved in Slack. Will have to quickly check if there were no unintended changes. Other than that this is ready for merging!

Could you open a new PR after this one with the checkstyle config? We can continue the discussion there regarding the rules we want to enforce and how. Thanks for your time and effort!

mureinik:
@TimvdLippe Agreed. I'll open a new PR for the checkstyle stuff after this one is merged, so I can easily rebase on top of it.

@szczepiq Sure, I'll take a look at #685 - looks interesting.

szczepiq:
Great work! We discussed it on slack. Clean whitespace / checkstyle will help with git workflows for cleaner history.

@TimvdLippe, do we want this also in release/2.x for easier merges?

TimvdLippe:
@szczepiq I am cherry-picking it into release/2.x atm

PR #932, opened 2017-02-05
Checkstyle conf. to enforce whitespace rules

### LINKED ISSUE #903: Review usage of linter during build
Given the amount of stylistic changes lately in primarily @bric3 PRs, I think we should consider installing a linter to make sure the style is enforced project wide. In my opinion this linter should be extremely lenient, just enforce some very simple rules.

We can look into configuring Checkstyle such that it checks a handful of properties. We can always later decide to make the linter more strict.

mureinik:
After PR #928 cleaned up some whitespace issues in the codebase (see
details there), this patch adds the checkstyle plugin to build to
enforce the standard established there:
1. A line may not end with trailing whitespaces
2. A file must end with the newline character

Note that this PR by no means intends to provide a full checkstyle configuration to the project.
It merely introduces the framework and two basic checks discussed in PR #928. Additional patches could easily add (or remove) additional checks as the dev community sees fit.

Usage note:
With this change applied, checkstyle would be run as part fo the gradle build. If a line in a `.java` source file ends with trailing whitespaces, the build will fail and an error message would be printed to the console. Additionally, an HTML report with all the errors would be generated. 

E.g.:

```
[ant:checkstyle] /home/amureini/src/git/mockito/subprojects/android/src/main/java/org/mockito/android/internal/creation/AndroidByteBuddyMockMaker.java:14: Line has trailing spaces.
:android:checkstyleMain FAILED

* What went wrong:
Execution failed for task ':android:checkstyleMain'.
> Checkstyle rule violations were found. See the report at: file:///home/amureini/src/git/mockito/subprojects/android/build/reports/checkstyle/main.html

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED
```

Fixes #903

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC
+    "-//Puppy Crawl//DTD Check Configuration 1.2//EN"
+    "http://www.puppycrawl.com/dtds/configuration_1_2.dtd">
+<module name="Checker">
+    <!--
+     Different editors and IDEs may insert or remove trailing
+     whitespaces at the end of source lines or newline characyers
+     at the end of source files.
+
+     This changes have no actual meaning, but may cause visual spam
+     when [re]viewing diffs.
+     Having checkstyles rules to strictly enforce the defacto
+     standards of the project helps avoid such issues and make the
+     review process more efficient.
+
+     See also https://github.com/mockito/mockito/issues/903
+     -->
+    <module name="NewlineAtEndOfFile"/>
+    <module name="RegexpSingleline">
+        <!-- \s matches whitespace character, $ matches end of line. -->
+        <property name="format" value="\s+$"/>
+        <property name="message" value="Line has trailing spaces."/>
+    </module>
+</module>
+
--- a/gradle/java-library.gradle
+++ b/gradle/java-library.gradle
@@ -28,5 +28,13 @@ test {
     }
 }
 
+tasks.withType(Checkstyle) {
+    reports {
+        xml.enabled false
+        html.enabled true
+        html.stylesheet resources.text.fromFile("$rootDir/config/checkstyle/checkstyle.xsl")
+    }
+}
+
 assert project.hasProperty("artifactName") : "Please configure 'ext.artifactName' for project that wants to use 'java-library.gradle' plugin"
 archivesBaseName = project.ext.artifactName
--- a/build.gradle
+++ b/build.gradle
@@ -35,6 +35,10 @@ allprojects {
         options.warnings = false
         options.encoding = 'UTF-8'
     }
+    apply plugin: 'checkstyle'
+    checkstyle {
+       configFile = rootProject.file('config/checkstyle/checkstyle.xml')
+    }
 }
 
 configurations {
--- a/config/checkstyle/checkstyle.xsl
+++ b/config/checkstyle/checkstyle.xsl
@@ -1,0 +1,207 @@
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0">
+	<xsl:output method="html" indent="yes"/>
+	<xsl:decimal-format decimal-separator="." grouping-separator="," />
+
+	<xsl:key name="files" match="file" use="@name" />
+
+	<!-- Checkstyle XML Style Sheet by Rolf Wojtech <rolf@wojtech.de>                   -->
+	<!-- (based on checkstyle-noframe-sorted.xsl by Stephane Bailliez                   -->
+	<!--  <sbailliez@apache.org> and sf-patch 1721291 by Leo Liang)                     -->
+	<!-- Changes: 																								                      -->
+	<!--  * Outputs seperate columns for error/warning/info                             -->
+	<!--  * Sorts primarily by #error, secondarily by #warning, tertiary by #info       -->
+	<!--  * Compatible with windows path names (converts '\' to '/' for html anchor)    -->
+	<!--                                                                                -->
+	<!-- Part of the Checkstyle distribution found at http://checkstyle.sourceforge.net -->
+	<!-- Usage (generates checkstyle_report.html):                                      -->
+	<!--    <checkstyle failonviolation="false" config="${check.config}">               -->
+	<!--      <fileset dir="${src.dir}" includes="**/*.java"/>                          -->
+	<!--      <formatter type="xml" toFile="${doc.dir}/checkstyle_report.xml"/>         -->
+	<!--    </checkstyle>                                                               -->
+	<!--    <style basedir="${doc.dir}" destdir="${doc.dir}"                            -->
+	<!--            includes="checkstyle_report.xml"                                    -->
+	<!--            style="${doc.dir}/checkstyle-noframes-severity-sorted.xsl"/>        -->
+
+	<xsl:template match="checkstyle">
+		<html>
+			<head>
+				<style type="text/css">
+					.bannercell {
+					border: 0px;
+					padding: 0px;
+					}
+					body {
+					margin-left: 10;
+					margin-right: 10;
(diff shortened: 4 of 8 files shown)


gradle/java-library.gradle
```diff
@@ -1,4 +1,5 @@
 apply plugin: 'java'
+apply plugin: 'checkstyle'
```
TimvdLippe: Don't we also need the dev dependency for checkstyle?

TimvdLippe:
Oh and last thing: please include a note in the README on how to run the linter.

gradle/java-library.gradle
mureinik: Not sure what you mean by this. Could you please elaborate?

mureinik:
@TimvdLippe The checkstyle plugin runs its checks as part of the gradle build, e.g. when you run `gradle build`, `gradle check` or `gradle ciBuild`, so you'll get it automatically in Travis too. As it's a seemless part of the build, I', not sure what exactly should be added to the README.

TimvdLippe:
I just discovered that #928 was merged into master instead of release/2.x. Therefore cherry-picking that PR to release/2.x. Could you update this PR to point to release/2.x?

mureinik:
@TimvdLippe I must be misunderstanding something in the dev process here. Isn't the right procedure to first merge into master, and then cherry-pick to 2.x?

TimvdLippe:
No actually we periodically merge release/2.x into master. See https://github.com/mockito/mockito/blob/release/2.x/.github/CONTRIBUTING.md#contributing-to-mockito for more information

TimvdLippe:
release/2.x has been updated with your commit. Please update this PR to point to release/2.x, you should not have any merge issues.

TimvdLippe:
You should probably cherry pick your commit from your old branch onto a new branch which originates from release/2.x. Then you dont get all these master commits :smile:

mureinik:
@TimvdLippe I read that, but somehow didn't register the fact that you're merging 2.x into master instead of the other way round. Need to get used to that. I've redid this change locally, and will update the PR shortly once I'm sure my local build passes. Sorry for the mess.

TimvdLippe:
@mureinik No problem at all! We are very happy that you contribute to Mockito :tada:

mureinik:
@TimvdLippe should be OK now - could you take a look please?

TimvdLippe:
Okay this is working for me! One thing I would like to see, not sure if that is possible: The report generated at `build/reports/checkstyle/main.html` contains only files with 0 errors. If there are errors, the build fails (yay!) but no report is generated. Could you configure checkstyle such that the report is always generated and only shows issues? E.g. it does not list classes that have no issues.

mureinik:
@TimvdLippe good idea. I'm not an expert of checkstyle's gradle plugin, but let me play around with it a bit and see if I make it work like you suggest.

mureinik:
@TimvdLippe I couldn't find a way to generate a report only with the violating files, but I also can't reproduce the behavior of the report not being generated if there are failures, as you describe. I run `./gradlew clean build`, and get a report with all the files, both those with violations and those without. 

See the attachment with a patch that causes a checkstyle violation (by adding a bunch of spaces at the end of a source line), and the report generated with it present, showing 1 error:
[checkstyle-error-and-report.zip](https://github.com/mockito/mockito/files/755500/checkstyle-error-and-report.zip)

TimvdLippe:
@mureinik Oh right yes it is working now. It shows the error in the table. I will try to find an option to not generate the HTML for 0 errors, other than that LGTM!

TimvdLippe:
Sigh, no that is not possible. Creating an issue for Checkstyle and hope they will include it.

TimvdLippe:
Could you please apply the following diff to your PR? [diff.txt](https://github.com/mockito/mockito/files/758320/diff.txt)

This removes all files which do not have any warnings or errors :smile:

mureinik:
Having some trouble building release/2.x, will revisit tomorrow when I have some more bandwidth

mureinik:
I rebased on top of release/2.x to get PR #934 so I could build locally with JDK 1.8. During this rebase I noticed that some new trailing white spaces slipped in, so I removed them in another patch and build this patch, with the @TimvdLippe's suggested xslt magic on top of it.

gradle/java-library.gradle
szczepiq: @TimvdLippe, we don't need any extra dependency. checkstyle plugin is a part of Gradle distribution. Good question!

subprojects/android/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC
+    "-//Puppy Crawl//DTD Check Configuration 1.2//EN"
+    "http://www.puppycrawl.com/dtds/configuration_1_2.dtd">
+<module name="Checker"/>
```
szczepiq: @mureinik, do we need separate checker configurations per submodule? I'd rather us have a single consistent checkstyle rule for all java code.

config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC
+    "-//Puppy Crawl//DTD Check Configuration 1.2//EN"
+    "http://www.puppycrawl.com/dtds/configuration_1_2.dtd">
+<module name="Checker">
+    <module name="NewlineAtEndOfFile"/>
+    <module name="RegexpSingleline">
+        <!-- \s matches whitespace character, $ matches end of line. -->
```
szczepiq: Can you write a bit of xml comment here that we added the validation to make the git blame / git diff workflows easier? I suspect this rule can be annoying to engineers when the builds start failing (including me ;). It's nice to see the reminder in the code why we decided to apply those rules.

raphw:
Not a big priority for me, but all in all, it makes sense.

subprojects/android/config/checkstyle/checkstyle.xml
mureinik: I tried to reduce the scope of the change as much as possible, but you're right - this is a bad idea, and we should use one conf file for the entire project. I'll update the PR.

config/checkstyle/checkstyle.xml
mureinik: @szczepiq will do, thanks for the idea.

mureinik:
Answered @szczepiq 's change requests:

- Added details to the PR description wrt how this change is used (tl;dr - the build will fail if a checkstyle violation is introduced).
- Had all the subprojects use the same `checkstyle.xml` file
- Added some explanation inline in checkstyle.xml on why we have this mechanism

gradle/java-library.gradle
```diff
@@ -28,5 +28,13 @@ test {
     }
 }
 
+tasks.withType(Checkstyle) {
+    reports {
+        xml.enabled false
+        html.enabled true
+        html.stylesheet resources.text.fromFile("$rootDir/config/checkstyle/checkstyle.xsl")
```
bric3: Is it necessary to have a stylesheet ?

TimvdLippe: It makes searching for errors a lot easier. Else it would be thousands of rows with 0 information

mureinik: Yup, what @TimvdLippe said. Actually, he authored this style sheet and asked me to include it in the patch.

TimvdLippe:
@bric3 agreed that GitHub does not nicely integrate, but regular diff and git blame are affected too.

Given everyone is in semi-favor, I am merging this change. We can revisit this decision in a month or two to evaluate whether it is worth it :)

bric3:
Yes but this is still configurable via the command line.

* git blame

    > ```
    > -w
    >    Ignore whitespace when comparing the parent's version and the child's to find where the lines came from.
    > ```

* git diff
    > ```
    > -w, --ignore-all-space
    >      Ignore whitespace when comparing lines. This ignores differences even if one line has whitespace where the other line has none.
    > ```

PR #933, opened 2017-02-06
Support for return type 'Void' in DoesNothingAnswer

### LINKED ISSUE #927: Mockito can't unbox Void
Hi,

The following [bug](https://github.com/powermock/powermock/issues/708) was created for PowerMock:  

> PowerMockito doesn't seem to unbox java.lang.Void.
> I am writing a test for my code that uses Google cloud storage API to delete a file. Google API returns Void for deleting a file. Code works fine but when I instruct my test as PowerMockito.doNothing().when(delete).execute(); It complains that execute has a return value of Void.

But PowerMock just delegates call to Mockito in this case and Mockito throws exception: 

```
Only void methods can doNothing()!
Example of correct use of doNothing():
    doNothing().
    doThrow(new RuntimeException())
    .when(mock).someVoidMethod();
Above means:
someVoidMethod() does nothing the 1st time but throws an exception the 2nd time is called
```

I have doubts about expected behaviour. For one side, method return something, but it return a `Void` which always is just `Void` and nothing could be changed. 

Anyway, it could be fixed only on Mockito side (without dirty hacks).

ChristianSchwarz:
fixes #927 
This PR allows to use `doNothing(..)` for methods with return type `java.lang.Void` like void-methods.  

To solve this `InvocationInfo.isVoid()` checks now the return type for the pseudo-type of the keyword `void` or `Void`.


src/test/java/org/mockito/internal/stubbing/answers/DoesNothingTest.java
```diff
@@ -4,25 +4,59 @@
...
+    private Invocation invocation_void;
+    private Invocation invocation_String;
 
-public class DoesNothingTest {
+    @Before
+    public void init(){
+        mock = mock(IMethods.class);
+        
+        mock.voidMethod();
+        invocation_Void =getLastInvocation();
```
TimvdLippe: Spacing around `=`

src/main/java/org/mockito/internal/stubbing/answers/InvocationInfo.java
```diff
@@ -38,8 +38,13 @@ public boolean isValidReturnType(Class<?> clazz) {
         }
     }
 
+    /**
+     * Returns {@code true} is the return type is {@link Void} or represents the pseudo-type to the keyword {@code void}.
```
szczepiq: Very nice! Thank you for updating the documentation diligently.

PR #934, opened 2017-02-06
Fix JavaDoc generation on recent JDK 8 builds

### LINKED ISSUE #894: Javadoc generation fails with jdk1.8.0_121
When I set the jdk to jdk1.8.0_121 the build fails javadoc error. 

Full log including steps to reproduce:

```
~/mockito/src$ export JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk1.8.0_121.jdk/Contents/Home/
~/mockito/src$ ./gradlew javadoc -i


...

javadoc: error - Argument for -top contains JavaScript.
Use --allow-script-in-comments to allow use of JavaScript.
1 error
100 warnings
Problems generating Javadoc.

...

:javadoc FAILED
:javadoc (Thread[Task worker,5,main]) completed. Took 1.936 secs.

* What went wrong:
Execution failed for task ':javadoc'.
> Javadoc generation failed. Generated Javadoc options file (useful for troubleshooting): '/Users/sfaber/mockito/src/build/tmp/javadoc/javadoc.options'

* Try:
Run with --stacktrace option to get the stack trace. Run with --debug option to get more log output.

BUILD FAILED
```

TimvdLippe:
Syntax was a bit tricky, but this appears to generate the correct javadoc.options file for me.

Fixes #894

### DIFF
--- a/gradle/mockito-core/javadoc.gradle
+++ b/gradle/mockito-core/javadoc.gradle
@@ -66,7 +66,17 @@ javadoc {
     """.replaceAll(/\r|\n|[ ]{8}/, ""))
     if (JavaVersion.current().isJava8Compatible()) {
         options.addStringOption('Xdoclint:none', '-quiet')
+
+        def minorVersion = System.getProperty("java.version") =~ /1.8.0_(.*)/
+
+        // Since JDK 8 version 121 scripts are not allowed in comments. However,
+        // earlier version throw an error if an unknown option is passed.
+        // Therefore only add this option if the JDK is actually correct.
+        if (minorVersion.size() > 0 && Integer.valueOf(minorVersion[0][1]) >= 121) {
+            options.addBooleanOption('-allow-script-in-comments', true)
+        }
     }
+
 
     doLast {
         copy {


TimvdLippe:
Hm probably also need a conditional for those JDKs that do not have this option :sob:

TimvdLippe:
I am merging this now as the workflow of multiple users (including me) is severely complicated. Should be an easy change.

gradle/mockito-core/javadoc.gradle
```diff
@@ -66,8 +66,18 @@ javadoc {
     """.replaceAll(/\r|\n|[ ]{8}/, ""))
     if (JavaVersion.current().isJava8Compatible()) {
         options.addStringOption('Xdoclint:none', '-quiet')
+
+        def minorVersion = System.getProperty("java.version") =~ /1.8.0_(.*)/
+
+        // Since JDK 8 version 121 scripts are not allowed in comments. However,
```
szczepiq: Thank you for fixing the problem and documenting it diligently!

PR #935, opened 2017-02-06
New feature - enable mocking using constructor arguments

### LINKED ISSUE #685: Support constructor parameters for spying on abstract classes

mureinik:
To quote the requirement (see in-depth design in #685):
> We already support spying on abstract classes by allowing
parameterless constructor. However, there is no support for
constructor parameters.
This has been asked about long time ago. Mockito API is not robust
enough if it supports mocking with constructor but not when one has
any constructor parameters.

This patch enhances the `MockSettings#useConstrctor()` method (as suggested in the alternative API comment in the issue) and adds
optional ellipsis arguments that are passed to the constructor.

The patch streamlines the creation of mocks via constructors to a
single flow, where using a no-arg constructor or an enclosing class
are just private cases of this flow, and don't require their own
special treatment.

### DIFF
--- a/src/main/java/org/mockito/MockSettings.java
+++ b/src/main/java/org/mockito/MockSettings.java
@@ -240,19 +240,25 @@ public interface MockSettings extends Serializable {
      * OtherAbstract spy = mock(OtherAbstract.class, withSettings()
      *   .useConstructor().defaultAnswer(CALLS_REAL_METHODS));
      *
+     * //Mocking an abstract class with constructor arguments
+     * SomeAbstract spy = mock(SomeAbstract.class, withSettings()
+     *   .useConstructor("arg1", 123).defaultAnswer(CALLS_REAL_METHODS));
+     *
      * //Mocking a non-static inner abstract class:
      * InnerAbstract spy = mock(InnerAbstract.class, withSettings()
      *   .useConstructor().outerInstance(outerInstance).defaultAnswer(CALLS_REAL_METHODS));
      * </code></pre>
      *
+     * @param args The arguments to pass to the constructor. Not passing any arguments means that a parameter-less
+     *             constructor will be called
      * @return settings instance so that you can fluently specify other settings
-     * @since 1.10.12
+     * @since 2.7.14 (useConstructor with no arguments was supported since 1.10.12)
      */
     @Incubating
-    MockSettings useConstructor();
+    MockSettings useConstructor(Object... args);
 
     /**
-     * Makes it possible to mock non-static inner classes in conjunction with {@link #useConstructor()}.
+     * Makes it possible to mock non-static inner classes in conjunction with {@link #useConstructor(Object...)}.
      * <p>
      * Example:
      * <pre class="code"><code class="java">
--- a/src/main/java/org/mockito/internal/creation/MockSettingsImpl.java
+++ b/src/main/java/org/mockito/internal/creation/MockSettingsImpl.java
@@ -9,6 +9,7 @@ import org.mockito.MockSettings;
 import static org.mockito.internal.exceptions.Reporter.*;
 import org.mockito.internal.creation.settings.CreationSettings;
 import org.mockito.internal.debugging.VerboseMockInvocationLogger;
+import org.mockito.internal.util.Checks;
 import org.mockito.internal.util.MockCreationValidator;
 import org.mockito.internal.util.MockNameImpl;
 import org.mockito.listeners.InvocationListener;
@@ -18,6 +19,8 @@ import org.mockito.mock.SerializableMode;
 import org.mockito.stubbing.Answer;
 
 import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -30,6 +33,7 @@ public class MockSettingsImpl<T> extends CreationSettings<T> implements MockSett
     private static final long serialVersionUID = 4475297236197939569L;
     private boolean useConstructor;
     private Object outerClassInstance;
+    private Object[] constructorArgs;
 
     public MockSettings serializable() {
         return serializable(SerializableMode.BASIC);
@@ -95,8 +99,12 @@ public class MockSettingsImpl<T> extends CreationSettings<T> implements MockSett
         return this;
     }
 
-    public MockSettings useConstructor() {
+    public MockSettings useConstructor(Object... constructorArgs) {
+        Checks.checkNotNull(constructorArgs,
+            "constructorArgs",
+            "If you need to pass null, please cast it to the right type, e.g.: useConstructor((String) null)");
         this.useConstructor = true;
+        this.constructorArgs = constructorArgs;
         return this;
     }
 
@@ -111,6 +119,16 @@ public class MockSettingsImpl<T> extends CreationSettings<T> implements MockSett
 
     public Object getOuterClassInstance() {
         return outerClassInstance;
+    }
+
+    public Object[] getConstructorArgs() {
+        if (outerClassInstance == null) {
+            return constructorArgs;
+        }
+        List<Object> resultArgs = new ArrayList<Object>(constructorArgs.length + 1);
+        resultArgs.add(outerClassInstance);
+        resultArgs.addAll(Arrays.asList(constructorArgs));
+        return resultArgs.toArray(new Object[constructorArgs.length + 1]);
     }
 
     public boolean isStubOnly() {
--- a/src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
+++ b/src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
@@ -5,59 +5,110 @@
 package org.mockito.internal.creation.instance;
 
 import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.util.Arrays;
+import java.util.LinkedList;
+import java.util.List;
+
+import org.mockito.internal.util.Primitives;
 import org.mockito.internal.util.reflection.AccessibilityChanger;
 
 import static org.mockito.internal.util.StringUtil.join;
 
 public class ConstructorInstantiator implements Instantiator {
 
-    private final Object outerClassInstance;
+    /**
+     * Whether or not the constructors used for creating an object refer to an outer instance or not.
+     * This member is only used to for constructing error messages.
+     * If an outer inject exists, it would be the first ([0]) element of the {@link #constructorArgs} array.
+     */
+    private final boolean hasOuterClassInstance;
+    private final Object[] constructorArgs;
 
-    public ConstructorInstantiator(Object outerClassInstance) {
-        this.outerClassInstance = outerClassInstance;
+    public ConstructorInstantiator(boolean hasOuterClassInstance, Object... constructorArgs) {
+        this.hasOuterClassInstance = hasOuterClassInstance;
+        this.constructorArgs = constructorArgs;
     }
 
     public <T> T newInstance(Class<T> cls) {
-        if (outerClassInstance == null) {
-            return noArgConstructor(cls);
-        }
-        return withParams(cls, outerClassInstance);
+        return withParams(cls, constructorArgs);
     }
 
-    private static <T> T withParams(Class<T> cls, Object... params) {
+    private <T> T withParams(Class<T> cls, Object... params) {
+        List<Constructor<?>> matchingConstructors = new LinkedList<Constructor<?>>();
         try {
-            //this is kind of over-engineered because we don't need to support more params
-            //however, I know we will be needing it :)
             for (Constructor<?> constructor : cls.getDeclaredConstructors()) {
                 Class<?>[] types = constructor.getParameterTypes();
                 if (paramsMatch(types, params)) {
-                    return invokeConstructor(constructor, params);
+                    matchingConstructors.add(constructor);
                 }
+            }
+
+            if (matchingConstructors.size() == 1) {
+                return invokeConstructor(matchingConstructors.get(0), params);
             }
         } catch (Exception e) {
             throw paramsException(cls, e);
         }
-        throw noMatchingConstructor(cls);
+        if (matchingConstructors.size() == 0) {
+            throw noMatchingConstructor(cls);
+        } else {
+            throw multipleMatchingConstructors(cls, matchingConstructors);
+        }
     }
 
     @SuppressWarnings("unchecked")
-    private static <T> T invokeConstructor(Constructor<?> constructor, Object... params) throws java.lang.InstantiationException, IllegalAccessException, java.lang.reflect.InvocationTargetException {
+    private static <T> T invokeConstructor(Constructor<?> constructor, Object... params) throws java.lang.InstantiationException, IllegalAccessException, InvocationTargetException {
         AccessibilityChanger accessibility = new AccessibilityChanger();
         accessibility.enableAccess(constructor);
         return (T) constructor.newInstance(params);
     }
 
-    private static <T> InstantiationException paramsException(Class<T> cls, Exception cause) {
-        return new InstantiationException(
-                join("Unable to create instance of '" + cls.getSimpleName() + "'.",
-                     "Please ensure that the outer instance has correct type and that the target class has 0-arg constructor."),
-                cause);
+    private InstantiationException paramsException(Class<?> cls, Exception e) {
+        return new InstantiationException(join(
+                "Unable to create instance of '" + cls.getSimpleName() + "'.",
+                "Please ensure the target class has " + constructorArgsString() + " and executes cleanly.")
+                , e);
     }
 
-    private static <T> InstantiationException noMatchingConstructor(Class<T> cls) {
-        return new InstantiationException(
-                join("Unable to create instance of '" + cls.getSimpleName() + "'.",
-                     "Unable to find a matching 1-arg constructor for the outer instance.")
+    private String constructorArgTypes() {
+        int argPos = 0;
+        if (hasOuterClassInstance) {
+            ++argPos;
+        }
+        String[] constructorArgTypes = new String[constructorArgs.length - argPos];
+        for (int i = argPos; i < constructorArgs.length; ++i) {
+            constructorArgTypes[i - argPos] = constructorArgs[i] == null ? null : constructorArgs[i].getClass().getName();
+        }
+        return Arrays.toString(constructorArgTypes);
+    }
+
+    private InstantiationException noMatchingConstructor(Class<?> cls) {
+        String constructorString = constructorArgsString();
+        String outerClassString = "";
+        if (hasOuterClassInstance) {
+            outerClassString = " and that it's indeed an inner class of the passed instance of type " +
+                constructorArgs[0].getClass().getName();
+        }
+        return new InstantiationException(join("Unable to create instance of '" + cls.getSimpleName() + "'.",
+                "Please ensure that the target class has " + constructorString + outerClassString + ".")
+                , null);
+    }
+
+    private String constructorArgsString() {
+        String constructorString;
+        if (constructorArgs.length == 0 || (hasOuterClassInstance && constructorArgs.length == 1)) {
+            constructorString = "a 0-arg constructor";
+        } else {
+            constructorString = "a constructor that matches these argument types: " + constructorArgTypes();
+        }
+        return constructorString;
+    }
+
+    private InstantiationException multipleMatchingConstructors(Class<?> cls, List<Constructor<?>> constructors) {
+        return new InstantiationException(join("Unable to create instance of '" + cls.getSimpleName() + "'.",
+                "Multiple constructors could be matched to arguments of types " + constructorArgTypes() + ":",
+                join("", constructors))
                 , null);
     }
 
@@ -66,21 +117,15 @@ public class ConstructorInstantiator implements Instantiator {
             return false;
         }
         for (int i = 0; i < params.length; i++) {
-            if (!types[i].isInstance(params[i])) {
+            if (params[i] == null) {
+                if (types[i].isPrimitive()) {
+                    return false;
+                }
+            } else if ((!types[i].isPrimitive() && !types[i].isInstance(params[i])) ||
+                    (types[i].isPrimitive() && !types[i].equals(Primitives.primitiveTypeOf(params[i].getClass())))) {
                 return false;
             }
         }
         return true;
-    }
-
-    private static <T> T noArgConstructor(Class<T> cls) {
-        try {
-            return invokeConstructor(cls.getDeclaredConstructor());
-        } catch (Throwable t) {
-            throw new InstantiationException(join(
-                    "Unable to create instance of '" + cls.getSimpleName() + "'.",
-                    "Please ensure it has 0-arg constructor which invokes cleanly."),
-                                             t);
-        }
     }
 }
--- a/src/main/java/org/mockito/internal/creation/instance/DefaultInstantiatorProvider.java
+++ b/src/main/java/org/mockito/internal/creation/instance/DefaultInstantiatorProvider.java
@@ -12,8 +12,8 @@ public class DefaultInstantiatorProvider implements InstantiatorProvider {
     private final static Instantiator INSTANCE = new ObjenesisInstantiator();
 
     public Instantiator getInstantiator(MockCreationSettings<?> settings) {
-        if (settings != null && settings.isUsingConstructor()) {
-            return new ConstructorInstantiator(settings.getOuterClassInstance());
+        if (settings != null && settings.getConstructorArgs() != null) {
+            return new ConstructorInstantiator(settings.getOuterClassInstance() != null, settings.getConstructorArgs());
         } else {
             return INSTANCE;
         }
--- a/src/main/java/org/mockito/mock/MockCreationSettings.java
+++ b/src/main/java/org/mockito/mock/MockCreationSettings.java
@@ -73,6 +73,18 @@ public interface MockCreationSettings<T> {
     boolean isUsingConstructor();
 
     /**
+     * Used when arguments should be passed to the mocked object's constructor, regardless of whether these
+     * arguments are supplied directly, or whether they include the outer instance.
+     *
+     * @return An array of arguments that are passed to the mocked object's constructor. If
+     * {@link #getOuterClassInstance()} is available, it is prepended to the passed arguments.
+     *
+     * @since 2.7.14
+     */
+    @Incubating
+    public Object[] getConstructorArgs();
+
+    /**
      * Used when mocking non-static inner classes in conjunction with {@link #isUsingConstructor()}
      *
      * @return the outer class instance used for creation of the mock object via the constructor.
--- a/src/test/java/org/mockito/internal/creation/instance/ConstructorInstantiatorTest.java
+++ b/src/test/java/org/mockito/internal/creation/instance/ConstructorInstantiatorTest.java
@@ -28,25 +28,52 @@ public class ConstructorInstantiatorTest extends TestBase {
         }
     }
 
+    static class SomeClass3 {
+
+        SomeClass3(int i) {
+
+        }
+    }
+
     @Test
     public void creates_instances() {
-        assertEquals(new ConstructorInstantiator(null).newInstance(SomeClass.class).getClass(), SomeClass.class);
+        assertEquals(new ConstructorInstantiator(false, new Object[0]).newInstance(SomeClass.class).getClass(), SomeClass.class);
     }
 
     @Test
     public void creates_instances_of_inner_classes() {
-        assertEquals(new ConstructorInstantiator(this).newInstance(SomeInnerClass.class).getClass(), SomeInnerClass.class);
-        assertEquals(new ConstructorInstantiator(new ChildOfThis()).newInstance(SomeInnerClass.class).getClass(), SomeInnerClass.class);
+        assertEquals(new ConstructorInstantiator(true, this).newInstance(SomeInnerClass.class).getClass(), SomeInnerClass.class);
+        assertEquals(new ConstructorInstantiator(true, new ChildOfThis()).newInstance(SomeInnerClass.class).getClass(), SomeInnerClass.class);
+    }
+
+    @Test
+    public void creates_instances_with_arguments() {
+        assertEquals(new ConstructorInstantiator(false, "someString").newInstance(SomeClass2.class).getClass(), SomeClass2.class);
+    }
+
+    @Test
+    public void creates_instances_with_null_arguments() {
+        assertEquals(new ConstructorInstantiator(false, new Object[]{null}).newInstance(SomeClass2.class).getClass(), SomeClass2.class);
+    }
+
+    @Test
+    public void creates_instances_with_primitive_arguments() {
+        assertEquals(new ConstructorInstantiator(false, 123).newInstance(SomeClass3.class).getClass(), SomeClass3.class);
+    }
+
+    @Test(expected = InstantiationException.class)
+    public void fails_when_null_is_passed_for_a_primitive() {
+        assertEquals(new ConstructorInstantiator(false, new Object[]{null}).newInstance(SomeClass3.class).getClass(), SomeClass3.class);
     }
 
     @Test
     public void explains_when_constructor_cannot_be_found() {
         try {
-            new ConstructorInstantiator(null).newInstance(SomeClass2.class);
+            new ConstructorInstantiator(false, new Object[0]).newInstance(SomeClass2.class);
             fail();
         } catch (InstantiationException e) {
             assertThat(e).hasMessageContaining("Unable to create instance of 'SomeClass2'.\n" +
-                    "Please ensure it has 0-arg constructor which invokes cleanly.");
+                    "Please ensure that the target class has a 0-arg constructor.");
         }
     }
 }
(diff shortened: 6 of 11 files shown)


szczepiq:
Wow! Thank you for swift contribution! I will review once I have a bandwidth.

mureinik:
Apologies, the previous push had a couple of failing tests due to a missed file. I've updated it and it should work now.

mureinik:
Question: This patch was built on top of master, for the next release. Should I have built it on top of release/2.x?

TimvdLippe:
@mureinik Yes please 😄

mureinik:
@TimvdLippe done.

It's slightly less elegant in 2.x, as the source compatibility level is 1.6, not 1.8 (so no `Stream`s or `Optional`s), but the idea remains the same.
Feedback more than ready.

src/main/java/org/mockito/internal/creation/MockSettingsImpl.java
```diff
@@ -95,8 +98,9 @@ public MockSettings defaultAnswer(Answer defaultAnswer) {
         return this;
     }
 
-    public MockSettings useConstructor() {
+    public MockSettings useConstructor(Object... constructorArgs) {
```
ChristianSchwarz: @mureinik 
We should ensure that the vararg-array is not null. You can use the class Checks here.

src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
```diff
@@ -47,18 +51,11 @@ public ConstructorInstantiator(Object outerClassInstance) {
...
-                     "Please ensure that the outer instance has correct type and that the target class has 0-arg constructor."),
-                cause);
-    }
-
-    private static <T> InstantiationException noMatchingConstructor(Class<T> cls) {
-        return new InstantiationException(
-                join("Unable to create instance of '" + cls.getSimpleName() + "'.",
-                     "Unable to find a matching 1-arg constructor for the outer instance.")
-                , null);
+    private <T> InstantiationException paramsException(Class<T> cls, Exception e) {
```
ChristianSchwarz: We can leave the type parameter out and reduce `Class<T>` to `Class<?>`.

src/main/java/org/mockito/internal/creation/instance/DefaultInstantiatorProvider.java
```diff
@@ -12,8 +12,8 @@
     private final static Instantiator INSTANCE = new ObjenesisInstantiator();
 
     public Instantiator getInstantiator(MockCreationSettings<?> settings) {
-        if (settings != null && settings.isUsingConstructor()) {
-            return new ConstructorInstantiator(settings.getOuterClassInstance());
+        if (settings != null && settings.getConstructorArgs() != null) {
```
ChristianSchwarz: A check for `settings.getConstructorArgs() != null` is not required if we check that null can't be passed to `MockSetting.useConstructor(Object ...args)` as described above.

src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
```diff
@@ -5,26 +5,30 @@
...
 public class ConstructorInstantiator implements Instantiator {
 
-    private final Object outerClassInstance;
+    private final Object[] constructorArgs;
+
+    public ConstructorInstantiator(Object... constructorArgs) {
+        if (constructorArgs != null) {
+            this.constructorArgs = constructorArgs;
+        } else {
+            this.constructorArgs = new Object[0];
```
ChristianSchwarz: It should not be possible  to pass a null vararg-array here, as this probably indicates an bug upper in the call hierarchy. We should better check the precondition here instead of correcting the input. WDYT?

src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
```diff
@@ -47,18 +51,11 @@ public ConstructorInstantiator(Object outerClassInstance) {
...
-                     "Unable to find a matching 1-arg constructor for the outer instance.")
-                , null);
+    private <T> InstantiationException paramsException(Class<T> cls, Exception e) {
+        return new InstantiationException(join(
+                "Unable to create instance of '" + cls.getSimpleName() + "'.",
+                "Please ensure that it has a constructor that matches these arguments: " + Arrays.toString(constructorArgs))
+                , e);
     }
 
     private static boolean paramsMatch(Class<?>[] types, Object[] params) {
```
ChristianSchwarz: I am not sure if this alogrithm here works as expected. The JVM would choose the most specific constructor, but we choose the first matching. This can cause some trouble that is hard to detect. 

Assume we have two `String` arguments and get constructor's in this order:
1. MyClass(Object a, Object b)
2. MyClass(String a, String b)

Since `String` is an instance of `Object `we would choose the first constructor, whereas the JVM would choose the second. This should be fixed!

mureinik:
@ChristianSchwarz thanks for the comments - will take a look tomorrow, hopefully, when I have a bit more bandwidth.

src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
szczepiq: I like this suggestion!

src/main/java/org/mockito/MockSettings.java
```diff
@@ -249,10 +249,10 @@
      * @since 1.10.12
```
szczepiq: Please add documentation (e.g. no parameters means that Mockito will look for parameter-less constructor, etc.) and change the @since tag.

src/main/java/org/mockito/internal/creation/MockSettingsImpl.java
szczepiq: +1

@mureinik, can you add a test case for following scenarios:

1. There is a public constructor that takes single parameter and the user creates mock with useConstructor(null).
2. There are multiple overloaded public constructors that takes single parameter of different types, the user creates mock with useConstructor(null)

src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
szczepiq: I suggest to add test cases for this scenario and choose the solution that makes Mockito most intuitive. Thanks a lot @ChristianSchwarz for bringing this up!

src/main/java/org/mockito/mock/MockCreationSettings.java
```diff
@@ -73,6 +73,16 @@
...
 
     /**
+     * Used when arguments should be passed to the mocked object's constructor, regardless of whether these
+     * arguments are supplied directly, or whether they include the outer instance.
+     *
+     * @return An array of arguments that are passed to the mocked object's constructor. If
+     * {@link #getOuterClassInstance()} is available, it is prepended to the passed arguments.
+     */
+    @Incubating
+    public Object[] getConstructorArgs();
```
szczepiq: Great new public method! Thank you for growing the API!

I really like the idea that it automatically includes outer class instance and that it is diligently document here. Nice work.

Let's add @since tag.

src/test/java/org/mockito/internal/creation/instance/ConstructorInstantiatorTest.java
```diff
@@ -46,7 +46,7 @@ public void explains_when_constructor_cannot_be_found() {
             fail();
         } catch (InstantiationException e) {
             assertThat(e).hasMessageContaining("Unable to create instance of 'SomeClass2'.\n" +
-                    "Please ensure it has 0-arg constructor which invokes cleanly.");
+                    "Please ensure that it has a constructor that matches these arguments: []");
```
szczepiq: I think that previous exception was a bit clearer. Let's keep the logic that presents cleaner exception even at the cost of a little bit of conditional complexity. In Mockito we strive to have super clean exception messages because that's where users need most help - when troubleshooting.

src/test/java/org/mockitousage/constructor/CreatingMocksWithConstructorTest.java
```diff
@@ -52,6 +55,12 @@ public void can_spy_abstract_classes() {
     }
 
     @Test
+    public void can_spy_abstract_classes_with_constructor_args() {
```
szczepiq: Thank you very much for adding test coverage! I request more :)

As pointed out in earlier code review comments, we need coverage for ambiguous, overloaded constructor use cases. Perhaps they need better exception messages and intuitive behavior?

src/test/java/org/mockitousage/constructor/CreatingMocksWithConstructorTest.java
```diff
@@ -83,7 +92,7 @@ public void mocking_inner_classes_with_wrong_outer_instance() {
             fail();
         } catch (MockitoException e) {
             assertThat(e).hasMessage("Unable to create mock instance of type 'InnerClass'");
-            assertThat(e.getCause()).hasMessageContaining("Unable to find a matching 1-arg constructor for the outer instance.");
+            assertThat(e.getCause()).hasMessageContaining("Please ensure that it has a constructor that matches these arguments: [foo]");
```
szczepiq: I think it is more useful to the user to see the types of arguments rather than the value of the argument.

If we want to push this further, we can even inspect the existing constructor arguments for the class and list them out to the user. I don't find it necessary to merge this change, though. Food for thought.

src/main/java/org/mockito/MockSettings.java
szczepiq: Let's also update the description in main [Mockito Javadoc](https://static.javadoc.io/org.mockito/mockito-core/2.7.5/org/mockito/Mockito.html#spying_abstract_classes).

mureinik:
@szczepiq @ChristianSchwarz thanks for your comments. I was a bit swamped last week and didn't manage to address the earlier ones, will look into all of them this week - hopefully, tomorrow or Tuesday.

mureinik:
Rebased in order to apply PR #934 so I can build locally and get the CI passing, no other changes applied yet.

src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
mureinik: Good point, will do

src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
mureinik: @ChristianSchwarz @szczepiq there's no new code here - this is the same code used originally for the enclosing type usecase, so although this is indeed an issue, it's not a new one, and thus I think it should be fixed in another patch (although probably still part of this PR.

The "strictest" solution, IMHO, would be to actually pass a list of arguments types - `useConstructor(Class[] types, Objcet[] args)`.

Alternatively, we can do something implicit:
- if an argument list matches a strictly more "specific" constructor, use it
- if you can't decide (e.g., the argument list is `{"String1", "String2"}`, and there are two constructors `MyClass(Object, String)` and `MyClass(String, Object)`, throw an exception.

What do you guys think?

src/main/java/org/mockito/internal/creation/MockSettingsImpl.java
mureinik: @ChristianSchwarz - good point, will do. Thanks!

@szczepiq:
1. Good idea, will do. Thanks!
2. Before writing tests, I think we need to decide what the correct behavior would be. IMHO, this should just throw an exception, the same way java would refuse to compile `new MyClass(null)` if you had both a `MyClass(Object)` and a `MyClass(String)` constructor. What do you think?

src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
mureinik: Agreed. `ConstructorInstantiatorTest` passes `null` to it in a couple of cases, but you're right - the test should evolve together with the code, especially for internal classes. I'll change the tests to use empty arrays.

src/main/java/org/mockito/internal/creation/instance/DefaultInstantiatorProvider.java
mureinik: IIUC, this is incorrect. This check is analogous to the `isUsingConstructor()` check - the `settings` object has all sorts of properties that aren't relevant here, so even if a non-null `settings` is passed, it doesn't mean it would have constructor arguments.

src/main/java/org/mockito/MockSettings.java
mureinik: will do, thanks.

src/main/java/org/mockito/mock/MockCreationSettings.java
mureinik: will do, thanls

src/test/java/org/mockito/internal/creation/instance/ConstructorInstantiatorTest.java
mureinik: With this change there's no longer any distinction between a 0-arg constructor, a 1-arg constructor or a 12-arg constructor, so I'm not sure I can retain this exception messages. 

Perhaps the 0-arg constructor can be singled out in `ConstructorInstantiator#paramsException(Class ,Exception)`, but that's probably the only one. Would you recomend that?

src/test/java/org/mockitousage/constructor/CreatingMocksWithConstructorTest.java
mureinik: Agreed - but let's first decide what the correct behavior is and then write tests for it :-)

src/test/java/org/mockitousage/constructor/CreatingMocksWithConstructorTest.java
mureinik: Will change it to print the arg types instead. Adding the available constructors should be pretty easy if you think it's appropriate.

mureinik:
Uploaded an updated patch. Changes since last revision (in response to comments above):
- Rebased on top of current 2.x branch so that it can build locally with Java 8
- Added a null check in `useConstrutor`, allowing to clear up some other null checks
- Added documentation, javadoc and since tags
- Added some more test cases.

Open issues:
- Handling ambiguity - do we try to infer the consturctor's argument types from the argument list? If so, how do we handle ambiguity. Or actually - do we? Isn't it easier to throw up an exception?
- Exception messages, as detailed above

And of course, any other comments are more than welcome.

src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
```diff
@@ -47,40 +48,43 @@ public ConstructorInstantiator(Object outerClassInstance) {
...
-    private static <T> T noArgConstructor(Class<T> cls) {
-        try {
-            return invokeConstructor(cls.getDeclaredConstructor());
-        } catch (Throwable t) {
-            throw new InstantiationException(join(
-                    "Unable to create instance of '" + cls.getSimpleName() + "'.",
-                    "Please ensure it has 0-arg constructor which invokes cleanly."),
-                                             t);
-        }
+    private static final Map<Class<?>, Class<?>> PRIMITIVES_TO_WRAPPERS = new HashMap<Class<?>, Class<?>>();
```
szczepiq: Can we reuse existing code? For example Primitives class?

src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
```diff
@@ -5,26 +5,27 @@
...
-        }
-        return withParams(cls, outerClassInstance);
+        return withParams(cls, constructorArgs);
     }
 
-    private static <T> T withParams(Class<T> cls, Object... params) {
+    private <T> T withParams(Class<T> cls, Object... params) {
         try {
             //this is kind of over-engineered because we don't need to support more params
             //however, I know we will be needing it :)
```
szczepiq: Is the code comment up-to-date? :)

src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
```diff
@@ -47,40 +48,43 @@ public ConstructorInstantiator(Object outerClassInstance) {
...
         if (params.length != types.length) {
             return false;
         }
         for (int i = 0; i < params.length; i++) {
-            if (!types[i].isInstance(params[i])) {
+            if (params[i] == null) {
+                if (types[i].isPrimitive()) {
+                    return false;
+                }
+            } else if ((!types[i].isPrimitive() && !types[i].isInstance(params[i])) ||
```
szczepiq: Please ensure we have sufficient test coverage for all combinations of above. It's kind of hard to identify that by just looking at the review in GitHub.

src/test/java/org/mockitousage/constructor/CreatingMocksWithConstructorTest.java
```diff
@@ -70,7 +115,7 @@ public void exception_message_when_constructor_not_found() {
             fail();
         } catch (MockitoException e) {
             assertThat(e).hasMessage("Unable to create mock instance of type 'HasConstructor'");
-            assertThat(e.getCause()).hasMessageContaining("0-arg constructor");
+            assertThat(e.getCause()).hasMessageContaining("Please ensure that it has a constructor that matches these argument types: []");
```
szczepiq: In your PR all unhappy scenarios are handled by generic exception message "Please ensure that it has a constructor that matches these argument types".

I'd like to see fine tuning of the exception message for each scenario because that's how we delight our users! @Brice was working on tuning those exception messages last month, based on feedback/questions from the community. I'd rather not revert his efforts.

I think we have 2 variants:

1. 0-arg constructor (A) or n-arg constructor (B)
2. static/top-level class mocking (X) or inner class mocking (Y)

Here're my suggestions for exception messages:

AX - "Target class does not have 0-arg constructor. Either add 0-arg constructor or specify required constructor arguments."
BX - "Target class does not have matching constructor. Are supplied constructor args correct: []?"
AY - "Is provided outer instance an object of the right type?"
BY - "Target class does not have matching constructor. Is provided outer instance an object of the right type? Are supplied constructor args correct: []?"

In BY, I suggest to get rid of the 1st argument (outer instance type) because most users do not know that java generates that constructor argument automatically.

Above are suggestions - feel free to change it as you work on it.
When we work on fine tuning the exception messages, we write unit tests for scenarios, make them fail, read the messages and asses if they need further tuning, e.g. we check if the exception message accurately describes the scenario and if it informs the user what are the steps to resolve the problem.

Thank you for your patience and improving this PR!!!

src/test/java/org/mockitousage/constructor/CreatingMocksWithConstructorTest.java
szczepiq: Options for ambiguity:

1. use 1st constructor that matches
2. throw an exception

I'd say we go for option 1 because it's simpler to implement and seems more convenient (works out of the box for certain use cases).
If we go for option 2 we need to inform the user what to do to resolve the problem (for example document and suggest @fluentfuture idea of creating an inner implementation).

src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
mureinik: Good point - I'll remove it.

src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
mureinik: I'll add a couple more tests to `ConstructorInstantiatorTest` to make this more explicit, thanks.

ChristianSchwarz:
> Options for ambiguity:
> 
>    1. use 1st constructor that matches
>    2. throw an exception

3. use the constructor the compiler would choose

[Class.getConstructor(Class<?> argTypes)](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getConstructor-java.lang.Class...-) should do the trick

src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
mureinik: Wasn't aware of it - I'm still finding my way through the source code. Good idea, will do.

mureinik:
WRT handling ambiguity, I'm starting to think that throwing an exception is the only viable option here. To recap the suggestions:

> use 1st constructor that matches

`Class#getConstructors` does not guarantee anything about the order of constructors returned, making this behavior very fragile - not only could adding new constructors break existing tests, running the same tests from different JDKs could, potentially, produce different results.

> use the constructor the compiler would choose. Class.getConstructor(Class<?> argTypes) should do the trick

Unless I've completely misunderstood the suggestion, this won't work. `getConstrucor(Class<?>...)` matches the static types the constructor expects. Since we're only providing the arguments themselves, all we have is the runtime types. The following example should illustrate my point (uploaded as a `.txt` file due to github's restructions - just rename it to `.java` if you want to compile and run it):
[ConstructorFun.java.txt](https://github.com/mockito/mockito/files/784964/ConstructorFun.java.txt)

Unless anyone has a better idea, I'll amend this patch to throw an exception if there are two potential constructors that match the given argument list.

ChristianSchwarz:
@mureinik Sorry my fault, it was pretty  sure getConstructor() would make it. 

FYI: Apache-Commons-Lang has [ConstructorUtils](https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/reflect/ConstructorUtils.java#L244)
, but I guess it won't work if null arguments are involved.

mureinik:
@ChristianSchwarz `ConstructorUtils` offers a slight improvement on our suggested first approach. To quote the documentation: 

>First it checks if there is a constructor matching the exact signature. If not then all the constructors of the class are checked to see if their signatures are assignment-compatible with the parameter types. The first assignment-compatible matching constructor is returned.

We could implement such a behavior, but IMHO, it still suffers from all the problems I listed above.

mureinik:
CI seems to complain about this:
```
import static org.mockito.internal.util.StringJoiner.join;
                                       ^
  symbol:   class StringJoiner
  location: package org.mockito.internal.util
/home/travis/build/mockito/mockito/src/main/java/org/mockito/internal/util/Platform.java:11: error: static import only from classes and interfaces
import static org.mockito.internal.util.StringJoiner.join;
^
/home/travis/build/mockito/mockito/src/main/java/org/mockito/internal/util/Platform.java:55: error: cannot find symbol
            description = join(
                          ^
  symbol:   method join(String,String,String,String,String,String,String)
  location: class Platform
```

My patch does contain any such code.
Is Travis automatically rebasing it?

mureinik:
New patch includes:

- Made a decision about multiple matching constructors (An exception is thrown. In the discussion above we couldn't find any better idea).
- Improved error messages. Specifically, distinguished the 0-arg consturctor usecase (looking at the discussion above, I can't find any way that distinguighing an inner class would produce better messages).
- Added tests for above scenarios, and beefed up existing tests. Once PR #959 or something equivalent is merged, and the the build passes again ,Travis should produce coverage details on this patch. If anything is missing, I'll add tests to cover it.

As always, feedback more than welcome.

mureinik:
Latest push was rebased on top on recent build-fix in order to force CI to run and generate coverage report

mureinik:
The Travis build is done, but it doesn't report back to this page... What am I missing?

TimvdLippe:
Probably an issue with GitHub/Travis, it ain't your issue 😄

mureinik:
@TimvdLippe I figured as much. I wonder, though, if there's any way to force it to rerun. I'd really like to see the CodeCov report for this patch.

TimvdLippe:
I have restarted one of the jobs. Hopefully that should flush CodeCov/Travis

mureinik:
@szczepiq we kinda lost speed on this - would you have the time to take a look at the revised patch?

szczepiq:
@mureinik, thank you for your patience! I've been on holidays in Mexico and returned this weekend. I will review by the end of tomorrow and most likely merge :)

src/test/java/org/mockitousage/constructor/CreatingMocksWithConstructorTest.java
```diff
@@ -52,11 +58,51 @@ public void can_spy_abstract_classes() {
...
+
+    @Test
+    public void can_spy_abstract_classes_with_constructor_array_of_nulls() {
+        AbstractMessage mock = mock(AbstractMessage.class, withSettings().useConstructor(new Object[]{null}).defaultAnswer(CALLS_REAL_METHODS));
+        assertNull(mock.getMessage());
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void can_spy_abstract_classes_with_null_varargs() {
+        mock(AbstractMessage.class, withSettings().useConstructor(null).defaultAnswer(CALLS_REAL_METHODS));
```
szczepiq: Can you also add test that ensures that this work:

```java
mock(AbstractMessage.class, withSettings().useConstructor((String) null).defaultAnswer(CALLS_REAL_METHODS));
```

src/test/java/org/mockitousage/constructor/CreatingMocksWithConstructorTest.java
```diff
@@ -52,11 +58,51 @@ public void can_spy_abstract_classes() {
...
+        assertEquals("7", mock.getMessage());
+    }
+
+    @Test
+    public void can_spy_abstract_classes_with_constructor_array_of_nulls() {
+        AbstractMessage mock = mock(AbstractMessage.class, withSettings().useConstructor(new Object[]{null}).defaultAnswer(CALLS_REAL_METHODS));
+        assertNull(mock.getMessage());
+    }
+
+    @Test(expected = IllegalArgumentException.class)
```
szczepiq: Can you assert on exception message? I'd like the exception message to tell the user what to do if he needs to pass null. For example:

>"If you need to pass null, please cast it to the right type, example:
useConstructor((String) null)"

src/test/java/org/mockitousage/constructor/CreatingMocksWithConstructorTest.java
```diff
@@ -70,7 +116,25 @@ public void exception_message_when_constructor_not_found() {
...
+    static class MultipleConstructors {
+        MultipleConstructors(Integer x) {}
+        MultipleConstructors(Number x) {}
+    }
+
+    @Test
+    public void exception_message_when_multiple_constructors_found() {
+        try {
+            //when
+            MultipleConstructors mock = mock(MultipleConstructors.class, withSettings().useConstructor(123).defaultAnswer(CALLS_REAL_METHODS));
```
szczepiq: 123 is integer so we should be able to find a matching constructor here, no?

1. I'd like to see the test coverage for those scenarios:

```java
class A {}
class B extends A {}
class C extends A {}

//given constructors:
SomeClass(A a) {}
SomeClass(B b) {}
SomeClass(C c) {}

//this should work:
SomeClass(a);
SomeClass(b);
SomeClass(c);
```

2. Also, I'd like to see coverage for truly ambiguous constructors:

```java
class A {}
class B extends A {}
class C extends B {}

//given constructors
SomeClass(A a) {}
SomeClass(B b) {}

//this is ambiguous and should fail
SomeClass(c)

//not ambiguous and should work:
SomeClass(a)
SomeClass(b)
```

src/test/java/org/mockitousage/constructor/CreatingMocksWithConstructorTest.java
```diff
@@ -83,7 +147,7 @@ public void mocking_inner_classes_with_wrong_outer_instance() {
             fail();
         } catch (MockitoException e) {
             assertThat(e).hasMessage("Unable to create mock instance of type 'InnerClass'");
-            assertThat(e.getCause()).hasMessageContaining("Unable to find a matching 1-arg constructor for the outer instance.");
+            assertThat(e.getCause()).hasMessageContaining("Please ensure that the target class has a constructor that matches these argument types: [java.lang.String].");
```
szczepiq: In this scenario we should inform the user somehow that the provided outer object is most likely incorrect because we cannot find matching constructor.

szczepiq:
Also, can you rebase? There are some conflicts. Thanks!

src/test/java/org/mockito/internal/creation/instance/ConstructorInstantiatorTest.java
```diff
@@ -40,13 +47,33 @@ public void creates_instances_of_inner_classes() {
...
+    public void creates_instances_with_null_arguments() {
+        assertEquals(new ConstructorInstantiator(new Object[]{null}).newInstance(SomeClass2.class).getClass(), SomeClass2.class);
+    }
+
+    @Test
+    public void creates_instances_with_primitive_arguments() {
+        assertEquals(new ConstructorInstantiator(123).newInstance(SomeClass3.class).getClass(), SomeClass3.class);
+    }
+
+    @Test(expected = InstantiationException.class)
```
mureinik: Based on the feedback above, the error messages refer to the passed argument types, not their values. `null`s, unfortunately, don't have types, so there's no way I can conclude it.  I'd be happy to improve this, but I'm not quite sure what you're suggesting here. Could you elaborate please?

src/test/java/org/mockitousage/constructor/CreatingMocksWithConstructorTest.java
mureinik: Good idea! `Checks.checkNull` doesn't support this kind of error reporting, unfortunately, and I don't want to mess up this patch with adding it (and its tests). I'll add another patch (in this PR, of course), to do so and then use it in this patch.

src/test/java/org/mockitousage/constructor/CreatingMocksWithConstructorTest.java
mureinik: Good idea, will do.

src/test/java/org/mockitousage/constructor/CreatingMocksWithConstructorTest.java
mureinik: > 123 is integer so we should be able to find a matching constructor here, no?
But `Integer` is a subclass of `Number`, so both constructors would match. As per the discussion above, I think the safest approach would be to fail here.

In the first scenario you suggested, the only unambiguous call is `SomeClass(a)`. 
`SomeClass(b)` matches two constructors - `SomeClass(A)` and `SomeClass(B)`, and the same argument goes for `SomeClass(c)`. In fact, this is the same usecase as the `Integer` vs. `Number` above.

By the same reasoning, in the second scenario you suggest,  `SomeClass(c)` and `SomeClass(b)` are both ambiguous, but `SomeClass(a)` is not. I'll add a couple of tests to make this clearer. Again, this is the same usecase as `Integer` and `Number`.

In retrospect, this isn't a very clear test, as there's too much going on here - autoboxing and inheritence, all mixed together. I'll replace this test case with something more explicit to get the point across.

src/test/java/org/mockitousage/constructor/CreatingMocksWithConstructorTest.java
mureinik: Good point, I'll try to improve the error message.

mureinik:
Updated the PR with new patches as per @szczepiq 's review above. This revision contains:

- Improved error messages. Note that as discussed above, new patches were added to extend `Checks`' functionality. While this is needed for the error messages in this patch, it is not intrinsic to it, and may prove useful in other future work too.
- Added missing test cases, and made the existing ones more explicit
- Rebased against `relesae/2.x` and solved merge conflicts in `Mockito.java`. `@since` tags and documentation entries were updated to refer to 2.7.14 (which this patch will hopefully be a part of) instead of 2.7.12 which has already been released.

szczepiq:
Reviewing...

PR #942, opened 2017-02-12
Allow spying on interfaces so that it is convenient to work with Java 8 default methods

### LINKED ISSUE #680: Allow spying on interfaces so that it is convenient to work with Java 8 default methods
## Problem

While Mockito supports and stubs by default the _default methods_ in interfaces, the stubbing code can be a bit non intuitive, e.g.

``` java
interface DM {
    int contract();
    default int default_contract() { return contract() + 1; }
}

DM dm = mock(DM.class);
given(dm.contract()).willReturn(2);

// necessary otherwise default method is stubbed
given(dm.default_contract()).willCallRealMethod();

assertThat(dm.default_contract()).isEqualTo(3);
```

This behavior is unintuitive to users who expect the default methods to trigger real implementation (callRealMethod() by default). See also user report at #940.

## Suggested plan

:1st_place_medal: Contributions are welcome!

- [x] relax validation and allow interfaces to be spied. This way users can invoke spy(SomeInterface) or use @Spy with interfaces. This way, we don't need to mark default methods with "callRealMethod".
- [x] ensure test coverage for mocking
  - interfaces with and without default methods
  - concrete classes that extend interface with default methods (perhaps already covered)
- [x] document this use case. On main Mockito javadoc the use can search for "default" and find information about default methods behavior. Suggested by user at #940.
- [ ] create a separate ticket for Mockito 3 ("2.* incompatible" label) to discuss whether we should change the default behavior for defender/default methods. Perhaps they should automatically call real method regardless if someone uses spy() or mock() with the interface. Also we should consider mocking/spying on concrete classes that extend from interface with default methods.

## Discontinued ~~original plan~~

Below idea was discontinued:

Replace `DM` by `Map`, `default_contract()` by `getOrDefault()`, `contract()` by `get()` or `containsKey()` and you have a problem with designs that are used in the JDK itself.

I think mockito can improve on this by configuring the mock to invoke concrete default methods rather than stubbing them. This could be done the following way (_api naming in progress_) : 
- `mock(DM.class, USE_DEFAULT_METHODS)`
  
  The issue with that approach is that a default answer is mutually exclusive with other answers. e.g. if one wants to use `RETURNS_SMART_NULLS` and default methods this cannot work with the current design.
- `mock(DM.class, withSettings().useDefaultMethods())`
  
  This approach is interesting as it allows to configure the behaviour with possibly any answer. However this may require some changes with our internal answers, not a deal breaker though.

fluentfuture:
Fixes #680

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -64,7 +64,7 @@ import org.mockito.verification.*;
  *      <a href="#27">27. Delegate calls to real instance (Since 1.9.5)</a><br/>
  *      <a href="#28">28. <code>MockMaker</code> API (Since 1.9.5)</a><br/>
  *      <a href="#29">29. BDD style verification (Since 1.10.0)</a><br/>
- *      <a href="#30">30. Spying or mocking abstract classes (Since 1.10.12)</a><br/>
+ *      <a href="#30">30. Spying or mocking abstract classes (Since 1.10.12) and Java 8 default methods (Since release 2.x)</a><br/>
  *      <a href="#31">31. Mockito mocks can be <em>serialized</em> / <em>deserialized</em> across classloaders (Since 1.10.0)</a></h3><br/>
  *      <a href="#32">32. Better generic support with deep stubs (Since 1.10.0)</a></h3><br/>
  *      <a href="#32">33. Mockito JUnit rule (Since 1.10.17)</a><br/>
@@ -1008,7 +1008,7 @@ import org.mockito.verification.*;
  *
  *
  *
- * <h3 id="30">30. <a class="meaningful_link" href="#spying_abstract_classes" name="spying_abstract_classes">Spying or mocking abstract classes (Since 1.10.12)</a></h3>
+ * <h3 id="30">30. <a class="meaningful_link" href="#spying_abstract_classes" name="spying_abstract_classes">Spying or mocking abstract classes (Since 1.10.12) and Java 8 default methods (Since release 2.x)</a></h3>
  *
  * It is now possible to conveniently spy on abstract classes. Note that overusing spies hints at code design smells (see {@link #spy(Object)}).
  * <p>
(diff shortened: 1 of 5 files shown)


src/test/java/org/mockitousage/annotation/SpyAnnotationTest.java
```diff
@@ -59,19 +59,16 @@ public void should_init_spy_and_automatically_create_instance() throws Exception
     }
 
     @Test
-    public void should_prevent_spying_on_interfaces() throws Exception {
+    public void should_allow_spying_on_interfaces() throws Exception {
```
szczepiq: Nice! Thank you for updating the test!

szczepiq:
@TimvdLippe, I'm ok with not having dedicated coverage for Java8 in release/2.x. We don't have a way of doing it right now so I would not want to block this PR.

TimvdLippe:
We did have some tests for Java 8 in release 2. So it can be done. But lets drop it as requirement for this PR.

fluentfuture:
Thanks for review!

Let me know if the javadoc change looks okay to you.

src/main/java/org/mockito/Mockito.java
```diff
@@ -1008,7 +1008,7 @@
  *
  *
  *
- * <h3 id="30">30. <a class="meaningful_link" href="#spying_abstract_classes" name="spying_abstract_classes">Spying or mocking abstract classes (Since 1.10.12)</a></h3>
+ * <h3 id="30">30. <a class="meaningful_link" href="#spying_abstract_classes" name="spying_abstract_classes">Spying or mocking abstract classes (Since 1.10.12) and Java 8 default methods (Since release 2.x)</a></h3>
```
szczepiq: Nice and clean! Thank you!

szczepiq:
Restarted the Travis CI jobs and will merge as soon as they are green!

PR #975, opened 2017-03-05
Corrected a comment in exceptions package-info

L-KID:
In the description comment of exceptions package-info.java, it wrote as 'Mockito configuration utilities', which should be 'Mockito exceptions utilities'.


L-KID:
Thanks for your review! I've cleaned up the commit history. @TimvdLippe

PR #979, opened 2017-03-06
Fixes #974: Fix to get correct stubbing location with inline mocking

### LINKED ISSUE #974: MockitoJUnitRunner.StrictStubs does not detect 'Unnecessary Stubbing' when inline mock maker is enabled
- mockito-core 2.7.13
- OpenJDK 1.8.0_121
- Ubuntu 16.10
- Enable `mock-maker-inline` (or use `mockito-inline` artifact)

If running the following test, 'Unnecessary Stubbing' is not detected.

```java
@RunWith(MockitoJUnitRunner.StrictStubs.class)
public class MockitoTest {

    @Mock
    ArrayList<String> mock;

    @Test
    public void should_pass() {
        when(mock.get(0)).thenReturn("foo");
        assertEquals("foo", mock.get(0));
    }

    @Test
    public void should_detect_unnecessary_stubbing() {
        when(mock.get(1)).thenReturn("foo");
    }

}
```

In the above code, 'Unnecessary Stubbing' is detected by changing ArrayList to List.
```java

    @Mock
//    ArrayList<String> mock;
    List<String> mock;

```

This problem does not occur with default mock maker.

tmurakami:
- Fix StackTraceFilter to support inline mocking
- Add one test to StackTraceFilterTest
- Add StubbingLocationTest into inline subproject

This PR fixes #974.

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/InterceptedInvocation.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/InterceptedInvocation.java
@@ -42,14 +42,15 @@ class InterceptedInvocation implements Invocation, VerificationAwareInvocation {
                                  MockitoMethod mockitoMethod,
                                  Object[] arguments,
                                  SuperMethod superMethod,
+                                 Location location,
                                  int sequenceNumber) {
         this.mock = mock;
         this.mockitoMethod = mockitoMethod;
         this.arguments = ArgumentsProcessor.expandVarArgs(mockitoMethod.isVarArgs(), arguments);
         this.rawArguments = arguments;
         this.superMethod = superMethod;
+        this.location = location;
         this.sequenceNumber = sequenceNumber;
-        location = new LocationImpl();
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


tmurakami:
To solve the issue #974, I thought that if one stack trace element is the method frame representation of `MockMethodAdvice#handle`, the next element must be skipped.
Since the `StackTraceCleaner` API cannot handle the two related elements, I abandoned to move this logic into `DefaultStackTraceCleaner`.

> Can you describe how did you test this change?

I ran `./gradlew clean build` with this PR and confirmed that there was no problem.
Please point out me if you need additional tests.

tmurakami:
@TimvdLippe, thank you for your review.

I did not know that mock maker can be switched by environmental variables.
Tests failed with inline-mock-maker, so I'm working on it.

tmurakami:
I've just updated.

I ran the following command on release/2.x and this PR.

```
$ git branch > log.txt;
$ MOCK_MAKER=mock-maker-inline ./gradlew -Dorg.gradle.parallel=false -x removeTestResources --continue clean ciBuild 2>&1 | tee -a ~/log.txt;
$ ./gradlew removeTestResources
```

On release/2.x ([log](https://github.com/mockito/mockito/files/852805/release_2_x.txt)):
```
1830 tests completed, 8 failed, 60 skipped
```

On this PR ([log](https://github.com/mockito/mockito/files/852806/fix_issue_974.txt)):
```
1830 tests completed, 8 failed, 60 skipped
```

The same tests failed.
([This](https://github.com/mockito/mockito/files/852808/diff.txt) is difference of two logs)
There is no failure caused by this PR.

Issue #911, opened 2017-01-29
Mockito Continuous Delivery Pipeline 2.0


mockitoguy:
Proposed changes on easy-to-grasp diagrams are documented on [Mockito blog article](http://blog.mockito.org/2017/02/mockito-continuous-delivery-pipeline-2.0.html).

## Current release model

Every code change results in a new version in Central repository (not snapshot, a real version). More information on [Continuous Delivery in Mockito](https://github.com/mockito/mockito/wiki/Continuous-Delivery-Overview).

Given high rate of new versions, the community and current tooling is not comfortable in taking in new Mockito versions at fast pace. More feedback and the discussion in issue #618.

## Proposed changes

Release every change but not all releases go to Central. Push to standard repository (JCenter / Maven Central) only substantial releases. Push remaining versions to less prominent, but still public space for early adopters.

Glossary: What are major, minor, patch versions? Explanatory example: Version 2.6.5 has major=2, minor=6, patch=5.

1. Every merged pull request produces new version that lands in Bintray "mockito/maven" repo.
2. Some versions we produce are "notable versions", by default those are new major (3.0.0, 4.0.0) and new minor versions (3.1.0, 3.2.0). In contrast to regular new versions, the notable versions are included in well-known central repositories: JCenter and Maven Central.

For more details, see the wiki page paragraph on [continuous delivery](https://github.com/mockito/mockito/wiki/Continuous-Delivery-Details#cd).

## Action list
- [x] - Configure 2.7.22 as last notable version because it was the last version published to central
- Release notes:
  - [ ] - detailed release notes link to notable release notes and vice versa
  - [x] - release notes inform where the jars can be found (https://github.com/mockito/mockito-release-tools/issues/4)
  - [ ] - link to wiki documentation
- [ ] - Publish next notable version
- [x] - procedure of promoting to notable release is documented - https://github.com/mockito/mockito/blob/release/2.x/README.md#how-to-release-new-version
- [x] - the new release model is comprehensively documented - https://github.com/mockito/mockito/wiki/Continuous-Delivery-Details#cd

Optional:

- [ ] - release notes have clean format (https://github.com/mockito/mockito-release-tools/issues/3)
- [ ] - notable release (target repo) can be forced by commit message

## Implementation details

- https://github.com/mockito/mockito/wiki/Continuous-Delivery-Details#cd
- https://github.com/mockito/mockito/blob/release/2.x/README.md#how-to-release-new-version

TimvdLippe:
[...]

PR #980, opened 2017-03-07
Fixes #976: Resolve ambiguous constructors

### LINKED ISSUE #976: Make MockSettings.useConstructor able to choose the right constructor
## Use case

We recently added a very nice new feature #935 - ability to mock abstract classes / use real constructor for creating mocks. Currently we fail fast when multiple constructors are found matching provided arguments. However, Mockito could potentially do a better job identifying which constructor to use. Example scenario:

```java
class A {}
class B extends A {}
class C extends B {}

class Foo {
  Foo(A a) {};
  Foo(B b) {};
}

//Below fails with current implementation:
mock(Foo, withSettings().useConstructor(new A()));
mock(Foo, withSettings().useConstructor(new B()));
//However, we could make above work because the constructor argument are not really ambiguous

//Example use case of ambiguous constructor that we can remain failing on
mock(Foo, withSettings().useConstructor(new C()));
```

## Implementation

To get started, see CreatingMocksWithConstructorTest class. Unless the code changes, the interesting test method is [this one](https://github.com/mockito/mockito/blob/release/2.x/src/test/java/org/mockitousage/constructor/CreatingMocksWithConstructorTest.java#L150).

mureinik:
With the current code (introduced in Mockito 2.7.14 by commit 6a82c03), calling `MockSettings.useConstructor` with an argument list that would be applicable to more than one constructor would fail with an `org.mockito.internal.creation.instance.InstantiationException`.

This behavior, however, is suboptimal, as described in issue #976, as it makes `useConstructor` less robust than the Java compiler, which is able to resolve such ambiguities.
With this patch, Mockito will attempt to match the constructor with the most specific parameter types. A constructor X is considered more specific than a constructor Y if:

1. They are both applicable to the given argument list
2. Constructor X has at least one parameter which is a further specialization of the corresponding parameter of constructor Y (i.e. `paramX.isAssignableFrom(paramY)`).
3. Constructor Y has no parameter which is a further specialization of the corresponding parameter of constructor X, as defined above.

E.g., consider the following class:

    public class SomeClass {
        SomeClass(Object o) {}
        SomeClass(String s) {}
    }

Without this patch, calling

    mock(SomeClass.class, withSettings().useConstructor("string!"))

would fail. With this patch, such a call would invoke the `SomeClass(String)` constructor.

As noted above, this PR fixes issue #976.

### DIFF
--- a/src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
+++ b/src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
@@ -40,7 +40,7 @@ public class ConstructorInstantiator implements Instantiator {
             for (Constructor<?> constructor : cls.getDeclaredConstructors()) {
                 Class<?>[] types = constructor.getParameterTypes();
                 if (paramsMatch(types, params)) {
-                    matchingConstructors.add(constructor);
+                    evaluateConstructor(matchingConstructors, constructor);
                 }
             }
 
@@ -129,5 +129,56 @@ public class ConstructorInstantiator implements Instantiator {
             }
         }
         return true;
+    }
+
+    /**
+     * Evalutes {@code constructor} against the currently found {@code matchingConstructors} and determines if
+     * it's a better match to the given arguments, a worse match, or an equivalently good match.
+     * <p>
+     * This method tries to emulate the behavior specified in
+     * <a href="https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.12.2">JLS 15.12.2. Compile-Time
+     * Step 2: Determine Method Signature</a>. A constructor X is deemed to be a better match than constructor Y to the
+     * given argument list if they are both applicable, constructor X has at least one parameter than is more specific
+     * than the corresponding parameter of constructor Y, and constructor Y has no parameter than is more specific than
+     * the corresponding parameter in constructor X.
+     * </p>
+     * <p>
+     * If {@code constructor} is a better match than the constructors in the {@code matchingConstructors} list, the list
+     * is cleared, and it's added to the list as a singular best matching constructor (so far).<br/>
+     * If {@code constructor} is an equivalently good of a match as the constructors in the {@code matchingConstructors}
+     * list, it's added to the list.<br/>
+     * If {@code constructor} is a worse match than the constructors in the {@code matchingConstructors} list, the list
+     * will remain unchanged.
+     * </p>
+     *
+     * @param matchingConstructors A list of equivalently best matching constructors found so far
+     * @param constructor The constructor to be evaluated against this list
+     */
+    private void evaluateConstructor(List<Constructor<?>> matchingConstructors, Constructor<?> constructor) {
+        boolean newHasBetterParam = false;
+        boolean existingHasBetterParam = false;
+
+        Class<?>[] paramTypes = constructor.getParameterTypes();
+        for (int i = 0; i < paramTypes.length; ++i) {
+            Class<?> paramType = paramTypes[i];
+            if (!paramType.isPrimitive()) {
+                for (Constructor<?> existingCtor : matchingConstructors) {
+                    Class<?> existingParamType = existingCtor.getParameterTypes()[i];
+                    if (paramType != existingParamType) {
+                        if (paramType.isAssignableFrom(existingParamType)) {
+                            existingHasBetterParam = true;
+                        } else {
+                            newHasBetterParam = true;
+                        }
+                    }
+                }
+            }
+        }
+        if (!existingHasBetterParam) {
+            matchingConstructors.clear();
+        }
+        if (newHasBetterParam || !existingHasBetterParam) {
+            matchingConstructors.add(constructor);
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


mureinik:
I wonder why Codecov has no previous report to compare against. Do any of the maintainers have a clue?

TimvdLippe:
@mureinik Because commits by the delivery drone do `[ci skip]` and therefore do not trigger a code coverage report.

mureinik:
@TimvdLippe That means that any PR that's rebased on top of a release won't have the benefit of a comparative CodeCov report. I get why the delivery drone should skip CI, but this is an unfortunate side effect. I'll bring this up on the mailing list - this PR probably isn't the right place to discuss it.

And thanks for the explanation!

TimvdLippe:
@mureinik please see #493 for reference.

mureinik:
@szczepiq ping? Will you have some time to take a look at this please?

szczepiq:
@mureinik, I'm very keen to review this change. Thank you for your submission!

I'm allocating most of my time to https://github.com/mockito/mockito/issues/911 currently but I promise to review (and hopefully merge) by the end of this weekend.

Thanks for patience with us :)

mureinik:
Rebased on top of 2.7.17 and added `CreatingMocksWithConstructorTest#can_spy_ambiguius_constructor_with_primitive` as per @TimvdLippe's suggestion above.

Note that this API does support implicit promotions (e.g., passing an `int` in `useConstructors()` varargs will autobox it to an `Integer`. This integer can't be matched to a constructor expecting a `long` primitive or a `Long` wrapper), in the same fashion that the `java.lang.reflect` APIs don't support it.

src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
```diff
@@ -130,4 +127,32 @@ private static boolean paramsMatch(Class<?>[] types, Object[] params) {
         }
         return true;
     }
+
+    private void evaluateConstructor(List<Constructor<?>> matchingConstructors, Constructor<?> constructor) {
```
szczepiq: Can you add a javadoc describing this method? It has non trivial complexity and it would be handy to understand why we are implementing the algorithm this way.

src/test/java/org/mockitousage/constructor/CreatingMocksWithConstructorTest.java
```diff
@@ -137,45 +137,90 @@ public void exception_message_when_constructor_not_found() {
     static class ExtendsExtendsBase extends ExtendsBase {}
```
szczepiq: I love the new tests. Thank you for adding comprehensive coverage!!!

src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
```diff
@@ -107,10 +107,7 @@ private String constructorArgsString() {
     private InstantiationException multipleMatchingConstructors(Class<?> cls, List<Constructor<?>> constructors) {
         return new InstantiationException(join("Unable to create instance of '" + cls.getSimpleName() + "'.",
                 "Multiple constructors could be matched to arguments of types " + constructorArgTypes() + ":",
-                join("", " - ", constructors),
-                "If you believe that Mockito could do a better join deciding on which constructor to use, please let us know.",
-                "Ticket 685 contains the discussion and a workaround for ambiguous constructors using inner class.",
-                "See https://github.com/mockito/mockito/issues/685"
+                join("", " - ", constructors)
```
szczepiq: Let's keep the extra information. I think it is useful to:
 - inform about the inner class pattern for working around the ambiguous constructors
 - ask for feedback - it does not hurt to do that

You can keep the original or modify it if you want. I'd like to keep at least the hint about inner class.

mureinik: I originally took this message as a "workaround" until this PR is implemented, but you're right, let's keep it. Note there's a typo there ("join" instead of "job"). I'll fix that while I'm at it.

src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
mureinik: Good point, thanks.
I'll add some javadoc to explain this.

mureinik:
Pushed updated branch. Change from previous push:

- Returned the error message regarding issue #685, as per @szczepiq review above
- Fixed a Freudian slip in the aforementioned error message ("a better **join**" -> "a better **job**")
- Added javadoc explaining the logic of `org.mockito.internal.creation.instance.ConstructorInstantiator#evaluateConstructor`, as per @szczepiq  review.

mureinik:
The build failure seems to be a TravisCI issue, not an issue with the code itself, as this revision introduced only minor changes (added javadoc and fixed a typo) compared to the previous one, and some of the Travis checks are passing. 

The error:

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

@szczepiq @TimvdLippe Is there any way to re-trigger Travis on this PR?

TimvdLippe:
I have restarted the builds

PR #983, opened 2017-03-08
Break cyclical compile time dependency on hideRecursiveCall

### LINKED ISSUE #982: Break cyclical compile time dependency on hideRecursiveCall
I am working on upgrading the version of mockito in AOSP to 2.7.13 using byte buddy and byte buddy android rather than dexmaker.

As part of that I need to exclude a number of classes which will not compile on Android because it does not support Java agents/instrumentation. The mockito code has always been very well organized and made that very simple to do but in the latest version I found a minor issue.

The files I need to exclude are:
- src/main/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMaker.java
- src/main/java/org/mockito/internal/creation/bytebuddy/InlineBytecodeGenerator.java
- src/main/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMaker.java

Unfortunately, when I do that I get a compile error because MockMethodAdvice is trying to use hideRecursiveCall from InlineByteBuddyMockMaker. I looked at the code and it seems as though InlineByteBuddyMockMaker depends on InlineBytecodeGenerator which depends on MockMethodAdvice which depends on hideRecursiveCall from InlineByteBuddyMockMaker.

The only places that actually use hideRecursiveCall are MockMethodAdvice and InlineByteBuddyMockMakerTest. That depends on InlineByteBuddyMockMaker which depends on InlineBytecodeGenerator which depends on MockMethodAdvice which depends on hideRecursiveCall from InlineByteBuddyMockMaker.

So, moving hideRecursiveCall from InlineByteBuddyMockMaker to MockMethodAdvice breaks a dependency cycle and allows the InlineByteBuddyMockMaker.java file to be excluded without problem. Also, MockMethodAdvice seems to be a better place for it.

If you are happy with this proposal or can suggest a better place for hideRecursiveCall then I am happy to submit a fix for this.

paulduffin:
Fixes #982.

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMaker.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMaker.java
@@ -311,23 +311,4 @@ public class InlineByteBuddyMockMaker implements ClassCreatingMockMaker {
         };
     }
 
-    static Throwable hideRecursiveCall(Throwable throwable, int current, Class<?> targetType) {
-        try {
-            StackTraceElement[] stack = throwable.getStackTrace();
-            int skip = 0;
-            StackTraceElement next;
-            do {
-                next = stack[stack.length - current - ++skip];
-            } while (!next.getClassName().equals(targetType.getName()));
-            int top = stack.length - current - skip;
-            StackTraceElement[] cleared = new StackTraceElement[stack.length - skip];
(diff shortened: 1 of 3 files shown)


raphw:
LGTM. Probably a left over from the early phase when I refactored the mock makers.

TimvdLippe:
Nice, merging! Thanks for your quick reviews @raphw !

PR #986, opened 2017-03-14
Made CaturingMatcher threadsafe

SimY4:
In my tests I faced two problems with current capturing matcher.
1. It exposes its internal store through `#getAllValues()`.
2. Internal data store in CapturingMatcher is not threadsafe. Exposing internal store when interraction with mocked object isn't finished may produce all kind of weird exceptions in tests.

### DIFF
--- a/src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
+++ b/src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
@@ -11,13 +11,19 @@ import static org.mockito.internal.exceptions.Reporter.noArgumentValueWasCapture
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 @SuppressWarnings("unchecked")
 public class CapturingMatcher<T> implements ArgumentMatcher<T>, CapturesArguments, VarargMatcher, Serializable {
 
-    private final List<Object> arguments = Collections.synchronizedList(new ArrayList<Object>());
+    private final List<Object> arguments = new ArrayList<Object>();
+
+    private final ReadWriteLock lock = new ReentrantReadWriteLock();
+    private final Lock readLock = lock.readLock();
+    private final Lock writeLock = lock.writeLock();
 
     public boolean matches(Object argument) {
         return true;
@@ -28,20 +34,33 @@ public class CapturingMatcher<T> implements ArgumentMatcher<T>, CapturesArgument
     }
 
     public T getLastValue() {
-        synchronized (arguments) {
+        readLock.lock();
+        try {
             if (arguments.isEmpty()) {
                 throw noArgumentValueWasCaptured();
             }
 
             return (T) arguments.get(arguments.size() - 1);
+        } finally {
+            readLock.unlock();
         }
     }
 
     public List<T> getAllValues() {
-        return Arrays.asList((T[]) arguments.toArray());
+        readLock.lock();
+        try {
+            return Arrays.asList((T[]) arguments.toArray());
+        } finally {
+            readLock.unlock();
+        }
     }
 
     public void captureFrom(Object argument) {
-        this.arguments.add(argument);
+        writeLock.lock();
+        try {
+            this.arguments.add(argument);
+        } finally {
+            writeLock.unlock();
+        }
     }
 }
--- a/src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
+++ b/src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
@@ -9,13 +9,15 @@ import org.mockito.ArgumentMatcher;
 import static org.mockito.internal.exceptions.Reporter.noArgumentValueWasCaptured;
 
 import java.io.Serializable;
-import java.util.LinkedList;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 
 @SuppressWarnings("unchecked")
 public class CapturingMatcher<T> implements ArgumentMatcher<T>, CapturesArguments, VarargMatcher, Serializable {
 
-    private final LinkedList<Object> arguments = new LinkedList<Object>();
+    private final List<Object> arguments = Collections.synchronizedList(new ArrayList<Object>());
 
     public boolean matches(Object argument) {
         return true;
@@ -26,16 +28,17 @@ public class CapturingMatcher<T> implements ArgumentMatcher<T>, CapturesArgument
     }
 
     public T getLastValue() {
-        if (arguments.isEmpty()) {
-            throw noArgumentValueWasCaptured();
+        synchronized (arguments) {
+            if (arguments.isEmpty()) {
+                throw noArgumentValueWasCaptured();
+            }
+
+            return (T) arguments.get(arguments.size() - 1);
         }
-
-        return (T) arguments.getLast();
-
     }
 
     public List<T> getAllValues() {
-        return (List) arguments;
+        return Arrays.asList((T[]) arguments.toArray());
     }
 
     public void captureFrom(Object argument) {
--- a/src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
+++ b/src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
@@ -10,7 +10,6 @@ import static org.mockito.internal.exceptions.Reporter.noArgumentValueWasCapture
 
 import java.io.Serializable;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
@@ -49,7 +48,7 @@ public class CapturingMatcher<T> implements ArgumentMatcher<T>, CapturesArgument
     public List<T> getAllValues() {
         readLock.lock();
         try {
-            return Arrays.asList((T[]) arguments.toArray());
+            return new ArrayList<T>((List) arguments);
         } finally {
             readLock.unlock();
         }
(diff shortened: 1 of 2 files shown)


SimY4:
@TimvdLippe sorry about that. Fixed tests, all green now.

src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
```diff
@@ -26,16 +28,17 @@ public String toString() {
...
+                throw noArgumentValueWasCaptured();
+            }
 
+            return (T) arguments.get(arguments.size() - 1);
+        }
     }
 
     public List<T> getAllValues() {
-        return (List) arguments;
+        return Arrays.asList((T[]) arguments.toArray());
```
TimvdLippe: This feels very hacky. I think instead we should update `arguments` to `List<T>` instead.

src/test/java/org/mockito/internal/matchers/CapturingMatcherTest.java
```diff
@@ -52,4 +54,20 @@ public void should_scream_when_nothing_yet_captured() throws Exception {
...
+        final CapturingMatcher<String> m = new CapturingMatcher<String>();
+
+        //when
+        m.captureFrom("concurrent access");
+        Iterator<String> iterator = m.getAllValues().iterator();
+        m.captureFrom("concurrent access");
+
+        //then
+        assertThat(iterator.hasNext()).isTrue();
+        assertThat(iterator.next()).isEqualTo("concurrent access"); // Potential ConcurrentModificationException
```
TimvdLippe: Is it certain that without this fix the Exception is always thrown? If not, it would be nice to make this deterministic (although that probably is hard to achieve :cry: )

SimY4: I thins particular test that I added - yes. It simulates the case of unsafe publication of internal storage when interaction with Captor is not finished yet. Here I extracted internal state and created iterator over it. Next captured value will mark iterator as dirty and calling `next()` method on it will cause ConcurrentModificationException.

So in unchanged code it will fail in 100% of the times.

But in my tests I face other errors as well, like NullPointerException on `getLastItem()` but you're right, it's very hard to reproduce it in test.

src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
SimY4: I'm happy to change it to `List<T>`. But that may cause weird class cast exceptions during Captor interaction during test. In current approach class cast may only happen when you work with Captor directly through its API.

SimY4: @ChristianSchwarz why not? Altering the list is probably too extreme, but holding a iterator as I did in test, while interaction with the mock wasn't finished yet will cause problems.

src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
```diff
@@ -26,16 +28,17 @@ public String toString() {
     }
 
     public T getLastValue() {
-        if (arguments.isEmpty()) {
-            throw noArgumentValueWasCaptured();
-        }
-
-        return (T) arguments.getLast();
+        synchronized (arguments) {
```
SimY4: @ChristianSchwarz because `.add` method on synchronized collection is guaranteed to be serialized - no need to additional synchronization.

src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
ChristianSchwarz: The user may alter the returned list but in fact it is not intended by the API. Therefore it should be forebidden e.g. by an returning a copy or an readonly view.

SimY4: @ChristianSchwarz Sorry, I didn't understand your original comment correctly. Yes, I agree, it should be forbidden to change the contents of this collection and returning immutable collection here is the right thing to do. In fact, I'm already doing so - `Arrays.asList()` factory creates immutable lists. You can't alter the contents of a list produced by this factory.

ChristianSchwarz: @SimY4 
Note that you must wrap this code in an synchronized block. Otherwise there is sitll room for a race condition when an argument is added in `captureFrom(..)`.

src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
ChristianSchwarz: @SimY4 
It doesn't work cause the synchronized List uses its own lock. A ReadWriteLock is a better choice here or a least a dedicated lock-Object, using a List as a lock violates seperation of concerns and should be avoided.

SimY4: @ChristianSchwarz just an FYI - it is [recommended in java documentation](https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#synchronizedList(java.util.List)) to synchronise on a list instance to perform operations on it.

SimY4: If you think about it, it makes perfect sense. When you publish your collection, you want to make sure that everyone use the same lock when accessing your collection. That mean you have to pass the lock instance along with your collection reference. But since every object can be a synchronization primitive, you don't need to do that - you just synchronize on a collection itself to make operations serial.

ChristianSchwarz: I don't understand how you prevent a race condition if you only synchronize the read/getLastValue() operation but not not the write/captureFrom() operation. Again the synchronizedList uses it's own private lock object to make single operations atomar.
It is still possible that one Thread add a new argument while an other Thread is in `getLastValue()` after `arguments.size(..)` and before  `arguments.get(..)`.

ChristianSchwarz:
@SimY4 
Your implementation doesn't fix the test you provided in #1037. 

> 2. Internal data store in CapturingMatcher is not threadsafe. Exposing internal store when interraction with mocked object isn't finished may produce all kind of weird exceptions in tests.

Can you explain what exactly goes wrong?

src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
SimY4: @ChristianSchwarz that makes sence. Sorry for taking me so much time to understand that. I introduced locks to prevent this, can you please have a look?

ChristianSchwarz:
@SimY4 Why do we need to synchonize/guard the code at all? The list is final so i wonder what can go wrong here.

SimY4:
@ChristianSchwarz Let me quickly describe my setup. I'm writing an integration test using spring test facilities (which uses mockito underneath) to mock communications with external systems.
This test intensively and concurrently hits argument captor of one of the created mocks.
Test conditions are getting verified several time during the lifetime of the test lifecycle (using Awaitility).

Under this conditions the usage of argument captor produces weird exceptions every once in a while, like `ArrayIndexOutOfBoundException`.

src/test/java/org/mockito/internal/matchers/CapturingMatcherTest.java
szczepiq: This test is very smart, thank you!

src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
```diff
@@ -9,13 +9,21 @@
...
 import java.util.List;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 @SuppressWarnings("unchecked")
 public class CapturingMatcher<T> implements ArgumentMatcher<T>, CapturesArguments, VarargMatcher, Serializable {
 
-    private final LinkedList<Object> arguments = new LinkedList<Object>();
+    private final List<Object> arguments = new ArrayList<Object>();
```
szczepiq: Good idea to use array list because the matcher offers API to access random elements by index, so your improvement makes the code faster :)

src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
```diff
@@ -9,13 +9,21 @@
...
+
+    private final ReadWriteLock lock = new ReentrantReadWriteLock();
```
szczepiq: I think that it is a good idea to use this kind of lock. I also thought about using [CopyOnWriteArrayList](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CopyOnWriteArrayList.html) but your implementation is probably better, because in extreme cases there can be many writes and little reads.

src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
```diff
@@ -26,19 +34,33 @@ public String toString() {
...
+        } finally {
+            readLock.unlock();
+        }
     }
 
     public List<T> getAllValues() {
-        return (List) arguments;
+        readLock.lock();
+        try {
+            return Arrays.asList((T[]) arguments.toArray());
```
szczepiq: Why not just return ```new ArrayList()``` or ```new LinkedList()```? This way, we would be a little bit more backwards compatible in some edge cases (like somebody adding stuff the list returned by the capturing matcher).

szczepiq:
I reviewed the change and I like it. @mockito/developers, are there any concerns with this PR? @TimvdLippe,  @ChristianSchwarz - thank you guys for reviewing the fix! Are you happy with explanations provided by @SimY4? I am and I'm happy to merge.

@SimY4 - thank you for contribution! It's nice that you submitted PR to Mockito to fix the root cause even though you already have a workaround in your test code. I suggested one modification but otherwise I'm happy to merge.

src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
SimY4: @szczepiq Hi! Thanks for reviewing these changes. I'll change this code to new ArrayList instantiation.

szczepiq:
Sounds good!

src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
szczepiq: Sounds good!

szczepiq:
Nice! Thank you for merging.

PR #988, opened 2017-03-16
Fix #952: fix invalid examples in javadoc of AnswerX and VoidAnswerX.

### LINKED ISSUE #952: Javadoc, Answer1 code example does not compile.
I believe that all interfaces: Answer1, Answer2 and so on have bad code example in javadoc.

For example Answer1 has this: 

`   when(mock.someMethod(anyString())).thenAnswer(new Answer<Integer, String>() {
       Integer answer(String arg) {
           return arg.length();
       }
   });`

which does not compile. 
First: Answer1 should be there instead of Answer 
Second: thenAnswer method does not accept Answer1 

BTW. I personally consider naming classes or interfaces with digits as last character as poorly named construct. 
I know that there is a "fashion" right now to give names like in some functional languages, but still bothers me as something that is not aesthetic at all. Just a thought.

SaiyiW:
Hi,
I read in [#952](https://github.com/mockito/mockito/issues/952) that examples in interfaces AnswerX(Answer1, Answer2, etc.) and VoidAnswerX do not work. Thanks to an example correction by  Christian Schwarz in  the comment of the issue, I made some modifications in those files. If this is of any help, I'd be so happy! But do not hesitate to tell me if I did anything wrong! Sorry up front if I bring any inconvenience to you!

Below are some thoughts of mine during and after the correction work:  
- I've tried those examples in my own simple tests, luckily they all work fine. 
- I noticed that Christian also gave an example on using the interface with lambda expression. I've also worked out the "lambda" versions regarding AnswerX interfaces but not with VoidAnswerX(at least not with the same case, the throwing exception usage, in the current example). So I decide not to include the lambda version in AnswerX, either(for consistency). But if you want them I'm happy to add them(probably with your kind advice)! 
- In the examples in VoidAnswerX I have to also import org.mockito.stubbing.VoidAnswerX to build the test successfully, which appears a little bit strange:-(  

Thanks for your review!!  

Regards,
Saiyi

### DIFF
--- a/src/main/java/org/mockito/stubbing/Answer.java
+++ b/src/main/java/org/mockito/stubbing/Answer.java
@@ -13,15 +13,15 @@ import org.mockito.invocation.InvocationOnMock;
  * Example of stubbing a mock with custom answer:
  *
  * <pre class="code"><code class="java">
- * import static org.mockito.AdditionalAnswers.answer;
  *
- * when(mock.someMethod(anyString())).then(answer(
+ * when(mock.someMethod(anyString())).thenAnswer(
  *     new Answer() {
  *         public Object answer(InvocationOnMock invocation) {
  *             Object[] args = invocation.getArguments();
  *             Object mock = invocation.getMock();
  *             return "called with arguments: " + Arrays.toString(args);
- *         }}));
+ *         }
+ * });
  *
  * //Following prints "called with arguments: [foo]"
  * System.out.println(mock.someMethod("foo"));
--- a/src/main/java/org/mockito/stubbing/Answer.java
+++ b/src/main/java/org/mockito/stubbing/Answer.java
@@ -13,13 +13,15 @@ import org.mockito.invocation.InvocationOnMock;
  * Example of stubbing a mock with custom answer:
  *
  * <pre class="code"><code class="java">
- * when(mock.someMethod(anyString())).thenAnswer(new Answer() {
- *     Object answer(InvocationOnMock invocation) {
- *         Object[] args = invocation.getArguments();
- *         Object mock = invocation.getMock();
- *         return "called with arguments: " + Arrays.toString(args);
- *     }
- * });
+ * import static org.mockito.AdditionalAnswers.answer;
+ *
+ * when(mock.someMethod(anyString())).then(answer(
+ *     new Answer() {
+ *         public Object answer(InvocationOnMock invocation) {
+ *             Object[] args = invocation.getArguments();
+ *             Object mock = invocation.getMock();
+ *             return "called with arguments: " + Arrays.toString(args);
+ *         }}));
  *
  * //Following prints "called with arguments: [foo]"
  * System.out.println(mock.someMethod("foo"));
--- a/src/main/java/org/mockito/stubbing/Answer1.java
+++ b/src/main/java/org/mockito/stubbing/Answer1.java
@@ -14,13 +14,15 @@ import org.mockito.Incubating;
  * Example of stubbing a mock with this custom answer:
  *
  * <pre class="code"><code class="java">
+ *
  * import static org.mockito.AdditionalAnswers.answer;
  *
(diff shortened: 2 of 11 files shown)


TimvdLippe:
Please read the CONTRIBUTING.md regarding our style guide. The build is failing, because you do not adhere to this guide.

SaiyiW:
@TimvdLippe I'm really sorry I didn't read the guide carefully.. Now maybe the style is okay? Thanks for your kind reminding!

src/main/java/org/mockito/stubbing/Answer.java
```diff
@@ -13,13 +13,15 @@
...
- *     Object answer(InvocationOnMock invocation) {
- *         Object[] args = invocation.getArguments();
- *         Object mock = invocation.getMock();
- *         return "called with arguments: " + Arrays.toString(args);
- *     }
- * });
+ * import static org.mockito.AdditionalAnswers.answer;
+ *
+ * when(mock.someMethod(anyString())).then(answer(
+ *     new Answer() {
```
TimvdLippe: This one does not compile, as `Answer` does not have the correct generics.

src/main/java/org/mockito/stubbing/Answer1.java
```diff
@@ -14,11 +14,13 @@
...
- *         return arg.length();
- *     }
- * });
+ * import static org.mockito.AdditionalAnswers.answer;
+ *
+ * when(mock.someMethod(anyString())).then(answer(
+ *     new Answer1&lt;Integer, String&gt;() {
+ *         public Integer answer(String arg) {
+ *             return arg.length();
+ *         }}));
```
TimvdLippe: Please put back the original location of the brackets. This is also true for all other changes

src/main/java/org/mockito/stubbing/VoidAnswer4.java
```diff
@@ -14,15 +14,18 @@
...
+ *     new VoidAnswer4&lt;String, Integer, String, Character&gt;() {
+ *         public void answerVoid(String msg, Integer count, String another, Character c) throws Exception {
  *             throw new Exception(String.format(msg, another, c, count));
- *         }
- *     });
+ *         }})).when(mock).someMethod(anyString(), anyInt(), anyString(), anyChar());
  *
  * //Following will raise an exception with the message "ka-boom <3"
- * mock.someMethod("%s-boom %c%d", 3, "ka", '&lt;');
+ * mock.someMethod("%s-boom %c%d", 3, "ka", '<');
```
TimvdLippe: Is this intended?

src/main/java/org/mockito/stubbing/VoidAnswer5.java
```diff
@@ -14,15 +14,18 @@
...
+ *     new VoidAnswer5&lt;String, Integer, String, Character, String&gt;() {
+ *         public void answerVoid(String msg, Integer count, String another, Character c, String subject) throws Exception {
  *             throw new Exception(String.format(msg, another, c, count, subject));
- *         }
- *     });
+ *         }})).when(mock).someMethod(anyString(), anyInt(), anyString(), anyChar(), anyString());
  *
  * //Following will raise an exception with the message "ka-boom <3 mockito"
- * mock.someMethod("%s-boom %c%d %s", 3, "ka", '&lt;', "mockito");
+ * mock.someMethod("%s-boom %c%d %s", 3, "ka", '<', "mockito");
```
TimvdLippe: Is this intended?

SaiyiW:
@TimvdLippe Thanks for your review. I have made some changes according to your feedback. Plus, I found I somehow miswrote the name of public method when creating new instance of VoidAnswerX(). Now I have corrected them. Please tell me if there are still problems:blush:

TimvdLippe:
I have removed some whitespace in the docs and manually merged per https://github.com/mockito/mockito/commit/dcb21e0f17c783e99254e8486c115d6a0251a793 Thanks!

PR #991, opened 2017-03-17
Add BaseArgumentMatcher to ease migration from pre 2.x

paulduffin:
Pre 2.x ArgumentMatcher was a class that provided a default
implementation of describeTo(Description). The new
BaseArgumentMatcher class provides an equivalent implementation
of toString() making it easier for pre 2.x code that did not
override the describeTo(Description) method to be migrated.

### DIFF
--- a/src/main/java/org/mockito/ArgumentMatcher.java
+++ b/src/main/java/org/mockito/ArgumentMatcher.java
@@ -89,17 +89,24 @@ package org.mockito;
  *
  * All existing custom implementations of <code>ArgumentMatcher</code> will no longer compile.
  * All locations where hamcrest matchers are passed to <code>argThat()</code> will no longer compile.
- * There are 2 approaches to fix the problems:
+ * There are 3 approaches to fix the problems:
  * <ul>
- * <li>a) Refactor the hamcrest matcher to Mockito matcher:
+ * <li>Refactor the hamcrest matcher to Mockito matcher:
  * Use "implements ArgumentMatcher" instead of "extends ArgumentMatcher".
  * Then refactor <code>describeTo()</code> method into <code>toString()</code> method.
  * </li>
+ * <li>If you need to separate the changes to the code from the switching of the Mockito versions,
+ * e.g. if you have a large repository and cannot make a single atomic change to switch Mockito
+ * version and fix any issues it causes, then you should change any of your classes that extend
+ * {@link ArgumentMatcher} directly to extend {@link BaseArgumentMatcher} instead as it is
+ * compatible across 1.x and 2.x versions. You will need to refactor the <code>describeTo()</code>
+ * method into the <code>toString()</code> method and the <code>matches(Object)</code> method into
+ * the <code>matchesSafely(T)</code> method.
+ * </li>
  * <li>
- * b) Use <code>org.mockito.hamcrest.MockitoHamcrest.argThat()</code> instead of <code>Mockito.argThat()</code>.
+ * Use <code>org.mockito.hamcrest.MockitoHamcrest.argThat()</code> instead of <code>Mockito.argThat()</code>.
  * Ensure that there is <a href="http://hamcrest.org/JavaHamcrest/">hamcrest</a> dependency on classpath
  * (Mockito does not depend on hamcrest any more).
- *
  * </li>
  * </ul>
  * What option is right for you? If you don't mind compile dependency to hamcrest
@@ -125,4 +132,12 @@ public interface ArgumentMatcher<T> {
      * @return true if this matcher accepts the given argument.
      */
     boolean matches(T argument);
+
+    /**
+     * Provide a meaningful description of the matcher in error messages.
+     *
+     * @return a description of the values that the matcher matches, e.g. the
+     * {@link ArgumentMatchers#anyString()} matcher returns {@code "<any string>"}.
+     */
+    String toString();
 }
--- a/subprojects/migration/src/main/java/org/mockito/migration/MockitoV1CompatibilityMatcher.java
+++ b/subprojects/migration/src/main/java/org/mockito/migration/MockitoV1CompatibilityMatcher.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2017 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.migration;
+
+import java.lang.reflect.Method;
+import org.mockito.ArgumentMatcher;
+import org.mockito.Incubating;
+
+/**
+ * Base class for classes that implement {@link ArgumentMatcher}.
+ * <p>
+ * This was provided to simplify migration from pre 2.x implementations; it resolves the conflict
+ * caused by the signature of the {@code matches(...)} method changing from {@code matches(Object)}
+ * in pre 2.x to {@code matches(T)} by adding a new {@link #matchesSafely(T)} method which
+ * developers implement instead. This checks the type of the
+ * argument beforehand and returns false if it is not of the correct type to avoid a
+ * {@link ClassCastException} being thrown.
+ * <p>
+ * It does not provide a default implementation of {@link #toString()} as if no {@link #toString()}
+ * method is provided then it defaults to the pre 2.x behavior of creating a decamelized form of the
+ * class name.
+ *
+ * @deprecated this was only intended to ease migration from 1.x to 2.x please implement
+ * {@link ArgumentMatcher} directly.
+ */
+@Incubating
+@Deprecated
+public abstract class MockitoV1CompatibilityMatcher<T> implements ArgumentMatcher<T> {
+
+    /**
+     * Informs if this matcher accepts the given argument.
+     * <p>
+     * It is marked as final to prevent it being overridden because the method clashes with the
+     * equivalent method in the 1.x version of {@link ArgumentMatcher}. Implement
+     * {@link #matchesSafely(Object)} instead.
+     *
+     * @param argument the argument
+     * @return true if this matcher accepts the given argument.
+     */
+    @Override
+    public final boolean matches(Object argument) {
+        Class<T> argumentType = getArgumentType(this);
+        return argumentType.isInstance(argument) && matchesSafely(argumentType.cast(argument));
+    }
+
+    /**
+     * Informs if this matcher accepts the given argument.
+     * <p>
+     * The method should <b>never</b> assert if the argument doesn't match. It
+     * should only return false.
+     *
+     * @param argument the argument
+     * @return true if this matcher accepts the given argument.
+     */
+    public abstract boolean matchesSafely(T argument);
+
+    /**
+     * Returns the type of {@link ArgumentMatcher#matches(Object)} of the given
+     * {@link ArgumentMatcher} implementation.
+     */
+    @SuppressWarnings("unchecked")
+    private static <T> Class<T> getArgumentType(ArgumentMatcher<T> argumentMatcher) {
+        Method[] methods = argumentMatcher.getClass().getMethods();
+
+        for (Method method : methods) {
+            if (isMatchesMethod(method)) {
+                return (Class<T>) method.getParameterTypes()[0];
+            }
+        }
+        throw new NoSuchMethodError("Method '" + "matchesSafely" + "(T)' not found in ArgumentMatcher: "
+            + argumentMatcher + " !\r\n Please file a bug with this stack trace at: https://github.com/mockito/mockito/issues/new ");
+    }
+
+    /**
+     * Returns <code>true</code> if the given method is
+     * {@link MockitoV1CompatibilityMatcher#matchesSafely(Object)}
+     */
+    private static boolean isMatchesMethod(Method method) {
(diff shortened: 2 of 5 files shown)


paulduffin:
I am currently working on migrating Android code to use mockito 2.7.13. We have a lot of implementations of ArgumentMatcher that relied on the default implementation of describeTo(Description) provided by the 1.10.19 ArgumentMatcher class which is not provided by the 2.x ArgumentMatcher interface (no default methods yet). This new class will aid in that migration.

TimvdLippe:
I am all for ease of migration, glad you are using Mockito in your Android project! Is there a particular reason it was not possible to create this class in your project and extend from there? Or did you do so, but you would like to upstream it for other developers?

paulduffin:
You misunderstand me, it is not an Android project, it is Android itself, i.e. AOSP.

We have a local patch to mockito that adds a compatibility class and we would rather not maintain a local class and it just seems like something that other people moving to 2.x might appreciate.

TimvdLippe:
Ah oke, all makes sense! This would be fine by me :)

szczepiq:
This is really nice improvement and I like the use case it solves. Thank you!

1. Before we merge it, I'd like us to discuss high level design
 - is this matcher implementation useful outside of the context of 2.x migration? I suspect it is. If that's the case, we need to surface this new implementation in Javadoc in other places
 - can we come up with a solution that does not require base class? Base classes are less convenient to work with and make the API slightly more complicated. They also need more documentation because users have 2 alternatives - use the interface of the base method. Can we detect if matcher has custom toString() in runtime? If so, we could use decamelized class name automatically if the instance does not have custom toString()
 - let's mark this API incubating in case we want to experiment with the idea I mentioned earlier

2. Feedback about the implementation
 - there are no tests
 - Javadoc needs to be updated in other places, too (some examples: ArgumentMatcher, argThat method, etc). Otherwise this class will be very hard to discover and users would not use it.
 - there is no @since tag
 - otherwise, nice & clean implementation. Thank you for neat javadocs and code comments!!!

paulduffin:
1. Ok
      - Yes it is useful outside of 2.x migration, at least as useful as
      the old ArgumentMatcher was.
      - I suppose that you could use reflection to detect whether the class
      has overridden the toString() method but that seems like quite a
      heavyweight solution. It would not work for any implementations that
      override and call super.toString(), e.g. to add additional information to
      the default implementation. If you targeted Java 8 I would say
that default
      method would be a better solution.
      - I am not quite sure what incubating means but that is fine with me.
      2. It was a quick proof of concept to illustrate what I was thinking
   not intended to be submitted as is.
      - Ok.
      - Can you give me a definitive list of the places to update and
      suggestions as to what to add to each? I find that otherwise we end up
      going backwards and forwards as I try and guess what you want.
      - Ok.
      - Thanks.

TimvdLippe:
You can denote this class as incubating by adding `@Incubating` to the declaration.

I would go for the following mentions:
- Let's mention this class in [this list](https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/ArgumentMatcher.java#L16)
- Let's mention this class in [2.1.0 migration guide](https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/ArgumentMatcher.java#L88)

That should be sufficient for now. This class is merely for ease of migration, I think we should discourage its usage if the user can afford to do so.

paulduffin:
I think I have messed up this PR. I usually work in Gerrit and it is better there to do an amend and a force push. Let me know what you want me to do. I am happy to close this off and do a new PR.

src/main/java/org/mockito/ArgumentMatcher.java
```diff
@@ -125,4 +132,12 @@
...
      */
     boolean matches(T argument);
+
+    /**
+     * Provide a meaningful description of the matcher in error messages.
+     *
+     * @return a description of the values that the matcher matches, e.g. the
+     * {@link ArgumentMatchers#anyString()} matcher returns {@code "<any string>"}.
+     */
+    String toString();
```
TimvdLippe: So my Java experience is probably lackluster here, but is adding this method to the interface a breaking change? Given that you didn't make other changes and the tests still pass, I assume it isn't. Just something I don't know 😛

src/main/java/org/mockito/BaseArgumentMatcher.java
```diff
@@ -0,0 +1,61 @@
...
+import org.mockito.internal.util.Decamelizer;
+
+/**
+ * Base class for classes that implement {@link ArgumentMatcher}.
+ *
+ * <p>This was provided to simplify migration from pre 2.x implementations; it provides a default
+ * implementation of {@link #toString()} that provides equivalent behavior to the pre 2.x version.
+ * It also resolves
+ */
+@Incubating
```
TimvdLippe: Do we also want to make this deprecated, to signal to developers that this is an interim/temporary solution?

TimvdLippe: Also, let's add a `@since 2.8` here

src/main/java/org/mockito/ArgumentMatcher.java
paulduffin: No it is not a breaking change. Any class that implements it will extend Object so it will get an implementation of that method. I simply added it to the interface to document that it is needed.

src/main/java/org/mockito/BaseArgumentMatcher.java
paulduffin: Actually, I was planning on adding a version of this in a new 1.x version, say 1.10.20, so that developers can first switch to 1.10.20. Make changes (hopefully mostly automatically using error prone) so that their code will be compatible with 1.10.20 and 2.7.19 (or whatever version it goes in). They can then switch from 1.10.20 to 2.7.19 and if we have covered every incompatibility then everything will work as intended.

If you don't think this will be useful in the longer term then maybe its name should be changed to make that clearer, e.g. CompatArgumentMatcher or something. Maybe even in a separate package (separate build artifact) so developers who don't need it don't have to pay the cost for it.

paulduffin:
@szczepiq you asked about whether the toString() method is needed and whether the default behavior could be provided separately. I just found out that it already is in org.mockito.internal.matchers.text.MatcherToString. It traverses the class hierarchy looking for a toString() method and if it cannot find one then it defaults to the decamelizer form.

Based on that I think it would be much better if these classes were in a separate package and maybe even in their own compat subproject.

paulduffin:
I improved the testing of BaseArgumentMatcher and found a problem with my previous type safe matching mechanism, fixing it made it a lot more complicated so I junked it and refactored TypeSafeMatching class to allow me to reuse it with BaseArgumentMatcher in order to provide similar capabilities.

TimvdLippe:
Oof this is getting out of hand pretty quickly :O We have to see if we can make this compatibility matcher easily, without modifying too much in core.

szczepiq:
Wow, thank you for trying to close down this use case!

Given that it's getting complicated we can choose to document this use case on a wiki, with a code sample to drop to the code. This is an edge case for incremental migrations for some users, so perhaps it does not have to be part of the main code? If you all +1 to include it in the code I suggest to make it really explicit in the API for clarity (use 'org.mockito.migration' package, call the matcher: MockitoV1CompatibilityMatcher, perhaps make it deprecated so that it's clear that users should not use it forever ;).

I'm +1 to keep the API small, avoid adding the matcher and document the migration path on the wiki. We can link to that wiki from the javadoc.

paulduffin:
@szczepiq, @TimvdLippe 

Ok, so change the name and package to make it clearer, to 'org.mockito.migration.MockitoV1CompatibilityMatcher'? Is that name acceptable to both of you?

Tim, would you prefer that I keep these changes completely isolated from the main code? That would involve duplicating a little bit of the code but I am happy to do that.

That raises the next question what about packaging? Should I add a subproject called migration into which I add this code? That would isolate it even more.

I was planning on adding the ErrorProne checks that will help automate migration to an errorProne subproject.

TimvdLippe:
As a Mockito maintainer, my main goal is easy upgrades for our users. Your contribution helps greatly with that! However, it seems that integrating it is more difficult than anticipated. Therefore  little duplication sounds better to me, than the extra changes in core.

I think a separate package is very good for this, as it signals to our users that the implementation is for migration only. We can add this to the package-info for example.

So all in all, if we can find a solution that is minimal, yet greatly improves the usability of the library: yes please. I think we are close with this PR, but some minor tweaks are required to keep the codebase as nice as it is now ;)

paulduffin:
Do you want me to put this in a subproject? e.g. subproject/migration? If so presumably it will have its own artifact as well.

TimvdLippe:
If you can do that without spending too much time/resources, I would be :+1: to that!

paulduffin:
I will need to backport this to a branch based off 1.10.19. Anyone upgrading from 1.10.19 would first get the 1.10.20 (or 1.10.19-migration) version which would contain a version of MockitoV1CompatibilityMatcher that they could extend instead of extending ArgumentMatcher directly. Could you create a new 1.10.x-migration branch that I could send some pull requests to?

FYI. I just found out that I have a load of code that depends on 1.9.5 so I will be looking at doing something similar to allow easy upgrading from 1.9.5 to 1.10.19.

TimvdLippe:
Oh and to keep it simple, you can probably copy your (great!) commit message in the documentation. That should be sufficient to inform other users.

src/main/java/org/mockito/ArgumentMatcher.java
```diff
@@ -89,17 +89,24 @@
...
  * All locations where hamcrest matchers are passed to <code>argThat()</code> will no longer compile.
- * There are 2 approaches to fix the problems:
+ * There are 3 approaches to fix the problems:
  * <ul>
- * <li>a) Refactor the hamcrest matcher to Mockito matcher:
+ * <li>Refactor the hamcrest matcher to Mockito matcher:
  * Use "implements ArgumentMatcher" instead of "extends ArgumentMatcher".
  * Then refactor <code>describeTo()</code> method into <code>toString()</code> method.
  * </li>
+ * <li>If you need to separate the changes to the code from the switching of the Mockito versions,
```
szczepiq: Given that the code changed and now we have a separate 'migration' module, can you update this part of the javadoc. I really like that you added option no. 3 to the list. Perhaps it's enough if the new option is one sentence / paragraph and a link to the compatibility matcher.

subprojects/migration/src/main/java/org/mockito/migration/MockitoV1CompatibilityMatcher.java
```diff
@@ -0,0 +1,84 @@
...
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.migration;
+
+import java.lang.reflect.Method;
+import org.mockito.ArgumentMatcher;
+import org.mockito.Incubating;
+
+/**
+ * Base class for classes that implement {@link ArgumentMatcher}.
```
szczepiq: Fantastic javadoc. Thank you!

Can you add information that folks need to depend on new artifact, e.g. 'mockito-migration'?

subprojects/migration/src/main/java/org/mockito/migration/MockitoV1CompatibilityMatcher.java
```diff
@@ -0,0 +1,84 @@
...
+ * developers implement instead. This checks the type of the
+ * argument beforehand and returns false if it is not of the correct type to avoid a
+ * {@link ClassCastException} being thrown.
+ * <p>
+ * It does not provide a default implementation of {@link #toString()} as if no {@link #toString()}
+ * method is provided then it defaults to the pre 2.x behavior of creating a decamelized form of the
+ * class name.
+ *
+ * @deprecated this was only intended to ease migration from 1.x to 2.x please implement
+ * {@link ArgumentMatcher} directly.
```
szczepiq: Let's add @since tag here. If you think that we're missing @since tag somewhere else, let's add it, too :)

TimvdLippe:
@paulduffin I think you can close this PR and we need to cherry-pick your release/1.x commit to release/2.x, right?

paulduffin:
The cherry pick won't work I will still need 2.x specific patches. However, if you want me to cherry pick from 1.x and then add the necessary changes on top as a separate commit and then create a pull request I would be happy to do that.

TimvdLippe:
@paulduffin so if I understand you correctly, this PR can be merged as-is and we need to publish the 1.x artifact from the current `release/1.x` branch?

paulduffin:
It could yes. I still have some changes that I need to make, error prone checkers/fixers to add but if you want to publish then feel free.

szczepiq:
It's been some time since we reviewed this PR :) What is the current status? @paulduffin, is the PR ready to go based on your understanding of the status?

paulduffin:
I am currently working with the error prone team to get some changes into
Error Prone to make it work better with open source code. I am also working
on Error Prone checks for detecting common problems with Mockito as well as
to aid in migration. I have not forgotten about this work, there is still
quite a lot to be done.

TimvdLippe:
I am closing this right now as we are not going to land this for 2.x. We will provide a better migration path for future breaking changes.

PR #995, opened 2017-03-21
verifyZeroInteractions guarantees zero interactions instead of just delegating to verifyNoMoreInteractions

### LINKED ISSUE #989: Misleading behavior of Mockito.verifyZeroInteractions()
## Problem

Mockito.verifyZeroInteractions() actually does not do what it advertises. The purpose of this ticket is to discuss whether to change the behavior of this method in Mockito 3.

## Details

Based on user feedback at #977, from @ffissore, "verifyZeroInteractions" API can be misleading. When you read that method in test code, you expect zero interactions with specified mocks. However, "verifyZeroInteractions" is only an alias to "verifyNoMoreInteractions" and in fact, it **does not** guarantee that there were zero interactions. Let's use examples to illustrate the problem

## Feedback needed

Please comment / vote / state your opinion about changing the behavior of "verifyZeroInteractions" so that it actually ensures that there were **zero** interactions with the mock (regardless if previously verified or not).

## Examples

Below examples illustrate current behavior of Mockito 2.x, in case you are puzzled what this ticket is about.

Example 1 - verifyNoMoreInteractions

```java
mock.foo();
verify(mock).foo();
verifyNoMoreInteractions(mock); //pretty clear I think
```

Example 2 - verifyZeroInteractions

```java
mock.foo();
verify(mock).foo();

verifyZeroInteractions(mock);
//above passes because verifyZeroInteractions == verifyNoMoreInteractions
//and we already have verified the only method invoked on the mock
```

Example 3 - verifyZeroInteractions with regular stubbing

```java
given(mock.foo()).willReturn("x");
assertEquals("x", mock.foo());
verify(mock).foo();

verifyZeroInteractions(mock);
//above passes because verifyZeroInteractions == verifyNoMoreInteractions
```

Example 4 - verifyZeroInteractions with new strict stubbing

```java
given(mock.foo()).willReturn("x");
assertEquals("x", mock.foo());

verifyZeroInteractions(mock);
//above passes because with strict stubbing
//stubbed method are implicitly verified when they are "used"
//this is one of key features of strict stubbing (DRY, don't repeat yourself)
```

ffissore:
Fixes #989

`verifyZeroInteractions` no longer delegates to `verifyNoMoreInteractions`
Instead, it fetches the number of invocations made to the mock and checks if
they are zero, failing otherwise

I'm not sure if the style of my code complies with mockito: please let me know what to change and where to look to make it more mockitoish

### DIFF
--- a/src/main/java/org/mockito/BDDMockito.java
+++ b/src/main/java/org/mockito/BDDMockito.java
@@ -259,6 +259,12 @@ public class BDDMockito extends Mockito {
          * @since 2.1.0
          */
         void shouldHaveNoMoreInteractions();
+
+        /**
+         * @see #verifyNoInteractions(Object...)
+         * @since 3.0.0
+         */
+        void shouldHaveNoInteractions();
     }
 
     private static class ThenImpl<T> implements Then<T> {
@@ -304,9 +310,19 @@ public class BDDMockito extends Mockito {
         /**
          * @see #verifyZeroInteractions(Object...)
          * @since 2.1.0
+         * @deprecated Since 3.0.0. Please migrate your code to {@link #shouldHaveNoInteractions()}
          */
+        @Deprecated
         public void shouldHaveZeroInteractions() {
             verifyZeroInteractions(mock);
+        }
+
+        /**
+         * @see #verifyNoInteractions(Object...)
+         * @since 3.0.0
+         */
+        public void shouldHaveNoInteractions() {
+            verifyNoInteractions(mock);
         }
 
         /**
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -2010,9 +2010,31 @@ public class Mockito extends ArgumentMatchers {
      * See examples in javadoc for {@link Mockito} class
      *
      * @param mocks to be verified
+     * @deprecated Since 3.0.0. Please migrate your code to {@link #verifyNoInteractions(Object...)}
      */
+    @Deprecated
     public static void verifyZeroInteractions(Object... mocks) {
         MOCKITO_CORE.verifyNoMoreInteractions(mocks);
+    }
+
+    /**
+     * Verifies that no interactions happened on given mocks.
+     * <pre class="code"><code class="java">
+     *   verifyNoInteractions(mockOne, mockTwo);
+     * </code></pre>
+     * This method will also detect invocations
+     * that occurred before the test method, for example: in <code>setUp()</code>, <code>&#064;Before</code> method or in constructor.
+     * Consider writing nice code that makes interactions only in test methods.
+     * <p>
+     * See also {@link Mockito#never()} - it is more explicit and communicates the intent well.
+     * <p>
+     * See examples in javadoc for {@link Mockito} class
+     *
+     * @since 3.0.0
+     * @param mocks to be verified
+     */
+    public static void verifyNoInteractions(Object... mocks) {
+        MOCKITO_CORE.verifyNoInteractions(mocks);
     }
 
     /**
--- a/src/main/java/org/mockito/internal/MockitoCore.java
+++ b/src/main/java/org/mockito/internal/MockitoCore.java
@@ -148,16 +148,15 @@ public class MockitoCore {
         assertMocksNotEmpty(mocks);
         mockingProgress().validateState();
         for (Object mock : mocks) {
-            try {
-                if (mock == null) {
-                    throw nullPassedToVerifyNoMoreInteractions();
-                }
-                InvocationContainer invocations = getMockHandler(mock).getInvocationContainer();
-                VerificationDataImpl data = new VerificationDataImpl(invocations, null);
-                noInteractions().verify(data);
-            } catch (NotAMockException e) {
+            if (mock == null) {
+                throw nullPassedToVerifyNoMoreInteractions();
+            }
+            if (!isMock(mock)) {
                 throw notAMockPassedToVerifyNoMoreInteractions();
             }
+            InvocationContainer invocations = getMockHandler(mock).getInvocationContainer();
+            VerificationDataImpl data = new VerificationDataImpl(invocations, null);
+            noInteractions().verify(data);
         }
     }
 
--- a/src/main/java/org/mockito/internal/MockitoCore.java
+++ b/src/main/java/org/mockito/internal/MockitoCore.java
@@ -19,6 +19,7 @@ import static org.mockito.internal.util.MockUtil.getMockHandler;
 import static org.mockito.internal.util.MockUtil.isMock;
 import static org.mockito.internal.util.MockUtil.resetMock;
 import static org.mockito.internal.util.MockUtil.typeMockabilityOf;
+import static org.mockito.internal.verification.VerificationModeFactory.noInteractions;
 import static org.mockito.internal.verification.VerificationModeFactory.noMoreInteractions;
 
 import java.util.Arrays;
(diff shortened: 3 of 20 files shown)


TimvdLippe:
You can always check if you comply with the code by running `./gradlew build`. That runs the test and checks the code style :)

ChristianSchwarz:
@ffissore 
Looks cool!  Can you add more tests?

szczepiq:
Thank you for submission! I have not looked into the implementation details, if nobody does picks up the review soon, I'll do it :)

Team, is the decision to change the behavior of "verifyZeroInteractions"? This is likely a disturbing change that will impact many projects migrating to Mockito v3. I'm +1 to this change but I have to acknowledge the impact ;)

TimvdLippe:
We would need to investigate the usage of this method. I think it is a logical change, but we have to prevent users from excessive code problems. They can of course always transfer everything to `verifyNoMoreInteractions` just before they are migrating, that shouldn't be a bit of a hassle.

ffissore:
@ChristianSchwarz I'll add more in the coming days
About the name: pls consider using a different name, `verifyNoInteractions` for example, and deprecating `verifyZeroInteractions` with an explanation

src/main/java/org/mockito/BDDMockito.java
```diff
@@ -304,12 +304,22 @@ public T should(InOrder inOrder, VerificationMode mode) {
         /**
          * @see #verifyZeroInteractions(Object...)
          * @since 2.1.0
+         * @deprecated Please migrate your code to {@link #shouldHaveNoInteractions()}
```
TimvdLippe: Let's add a `@deprecated since 3.0.0` here, so we don't accidently remove it in the wrong version (which we did previously)

src/test/java/org/mockitousage/bugs/VerifyNoInteractionsIsLikeVerifyNeverTest.java
```diff
@@ -0,0 +1,86 @@
...
+    Container container = new Container(collaborator);
+    when(collaborator.doStuff(42)).thenReturn("yeah");
+
+    String output = container.doStuff(42);
+    assertEquals("yeah", output);
+
+    verifyNoInteractions(collaborator);
+  }
+
+  @Test //(expected = NoInteractionsWanted.class)
```
TimvdLippe: Since we deprecate the method anyways, is this test really required?

ffissore:
As noted, in support to my proposal, I've changed my PR to include a new `verifyNoInteractions` and deprecated old `verifyZeroInteractions`. I've added some tests, and coverage will hopefully be happier.
@TimvdLippe I've added those "since 3.0.0" and removed my initial test class

src/main/java/org/mockito/Mockito.java
```diff
@@ -2010,12 +2010,33 @@ public static void verifyNoMoreInteractions(Object... mocks) {
...
+     * </code></pre>
+     * This method will also detect invocations
+     * that occurred before the test method, for example: in <code>setUp()</code>, <code>&#064;Before</code> method or in constructor.
+     * Consider writing nice code that makes interactions only in test methods.
+     * <p>
+     * See also {@link Mockito#never()} - it is more explicit and communicates the intent well.
+     * <p>
+     * See examples in javadoc for {@link Mockito} class
+     *
+     * @param mocks to be verified
```
TimvdLippe: This still needs a `@since`

src/test/java/org/mockitousage/basicapi/ResetTest.java
```diff
@@ -63,6 +63,13 @@ public void shouldRemoveAllInteractions() throws Exception {
     }
 
     @Test
+    public void shouldRemoveAllInteractions2() throws Exception {
```
TimvdLippe: I would rather not name it with `2`. Probably better as `should_remove_all_interactions_verify_no_interactions()`

src/test/java/org/mockitousage/basicapi/ResetTest.java
```diff
@@ -79,6 +86,14 @@ public void shouldStubbingNotBeTreatedAsInteraction() {
     }
 
     @Test
+    public void shouldStubbingNotBeTreatedAsInteraction2() {
```
TimvdLippe: Same here

src/test/java/org/mockitousage/misuse/InvalidUsageTest.java
```diff
@@ -104,6 +109,18 @@ public void shouldNotMockObjectMethodsOnInterface() throws Exception {
         verifyZeroInteractions(inter);
     }
 
+    @Test
+    public void shouldNotMockObjectMethodsOnInterface2() throws Exception {
```
TimvdLippe: Same here

src/test/java/org/mockitousage/misuse/InvalidUsageTest.java
```diff
@@ -113,4 +130,15 @@ public void shouldNotMockObjectMethodsOnClass() throws Exception {
 
         verifyZeroInteractions(clazz);
     }
+
+    @Test
+    public void shouldNotMockObjectMethodsOnClass2() throws Exception {
```
TimvdLippe: Same here

src/test/java/org/mockitousage/verification/NoMoreInteractionsVerificationTest.java
```diff
@@ -126,6 +139,25 @@ public void shouldVerifyOneMockButFailOnOther() throws Exception {
         } catch (NoInteractionsWanted e) {}
     }
 
+    @Test
+    public void shouldVerifyOneMockButFailOnOther2() throws Exception {
```
TimvdLippe: Same here

src/test/java/org/mockitousage/stubbing/BasicStubbingTest.java
```diff
@@ -67,6 +67,14 @@ public void should_stubbing_not_be_treated_as_interaction() {
     }
 
     @Test
+    public void should_stubbing_not_be_treated_as_interaction2() {
```
TimvdLippe: Same here

src/test/java/org/mockitousage/stubbing/BasicStubbingTest.java
```diff
@@ -79,6 +87,18 @@ public void unfinished_stubbing_cleans_up_the_state() {
     }
 
     @Test
+    public void unfinished_stubbing_cleans_up_the_state2() {
```
TimvdLippe: Same here

ffissore:
Applied requested modifications and squashed commits

ffissore:
I don't know what went wrong on travis, tests pass on my machine using openjdk 8. Ping me if I can help

TimvdLippe:
These are flaky tests, so it seems fine :)

src/main/java/org/mockito/internal/MockitoCore.java
```diff
@@ -137,6 +138,29 @@ public void verifyNoMoreInteractionsInOrder(List<Object> mocks, InOrderContext i
...
+        mockingProgress().validateState();
+        for (Object mock : mocks) {
+            try {
+                if (mock == null) {
+                    throw nullPassedToVerifyNoMoreInteractions();
+                }
+                InvocationContainer invocations = getMockHandler(mock).getInvocationContainer();
+                VerificationDataImpl data = new VerificationDataImpl(invocations, null);
+                noInteractions().verify(data);
+            } catch (NotAMockException e) {
```
ChristianSchwarz: @ffissore We should avoid this catch block by calling `MockUtil.isMock(..)`

ffissore:
Remove try catch block

PR #996, opened 2017-03-21
Add explicit mention for code style check

TimvdLippe:
To prevent confusion, let's include it in our CONTRIBUTING.md

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -82,6 +82,12 @@ For most editors, this should be automatically enforced by [EditorConfig](http:/
 Check if your editor has a built-in plugin or if you need to download one.
 IntelliJ has a built-in plugin, for Eclipse you need to download [this plugin](https://github.com/ncjones/editorconfig-eclipse#readme).
 
+If you want to check if your code complies to the style guide, you can run:
+
+* `./gradlew checkstyleMain` to check the main source code
+* `./gradlew checkstyleTest` to check the test source code
+* `./gradlew build` to check main and test source code, and run the tests
+
 ### Imports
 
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -82,6 +82,12 @@ For most editors, this should be automatically enforced by [EditorConfig](http:/
 Check if your editor has a built-in plugin or if you need to download one.
 IntelliJ has a built-in plugin, for Eclipse you need to download [this plugin](https://github.com/ncjones/editorconfig-eclipse#readme).
 
+If you want to check if your code complies to the style guide, you can run:
+
+* `./gradlew checkstyleMain` to check the main source code
+* `./gradlew checkstyleTest` to check the test source code
+* `./gradlew build` to check main and test source code, and run the tests
```
szpak: `./gradlew test` wouldn't be enough?

szpak: `./gradlew check` wouldn't be enough?

TimvdLippe: I ran that too and it did the same as `build`. E.g. both check code and run the tests

szpak: `check` is (a little bit) cheaper/faster - the artifacts should not be built (at least some of them as acceptance tests depends on some of them, but in general it's good to promote the faster possible way).

It's very similar to `mvn clean install` - a default command in Maven acknowledged by common wisdom. Most of the people in fact would be satisfied with `verify` which doesn't build and install artifacts (sometimes having 100MB+ assuming fat jars) in the local Maven repository.

TimvdLippe:
Merging as the review comment has been resolved.

PR #1001, opened 2017-03-23
Fix release notes badge

mureinik:
GitHub doesn't seem to render images from URLs with spaces properly,
breaking the release notes badge.
This change encodes the URL, replacing the space with `%20` to make it
render properly.

[ci skip]


szpak:
Good catch, thanks.

szczepiq:
Thank you very much for the fix and for the swift merge!!!

PR #1004, opened 2017-03-24
Fixes #1003: Allow reloading of plugins.

### LINKED ISSUE #1003: Allow reloading of plugins
I am working on https://github.com/powermock/powermock/issues/769. With custom extension PluginSwitch's we should be able to allow switches that change whether an extension is enabled or disabled in different tests (such as choosing different MockMakers, default or custom). PluginRegistry can facilitate this if it is not final and has a static method to refresh/recreate itself.

podarsmarty:
Fixes #1003: Allow reloading of plugins.

This is done by making the PluginRegistry non final and exposing a public static method in Plugins so that PluginSwitch's in the future can trigger a reload of plugins if it wants to disable/enable some of them.


podarsmarty:
oops messed up a script destroyed my repo's git history. fixing it now.

PR #1006, opened 2017-03-24
Fixes #1003: Allow reloading of plugins.

### LINKED ISSUE #1003: Allow reloading of plugins

podarsmarty:
Fixes #1003: Allow reloading of plugins.

This is done by making the PluginRegistry non final and exposing a public static method in Plugins so that PluginSwitch's in the future can trigger a reload of plugins if it wants to disable/enable some of them.

### DIFF
--- a/src/main/java/org/mockito/internal/configuration/plugins/Plugins.java
+++ b/src/main/java/org/mockito/internal/configuration/plugins/Plugins.java
@@ -14,7 +14,7 @@ import org.mockito.plugins.StackTraceCleanerProvider;
  */
 public class Plugins {
 
-    private static final PluginRegistry registry = new PluginRegistry();
+    private static PluginRegistry registry = new PluginRegistry();
 
     /**
      * The implementation of the stack trace cleaner
@@ -51,5 +51,12 @@ public class Plugins {
      */
     public static AnnotationEngine getAnnotationEngine() {
         return registry.getAnnotationEngine();
+    }
+
+    /**
+     * Refreshes the plugin registry and reloads plugins.
+     */
+    public static void refreshPluginRegistry() {
+        registry = new PluginRegistry();
     }
 }
--- a/src/test/java/org/mockito/internal/configuration/plugins/PluginsTest.java
+++ b/src/test/java/org/mockito/internal/configuration/plugins/PluginsTest.java
@@ -1,0 +1,25 @@
+package org.mockito.internal.configuration.plugins;
+
+import org.junit.Test;
+import org.mockito.plugins.MockMaker;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+
+public class PluginsTest {
+
+    @Test
+    public void refresh_plugin_registry_reloads_plugins() {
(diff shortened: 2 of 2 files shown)


TimvdLippe:
Hm, I am not sure about this change. Preferably I wouldn't want to make this change, as the PluginRegistry is supposed to run only once. Can't you make a delegate-`MockMaker` for your usecase, which dynamically determines which core `MockMaker` to delegate to?

podarsmarty:
hmm @TimvdLippe, That is essentially my work around for now in Powermock. What happens in the case where we have multiple MockMakers in the classpath? We wouldn't have control to make sure that the delegate is used since it would be whichever one ClassLoader.getResource() returns?

Is there something that depends on PluginRegistry not changing that I am missing? Or would you be more comfortable making our instance of mockmaker not final and only reloading that?

TimvdLippe:
We have had several issues with loading and illogical exceptions related to the Registry. Therefore I would like to avoid introducing multiple loading stages, where again stuff can go wrong.

That's my 2 cents though, let's see what the others have to say about this.

thekingnothing:
I also think that dynamic reloading plugins it is overhead and needless feature. I don't see any case when it is required. All tests in one test class always should be run with one MockMaker. If speak about case of using with PowerMock, then PowerMock reloads classes for each test class, so  it's easy to implement `DelegateMockMaker` to delegate to Mockito MockMaker when test is run without PowerMock and use PowerMockMockMaker is case if the test is started with using `PowerMockRunner`. 

It will be better to have a link to default/parent Mockito MockMaker. It will help implement delegate-`MockMaker` with logic same as ClassLoader logic. I mean parent-first for parent-last.

podarsmarty:
Writing a DelegateMockMaker is simple but what where would one put it?

Would we essentially make PowerMockMaker the delagate in PowerMock? What happens when a user has a 3rd MockMaker in their classpath? Now the user needs to write a delegate to another as well as writing a switch to turn off all other MockMakers. IMO that is needless overhead for someone trying to turn on/off something for different test classes. Right now PowerMock is not on ByteBuddy so b/c it is on the classpath, someone cannot take advantage of Mockito2 for tests that do not require PowerMock. They would be forced to write their own MockMaker and switch. If we cannot reload plugins, the switch is essentially just hardcoding things on and off. We are now forcing our users to understand implementation details of dependencies they need and to write delegates and switches.

IMO that is much less idea that allowing one to choose to dynamically reload plugins when they choose. It would not change default behavior. If one reloads halfway through a test class, then we may get some weird messaging but we already get that when one expects Mockito2 features.

TimvdLippe:
I am closing this PR for now. We are planning to address the MockMaker issues in Mockito 3, since we are now shipping multiple versions and there is no proper way to choose one for a test. For example, if just one test requires the InlineMockMaker. Stay tuned and thanks for your patience!

PR #1007, opened 2017-03-24
Fix test failing with Java 9-ea+146+

szpak:
Stack trace in JDK part contains also module name:

```
java.lang.AssertionError: 
Expecting:
 "[MockitoHint] 1. Unused -> at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)"
to match pattern:
 "\[MockitoHint\] 1\. Unused \-\> at [\w\.]+\.reflect\.NativeMethodAccessorImpl\.invoke0\(.*Native Method\)"
```

The new regexp handles both variants.

### DIFF
--- a/src/test/java/org/mockito/internal/junit/UnusedStubbingsTest.java
+++ b/src/test/java/org/mockito/internal/junit/UnusedStubbingsTest.java
@@ -7,8 +7,7 @@ import org.mockito.internal.stubbing.answers.DoesNothing;
 import org.mockito.internal.util.SimpleMockitoLogger;
 import org.mockitoutil.TestBase;
 
-import java.util.List;
-import java.util.regex.Pattern;
+import java.util.Collections;
 
 import static java.util.Arrays.asList;
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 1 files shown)


szpak:
Tested with Java 8 and Java 9-ea+146. Newer Java 9 versions (up to 9-ea+162) throws `ExceptionInInitializerError` in Gradle.

TimvdLippe:
Hm, hopefully known Gradle bug. If anyone can test and verify, that would be nice. Else I will merge this tomorrow :)

TimvdLippe:
Never mind, merged already :joy:

szpak:
Yup, known bug - https://github.com/gradle/gradle/issues/1095.

szpak:
Although there seems to be a workaround: https://github.com/gradle/gradle/issues/719#issuecomment-289083209 (just as a reference).

szczepiq:
@szpak, thanks a lot for the fix!!! BTW. do you know if we can set up travis build with java 9?

szpak:
@szczepiq In fact Mockito master is [already built](https://travis-ci.org/mockito/mockito/builds/214828895) with Java 9. Just not that fresh version to trigger mentioned issue. On the other hand, Gradle is not compatible with more recent Java 9 versions.

PR #1008, opened 2017-03-24
Sample usage of soft assertions

szpak:
Could be propagated by the way also in other similar places.


szpak:
That's what I meant by "by the way".  By the way of introducing related changes in tests :).

PR #1009, opened 2017-03-25
Clean up unused imports

mureinik:
During development work, orphaned, unused, imports sometime get left
behind in the code. At best, these imports are just redundant. At
worst, they introduce false dependencies in places there aren't
supposed to be any dependencies.

This patch removes all those unused imports and adds a Checkstyle
module to verify no such redundant imports will be introduced in the
future.


szpak:
Nice, thanks.

szczepiq:
Really nice. Thank you!!!

PR #1010, opened 2017-03-29
Bumped version to fix CI build

szczepiq:
Manually bumped version to unstuck the release process. This should help and get Travis CI jobs back in shape. More details about the problem and our long term plan - see mailing list discussion: https://groups.google.com/forum/#!topic/mockito/TTMuY5KdbjU


szczepiq:
I'll merge without waiting for review this since it's only version bump and we need it to fix the build.

Issue #1005, opened 2017-03-24
Test fails after enabling feature for mocking final classes and methods


geralt-encore:
I am mocking [GoogleSignInAccount](https://developers.google.com/android/reference/com/google/android/gms/auth/api/signin/GoogleSignInAccount) with the latest Mockito version:
```
GoogleSignInAccount mockGoogleAccount = mock(GoogleSignInAccount.class);
    when(mockGoogleAccount.getId()).thenReturn("id");
    when(mockGoogleAccount.getEmail()).thenReturn("email@email.com");
    when(mockGoogleAccount.getFamilyName()).thenReturn("family name");
    when(mockGoogleAccount.getGivenName()).thenReturn("given name");
```

After enabling mocking of final classes/methods test fails with this error (which is not really helpful):
```
org.mockito.exceptions.misusing.MissingMethodInvocationException: 
when() requires an argument which has to be 'a method call on a mock'.
For example:
    when(mock.getArticles()).thenReturn(articles);

Also, this error might show up because:
1. you stub either of: final/private/equals()/hashCode() methods.
   Those methods *cannot* be stubbed/verified.
   Mocking methods declared on non-public parent classes is not supported.
2. inside when() you don't call method on mock but on some other object.

	at com.app.profile.data.ProfileRepositoryTest.shouldFailRegisterIfEmailNotRegisteredAndUnknownGenderGoogleAccount(ProfileRepositoryTest.java:870)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
[...]

PR #1012, opened 2017-03-30
Do not suppress instrumentation exceptions

raphw:
Rather propagate them to mock maker that attempts instrumentation. Addresses #1005.


TimvdLippe:
@raphw fix seem fine by me, but it does not fix #1005 right?

raphw:
No, it does not fix it. Some classes are broken when it comes to their internal structure. I do not see how we can instrument classes that were messed up by ProGuard, unfortunately.

PR #1014, opened 2017-03-31
Clean up the code with diamond

mureinik:
Utilize Java 7's diamond operator (`<>`) to clean up the code.


bric3:
We can't accept this PR as mockito 2 is still built around Java 6. Such work may be performed on master because mockito 3 will be Java 8 minimum

mureinik:
That was stupid of me, apologies.

I'm surprised that Gradle didn't fail the build, even with `sourceCompatibility` and `targetCompatibility` set to 1.6.

bric3:
No problem ;)
It's nice to see any pull request, thanks for the interest. If you want to clean up things you can   still    target master.

Issue #99, opened 2014-10-08
RETURNS_DEEP_STUBS automatically tries to create serializable mocks


joerg-pfruender-hp:
I am just migrating from mockito 1.9.5 to 1.10.5

The following code runs fine with version 1.9.5. but breaks now:

``` java

  @Test
  public void test() {
    ToBeMocked mock = mock(ToBeMocked.class, RETURNS_DEEP_STUBS);
    assertThat(mock.getSomething()).isNotNull();
  }

  public static class ToBeMocked {

    NotSerializableReturnValue field1;

    public ToBeMocked(NotSerializableReturnValue field1) {
      this.field1 = field1;
    }

    public NotSerializableReturnValue getSomething() {
      return field1;
    }
  }

  public static class NotSerializableReturnValue {

    String field1 = "";

    public NotSerializableReturnValue(String field1) {
      this.field1 = field1;
    }

    public String getSomething2() {
      return field1;
    }
  }
```

org.mockito.exceptions.base.MockitoException: 
You are using the setting 'withSettings().serializable()' however the type you are trying to mock 'NotSerializableReturnValue'
do not implement Serializable AND do not have a no-arg constructor.

PR #1015, opened 2017-03-31
Fixes #992: Backport MockitoV1CompatibilityMatcher from 2.x to aid in migration to 2.x

### LINKED ISSUE #992: Help in people migrating from one version to the next
After having gone through the process of migrating a reasonably sized body of code from 1.10.19 to 2.7.13 I thought it would be helpful to provide some feedback.

The way that I work is generally upgrade the code in question, in this case mockito, and then:

1. Build the code
1. Fix build failures
1. Repeat from step 1 until build works.

While this is an effective, albeit time consuming and tedious, process it can often be difficult to know what to use when something breaks. It also does not help when there are behavioral changes that need to be checked manually by the developer. e.g. The change to anyX() and any(SomeType.class) matchers to not accept null would not cause compilation errors, may not even cause tests to fail but could still cause some tests to be incorrect.

What would be absolutely awesome is if as you were developing you built up a library of [error prone checks](http://errorprone.info/docs/installation) that could be used by developers to not only identify the code that would be affected by the upgrade but in some cases thanks to [error prone's patching](http://errorprone.info/docs/patching) capabilities and [refaster rules](http://errorprone.info/docs/refaster) could actually fix the code for them as well.

One thing that I find really essential when doing the upgrade is to try and keep the changes to our code separate from the actual changes to the code we're upgrading. e.g. Where necessary when upgrading from 1.10.19 to 2.x I backported some changes from 2.x (e.g. MockitoHamcrest) to allow me to change our code to use the new 2.x way of doing things (e.g. handling hamcrest based matchers) before I actually upgraded.

Where that was not possible, e.g. with the changes to ArgumentMatcher, I created special compatibility classes in 1.10.19 that I changed the code to use and then added equivalent classes in 2.7.13 so that the code would continue to compile against it. Once the upgrade was completed the code was changed again to use standard 2.7.13 code.

I went from 1.10.19 to 2.7.13 rather than smaller increments mainly because the overhead of doing the upgrade was so high that it was not worth doing it in smaller steps. That is also why we delayed doing the upgrade for as long as we did which necessitated us making a big jump. I believe that using error prone can reduce the cost and risk of the upgrade to the point where developers can start doing it so regularly that they are never more than a few releases behind head.

If the cost of upgrade was trivial and you followed the principle that you must deprecate and provide a replacement in at least one version before you remove or change an API element then it would not be necessary to create special compatibility classes as you could just upgrade from one release to the next to the next.

Writing and testing the checkers and the refaster rules will take some time but they will bring huge benefits to users and also to you. After all if the cost of upgrading was low then you can contemplate making improvements that you would otherwise avoid due to the impact on developers.

I am quite happy to help and I know that the error prone developers would be interested in someone from outside Google using error prone like this.

paulduffin:
Fixes #992: Backport MockitoV1CompatibilityMatcher from 2.x to aid in migration to 2.x

The 2.x version of ArgumentMatcher<T> differs quite
significantly from the current version:
1) It is an interface not a class and no longer extends
   org.hamcrest.Matcher.
2) Its main method is matches(T) not matches(Object).
3) It uses the toString() method to construct a description and
   not the describeTo(Description) method.

This class supports the following strategy for migrating
implementations of ArgumentMatcher from 1.10.19 to 2.x:
1) First migrate to a 1.x version that has
   MockitoV1CompatibilityMatcher change in it. That should be
   relatively simple.
2) Then modify all existing implementation of ArgumentMatcher
   to extend MockitoV1CompatibilityMatcher.
3) Switch to use a version of 2.x that contains this change.
   Everything should compile as usual.
4) Modify the classes that extend MockitoV1CompatibilityMatcher
   to implement ArgumentMatcher interface directly.

The above strategy is overkill when migrating a small project
that can be updated atomically but it has proven to be a good
approach when upgrading a large project spread across multiple
git repositories. This is added in part because of the
experience doing exactly that with the Android Open Source
Platform.

check list

 - [ x PR should be motivated, i.e. what does it fix, why, and if relevant how

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,5 +1,6 @@
 include 'testng'
 include 'extTest'
+include 'migration'
 
 rootProject.name = 'mockito'
 
--- a/subprojects/migration/migration.gradle
+++ b/subprojects/migration/migration.gradle
@@ -1,0 +1,9 @@
+apply plugin: 'java'
+description = "Compatibility classes to aid in migration to 2.x"
+
+dependencies {
+    compile "org.hamcrest:hamcrest-core:1.3"
+    compile project.rootProject
+
+    testCompile "junit:junit:4.10"
+}
--- a/subprojects/migration/src/main/java/org/mockito/migration/MockitoV1CompatibilityMatcher.java
+++ b/subprojects/migration/src/main/java/org/mockito/migration/MockitoV1CompatibilityMatcher.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2017 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.migration;
+
+import java.lang.reflect.Method;
+import org.mockito.ArgumentMatcher;
+import org.mockito.Incubating;
+
+/**
+ * Base class for classes that implement {@link ArgumentMatcher}.
+ * <p>
+ * This was provided to simplify migration from pre 2.x implementations; it resolves the conflict
+ * caused by the signature of the {@code matches(...)} method changing from {@code matches(Object)}
+ * in pre 2.x to {@code matches(T)} by adding a new {@link #matchesSafely(T)} method which
+ * developers implement instead. This checks the type of the
+ * argument beforehand and returns false if it is not of the correct type to avoid a
+ * {@link ClassCastException} being thrown.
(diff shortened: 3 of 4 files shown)


paulduffin:
This is the second part of the #99

PR #1016, opened 2017-03-31
Fixes #991: Backport MockitoV1CompatibilityMatcher from 2.x to aid in…

paulduffin:
Fixes #991: Backport MockitoV1CompatibilityMatcher from 2.x to aid in migration to 2.x

The 2.x version of ArgumentMatcher<T> differs quite
significantly from the current version:
1) It is an interface not a class and no longer extends
   org.hamcrest.Matcher.
2) Its main method is matches(T) not matches(Object).
3) It uses the toString() method to construct a description and
   not the describeTo(Description) method.

This class supports the following strategy for migrating
implementations of ArgumentMatcher from 1.10.19 to 2.x:
1) First migrate to a 1.x version that has
   MockitoV1CompatibilityMatcher change in it. That should be
   relatively simple.
2) Then modify all existing implementation of ArgumentMatcher
   to extend MockitoV1CompatibilityMatcher.
3) Switch to use a version of 2.x that contains this change.
   Everything should compile as usual.
4) Modify the classes that extend MockitoV1CompatibilityMatcher
   to implement ArgumentMatcher interface directly.

The above strategy is overkill when migrating a small project
that can be updated atomically but it has proven to be a good
approach when upgrading a large project spread across multiple
git repositories. This is added in part because of the
experience doing exactly that with the Android Open Source
Platform.

check list

 - [ x PR should be motivated, i.e. what does it fix, why, and if relevant how

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,5 +1,6 @@
 include 'testng'
 include 'extTest'
+include 'migration'
 
 rootProject.name = 'mockito'
 
--- a/subprojects/migration/migration.gradle
+++ b/subprojects/migration/migration.gradle
@@ -1,0 +1,9 @@
+apply plugin: 'java'
+description = "Compatibility classes to aid in migration to 2.x"
+
+dependencies {
+    compile "org.hamcrest:hamcrest-core:1.3"
+    compile project.rootProject
+
+    testCompile "junit:junit:4.10"
+}
--- a/subprojects/migration/src/main/java/org/mockito/migration/MockitoV1CompatibilityMatcher.java
+++ b/subprojects/migration/src/main/java/org/mockito/migration/MockitoV1CompatibilityMatcher.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2017 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.migration;
+
+import java.lang.reflect.Method;
+import org.mockito.ArgumentMatcher;
+import org.mockito.Incubating;
+
+/**
+ * Base class for classes that implement {@link ArgumentMatcher}.
+ * <p>
+ * This was provided to simplify migration from pre 2.x implementations; it resolves the conflict
+ * caused by the signature of the {@code matches(...)} method changing from {@code matches(Object)}
+ * in pre 2.x to {@code matches(T)} by adding a new {@link #matchesSafely(T)} method which
+ * developers implement instead. This checks the type of the
+ * argument beforehand and returns false if it is not of the correct type to avoid a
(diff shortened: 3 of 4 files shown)


paulduffin:
I have a change that fixes that build issue that removes GitHubTicketFetcherTest - I thought it might have been something to do with the environment but it looks like it might be a consequence of changes to GitHub which you have not encountered because you have not built v1.10.19 recently.

paulduffin:
Just pushed the fix to this.

szczepiq:
Nice PR! It's ok to remove the fetcher test, thank you for doing this!

Let's push the change in 2.x first (https://github.com/mockito/mockito/pull/991) and then I'll test this PR. We have never released older version of Mockito when new version was present. It might be fun :)

paulduffin:
There are a few more changes to come for 1.10.x so I would not release quite yet, unless you want to make sure that everything is working fine and are happy to do it again at a later date.

One thing that I have noticed is that 1.10.19 uses Gradle 2.2.1. I had some problems with that so upgraded locally to 2.12 and it seems to work fine (had to remove OfflineCheckerTest though). So, if you needed to you (or I) could always upgrade 1.10.19 to use the same version of Gradle and build files as 2.x so that you can take advantage of the release process you already have for that.

TimvdLippe:
@paulduffin Please open a new PR for the new changes to Gradle. One thing to keep in mind is that this is currently being altered in #1018, so maybe wait a bit for these changes to land. Then, releasing a new artifact should be a lot easier

szczepiq:
I suggest we don't over invest in automating 1.x release, it's not going to be something we do again any time soon :)

paulduffin:
I am not sure what you want me to do @TimvdLippe. I have removed the GitHubTicketFetcherTest fix from this pull request. Do you want me to create a separate pull request after this has been accepted?

TimvdLippe:
Originally I wanted to upgrade the Gradle infrastructure for automatic releases, but let's disregard that for now. We are not planning to release many 1.x releases any longer. I will merge this PR for now (as the functionality works, we can disregard this 1 failing test). How many other changes are you planning to make? If you think you are finished, please ping us so we can release the new artifact. I would like to limit the new releases to 1, if we can do so :smile:

paulduffin:
I have a few changes for 1.x and also some for 1.9.x. I am currently working with the error prone team to try and get some improvements made there so lots of balls being juggled at the moment. I am going on holiday for a week but I will be working on this as soon as I get back.

TimvdLippe:
Okay thanks for the update. Just ping us when you are done and we can finish :) Thanks for all your effort!

Issue #1020, opened 2017-04-03
Use JDK 7 with Java 6 compatibility for building release artifacts


szczepiq:
We need community feedback if it is ok to change the JDK we use to build release artifacts for mockito. 

###  Problem

Currently Mockito is build on JDK 6. This ensures it will work with Java 6. However, this causes issues like "sun.security.validator.ValidatorException: PKIX path building failed" (https://github.com/mockito/mockito-release-tools/issues/60). This issues makes it problematic for us to setup up continuous delivery automation with Travis CI and Gradle.

### Question

Is ok to start building release artifacts of Mockito 2.x with JDK 7 (with Java 6 compatibility)? This will help with release automation and also is safer because Java 6 is no longer supported by Oracle.

Thank you in advance for answers!

TimvdLippe:
This would be a major breaking change, so I am not for this. Can we contact Travis to investigate what the issue is?

szczepiq:
Why would this be a breaking change? We would compile with Java 6 compatibility and generate bytecode that works on Java 6.

TimvdLippe:
Ah, I misunderstood. So we build on JDK7, but set it to be Java 6 compatibility. Afaik, Gradle does not complain about Java 6 incompatible changes on newer JDKs, but the risk is probably low.

Could we maybe remove the release tool from the JDK6 Travis build, thus releasing on JDK7, but still build on JDK6 on Travis too?

szpak:
@TimvdLippe https://github.com/xvik/gradle-animalsniffer-plugin can be use to detect constructions not compatible with Java 6.

TimvdLippe:
@szpak Ah cool, TIL! Yes then let's go ahead and adopt that plugin + switch to JDK7

szpak:
I will add that plugin in 2.x branch as a preparation to the switch.

szczepiq:
@szpak, animalsniffer looks interesting! Why do we need animal sniffer given that we can compile on JDK7 with **source** and **target** compatibility set to 1.6?

TimvdLippe:
Because these are actually not checked. See the recent PR of the diamond generic which locally had not installed the jdk6

TimvdLippe:
See this specific comment: https://github.com/mockito/mockito/pull/1014#issuecomment-290741441

szpak:
@szczepiq Source/target version result in byte code to be in the correct version/format. It doesn't check if given construction (e.g. a method) was available in a given JDK version.

[...]

PR #1021, opened 2017-04-05
Change release build to JDK 7 with binary compatibility for Java 6

szpak:
In order to fix problems with Java 6 builds (https://github.com/mockito/mockito-release-tools/issues/60) 
we have discussed the possibility of using JDK 7 for building release artifacts (#1020). We can build release artifacts with JDK7 and use animal sniffer to detect Java 6 incompatibilities.

This PR contains:
 - new Gradle build plugin to use animal sniffer tool (the plugin is not used by default for faster builds)
 - removal of Java6 build from Travis
 - setup the sniffer to detect compatibility issues when building release artifacts on Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ branches:
 install:
  - true
 script:
-  - ./gradlew ciBuild release -s
+  - ./gradlew ciBuild release -s -Pcompatibility
 after_success:
   - ./gradlew --stacktrace coverageReport && cp build/reports/jacoco/mockitoCoverage/mockitoCoverage.xml jacoco.xml || echo "Code coverage failed"
   - bash <(curl -s https://codecov.io/bash) || echo "Codecov did not collect coverage reports"
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,14 +26,10 @@ before_install:
   - echo "$TRAVIS_COMMIT_MESSAGE"
 
 jdk:
-  - openjdk6
+  - oraclejdk7
 
 matrix:
   include:
-  - jdk: openjdk6
-    env: SKIP_RELEASE=true MOCK_MAKER=mock-maker-inline
-  - jdk: oraclejdk7
-    env: SKIP_RELEASE=true
   - jdk: oraclejdk7
     env: SKIP_RELEASE=true MOCK_MAKER=mock-maker-inline
   - jdk: oraclejdk8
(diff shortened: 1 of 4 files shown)


szczepiq:
I'd rather avoid extra dependencies on the build, I'd rather keep things simple and run builds on Java6. Extra plugins / dependencies are stuff we have to maintain. Also, I don't know the quality, and source code of the plugin, and we are doing interesting things in our builds (shipping continuously to a lot of users :).

However, animalsniffer seems like a pretty cool idea :)

szczepiq:
I'm softening :) It looks like there is no way to get clean build on Java6. Asked Travis CI support for help. Let's see what they say. If I don't get anything workable, I'll be ok with merging animal sniffer.

BTW. I checked the dependency graph of the animal sniffer and is reasonable and clean (animal-sniffer-ant-tasks, animal-sniffer, asm-all). I poke around the plugin's codebase and it looks clean and simple.

Thanks @szpak. BTW. are you using this plugin on any of your projects?

szpak:
@szczepiq Yup, in those that I would like to be compatible with Java 7 (such as [gradle-pitest-plugin](https://github.com/szpak/gradle-pitest-plugin) and [gradle-nexus-staging-plugin](https://github.com/Codearte/gradle-nexus-staging-plugin)).

Regarding dependency graph in a case it was a problem you could make classpath dependency on that also conditional based in `-Pcompatibility`.

szczepiq:
@szpak, very nice plugins! We should work together on the release tooling for the Java ecosystem :)

I put animal sniffer under scrutiny and I'm getting ready to merge.
 - tested / diffed the binaries, they are the same
 - added Objects.equals() to the source code and it is neatly detected by sniffer:

```
:animalsnifferMain

[Undefined reference] org.mockito.(Mockito.java:1731)
  >> String java.util.Objects.toString(Object)

:animalsnifferMain FAILED

* What went wrong:
Execution failed for task ':animalsnifferMain'.
> 1 AnimalSniffer violations were found in 1 files. See the report at: file:////Users/sfaber/mockito/src/build/reports/animalsniffer/main.text
```

szczepiq:
I've changed the implementation a bit. I also removed unnecessary Java6 builds from Travis CI. Faster and less builds!!!

Please review / merge at will.

gradle/root/java6-compatibility.gradle
```diff
@@ -0,0 +1,12 @@
+//Use animal sniffer to detect Java6 incompatibilities
+//We can remove it when we drop support for Java8
```
szpak: Java8? Maybe Java6?

.travis.yml
```diff
@@ -26,14 +26,10 @@ before_install:
...
 jdk:
-  - openjdk6
+  - oraclejdk7
 
 matrix:
   include:
-  - jdk: openjdk6
-    env: SKIP_RELEASE=true MOCK_MAKER=mock-maker-inline
-  - jdk: oraclejdk7
-    env: SKIP_RELEASE=true
```
szpak: Have you intentionally removed Java 7 build without mock-maker-inline?

szpak: Nevermind...

szczepiq:
Thank you @szpak!!!

bric3:
Grat job @szpak !

PR #1022, opened 2017-04-05
MatchersTest numeric literals

mureinik:
Use numeric literals, which are more elegant, instead of casting int literals to other primitive types.


src/test/java/org/mockitousage/matchers/MatchersTest.java
```diff
@@ -73,7 +73,7 @@ public void and_overloaded() {
         when(mock.oneArg(and(eq((double) 1), eq((double) 1)))).thenReturn("3");
         when(mock.oneArg(and(eq((float) 1), eq((float) 1)))).thenReturn("4");
```
TimvdLippe: Do we also want to update these while we are at it?

szczepiq:
Seems like a good idea. Thank you!

src/test/java/org/mockitousage/matchers/MatchersTest.java
mureinik: Good point, thanks. Pushed updated patch.

PR #1027, opened 2017-04-11
Fix some issues reported by SonarQube

TimvdLippe:
Colleague students ran SonarQube on Mockito. These were some easy fixes I was able to do, no extreme value. I will open an issue regarding a different and potentially more serious issue.

### DIFF
--- a/src/main/java/org/mockito/internal/invocation/TypeSafeMatching.java
+++ b/src/main/java/org/mockito/internal/invocation/TypeSafeMatching.java
@@ -65,6 +65,6 @@ public class TypeSafeMatching implements ArgumentMatcherAction {
         if (method.isBridge()) {
             return false;
         }
-        return method.getName().equals("matches");
+        return "matches".equals(method.getName());
     }
 }
--- a/src/main/java/org/mockito/internal/junit/UnnecessaryStubbingsReporter.java
+++ b/src/main/java/org/mockito/internal/junit/UnnecessaryStubbingsReporter.java
@@ -25,7 +25,7 @@ public class UnnecessaryStubbingsReporter implements MockCreationListener {
 
     public void validateUnusedStubs(Class<?> testClass, RunNotifier notifier) {
         Collection<Invocation> unused = new UnusedStubbingsFinder().getUnusedStubbingsByLocation(mocks);
-        if (unused.size() == 0) {
+        if (unused.isEmpty()) {
             return; //whoa!!! All stubbings were used!
         }
 
--- a/src/main/java/org/mockito/internal/matchers/text/ArrayIterator.java
+++ b/src/main/java/org/mockito/internal/matchers/text/ArrayIterator.java
@@ -6,6 +6,7 @@ package org.mockito.internal.matchers.text;
 
 import java.lang.reflect.Array;
 import java.util.Iterator;
+import java.util.NoSuchElementException;
 
 /**
  * Inspired on hamcrest, internal package class,
@@ -33,6 +34,9 @@ class ArrayIterator implements Iterator<Object> {
     }
 
     public Object next() {
+        if (!hasNext()) {
+            throw new NoSuchElementException();
+        }
         return Array.get(array, currentIndex++);
     }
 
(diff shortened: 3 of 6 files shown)


szpak:
If we were ok with their rules we could enable SonarQube public instance integration - https://sonarqube.com/organizations/default/projects

TimvdLippe:
To be honest, the output was rather superficial. There were not a lot of issues, and mostly concerning code that was working properly but just was incorrectly statically analyzed. Only the exception issue in #1028 was interesting, the others not so much. Thus, enabling it for every PR does not seem useful for me

src/main/java/org/mockito/internal/invocation/TypeSafeMatching.java
```diff
@@ -65,6 +65,6 @@ private static boolean isMatchesMethod(Method method) {
         if (method.isBridge()) {
             return false;
         }
-        return method.getName().equals("matches");
+        return "matches".equals(method.getName());
```
ChristianSchwarz: @TimvdLippe in this case the flip add no value since the Method.getName() never returns null, it makes the code less readable

src/main/java/org/mockito/internal/util/ObjectMethodsGuru.java
```diff
@@ -17,12 +17,12 @@ public static boolean isToStringMethod(Method method) {
         MockitoMethod m = new DelegatingMethod(method);
         return m.getReturnType() == String.class &&
                m.getParameterTypes().length == 0 &&
-               m.getName().equals("toString");
+               "toString".equals(m.getName());
```
ChristianSchwarz: @TimvdLippe in this case the flip add no value since the Method.getName() never returns null, it makes the code less readable

src/main/java/org/mockito/internal/util/ObjectMethodsGuru.java
```diff
@@ -17,12 +17,12 @@ public static boolean isToStringMethod(Method method) {
...
     }
 
     public static boolean isCompareToMethod(Method method) {
         return Comparable.class.isAssignableFrom(method.getDeclaringClass())
-                && method.getName().equals("compareTo")
+                && "compareTo".equals(method.getName())
```
ChristianSchwarz: @TimvdLippe in this case the flip add no value since the Method.getName() never returns null, it makes the code less readable

src/main/java/org/mockito/internal/matchers/text/ArrayIterator.java
```diff
@@ -33,6 +34,9 @@ public boolean hasNext() {
     }
 
     public Object next() {
+        if (!hasNext()) {
```
szczepiq: Nice, it throws much better exception now when we forgot to check for next element.

It would be cool to add unit tests for this code :)

src/main/java/org/mockito/internal/junit/UnnecessaryStubbingsReporter.java
```diff
@@ -25,7 +25,7 @@
 
     public void validateUnusedStubs(Class<?> testClass, RunNotifier notifier) {
         Collection<Invocation> unused = new UnusedStubbingsFinder().getUnusedStubbingsByLocation(mocks);
-        if (unused.size() == 0) {
+        if (unused.isEmpty()) {
```
szczepiq: Thanks for this change and similar ones related to isEmpty! The code is more readable this way.

src/main/java/org/mockito/internal/invocation/TypeSafeMatching.java
szczepiq: Interesting feedback!

Personally, flip or no-flip has no difference to me for readability. String literal on LHS is in general safer so the change is good IMHO.

TimvdLippe:
Let's use the flipped version for now. Even if it will always return Strings now, it might change later. Merging, thanks for the input @ChristianSchwarz and @szczepiq !

src/main/java/org/mockito/internal/matchers/text/ArrayIterator.java
TimvdLippe: More unit tests is a bigger TODO. Let's create a clear issue to request the community to contribute?

PR #1030, opened 2017-04-12
Fixed notable version

szczepiq:
- 2.7.0 does not exist :)
- this fixes the release automation issue [reported by Travis CI](https://travis-ci.org/mockito/mockito/jobs/221198113#L709)


szczepiq:
Given that this is only a change in version metadata, I'm merging without waiting for review ;) This will fix releasing from Travis CI!

PR #1032, opened 2017-04-12
Support Kotlin suspending functions

elizarov:
Original user problem report:
https://discuss.kotlinlang.org/t/verifying-suspending-functions-with-mockito-or-alternatives/2492

The crux of the problem is that Kotlin compiler adds additional synthetic parameter
to suspending functions that is not visible in the source as explained in the design docs:
https://github.com/Kotlin/kotlin-coroutines/blob/master/kotlin-coroutines-informal.md#continuation-passing-style

Expanded test cases in this repo pass with this patch:
https://github.com/elizarov/CoroutinesMockitoPlayground


elizarov:
If you don't mind introducing dependency on Kotlin into your test suit, then it should be possible. The test case from here can be used: 
https://github.com/elizarov/CoroutinesMockitoPlayground/blob/master/src/test/kotlin/SuspendableTest.kt
https://github.com/elizarov/CoroutinesMockitoPlayground/blob/master/src/main/kotlin/Suspendable.kt

I have not figured out what would be the best place for it, though. Where would be the appropriate place to put it to?

elizarov:
Unfortunately, I had to make this patch slightly larger (introduce `MockitoMethod.isSuspend`). The original one relied on comparison of # of params and # of args and did not pass all the existing tests for varargs.

TimvdLippe:
Hm, we might be able to introduce Kotlin as a test-dependency, but that would be a big change. @mockito/core what do you think?

elizarov:
Ok. That is what I did. I added Kotlin as test dependency and introduced a test for the corresponding functionality. Works well on my machine. Let us see if it passes all CI tests.

elizarov:
I've simplified the code a little to minimize patch size. It passes CI build. However, it still does not pass codecov check and I fail to understand why. All new lines of code seem to be covered, yet Codecov reports that there is only 81.25% of diff covered. In particular, it reports only partial coverage of this line: https://codecov.io/gh/mockito/mockito/pull/1032/diff#D6-18 (line 18 in "SuspendMethod.java")

However, if I run `mockitoCoverage` gradle task locally and load the resulting "build/jacoco/test.exec" file into IDEA coverage analysis tool (Analyse | Show coverage data) then I see that this line in 100% covered which is consistent with the fact that analysis of "build/reports/tests/classes" directory shows that `org.mockito.kotlin.SuspendTest` is successfully executed and passes. Any advice?

TimvdLippe:
The failing status of CodeCov is okay. It can sometimes be a bit off 😄 

I am okay with adding Kotlin as test dependency, feels good to make sure Mockito works cross-language! Adding the `isSuspend` to the interface seems less desirable. Could you expand on why this is really required? My Kotlin knowledge is almost non-existant, so apologies if the question is very simple.

elizarov:
My adding `isSuspend` to an `MockitoMethod` interface is purely Mockito-related. I've tried to avoid doing it, but failed. Here's the story:

1. In order to support Kotlin suspending functions we need to strip out the last synthetic continuation parameter from the result of `MockitoMethod.getParameterTypes()`. Stripping it out makes all the other code in Mockito properly see the original Kotlin's signature of the Kotlin's suspending function (without the synthetic continuation parameters that was added during compilation)
2. We also need to strip the last argument from `Invocation.getArguments()`, but leave it in the `Invocation.getRawArguments()` so that method invocation can be properly forwarded.
To that end, I've changed the code of `InterceptedInvocation` constructor. It now has this line:

```
this.arguments = ArgumentsProcessor.expandArgs(mockitoMethod, arguments);
```

Now, `ArgumentsProcessor.expandArgs` implementation needs to somehow be able to figure that `mockitoMethod` was a suspending function whose last continuation parameter was stripped out. However, because it was already stripped out, it cannot just take a look at `MockitoMethod.getParameterTypes()` (the continuation is not there anymore). I've solved it by adding `MockitoMethod.isSuspend()`. The implementation of `ArgumentsProcessor.expandArgs` is where this method is used.

I've tried a different approach in `ArgumentsProcessor.expandArgs` implementation. Instead of this line:

```
if (method.isSuspend())
```

 I've tried to use 

```
if (args != null && args.length > method.getParameterTypes().length)
```

It makes my suspend tests pass,  but it breaks some vararg tests. In particular, the following two tests fail:

```
org.mockito.internal.stubbing.answers.ReturnsArgumentAtTest.should_be_able_to_return_the_last_parameter
org.mockito.internal.stubbing.answers.ReturnsArgumentAtTest.should_be_able_to_return_the_specified_parameter
```

I lack sufficient understanding on how varargs are supposed to be treated in Mockito architecture. 

It seems that maybe those two tests are just incorrect, since they just pass varargs parameters in already expanded form into the method invocation piggy-backing on the fact that the original implementation of `ArgumentsProcessor.expandVarArgs` only performs expansion if the last parameter is indeed an array. Is this a feature or just an implementation detail that should not be relied upon? 

If we can confirm that those two tests are indeed incorrect and fix them (by passing an array as the last parameter), then it opens the way to easily get rid of `MockitoMethod.isSuspend()` function in the way I've shown above -- we can simply change `ArgumentsProcessor` to always trim the array of parameters to the number of method arguments reported by `MockitoMethod`.

TimvdLippe:
Thanks for the very detailed writeup! These tests are ancient (written >5 years ago). I would be fine by passing an array instead, but would like to know what @bric3 thinks (as he is probably the most familiar with that code). Then we would be able to get rid of the method :smile:

elizarov:
I've added a separate commit getting rid of `MockitoMethod.isSuspending` and fixing `ReturnsArgumentAtTest`. Indeed, digging down into this test, I've became convinced that this is simply a bug in the test, where vararg params are not properly wrapped into array (see one-line fix on it in the commit). 

I suggest to squash commits when merging.

szczepiq:
THANKS A LOT for making Mockito work well with Kotlin! I'm reviewing right now.

build.gradle
```diff
@@ -17,6 +17,7 @@ buildscript {
 
 plugins {
     id 'com.gradle.build-scan' version '1.0'
+    id 'org.jetbrains.kotlin.jvm' version '1.1.1'
```
szczepiq: I really, really like the idea of adding tests in Kotlin! This makes Mockito coverage much better.

I do have one request. Can we move the Kotlin test to a separate subproject? I would feel safer if we avoid using Kotlin plugin for the Gradle subproject that builds main distribution of Mockito. I haven't scrutinized the Kotlin plugin but plugins can add dependencies and do other interesting things that can alter the binary or pom of the distribution.

Also, it's safer for the core tests if they are decoupled from Kotlin and run without Kotlin libraries on the classpath.

src/main/java/org/mockito/internal/creation/DelegatingMethod.java
```diff
@@ -31,7 +31,7 @@ public String getName() {
     }
 
     public Class<?>[] getParameterTypes() {
-        return method.getParameterTypes();
+        return SuspendMethod.trimSuspendParameterTypes(method.getParameterTypes());
```
szczepiq: Can we create field on the class that contains trimmed parameter types? This method is invoked frequently (with every invocation on mock), so why trimming every time :)

src/main/java/org/mockito/internal/creation/SuspendMethod.java
```diff
@@ -0,0 +1,18 @@
+package org.mockito.internal.creation;
+
+import java.util.Arrays;
+
+/**
+ * Utilities for Kotlin Continuation-Passing-Style suspending function, detecting and trimming last hidden parameter.
+ * See <a href="https://github.com/Kotlin/kotlin-coroutines/blob/master/kotlin-coroutines-informal.md#continuation-passing-style">Design docs for details</a>.
+ */
+public class SuspendMethod {
```
szczepiq: Thank you very much for adding comprehensive Javadoc with the context and link to Kotlin docs!

src/main/java/org/mockito/internal/invocation/ArgumentsProcessor.java
```diff
@@ -9,15 +9,25 @@
...
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 /**
  * by Szczepan Faber, created at: 3/31/12
  */
 public class ArgumentsProcessor {
+    // drops hidden synthetic parameters (last continuation parameter from Kotlin suspending functions)
```
szczepiq: Again, thanks for the code comments! Very useful in long term maintenance of the project.

src/test/kotlin/org/mockito/kotlin/SuspendTest.kt
```diff
@@ -0,0 +1,37 @@
...
+import kotlinx.coroutines.experimental.runBlocking
+import org.hamcrest.core.IsEqual
+import org.junit.Assert.assertThat
+import org.junit.Test
+import org.mockito.Mockito.*
+
+/**
+ * Test for mocking and verifying Kotlin suspending functions.
+ */
+class SuspendTest {
```
szczepiq: I love the Kotlin test. We should have more of them I think!

Thanks a lot, this is our first ever official test coverage for different language :D

szczepiq:
I pushed kotlin tests to separate submodule and pushed to branch "elizarov-coroutines". Do you want pull those changes and update the PR?

Alternatively, I'll just submit new PR from the new branch.

Thanks again for the fix!

elizarov:
I've updated this PR with all the recent stuff. Moving Kotlin test to a separate project looks great. Thanks a lot for that. I've also added few more tests to `SuspendTest` (now in `kotlinTest` module) to be 100% sure that interactions between suspending functions and other Kotlin features (namely - default parameters) work fine both for mocking interfaces and for mocking classes.  It all works like a charm.

TimvdLippe:
This is great, thanks for the update @szczepiq ! I wonder though why CodeCov is not picking up the coverage: https://codecov.io/gh/mockito/mockito/compare/4a79615d78b34cc6ba6a8995dcaef6fa1d9a3466...bf092378fc42f31efa054b4e5e555c10c33f5994/diff#D7-15 Must we reconfigure jacoco to also run on the Kotlin subproject of fix this?

TimvdLippe:
Oh I see we added a TODO for this one. I will extract an issue for that and merge this PR. @elizarov Thanks for your effort. If you have any future issues, we are delighted to integrate your fixes ❤️

szczepiq:
Awesome ;) Thank you guys for finishing it off!

subprojects/kotlinTest/src/test/kotlin/org/mockito/kotlin/SuspendTest.kt
```diff
@@ -0,0 +1,107 @@
...
+        verify(mockSuspendable).suspendFunction(2)
+    }
+
+    @Test
+    fun shouldMockSuspendInterfaceFunctionWithDefault() = runBlocking<Unit> {
+        val mockSuspendable: SuspendableInterface = mock(SuspendableInterface::class.java)
+        val callsSuspendable = CallsSuspendable(mockSuspendable)
+
+        `when`(mockSuspendable.suspendFunctionWithDefault()).thenReturn(2)
+        `when`(mockSuspendable.suspendFunctionWithDefault(intThat { it >= 2 })).thenReturn(4)
```
szpak: @elizarov I'm curious, wouldn't be easier to just use `given/willReturn` instead of quoted `when/thenReturn` in Kotlin code?

src/main/java/org/mockito/internal/creation/SuspendMethod.java
```diff
@@ -0,0 +1,18 @@
...
+    private static final String KOTLIN_CONTINUATION = "kotlin.coroutines.experimental.Continuation";
```
felipecsl: @elizarov will this break now that kotlin coroutines are stable?

PR #1033, opened 2017-04-12
Fix NullPointerException on TestName.getTestName

yyvess:
On case of  `MockitoAnnotations.initMocks(target)`  throw any exception `this.target` still null.
As impact to throw a **NullPointerException on all following tests** from the method `DefaultInternalRunner.testFinished`.

To avoid this issue `this.target` should be defined before calling `initMocks()`.

Original error :

>  Underlying exception : java.lang.UnsupportedOperationException: class redefinition failed: attempted to add a method
	at org.mockito.internal.runners.DefaultInternalRunner$1.withBefores(DefaultInternalRunner.java:39)
	at org.junit.runners.BlockJUnit4ClassRunner.methodBlock(BlockJUnit4ClassRunner.java:276)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.mockito.internal.runners.DefaultInternalRunner$1.run(DefaultInternalRunner.java:72)
`

Following errors :

> java.lang.NullPointerException 
at org.mockito.internal.junit.util.TestName.getTestName(TestName.java:15)
	at org.mockito.internal.junit.MismatchReportingTestListener.testFinished(MismatchReportingTestListener.java:33)
	at org.mockito.internal.runners.DefaultInternalRunner$1$1.testFinished(DefaultInternalRunner.java:60)
	at org.junit.runner.notification.SynchronizedRunListener.testFinished(SynchronizedRunListener.java:56)
`

### DIFF
--- a/src/main/java/org/mockito/internal/runners/DefaultInternalRunner.java
+++ b/src/main/java/org/mockito/internal/runners/DefaultInternalRunner.java
@@ -35,8 +35,8 @@ public class DefaultInternalRunner implements InternalRunner {
                 Mockito.framework().addListener(mockitoTestListener);
 
                 // init annotated mocks before tests
-                MockitoAnnotations.initMocks(target);
                 this.target = target;
+                MockitoAnnotations.initMocks(target);
                 return super.withBefores(method, target, statement);
             }
 


TimvdLippe:
I am not really sure if I understand the issue. Could you add a test to our testsuite to show what the problem is and to make sure that the issue has been resolved?

yyvess:
Simple step to reproduce the issue : 
 
1) On the test `org.mockito.internal.configuration.injection.ConstructorInjectionTest`, mock a static class 
` 
public class ConstructorInjectionTest {
    @Mock
    private System system;
` 

2) Run all mockito tests

You should have most of tests in failure inside to have only the  `ConstructorInjectionTest` as faild.

TimvdLippe:
@yyvess could you add these steps to the Test you talked about and add it to this PR?

yyvess:
@TimvdLippe If I do that it's will break the build :-( I just explain how to reproduce the issue.

I have try to create a testable case it's hard very hard ... 
First the bug occur only when mocks are initialized on @Before method.
Second the class DefaultInternalRunner is not designed to be testable whit the interclass BlockJUnit4ClassRunner.

TimvdLippe:
Okay, could you create a GitHub gist which reproduces the issue? I can turn it into a unit test then :)

yyvess:
Close this pr, I have open a new one (sorry for the noise)

PR #1035, opened 2017-04-13
MockitoTestListener must be remove on case test fail on the initialization.

yyvess:
On case of  `MockitoAnnotations.initMocks(target)`  throw any exception `this.target` still null.
As impact to throw a **NullPointerException on all next tests** on the method `DefaultInternalRunner.testFinished`.

What append, the method testFinished is never call on case of any error on the initialization phase as impact that the listener is not cleanup.
On next tests the method testFinished will notify the previous listener that have not been unregistered.

To avoid this issue the mockitoListener must be remove if test fail and never started.

> java.lang.NullPointerException 
at org.mockito.internal.junit.util.TestName.getTestName(TestName.java:15)
	at org.mockito.internal.junit.MismatchReportingTestListener.testFinished(MismatchReportingTestListener.java:33)
	at org.mockito.internal.runners.DefaultInternalRunner$1$1.testFinished(DefaultInternalRunner.java:60)
	at org.junit.runner.notification.SynchronizedRunListener.testFinished(SynchronizedRunListener.java:56)
`

### DIFF
--- a/src/main/java/org/mockito/internal/runners/DefaultInternalRunner.java
+++ b/src/main/java/org/mockito/internal/runners/DefaultInternalRunner.java
@@ -30,40 +30,50 @@ public class DefaultInternalRunner implements InternalRunner {
             private MockitoTestListener mockitoTestListener;
 
             protected Statement withBefores(FrameworkMethod method, Object target, Statement statement) {
-                // get new test listener and add add it to the framework
+                this.target = target;
+                // get new test listener and add it to the framework
                 mockitoTestListener = listenerSupplier.get();
                 Mockito.framework().addListener(mockitoTestListener);
-
                 // init annotated mocks before tests
                 MockitoAnnotations.initMocks(target);
-                this.target = target;
                 return super.withBefores(method, target, statement);
             }
 
             public void run(final RunNotifier notifier) {
                 RunListener listener = new RunListener() {
+                    private boolean started;
                     Throwable failure;
 
-                    public void testFailure(Failure failure) throws Exception {
-                        this.failure = failure.getException();
+                    @Override
+                    public void testStarted(Description description) throws Exception {
+                        started = true;
                     }
 
+                    @Override
+                    public void testFailure(Failure failure) throws Exception {
+                        this.failure = failure.getException();
+                        // If the test fails during the setup, `testFinished` is never invoked
+                        // Therefore, if we have not started, cleanup the testlistener
+                        if (!started && mockitoTestListener != null) {
+                            Mockito.framework().removeListener(mockitoTestListener);
+                        }
+                    }
+
+                    @Override
                     public void testFinished(Description description) throws Exception {
-                        super.testFinished(description);
                         try {
                             if (mockitoTestListener != null) {
                                 Mockito.framework().removeListener(mockitoTestListener);
                                 mockitoTestListener.testFinished(new DefaultTestFinishedEvent(target, description.getMethodName(), failure));
                             }
                             Mockito.validateMockitoUsage();
-                        } catch(Throwable t) {
+                        } catch (Throwable t) {
                             //In order to produce clean exception to the user we need to fire test failure with the right description
                             //Otherwise JUnit framework will report failure with some generic test name
                             notifier.fireTestFailure(new Failure(description, t));
                         }
                     }
                 };
-
                 notifier.addListener(listener);
                 super.run(notifier);
             }
--- a/src/test/java/org/mockito/internal/runners/DefaultInternalRunnerTest.java
+++ b/src/test/java/org/mockito/internal/runners/DefaultInternalRunnerTest.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright (c) 2017 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.internal.runners;
+
+import org.junit.Test;
+import org.junit.runner.Description;
+import org.junit.runner.notification.Failure;
+import org.junit.runner.notification.RunListener;
+import org.junit.runner.notification.RunNotifier;
+import org.mockito.Mock;
+import org.mockito.internal.junit.MockitoTestListener;
+import org.mockito.internal.junit.TestFinishedEvent;
+import org.mockito.internal.util.Supplier;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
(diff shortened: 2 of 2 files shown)


src/test/java/org/mockito/internal/runners/DefaultInternalRunnerTest.java
```diff
@@ -0,0 +1,71 @@
...
+    };
+
+    @Test
+    public void nominal() throws InitializationError {
+        InternalRunner runner = new DefaultInternalRunner(SuccessTest.class, supplier);
+        runner.run(failIfFailure());
+    }
+
+    @Test
+    public void second_test_must_success() throws InitializationError {
```
TimvdLippe: Please add a more descriptive name here. Probably something like `does_not_throw_NPE_when_setup_failed_after_second_run()`?

src/test/java/org/mockito/internal/runners/DefaultInternalRunnerTest.java
```diff
@@ -0,0 +1,71 @@
...
+        InternalRunner failRunner = new DefaultInternalRunner(TestFailOnInitialization.class, supplier);
+        failRunner.run(new RunNotifier());
+
+        InternalRunner successRunner = new DefaultInternalRunner(SuccessTest.class, supplier);
+        successRunner.run(failIfFailure());
```
TimvdLippe: Can we add assertions for these invocations? Would like to make sure the tests were actually executed as expected.

src/test/java/org/mockito/internal/runners/DefaultInternalRunnerTest.java
```diff
@@ -0,0 +1,71 @@
...
+    private final Supplier<MockitoTestListener> supplier = new Supplier<MockitoTestListener>() {
+        public MockitoTestListener get() {
+            return new NoOpTestListener();
+        }
+    };
+
+    @Test
+    public void nominal() throws InitializationError {
+        InternalRunner runner = new DefaultInternalRunner(SuccessTest.class, supplier);
+        runner.run(failIfFailure());
```
TimvdLippe: Please add an assertion here too.

src/test/java/org/mockito/internal/runners/DefaultInternalRunnerTest.java
```diff
@@ -0,0 +1,71 @@
...
+public class DefaultInternalRunnerTest {
+
+    private final Supplier<MockitoTestListener> supplier = new Supplier<MockitoTestListener>() {
+        public MockitoTestListener get() {
+            return new NoOpTestListener();
+        }
+    };
+
+    @Test
+    public void nominal() throws InitializationError {
```
TimvdLippe: `does_not_fail_when_tests_succeeds()`

src/main/java/org/mockito/internal/runners/DefaultInternalRunner.java
```diff
@@ -26,44 +26,50 @@
...
-
                 // init annotated mocks before tests
                 MockitoAnnotations.initMocks(target);
-                this.target = target;
                 return super.withBefores(method, target, statement);
             }
 
             public void run(final RunNotifier notifier) {
-                RunListener listener = new RunListener() {
+                final RunListener listener = new RunListener() {
```
TimvdLippe: Nit: not really needed

src/main/java/org/mockito/internal/runners/DefaultInternalRunner.java
```diff
@@ -26,44 +26,50 @@
     public DefaultInternalRunner(Class<?> testClass, final Supplier<MockitoTestListener> listenerSupplier) throws InitializationError {
         runner = new BlockJUnit4ClassRunner(testClass) {
 
-            public Object target;
+            private Object target;
```
TimvdLippe: This might be a breaking change. Let's revert it for now.

src/main/java/org/mockito/internal/runners/DefaultInternalRunner.java
```diff
@@ -26,44 +26,50 @@
...
                     Throwable failure;
+                    boolean started;
+
+                    public void testStarted(Description description) throws Exception {
+                        started = true;
+                    }
 
                     public void testFailure(Failure failure) throws Exception {
                         this.failure = failure.getException();
+                        // If the test have failed before to started remove listener
```
TimvdLippe: Let's make this a bit more clear:
```
// If the test fails during the setup, `testFinished` is never invoked
// Therefore, if we have not started, cleanup the testlistener
```

yyvess:
All remarks have been fixed.

src/main/java/org/mockito/internal/runners/DefaultInternalRunner.java
```diff
@@ -30,40 +30,50 @@ public DefaultInternalRunner(Class<?> testClass, final Supplier<MockitoTestListe
...
+                        // If the test fails during the setup, `testFinished` is never invoked
+                        // Therefore, if we have not started, cleanup the testlistener
+                        if (!started && mockitoTestListener != null) {
+                            Mockito.framework().removeListener(mockitoTestListener);
+                        }
                     }
 
+                    @Override
                     public void testFinished(Description description) throws Exception {
-                        super.testFinished(description);
```
TimvdLippe: Why did you remove this?

yyvess: Because the class RunListener is documented that 
have no implementation.
It more like a interface with defaut empty implementation.

https://github.com/junit-team/junit4/blob/master/src/main/java/org/junit/runner/notification/RunListener.java

TimvdLippe: Ah okay, that seems fine!

TimvdLippe:
Merging this one, seems like a solid fix. Thanks for your work!

szczepiq:
Wow. I was just reviewing changes for potential notable release and I found this PR. Thank you very much for fixing it. The problem was pretty deep in our code, nice job! I made it noteworthy and it will be included in notable release notes when we publish the next release to maven central ([more details](https://github.com/mockito/mockito/wiki/Continuous-Delivery-Details#cd)).

yyvess:
For the storry. This issue have been detected during a migration from Mockito 1.x / PowerMock to Mockito 2.x. 
After fixed small compilation issues, most of tests will failed.
Little bit desperate at first time. 
In reallity most of failures has related to this issue and only few tests requierd a fix to finalize the migration.
Tanks for your API.

PR #1038, opened 2017-04-14
Fixed glitches with notable releases

szczepiq:
Fixed a couple of glitches:
 - one of the notable versions was not published to Central. I don't plan to publish it to central because it was published to wrong Bintray package in the first place. It landed in "mockito-development" but it should land in "mockito".
 - fixed the use of extra property for notable releasing according to release tooling api (need to request the property from the root project).
 - consumed new version of release tools that contains host of internal improvements (mockito/mockito-release-tools#74)


szczepiq:
I'm merging if that's ok. If not, let me know :) This change is a configuration-only and does not influence binary artifacts.

PR #1040, opened 2017-04-16
Fix javadoc issues

mureinik:
This PR fixes several javadoc issues reported by IntelliJ IDEA. Two types of errors are not handled in this PR:

- References which gradle is unable to resolve - these references seem to be valid javadoc links, so there's either an issue with gradle itself of with the configuration this project is using, and needs a deeper look
- Javadoc tags missing descriptions, which require a deeper understanding of the methods in question in order to resolve.


TimvdLippe:
We have some other javadoc warnings too. Might want to take a stab at these too? :)

mureinik:
@TimvdLippe I was trying to keep the PR minimal, but sure, I'll give it a go and add it to the PR.

PR #1051, opened 2017-04-26
Fixes #1050 : Improve error if class not found

### LINKED ISSUE #1050: Can't mock error message is confusing when root cause is ClassNotFoundException
If you attempt to mock a class which has a dependency that isn't in the test classpath, you will get the standard message:
>Mockito can only mock non-private & non-final classes.
>If you're not sure why you're getting this error, please report to the mailing list.

You have to dig down through the stack trace to see that the root cause was a `ClassNotFoundException`. This can be very confusing for a user (it was for two people in my organization).

My suggestion is to detect if there's a `ClassNotFoundException` as a cause of the failure to create an instance of the class, and if so, report that, instead of the default message.

This will improve usability in Mockito, making users more productive, and help prevent future support requests.

tyrel:
If a `ClassNotFoundException` is the root cause of not being able to mock an object, present that information to the user, instead of saying it's probably because the class is private or final. This can avoid a lot of confusion when a user is trying to figure out why they can't mock an object.

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassByteBuddyMockMaker.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassByteBuddyMockMaker.java
@@ -100,6 +100,14 @@ public class SubclassByteBuddyMockMaker implements ClassCreatingMockMaker {
                     ""
             ), generationFailed);
         }
+        ClassNotFoundException classNotFoundException = findClassNotFoundExceptionIfAny(generationFailed);
+        if (classNotFoundException != null) {
+            throw new MockitoException(join(
+                "Mockito cannot mock this class: " + mockFeatures.getTypeToMock() + ",",
+                "because a dependent class is not in the classpath: " + classNotFoundException.getMessage(),
+                ""
+            ), generationFailed);
+        }
         throw new MockitoException(join(
                 "Mockito cannot mock this class: " + mockFeatures.getTypeToMock() + ".",
                 "",
@@ -114,6 +122,18 @@ public class SubclassByteBuddyMockMaker implements ClassCreatingMockMaker {
                 "",
                 "Underlying exception : " + generationFailed
         ), generationFailed);
+    }
+
+    private ClassNotFoundException findClassNotFoundExceptionIfAny(Throwable generationFailed) {
+        if (generationFailed == null) {
+            return null;
+        }
+
+        if (generationFailed instanceof ClassNotFoundException) {
+            return (ClassNotFoundException)generationFailed;
+        }
+
+        return findClassNotFoundExceptionIfAny(generationFailed.getCause());
     }
 
     private static String describeClass(Class<?> type) {


tyrel:
@TimvdLippe I thought about it for a while. The unit test class SubclassByteBuddyMockMakerTest.java doesn't have any existing tests that end up calling this method to verify any of its output. Codecov shows a little bit of coverage but it's not clear to me where that coverage is coming from.

If someone can point me to appropriate test classes I might be able to add a test for this, but it would be weird, because a dependency would need to be added to the test that's not available at test time, so I'm not sure how it would compile...

TimvdLippe:
Hm yes coverage is quite low in this respect, we should improve on this. The coverage that you are seeing is regarding the mocking of private classes tested at https://github.com/mockito/mockito/blob/e9516f42b1182ae675022c101954e721e20b2c72/src/test/java/org/mockitousage/spies/SpyingOnRealObjectsTest.java#L182

szczepiq:
Hey @tyrel! Thank you very much for the PR - we definitely want to improve Mockito exception messages because it really helps our users!

How did you validate that the new behavior helps? Can you provide us with a sample project that we can use for testing? (or steps to reproduce the problem). I was also thinking how can we test that. I think we have 3 options:
 - unit test
 - (maybe?) test with custom classloader, similarly how we do (SimplePerRealmReloadingClassLoader)
 - separate tiny test project under 'subprojects'

I'm more than happy to merge it if I can see it working :D, ideally with every invocation of Mockito build. Thank you for your patience and sorry for not getting enough traction on this PR for past 2 weeks!

szczepiq:
Friendly ping @tyrel :)

How we test the change? We don't require you to come up with fancy coverage for this (unless you figure out how to do it). I only want to see it working.

tyrel:
Yeah, it's kind of complicated. You have to have a dependency that exists at compile time but then is missing at test time. Our situation was rather complex, but basically, a class in the project you're testing has to depend on something that's not in the test-time classpath.

I've been trying to think of a simple way to simulate this without having to replicate the entire environment we have.... I think if you mock a `class A` that has a member of type `B`, and then you build the test, then remove `B `from the classpath before running the test, you'll get a `ClassNotFoundException` when Mockito tries to mock an `A`, since `B` isn't in the classpath. In our case the classpath was complex and an odd situation lead to it being different when tests build vs when tests run.

szczepiq:
Thank you for the hints!

>I think if you mock a class A that has a member of type B, and then you build the test, then remove B from the classpath before running the test, you'll get a ClassNotFoundException when Mockito tries to mock an A, since B isn't in the classpath.

I think this is something I will ask you or the community to contribute :) One of the core devs will try out your suggestion at some point when we have a chance. I the meantime if you or some other eager engineer can try to put together a sample project (even super hacky one :D) and then give us steps to reproduce that would be awesome!!!

ChristianSchwarz:
Mockito has some onboard utilities to create exotic classloader setups e.g `ClassLoaders`.  Hopefully this works, no gurantees, not tested...

```
package package_a;
class A extends package_b.B {
}
```

```
private static Classloader classLoaderWithoutPackageB = ClassLoaders.excludingClassLoader().withCodeSourceUrlOf(A.class).without("package_b").build();

    @BeforeClass
    public static void init() throws ClassNotFoundException {
        Class<?> type= classLoaderWithoutPackageB .loadClass(A.class.getName());
      
 
      mock(type); //should throw the expected exception
    }
```

tyrel:
Sure, I'll throw something together when I get time.

ChristianSchwarz:
This should be enough to test the use case:

```
public class Test {

    @Test
    public  void test() throws ClassNotFoundException {
        SubclassLoader loader = createSubclassLoader();
        MockCreationSettings settings = new MockSettingsImpl().build(Runnable.class);

        new SubclassByteBuddyMockMaker(loader).createMock(settings, dummyHandler());
    }


    private static SubclassLoader createSubclassLoader() {
        
        final ClassLoadingStrategy<ClassLoader> strategy = new ClassLoadingStrategy<ClassLoader>() {
            @Override
            public Map<TypeDescription, Class<?>> load(ClassLoader classLoader, Map<TypeDescription, byte[]> types) {
                throw new MockitoException("kaboom", new ClassNotFoundException("package.name.of.the.DependingClass"));
            }
        };
        
        return new SubclassLoader() {
            @Override
            public ClassLoadingStrategy<ClassLoader> getStrategy(Class<?> mockedType) {
                return strategy;
            }
        };
    }

    private static MockHandler dummyHandler() {
        return new MockHandler() {
            @Override
            public Object handle(Invocation invocation) {return null;}

            @Override
            public MockCreationSettings getMockSettings() {return null; }

            @Override
            public InvocationContainer getInvocationContainer() {return null;}
        };
    }   
}
```

raphw:
I like this idea of improving the error message but we need unit tests around this.

Can I suggest to use Byte Buddy for creating a class that refers to a class that does not exist? It is already on the class path for Mockito. You can create a type by `InstrumentedType.Default.subclass("DoesNotExist", Object.class)` and then refer to this `TypeDescription` from the Byte Buddy DSL.

This class can be loaded into a custom class loader (`WRAPPER` loading stategy) from which it can be attempted to be mocked what should trigger this exception.

TimvdLippe:

tyrel:
Sad. I was hoping someone would have written the tests.

TimvdLippe:
There is some guidance posted above to write a test, so if you are still interested in landing these changes, please create a new PR with a test based on the above guidance.

tyrel:
I know, but this is way out of my league. I just wanted to get a better error message.

Issue #984, opened 2017-03-09
Release commits do not trigger code coverage report


TimvdLippe:
Found out at https://github.com/mockito/mockito/pull/980#issuecomment-285124740
This was initially fixed in #493, but I think the adoption of the new mockito-release-tools repository introduced this regression. All pull requests now have a failing status as CodeCov is unable to provide any feedback.

Commits should instead of `[ci skip]` use `[ci skip-release]`. Therefore Travis will kick off, but the release mechanism is not triggered.

szczepiq:
Sorry for slow turnaround. It will get fixed soon!

szczepiq:
My understanding is that code cov comparisons need every commit to produce report. Otherwise codecov "cannot find base report".

Running build for every commit defeats the purpose of "ci skip" :) Can we work with codecov tooling to accommodate this? Even if we make automated commits trigger builds, we still want to use "ci skip" at times.

Do you advocate to use "ci skip release" in place of "ci skip" for mockito project in general?

TimvdLippe:
Asked upstream: https://github.com/codecov/support/issues/355#issuecomment-292707657

szczepiq:
Thank you for asking! I'll unassign myself from the ticket for the time being.

szczepiq:
Created ticket for it: https://github.com/mockito/mockito-release-tools/issues/92

TimvdLippe:
This was fixed some time ago, probably in #1052.

PR #1052, opened 2017-04-28
Fixed problems with releases

szczepiq:
- fixed the logic that determines if the release is notable or not.
We shipped a couple of releases to wrong Bintray repo but we should be ok :)
I plan to make this configuration simpler but now it has to be this way.
- bumped to new version of tools to pick up a couple of improvements
including the fix for #984
- configured commit message postfix that will ensure that
code coverage is collected after CI automated commits (see #984)
- disabled automatic publication because it's still not ready

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,4 +2,4 @@ org.gradle.daemon=true
 org.gradle.parallel=true
 
 #Declared the dependency here to avoid duplication across build.gradle files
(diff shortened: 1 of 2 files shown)


szczepiq:
Merging, this change does not alter the binaries / behavior of Mockito so I'll go ahead without formal review.

PR #1055, opened 2017-04-29
Bumped version of release tools

szczepiq:
- cleaner notable repository support
- fixed the commit message postfix so that it includes the build number


szczepiq:
I'll merge it if that's ok. The change does not alter the behavior nor the binaries of Mockito.

PR #1057, opened 2017-04-30
Ensured the release actually happens automatically

szczepiq:
New version of release tools make the release use 'dryRun' by default.
This way, we have less risk of accidental release.
I updated .travis.yml file to turn off dry run in release build step.

Please prioritize this merge over other changes
so that we can continue releasing automatically.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,6 @@ install:
 script:
   - ./gradlew build idea -s -PcheckJava6Compatibility
 after_success:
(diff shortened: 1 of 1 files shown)


TimvdLippe:
This is not working as intended. The builds are now in a loop. Apparently the `[ci skip-release]` is disregarded or overriden by this change?

TimvdLippe:
https://travis-ci.org/mockito/mockito/jobs/227355853#L446
```
:assertReleaseNeeded
  Release is needed: true
    - skip by env variable: false
    - skip by commit message: false
    - is pull request build:  false
    - is releasable branch:  true
```

PR #1058, opened 2017-04-30
Test code refactoring - updated assertj test dependency from 1.x to 2.6.0

epeee:
make use of assertj 2.6.0 since java 7 is used now for building mockito and assertj 1.x is no longer under development.


szczepiq:
Thank you for updating the version! It's nice that you can help us keeping good hygiene of versions - something easy to forget during day-to-day execution of the project :)

epeee:
You're welcome!

Issue #1059, opened 2017-04-30
Release automation endless loop issue


szczepiq:
Recently I introduced a problem (PR: https://github.com/mockito/mockito/pull/1057) that created endless loop with our releases. Fallout / clean up needed:
 - [x] - stop the endless loop by reverting the commit. Thank you @TimvdLippe spotting the issue and reverting the change!
 - [x] - remove unnecessary tags
 - [x] - discard versions/artifacts that were pushed to Bintray (not "published" yet, safe to delete)
 - [x] - delete unnecessary content from [release notes file](https://github.com/mockito/mockito/blob/release/2.x/doc/release-notes/official.md)
 - [x] - fix the bug in release automation code, pr submitted, waiting for review & merge: https://github.com/mockito/mockito-release-tools/pull/103
 - [x] - test & reapply PR #1057

szczepiq:
Waiting for review / merge: https://github.com/mockito/mockito-release-tools/pull/103

TimvdLippe:
Fix seems fine to me. Have not executed it locally though.

szpak:
I don't know if this is related, but the release process currently [fails](https://travis-ci.org/mockito/mockito/jobs/230865593#L796) with:
> remote: Invalid username or password.
on a git push.

TimvdLippe:
It is sadly unrelated. It seems that @szczepiq password was reset for some reason :sob:

szczepiq:
There is something weird going on with GitHub auth tokens. They seem to be expiring or something. I've tested locally a new token, updated Travis configuration and restarted Travis Job. We'll see if it helps.

szczepiq:
It was fixed!

PR #1062, opened 2017-05-01
Bumped version of tools

szczepiq:
Bumped version of tools, addresses issue #1059


szczepiq:
Merging if that's ok. It's important to get the fix in.

PR #1069, opened 2017-05-04
Bumped release tools to 0.8.4 - concise release notes

szczepiq:
1. Bumped version of tools.
- The main improvement that is picked up is concise format of release notes: https://github.com/mockito/mockito-release-tools/pull/108
- Also picked up some internal refactorings of release tools. The previous version number will be automatically bumped / added to version.properties file.


szczepiq:
Since those changes don't affect the binaries, I will allow myself to merge it!

PR #1075, opened 2017-05-07
Streamline assertj and junit4 usage in gradle files and update assertj (v2.7.0)

epeee:
It looks like I missed some occurrences of assertj-core 1.7.1 in #1058.
Therefore I thought it might be a good idea to add a `dependencies.gradle` file which defines certain dependencies (for now assertj and junit4).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,6 +33,8 @@ apply from: 'gradle/mockito-core/javadoc.gradle'
 apply from: 'gradle/mockito-core/license.gradle'
 apply from: 'gradle/mockito-core/testing.gradle'
 
+apply from: "gradle/dependencies.gradle"
+
 archivesBaseName = "mockito-core"
 
 allprojects {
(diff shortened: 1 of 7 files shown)


szczepiq:
I like this idea! @epeee do you want to take a stab and configure other dependencies like that? It makes the code neatly readable and enforces consistent versioning. Nice!

FYI: The version of 'dependencies.mockito-release-tools' lives in 'gradle.properties' file. It probably cannot be placed in 'dependencies.gradle' because that dependency is used in 'buildscript' section. This means it is needed when the *.gradle file is compiled, before any other plugin or 'apply from' is evaluated. I mentioning it in case you want to do some experiments in putting versions in 'gradle.properties' file :)

Changes that help us maintain the project (consistent versions) and make the build logic easier to read are always welcome. Thank you!

epeee:
@szczepiq thx for your feedback. Yes, I will do so (configure other dependencies like that).

PR #1079, opened 2017-05-12
Fixed #1065. Add information about doNothing() method to CannotStubVo…

### LINKED ISSUE #1065: Complement error message
Exception from [cannotStubVoidMethodWithAReturnValue](https://github.com/mockito/mockito/blob/aeed5fe3b06912c35a4fd70fa61b45e1feedc9a1/src/main/java/org/mockito/internal/exceptions/Reporter.java#L447) does not inform about `doNothing()` method. But I think in many cases it is what user want to do.

SeriyBg:


src/main/java/org/mockito/internal/exceptions/Reporter.java
```diff
@@ -448,6 +448,9 @@ public static MockitoException cannotStubVoidMethodWithAReturnValue(String metho
                 "'" + methodName + "' is a *void method* and it *cannot* be stubbed with a *return value*!",
                 "Voids are usually stubbed with Throwables:",
                 "    doThrow(exception).when(mock).someVoidMethod();",
+                "If you need to set the void method to do nothing you can use:",
+                "    doNothing().when(mock).someVoidMethod();",
+                "More in javadocs for Mockito.doNothing().",
```
TimvdLippe: `For more information, check out the javadocs for Mockito.doNothing()`

SeriyBg: Thank you, done

PR #1093, opened 2017-05-27
Updated build for latest release tools

szczepiq:
- The concept of 'notable release' was removed from the tooling to keep things simple before releasing 1.0 of release tools (http://shipkit.org).
- Removed the use of 'notable release' from Mockito build.
- Started using 'shipkit.gradle' file to configure release automation.


szczepiq:
I will merge that if that's ok. The change does not update binaries.

TimvdLippe:
With the removal of notable releases, will every patch version now again be pushed to maven central? How do we control what gets pushed to central and what does not?

PR #1094, opened 2017-05-27
update assertj (v2.8.0)

epeee:
assertj v2.8.0 basically contains a bugfix and a small breaking change, see http://joel-costigliola.github.io/assertj/assertj-core-news.html#assertj-core-2.8.0 for details.


szczepiq:
Thanks!

PR #1096, opened 2017-05-27
InOrderImpl constructor signature

mureinik:
Replace `List<? extends Object>` in the constructor's argument list with a simpler `List<?>`.
The `extends Object` adds nothing, and just serves to confuse the reader.


mureinik:
@TimvdLippe SFTM?

TimvdLippe:
Seems fine to me :smile:

szczepiq:
Nice little improvement. More improvements like that, applied consistently over time, will make Mockito awesome :) Thank you!

PR #1098, opened 2017-05-28
Moved some more dependencies to dependencies.gradle

epeee:
As a follow up of #1075 I moved another bunch of dependencies to dependencies.gradle.


szczepiq:
Nice change! We had a problem in the past when we change changed bytebuddy version in one spot but forgot to do it in another. Combining library versions into a single location really helps to prevent such problems. Plus it shows nice pattern for other teams to adopt on how to manage library versions cleanly.

Thank you!

epeee:
You're welcome!

PR #1099, opened 2017-05-28
update gradle to v3.5

epeee:
While working on #1098 I realized that mockito is still using gradle 2.x. Since java 7 is used now for building and a local build using gradle 3.5 was successful I think we should give 3.5 a try.


szpak:
Btw, I cleared up the Travis cache for 2.x (which had ~1,2GB) to prune stuff related to old Gradle versions.

szpak:
Re-triggered builds with 3.5 took similarly long to the times before an upgrade.

szczepiq:
>Btw, I cleared up the Travis cache for 2.x (which had ~1,2GB) to prune stuff related to old Gradle versions.

That's great! How did you do it?

szczepiq:
>The execution time increased by around 30 seconds per build

Thank you for investigating it! You're right, we cannot stay on Gradle 2. Gradle 4 will be released in few weeks from now and we cannot keep using version that is over 2 years old.

szczepiq:
@epeee, nice change! I could swear I have even created a ticket in our tracker to get us to Gradle 3.5 :)

szpak:
@szczepiq I've used [Travis UI](UIhttps://docs.travis-ci.com/user/caching#clearing-caches).

Btw, I clear up also caches for other branches just for sanity.

image

szczepiq:
>@szczepiq I've used Travis UI.

Damn, I did not know that Travis has UI to purge caches!

PR #1104, opened 2017-05-31
Renamed configuration extension 'releasing' into 'shipkit'

NagRock:
Bumped version of `mockito-release-tools` that contains renamed gradle extension to (`shipkit`)
Fixes https://github.com/mockito/mockito-release-tools/issues/172


szczepiq:
Thank you! It's a great when Mockito dogfoods latest version of release tools / shipkit!

I tested the change locally before the push and it was all good.

PR #1115, opened 2017-06-08
update to shipkit v0.8.72

epeee:
let's update to the most recent shipkit version.


szczepiq:
Thank you for the change! It's great that Mockito uses latest shipkit!

The change affects the release and I like to monitor the release afterwards I will merge it tomorrow instead of today.

szczepiq:
I have monitored the release and it was successful :)

Issue #1117, opened 2017-06-08
Answer with delay in mock or spy to improve testing of asynchronous code


myrle-krantz:
I'm currently using Mockito for component testing of microservices.  All code should be run, up to the boundary at which another microservice is called.  In the component test, the mocks return very quickly, but obviously in integration testing, the real calls return much more slowly.  Some threading problems are revealed by the slow returns that would have been revealed in the component test if the returns could be forced to come back more slowly.  Since integration tests are *painfully* slow to start, this would save me, the developer, considerable time in debugging this kind of issue once I discover it in the integration tests.

Because it's so easy to write, I've already solved it with a class called AnswerWithDelay.  It takes another answer as a parameter, and inserts a little sleep before it returns.

If desired, I can turn this little helper into a PR for mockito. Or perhaps there's already something there that does this, that I've overlooked.

Best Regards,
Myrle

```
@RunWith(SpringRunner.class)
@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT,
        classes = {AbstractRhythmTest.TestConfiguration.class},
        properties = {"rhythm.user=homer", "rhythm.beatCheckRate=500"}
)
public class AbstractRhythmTest {
  //...
  static class AnswerWithDelay<T> implements Answer<T> {
    private final int sleepyTime;
    private final Answer<T> answer;

    AnswerWithDelay(final int sleepyTime, final Answer<T> answer) {
      this.sleepyTime = sleepyTime;
      this.answer = answer;
    }

    @Override
    public T answer(final InvocationOnMock invocation) throws Throwable {
      TimeUnit.MILLISECONDS.sleep(sleepyTime);
      return answer.answer(invocation);
    }
  }

  Beat createBeat(
          final String applicationIdentifier,
          final String beatIdentifier,
          final int alignmentHour,
          final LocalDateTime expectedBeatTimestamp) throws InterruptedException {
    final String tenantIdentifier = tenantDataStoreContext.getTenantName();

    final Beat beat = new Beat();
    beat.setIdentifier(beatIdentifier);
    beat.setAlignmentHour(alignmentHour);

    Mockito.doAnswer(new AnswerWithDelay<>(2_000, new Returns(Optional.of(PermittableGroupIds.forApplication(applicationIdentifier))))).when(beatPublisherServiceSpy).requestPermissionForBeats(Matchers.eq(tenantIdentifier), Matchers.eq(applicationIdentifier));
[...]

PR #1119, opened 2017-06-10
Made it possible to delay the return of a mock or spy.  (#1117)

myrle-krantz:
This is a useful aid in debugging race conditions and other synchronicity problems.

### DIFF
--- a/src/main/java/org/mockito/internal/stubbing/answers/AnswersWithDelay.java
+++ b/src/main/java/org/mockito/internal/stubbing/answers/AnswersWithDelay.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2017 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.internal.stubbing.answers;
+
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+import org.mockito.stubbing.ValidableAnswer;
+
+import java.io.Serializable;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Returns as the provided answer would return, after delaying the specified amount.
+ *
+ * <p>The <code>sleepyTime</code> specifies how long, in milliseconds, to pause before
+ * returning the provided <code>answer</code>.</p>
+ *
+ * @see org.mockito.AdditionalAnswers
+ */
+public class AnswersWithDelay implements Answer<Object>, ValidableAnswer, Serializable {
+    private static final long serialVersionUID = 2177950597971260246L;
+
+    private final long sleepyTime;
+    private final Answer<Object> answer;
+
+    public AnswersWithDelay(final long sleepyTime, final Answer<Object> answer) {
+        this.sleepyTime = sleepyTime;
+        this.answer = answer;
+    }
+
+    @Override
+    public Object answer(final InvocationOnMock invocation) throws Throwable {
+        TimeUnit.MILLISECONDS.sleep(sleepyTime);
+        return answer.answer(invocation);
+    }
+
+    @Override
+    public void validateFor(final InvocationOnMock invocation) {
+        if (answer instanceof ValidableAnswer) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/mockito/internal/stubbing/answers/AnswersWithDelay.java
```diff
@@ -0,0 +1,37 @@
...
+  private final Answer<Object> answer;
+
+  public AnswersWithDelay(final long sleepyTime, final Answer<Object> answer) {
+    this.sleepyTime = sleepyTime;
+    this.answer = answer;
+  }
+
+  @Override
+  public Object answer(final InvocationOnMock invocation) throws Throwable {
+    TimeUnit.MILLISECONDS.sleep(sleepyTime);
```
TimvdLippe: Should there also be an async version? Just a thought, not sure if we have to

myrle-krantz: Which release should I document this as @since ?

myrle-krantz: Also, by "added in the documentation" where do you mean? In CHANGES.md? or have I missed somewhere else?

myrle-krantz: I'm not sure what you mean by an "async version"; can you please elaborate?

Without futher details, I'm guessing the answer to the question is "no", since all state in AnswersWithDelay is final.

Though on further thought, it might be interesting to trigger the return externally via another mock method call, rather then make it wait a static time -- that'd be a feature for another time though...

src/test/java/org/mockitousage/stubbing/StubbingWithAdditionalAnswersTest.java
```diff
@@ -86,11 +100,7 @@ public void will_execute_a_void_based_on_strongly_typed_one_parameter_function()
         final IMethods target = mock(IMethods.class);
 
         given(iMethods.simpleMethod(anyString()))
-                .will(answerVoid(new VoidAnswer1<String>() {
-                    public void answer(String s) {
-                        target.simpleMethod(s);
-                    }
-                }));
+                .will(answerVoid((VoidAnswer1<String>) target::simpleMethod));
```
TimvdLippe: Please revert these changes. While we build locally on Java 8, we target Java 6 and I would like to make sure it keeps compiling on these older jdks. Our Travis build sadly does not test on jdk 6 any longer as it was deprecated and broken in their containers :sob:

myrle-krantz:
Re: revert. No problem. I was applying the boy scout method and killing off warnings, but I wasn't sure if that change would be acceptable. I could tell you were targeting older java versions, but wasn't sure if that's important in this test too.

I've provided it without the java 8 dependent changes.

TimvdLippe:
Great, this looks good to me. Signing off on this one, but want one other pair of :eyes: to check this PR :smile:

matan129:
IMHO, the PRs title is misleading - in the first place I thought it meant delay on `Mockito.mock()`/`Mockito.spy()`, which is obviously useless. 

Just after reading the code I understood it means answers with delay.
I'd change the title to "Add Delayed Answers".

myrle-krantz:
Can't right now because of day job. Are the release notes drawn from PR titles?

szpak:
I changed a title in #1117. It should be used to generate release notes (as that PR has no labels).

By the way we will test a new CD mechanism - notable version release. Thanks for your contribution @myrle-krantz!

szpak:
My bad, I didn't spot that PR was raised against master not 2.x...

szpak:
Btw, release notes don't look very good. I created #1123.

szczepiq:
Nice contribution. Thank you!

myrle-krantz:
You're welcome, it was my pleasure: Given how intensively mockito helps me every day, this was the least I could do.

Thank you!
Myrle

PR #1121, opened 2017-06-12
Improve and develop APIs required for framework integrators

szczepiq:
### Problem

Mockito does not do static mocking at the moment, Powermockito does. Powermockito uses a lot of internal Mockito APIs exposing the community to issues and wasted productivity on version upgrades. We want Mockito APIs to be easily extensible by framework integrators such us Powermockito without compromising API simplicity and consistency.

All proposed changes are backwards compatible in accordance to Mockito's compatibility contract (in short: limited changes to org.mockito.* types, freedom to change org.mockito.internal.*).

### Public API changes

The full discussion on the use cases, implementation options and decisions is in #1110 ticket.

1. New public method for creating mock creation settings object:

```java
Mockito.withSettings().name(“my mock”).build(Foo.class);
```

2. Framework integrators can now provide custom MockHandler via ```MockMaker#getMockHandler()```. Previously, it we didn’t offer this capability because we didn’t have convincing use cases that it was needed.

### Public API considerations

1. Framework integrators may **not** provide custom implementation of ```InvocationContainer``` at this stage. This part of API is accessible via public ```MockMaker#getMockHandler().getInvocationContainer()``` method. There are no convincing use cases that exposing this new public API is needed.


src/main/java/org/mockito/internal/stubbing/InvocationContainerImpl.java
```diff
@@ -128,8 +129,8 @@ public void clearInvocations() {
         registeredInvocations.clear();
     }
 
-    public List<StubbedInvocationMatcher> getStubbedInvocations() {
-        return stubbed;
+    public List<Stubbing> getStubbedInvocations() {
+        return (List) stubbed;
```
TimvdLippe: This seems unfortunate. Can we change the type of stubbed?

szczepiq: Nope :( Please suggest a different approach if you know it!

ChristianSchwarz: java generics are invariant, we need a wild card here: `public List<Stubbing> getStubbedInvocations() `

szczepiq: We could certainly do that, thank you for the pointer! Problem is that it makes the public API a bit more complicated because the invariant wildcard leaks to the public interface.

src/main/java/org/mockito/internal/util/MockUtil.java
```diff
@@ -43,26 +43,29 @@ public static TypeMockability typeMockabilityOf(Class<?> type) {
...
-            MockHandler handler = mockMaker.getHandler(mock);
-            return (InternalMockHandler) handler;
+            return mockMaker.getHandler(mock);
         } else {
             throw new NotAMockException("Argument should be a mock, but is: " + mock.getClass());
         }
     }
 
+    public static InvocationContainerImpl getInvocationContainer(Object mock) {
+        return (InvocationContainerImpl) getMockHandler(mock).getInvocationContainer();
```
TimvdLippe: Why cast here?

src/main/java/org/mockito/internal/verification/VerificationDataImpl.java
```diff
@@ -5,22 +5,22 @@
...
 import static org.mockito.internal.exceptions.Reporter.cannotVerifyToString;
 import static org.mockito.internal.util.ObjectMethodsGuru.isToStringMethod;
 
-import java.util.List;
-
 public class VerificationDataImpl implements VerificationData {
 
     private final InvocationMatcher wanted;
-    private final InvocationContainer invocations;
+    private final InvocationContainerImpl invocations;
```
TimvdLippe: I dont think this is necessary?

src/test/java/org/mockito/internal/handler/InvocationNotifierHandlerTest.java
```diff
@@ -121,11 +120,9 @@ public void should_report_listener_exception() throws Throwable {
     public void should_delegate_all_MockHandlerInterface_to_the_parameterized_MockHandler() throws Exception {
         notifier.getInvocationContainer();
         notifier.getMockSettings();
-        notifier.setAnswersForStubbing(new ArrayList<Answer<?>>());
 
         verify(mockHandler).getInvocationContainer();
         verify(mockHandler).getMockSettings();
-        verify(mockHandler).setAnswersForStubbing(ArgumentMatchers.<Answer<?>>anyList());
```
TimvdLippe: This is thus a breaking change? :(

szczepiq: Thank you for review! Absolutely let's avoid breaking changes!

Why do you consider this a breaking change? Those changes are all in "org.mockito.internal" components. Is it because VerificationData API already leaks some internal APIs? :)

src/main/java/org/mockito/internal/util/MockUtil.java
szczepiq: I think that InvocationContainer cannot be practically made public. It's too coupled with Mockito impl. details. I wrote details in the commit message and high level information in the Javadoc.

Do you have suggestions how to make it better?

szczepiq:
Thank you @TimvdLippe for reviewing! The PR is still in progress and we definitely must avoid breaking changes.

src/main/java/org/mockito/internal/util/MockUtil.java
TimvdLippe: I just was wondering why we couldn't use the interface, but if the impl is too coupled then there is no other option.

src/test/java/org/mockito/internal/handler/InvocationNotifierHandlerTest.java
szczepiq: >This is thus a breaking change? :(

"setAnswersForStubbing" method was internal so this is not a breaking change. I removed this line from the test because we no longer need this method :)

Thank you for feedback!

src/main/java/org/mockito/internal/verification/VerificationDataImpl.java
szczepiq: InvocationContainer is a marker interface at this stage to hide the internal implementation from leaking to the public API. It's impractical to expose it as public API. Hence, we use the 'impl' everywhere in the codebase. For more info, take a look at the javadoc for InvocationContainer interface.

Hope that helps!

szczepiq:
FYI: I'm not far from calling this complete. I don't expect much code changes, rather:

- verify Powermock needs all the new public API added
- add missing documentation / tests for public types
- add Since tag

Despite above, I think the code is ready for being reviewed. If you have feedback, please let me know!

szczepiq:
This is complete. @mockito/developers, friendly request for review :)

For high level summary of changes to the public API, see the new section added to our "user guide" in main Mockito class.

src/test/java/org/mockito/StaticMockingExperimentTest.java
```diff
@@ -0,0 +1,223 @@
...
+
+    @Test
+    public void verify_static_method() throws Throwable {
+        //register staticMethod call on mock
+        Invocation invocation = Mockito.framework().getInvocationFactory().createInvocation(mock, withSettings().build(Foo.class), staticMethod, realMethod,
+            "some arg");
+        handler.handle(invocation);
+
+        //verify staticMethod on mock
+        verify(mock);
```
TimvdLippe: I think this should call the staticMethod? The same for the next verify

szczepiq: Thanks for feedback! I have added code comment to this test method explaining what and why it is implemented this way. Is it clearer now?

szczepiq: Here's the comment: https://github.com/mockito/mockito/pull/1121/commits/5be550bba197d76697d69a84d9c46f92e99088dd

TimvdLippe:
Yes that is a lot clearer!

szczepiq:
Any feedback? I would love to merge it in a couple of days :)

TimvdLippe:
Seems like this branch has conflicts and has a failing build. Other than this PR is approved!

raphw:
A lot to read but it look really good to me. Props to the documentation and compatibility efforts!

koyemanaung:
support for my account in feature.

Issue #1064, opened 2017-05-02
Cannot Mock Android's View class with mockito-android 2.8.9


bparent:
When using mockito and mockito-android 2.8.9  I cannot mock the View class from Android.

This is not an issue using dexmaker and mockito 1.10. This seems to work with dexmaker 2.2.0 (which uses mockito 2.2.0) as well.

Repro:
1. Create a new project using Android Studio
2. Add mockito and mockito-android to app/build.gradle
    ``` gradle 
      testCompile 'org.mockito:mockito-core:2.8.9'
      androidTestCompile 'org.mockito:mockito-android:2.8.9'
    ```
3. Try to mock `View.class` in `ExampleInstrumentedTest` example test
    `View view = Mockito.mock(View.class);`
4. Run the test which will fail with the following stacktrace.

Stacktrace
```java
org.objenesis.ObjenesisException: java.lang.reflect.InvocationTargetException
at org.objenesis.instantiator.android.Android18Instantiator.newInstance(Android18Instantiator.java:50)
at org.objenesis.ObjenesisBase.newInstance(ObjenesisBase.java:73)
at org.mockito.internal.creation.instance.ObjenesisInstantiator.newInstance(ObjenesisInstantiator.java:18)
at org.mockito.internal.creation.bytebuddy.SubclassByteBuddyMockMaker.createMock(SubclassByteBuddyMockMaker.java:47)
at org.mockito.android.internal.creation.AndroidByteBuddyMockMaker.createMock(AndroidByteBuddyMockMaker.java:35)
at org.mockito.internal.util.MockUtil.createMock(MockUtil.java:35)
at org.mockito.internal.MockitoCore.mock(MockitoCore.java:63)
at org.mockito.Mockito.mock(Mockito.java:1729)
at org.mockito.Mockito.mock(Mockito.java:1642)
at examplecom.mockitoviewmockfail.ExampleInstrumentedTest.useAppContext(ExampleInstrumentedTest.java:23)
at java.lang.reflect.Method.invoke(Native Method)
at java.lang.reflect.Method.invoke(Method.java:372)
at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
[...]

PR #1128, opened 2017-06-14
Improve detection on non-mockable types - addresses #1064

raphw:
Stronger assertions on class loader origin when choosing what methods are mockable.

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -4,7 +4,7 @@ ext {
 
 def versions = [:]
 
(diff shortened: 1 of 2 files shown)


szczepiq:
I've [asked the reporter](https://github.com/mockito/mockito/issues/1064#issuecomment-308754350) to verify the PR. Once it's done we should merge. Given that it is really hard to write tests around this functionality, let's make sure it is verified to help with our scenario.

@raphw, thank you for fixing the problem! Mockito is a very happy consumer of latest Bytebuddy :)

PR #1132, opened 2017-06-17
make use of shipkit v0.8.92

epeee:
- speed up of `fetchAllContributors` task:
we did some performance improvements which results in a task duration of ~7s compared to ~22 sec in older versions of the plugin.


szczepiq:
Thank you! Merging! Mockito on latest Shipkit is a way to go.

PR #1133, opened 2017-06-18
Simplified and documented .travis.yml file

szczepiq:
I would love to get review of this, especially from @szpak! Please review commit-by-commit as I have commented separately on each removal from travis file.

Let's get the house clean!


szczepiq:
Thank you for review! Merging.

PR #1141, opened 2017-07-05
Bumped shipkit, simplified Travis file, removed unnecessary code

szczepiq:
- Now when the release fails Mockito CI job will fail, too. This is really useful to acknowledge problems early. Previously, the release was a part of 'after_success' Travis configuration and any failures of release build were ignored.
- Replaced Shipkit's java plugin with Gradle's vanilla 'java' plugin. This way the build is simpler.
- Removed Shipkit configuration that has been discontinued ("publishAllJavaSubprojects"). Instead, I disabled the bintrayUpload for modules we don't want to publish at this point.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,9 @@ install:
  - true
 
 script:
+  # We are using && below on purpose
+  # ciPerformRelease must run only when the entire build has completed
+  # This guarantees that no release steps are executed when the build or tests fail
   - ./gradlew build idea -s -PcheckJava6Compatibility && ./gradlew ciPerformRelease
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,11 +35,9 @@ install:
  - true
 
 script:
-  - ./gradlew build idea -s -PcheckJava6Compatibility
+  - ./gradlew build idea -s -PcheckJava6Compatibility && ./gradlew ciPerformRelease
 
 after_success:
-  #Releases new version:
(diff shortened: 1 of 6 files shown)


szczepiq:
Friendly ping :) Any feedback?

.travis.yml
```diff
@@ -35,11 +35,9 @@ install:
  - true
 
 script:
-  - ./gradlew build idea -s -PcheckJava6Compatibility
+  - ./gradlew build idea -s -PcheckJava6Compatibility && ./gradlew ciPerformRelease
```
TimvdLippe: Cant this just be `./gradlew build idea -s -PcheckJava6Compatibility ciPerformRelease` ?

TimvdLippe: Also the other commands like assert and prepare are removed, I guess that is intentional?

wwilk: @TimvdLippe ciPerformRelease runs assertReleaseNeeded and ciReleasePrepare.

szczepiq: >Cant this just be ./gradlew build idea -s -PcheckJava6Compatibility ciPerformRelease ?

It's not that safe because Gradle does various optimizations to make the build faster. So we can be still running tests while some other Gradle task will be shipping libraries to Bintray :)

TimvdLippe: All right, makes sense. Maybe add a comment such that future developers don't "optimize" this

szczepiq: Good idea!

szczepiq: I added code comment, if there is no other feedback, I'll merge it!

Issue #86, opened 2014-09-25
New lightweight, stub-only mocks for scenarios where high performance is needed


mockitoguy:
It is possible to create "stub-only" mocks that don't track invocations and hence might be useful for scenarios where low memory footprint / high speed is desired.

``` java
Foo lightweight = mock(Foo.class, withSettings().stubOnly());
```

For more information please see: https://code.google.com/p/mockito/issues/detail?id=84

PR #1146, opened 2017-07-07
Enable stubOnly() on @Mock annotation

hkdennis2k:
Since Issue #86,  withSettings() has a API stubOnly() which speed up mocked stub a lots.

In my case, stubOnly() huge amount of test overhead, stacktrace and avoid invocation footprint list grow too large and cause OOM.

However,  the API is not available with `@Mock` annotation.
I had to move tens of mock annotation from `@Mock Foo foo;` into `Mockito,mock(Foo.class, withSetting().stubOnly())`.

```
@Mock(stubOnly=true)
public Foo lightweight;

```

Suggest adding the above api for convenience 

check list

### DIFF
--- a/src/main/java/org/mockito/Mock.java
+++ b/src/main/java/org/mockito/Mock.java
@@ -31,11 +31,12 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  *       &#064;Mock(name = "database") private ArticleDatabase dbMock;
  *       &#064;Mock(answer = RETURNS_MOCKS) private UserProvider userProvider;
  *       &#064;Mock(extraInterfaces = {Queue.class, Observer.class}) private  articleMonitor;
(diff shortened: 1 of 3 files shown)


ChristianSchwarz:
@hkdennis2k 
Good idea! Can you add some unit tests to ensure the implementation works.

TimvdLippe:
Ping for a high level unit test, then we can land this change 😄

ChristianSchwarz:
@TimvdLippe 
@szczepiq 
The high level test is here [AnnotationsTest.java#L81](https://github.com/hkdennis2k/mockito/blob/2d3a5604525f73d8ff235fa0b032c8639fea8765/src/test/java/org/mockitousage/annotation/AnnotationsTest.java#L81)

TimvdLippe:
Ah I see, I wanted a separate new test, but I see that the current one already violates that principle. Based on that, LGTM.

PR #1157, opened 2017-08-03
Add tests to verify and show that @Spy can be used to allow stubbing/verification of List parameters using varargs.

fluentfuture:
Hi, I thought about sending an email first but then figured that I'd probably put in the test code as the example anyway, which is what this PR is all about.

I'm not completely sure that there isn't already a better way to go about stubbing List params. If there are, I'm happy to learn! :)

### DIFF
--- a/src/test/java/org/mockitousage/annotation/SpyAnnotationTest.java
+++ b/src/test/java/org/mockitousage/annotation/SpyAnnotationTest.java
@@ -40,6 +40,9 @@ public class SpyAnnotationTest extends TestBase {
     @Spy
     InnerStaticClassWithoutDefinedConstructor staticTypeWithoutDefinedConstructor;
 
+    @Spy
+    MockTranslator translator;
+
     @Rule
     public final ExpectedException shouldThrow = ExpectedException.none();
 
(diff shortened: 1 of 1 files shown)


fluentfuture:
While the test code should be self explanatory, here's a more sales-friendly example.

Assuming the test wants to stub the following interface:
```java
interface BatchSender {
  Result send(List<UnitOfWork> batch);
}
```

And if `UnitOfWork` has a lot of uninteresting data that the test doesn't care about, it's typical that one wants to use matchers for stubbing.

But matchers don't work for list elements by default. One might craft some kind of `doAnswer()` statement to do this but it will be verbose and not static type safe.

`@Spy` provides an alternative. All one needs is to create a simple delegating abstract class:

```java
class FooTest {

  @Spy private MockBatchSender batchSender;

  @Test public void testBatch() {
    when(batchSender.send(same(work1), eq(work2), isNotNull())
        .thenReturn(OK);
    // Now invoke SUT that calls batchSender...
  }

  abstract static class MockBatchSender {
    @Override public final Result send(List<UnitOfWork> batch) {
       return send(batch.toArray(new UnitOfWork[0]));
    }

    // Tests can stub/verify using this varargs friendly method.
    abstract Result send(UnitOfWork... batch);
  }
}
```

TimvdLippe:
I am all for increasing our test coverage, but I am not sure what this test tries to achieve. Is there an existing bug that is uncovered by this test? (even though the build and thus the test is passing). Are you missing something in our documentation and should we thus add it there as an example?

fluentfuture:
This test code serves as proof that the usage pattern documented in this
wiki
<https://github.com/mockito/mockito/wiki/Using-Spies-%28and-Fakes%29#matchers-for-list-elements>
works
as advertised, and as a guard against regression.

PR #1158, opened 2017-08-04
Adds support for Java 9 in inline mock maker by Byte Buddy update.

raphw:
Byte Buddy now uses ASM 6 which enables full support for Java 9 in the inline mock maker.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,65 +1,45 @@
 # More details on how to configure the Travis build
 # https://docs.travis-ci.com/user/customizing-the-build/
 
-# Speed up build by leversging travis caches
-before_cache:
-  - rm -f $HOME/.gradle/caches/modules-2/modules-2.lock
+# Speed up build by leveraging travis caches
 cache:
   directories:
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
 
-# required for oraclejdk9
-sudo: false
-dist: trusty
-group: beta
+# Enabling container based infrastructure hoping it will help the build speed <= this didn't work well, so it's reverted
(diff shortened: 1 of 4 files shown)


TimvdLippe:
Woohoo! Can we fix the tests that we disabled on Java 9?

raphw:
Of course, do we already have a Java 9 build? Where did we ignore tests?

raphw:
Of course, I enabled them now. Do we have a Java 9 CI build somewhere already?

TimvdLippe:
We run current master on 9: https://github.com/mockito/mockito/blob/master/.travis.yml

raphw:
This can be merged now, Byte Buddy 1.7.4 tests successfully on a JDK9 and by merging this, Mockito should be fully ready for the next release.

TimvdLippe:
Could you update the travis.yml to also build on Java 9? That way we make sure it does not regress.

raphw:
Sure, do we already know how?

TimvdLippe:
Yes. Copy from https://github.com/mockito/mockito/blob/master/.travis.yml

raphw:
There are some Gradle issues when using the JDK9 build.

First of all, the build seems to run out of memory even on JDK8: https://stackoverflow.com/questions/38967991/why-are-my-gradle-builds-dying-with-exit-code-137

Also, something does not seem to work out with Javadoc generation on JDK9.

That is a shame as the "hard part" which is running the tests and building the project does work.

I suggest we wait with the build until the tooling is up to date with JDK9. In the meantime, Mockito supporting Java 9 properly is a good enough property even if the CI is not yet up to date.

Issue #1155, opened 2017-08-03
thenThrow(Class) should only throw checked exceptions that match the signature of the stubbed method


tmurakami:
Hi,

- mockito-core 2.8.47

```java
public class FooTest {

    @Test(expected = MockitoException.class)
    public void using_thenThrow_Throwable() {
        C mock = mock(C.class);
        // Set a checked exception that does not matche the method signature
        when(mock.doIt()).thenThrow(new Exception());
    }

    @Test(expected = MockitoException.class)
    public void using_thenThrow_Class() {
        C mock = mock(C.class);
        // Set a checked exception that does not matche the method signature
        when(mock.doIt()).thenThrow(Exception.class);
    }

    static class C {
        String doIt() {
            return null;
        }
    }

}
```

`using_thenThrow_Throwable()` passes, but `using_thenThrow_Class()` fails.

```
java.lang.AssertionError: Expected exception: org.mockito.exceptions.base.MockitoException
```

The Javadoc of `OngoingStubbing#thenThrow(Class)` says, *If the throwable class is a checked exception then it has to match one of the checked exceptions of the stubbed method signature*.
However, the result of `using_thenThrow_Class()` shows that it is not so.

ChristianSchwarz:
I am going to fix this. 

ThrowsExceptionClass must simply implement ValidableAnswer...

szczepiq:
Thank you for reporting. Indeed it is a bug.

PR #1162, opened 2017-08-12
Fix for #1155 thenThrow(Class) 

ChristianSchwarz:
Fix for #1155 
Mockito.thenThrow(Class) now fails when the given Throwable type is not declared on the stubbed method.


ChristianSchwarz:
The build fails for JDK7 seem like an CI problem: Execution failed for task ':fetchAllContributors'.

szczepiq:
>The build fails for JDK7 seem like an CI problem: Execution failed for task ':fetchAllContributors'.

I will look into that, thank you.

ChristianSchwarz:
@TimvdLippe could you please remove the checkstyle rule [NewlineAtEndOfFile] it is frustration to fix and adds zero value

TimvdLippe:
If you have an editor with editorconfig support, it is no manual work. Both Eclipse and Intellij support it. Intellij even right out of the box. Else you can manually add the newline. I think you configured your editor to revert that change somewhere in your preferences.

Anyways, editorconfig makes sure we avoid unnecessary noise when contributing, even if you are new to the project and unaware that it exists. Its thus more friendly to new developers and I dont want to regress in that regard :(

ChristianSchwarz:
The IDE added a newline, also I tried to add manually a newline, but check style still fails. I have no further time to investigated into it.
----
This PR fixes now cleans up the unfinished stubbing state  in case null was passed to thenThrow(Class)/doThrow(Class). This leaded to unpredictable results in test suites before.

TimvdLippe:
You are fixing the wrong files. It is StubbingWithThrowables test: https://travis-ci.org/mockito/mockito/jobs/266766666#L303

ChristianSchwarz:
Strange when I run it locally it complains about the 3 files of the last commit...

src/test/java/org/mockitousage/stubbing/StubbingWithThrowablesTest.java
```diff
@@ -90,12 +95,9 @@ public void shouldAllowSettingCheckedException() throws Exception {
 
         when(reader.read()).thenThrow(ioException);
 
-        try {
-            reader.read();
-            fail();
-        } catch (Exception e) {
-            assertEquals(ioException, e);
-        }
+        exception.expect(sameInstance(ioException));
```
TimvdLippe: Missing the ewuals here

src/test/java/org/mockitousage/stubbing/StubbingWithThrowablesTest.java
```diff
@@ -39,48 +52,40 @@ public void shouldStubWithThrowable() throws Exception {
...
 
         doThrow(expected).when(mock).clear();
-        try {
-            mock.clear();
-            fail();
-        } catch (Exception e) {
-            assertEquals(expected, e);
-        }
+
+        exception.expect(IllegalArgumentException.class);
```
TimvdLippe: Missing equals here

src/test/java/org/mockitousage/stubbing/StubbingWithThrowablesTest.java
```diff
@@ -39,48 +52,40 @@ public void shouldStubWithThrowable() throws Exception {
         IllegalArgumentException expected = new IllegalArgumentException("thrown by mock");
         when(mock.add("throw")).thenThrow(expected);
 
-        try {
-            mock.add("throw");
-            fail();
-        } catch (IllegalArgumentException e) {
-            assertEquals(expected, e);
-        }
+        exception.expect(sameInstance(expected));
```
TimvdLippe: Missing equals here

ChristianSchwarz: Can you give me some more infos? We expect the instance to be thrown that was passed to the stubbing, right?

src/test/java/org/mockitousage/stubbing/StubbingWithThrowablesTest.java
ChristianSchwarz: Lets better use: `exception.expect(sameInstance(expected));`

src/test/java/org/mockitousage/stubbing/StubbingWithThrowablesTest.java
TimvdLippe: Sorry, this one is correct. I misread

TimvdLippe:
@ChristianSchwarz Nice, thanks! While could seems good, could you add a test (could just copy the literal test from #1155) to also check that case? I couldn't spot it in the added tests.

ChristianSchwarz:
Sure I can add more tests, but what you mean by "that case"?

TimvdLippe:
`when(mock.doIt()).thenThrow(Exception.class);` E.g. with only a literal class passed on, expecting this to fail.

ChristianSchwarz:
I am not sure what you mean by "only a literal class passed on, expecting this to fail".

Maybe a stubbing with an checked Exception type where an other checked exception is declared, like this:

```
class ExceptionA extends Exception

class ExceptionB extends Exception

public void doIt() throws ExceptionA

when(mock.doIt()).thenThrow(ExceptionB.class) 
//^^ fails like "can't throw checked exception >ExceptionB< cause it is not declared"
```

TimvdLippe:
The original report states that the following test did not fail. I would like to see that literal test as a regression test in our test body:

```java
@Test(expected = MockitoException.class)
    public void using_thenThrow_Class() {
        C mock = mock(C.class);
        // Set a checked exception that does not matche the method signature
        when(mock.doIt()).thenThrow(Exception.class);
    }
```

TimvdLippe:
@ChristianSchwarz Thanks for all your work!

Issue #1152, opened 2017-07-26
[Kotlin] Mocking fails when one suspend function calls other suspend function


rciurkot:
When trying to mock class that has `suspend` function that calls other `suspend` function, it fails with exception: 
```
Exception in thread "main" org.mockito.exceptions.base.MockitoException:
Mockito cannot mock this class: class com.myapp.SuspendableClass.
Can not mock final classes with the following settings :
- explicit serialization (e.g. withSettings().serializable())
- extra interfaces (e.g. withSettings().extraInterfaces(...))

You are seeing this disclaimer because Mockito is configured to create inlined mocks.
You can learn about inline mocks and their limitations under item #39 of the Mockito class javadoc.

Underlying exception : org.mockito.exceptions.base.MockitoException: Could not modify all classes [class java.lang.Object, class com.myapp.SuspendableClass]
at com.myapp.DupaKt$main$1.doResume(dupa.kt:25)
at kotlin.coroutines.experimental.jvm.internal.CoroutineImpl.resume(CoroutineImpl.kt:54)
at kotlinx.coroutines.experimental.DispatchTask.run(CoroutineDispatcher.kt:120)
at kotlinx.coroutines.experimental.EventLoopBase$QueuedRunnableTask.run(EventLoop.kt:189)
at kotlinx.coroutines.experimental.EventLoopBase.processNextEvent(EventLoop.kt:129)
at kotlinx.coroutines.experimental.BlockingCoroutine.joinBlocking(Builders.kt:225)
at kotlinx.coroutines.experimental.BuildersKt.runBlocking(Builders.kt:150)
at kotlinx.coroutines.experimental.BuildersKt.runBlocking$default(Builders.kt:142)
at com.myapp.DupaKt.main(dupa.kt:24)
Caused by: org.mockito.exceptions.base.MockitoException: Could not modify all classes [class java.lang.Object, class com.myapp.SuspendableClass]
at net.bytebuddy.TypeCache.findOrInsert(TypeCache.java:138)
at net.bytebuddy.TypeCache$WithInlineExpunction.findOrInsert(TypeCache.java:346)
at net.bytebuddy.TypeCache.findOrInsert(TypeCache.java:161)
at net.bytebuddy.TypeCache$WithInlineExpunction.findOrInsert(TypeCache.java:355)
... 9 more
Caused by: java.lang.IllegalStateException:
Byte Buddy could not instrument all classes within the mock's type hierarchy

[...]

PR #1165, opened 2017-08-15
Introduce test for mock-maker-inline with Kotlin suspend functions

elizarov:
See issue #1152
This test should pass when byte-buddy is updated with https://github.com/raphw/byte-buddy/pull/332

### DIFF
--- a/subprojects/kotlinTest/src/test/kotlin/org/mockito/kotlin/SuspendTest.kt
+++ b/subprojects/kotlinTest/src/test/kotlin/org/mockito/kotlin/SuspendTest.kt
@@ -21,6 +21,15 @@ class SuspendTest {
         suspend open fun suspendClassFunctionWithDefault(x: Int = 1): Int = error("not implemented")
     }
 
+    class FinalSuspendableClass {
+        suspend fun fetch(): Int {
+            fetchConcrete()
+            return 2
+        }
+
+        suspend fun fetchConcrete() = 1
+    }
+
     open class CallsSuspendable(private val suspendable: SuspendableInterface) {
         fun callSuspendable(x: Int) = runBlocking {
             suspendable.suspendFunction(x)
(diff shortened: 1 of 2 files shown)


elizarov:
I've update this PR with Kotlin 1.1.4-3 that fixes problem in #1152 and makes the test pass

raphw:
Awesome, thanks, we will merge this shortly. Improving Byte Buddy to handle this scenario better is on my list of things to do but this makes the issue less pressing. I think its good for Kotlon with regards to compatibility with tooling anyways!

TimvdLippe:
Travis seems unhappy on jdk7. Not sure what is going on, but need to verify it is working on 7 before merging.

raphw:
It is because the Kotlin plugin being used seems to require Java 8.

That is too bad, we probably need to add a rule to only build the test on Java 8+ VMs.

elizarov:
Yes. Kotlin compiler infrastructure had been recently moved to JDK8. The produced bytecode still targets JDK6 by default, but the compiler and all the tools around it require JDK8.

TimvdLippe:
> That is too bad, we probably need to add a rule to only build the test on Java 8+ VMs.

A shame, but better than nothing. @elizarov if you have the time, would you be able to include that in this PR to make Travis happy again? :smile:

szczepiq:
Kotlin 1.1.2 requires JDK 8 ([JetBrains blog post](https://blog.jetbrains.com/kotlin/2017/04/kotlin-1-1-2-is-out/)). Let's skip building/running Kotlin integration test in Java7 builds. Perhaps it is enough to conditionally include "kotlinTest" subproject in settings.gradle, based on the version of Java?

@szpak, you did good cool stuff with our Gradle automation, do you want to take a stab at it?

TimvdLippe:
We are no longer building on jdk7 per https://github.com/mockito/mockito/commit/da3f20eeb4db9fc78249d1b95a2548e68e9908a4 @elizarov could you rebase to retrigger Travis, then we can merge 🎉

elizarov:
Rebased

PR #1167, opened 2017-08-16
Fixes #1166: Fix to be able to stub the Kotlin function with any checked exception

### LINKED ISSUE #1166: [Kotlin] thenThrow(Throwable) should be able to stub the method with any checked exceptions
Hi,

Kotlin does not have checked exceptions.
https://kotlinlang.org/docs/reference/exceptions.html#checked-exceptions

However, `ThrowsException#validateFor(InvocationOnMock)` does not allow the checked exception that does not match the method signature.
Therefore, the following test fails.

```kotlin
class MyTest {

    @Test(expected = MyException::class)
    fun test() {
        val mock = Mockito.mock(I::class.java)
        Mockito.`when`(mock.doIt()).thenThrow(MyException()) // This line throws a MockitoException
        mock.doIt()
    }

    interface I {
        fun doIt(): String
    }

    class MyException : Exception()

}
```

I'm going to file the PR to fix this.

tmurakami:
If the stubbed method is of Kotlin, skip the signature check.

This PR fixes the issue #1166.

### DIFF
--- a/src/main/java/org/mockito/internal/stubbing/answers/ThrowsException.java
+++ b/src/main/java/org/mockito/internal/stubbing/answers/ThrowsException.java
@@ -6,6 +6,7 @@ package org.mockito.internal.stubbing.answers;
 
 import java.io.Serializable;
 import org.mockito.internal.exceptions.stacktrace.ConditionalStackTraceFilter;
+import org.mockito.internal.util.KotlinUtil;
 import org.mockito.internal.util.MockUtil;
 import org.mockito.invocation.InvocationOnMock;
 import org.mockito.stubbing.Answer;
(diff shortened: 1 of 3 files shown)


tmurakami:
This implementation is not sufficient.
For example, stubbing a **Java** method that does not have a checked exception in the **Kotlin** test will not work well.

```kotlin
class MyKotlinTest {

    @Test(expected = MyException::class)
    fun test() {
        // Mock a Java interface
        val mock = Mockito.mock(Runnable::class.java)
        // A MockitoException is thrown with the code below...
        // I think that the MyException should be thrown because Kotlin does not
        // have checked exceptions.
        Mockito.`when`(mock.run()).thenThrow(MyException())
        mock.run()
    }

    class MyException : Exception()

}
```

Also, stubbing a **Kotlin** function will fail in the following **Java** test:

```kotlin
interface MyKotlinInterface {
    fun doIt(): String
}
```

```java
public class MyJavaTest {

    @Test(expected = MockitoException.class)
    public void test() {
        // Mock a Kotlin interface
        MyKotlinInterface mock = Mockito.mock(MyKotlinInterface.class);
        // No MockitoException is thrown...
        // In this case, should not throw the MyException unless the Kotlin
        // function is marked with the `@kotlin.jvm.Throws` annotation.
        Mockito.when(mock.doIt()).thenThrow(new MyException());
        mock.doIt();
    }

    private static class MyException extends Exception {
    }

}
```

It is difficult to deal with these cases.
I will close this PR.
I apologize for bothering you.

PR #1169, opened 2017-08-18
Updated to latest Shipkit, tidied up configuration

szczepiq:
- Updated to pick up latest Shipkit. There is one particular bugfix to pick up (Shipkit's [release notes](https://github.com/mockito/shipkit/blob/master/docs/release-notes.md)).
- Removed unnecessary code from Gradle's build script files.

Tested by running "./gradlew testRelease". All good!


szczepiq:
Thanks for review!

PR #1170, opened 2017-08-20
Update objenesis version 2.5 -> 2.6 (Java9, bugfixes)

PascalSchumacher:
To quote from http://objenesis.org/notes.html:

Version 2.6 (2017-06-20)
- App Engine Java 8 Runtime support
- Improve Java 9 support
- Clarify serialization strategy behavior enhancement
- SunReflectionFactoryInstantiator not working if parent class constructor isn't public
- Validate Android O
- Drop Java 5 support

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -15,6 +15,6 @@ libraries.bytebuddy = "net.bytebuddy:byte-buddy:${versions.bytebuddy}"
 libraries.bytebuddyagent = "net.bytebuddy:byte-buddy-agent:${versions.bytebuddy}"
 libraries.bytebuddyandroid = "net.bytebuddy:byte-buddy-android:${versions.bytebuddy}"
 
-libraries.objenesis = 'org.objenesis:objenesis:2.5'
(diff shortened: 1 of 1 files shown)


szczepiq:
This looks like a very good change (Java9!). I'd like to ship never version to maven central as soon as this is merged and probably move the minor version. @mockito/developers, are you ok with this?

TimvdLippe:
Shall we wait to finish the PowerMock integration and then publish a new
version? We should be close there too, so we can combine it :)

szczepiq:
> We should be close there too, so we can combine it :)

We could do that.

The reason I thought about shipping separate version is because we don't know if Objenesis 2.6 is fully compatible to our customers. It would be nice to flesh out problems as soon as possible and minimize the changes included between versions to speed up any troubleshooting/bisecting.

szczepiq:
Powermock is close for sure! It'd say it still can be few more weeks to polish Mockito API, though :)

szczepiq:
@PascalSchumacher, thank you very much! Merging!

Issue #1179, opened 2017-08-28
Mockito 2.9.0 is significantly slower


jeffdcamp:
We just tried to upgrade to Mockito 2.9.0 (from 2.8.47) and found that our individual tests went from about 1 second to run per test to about 24 seconds to run per test.

We tried to find which release version causes the tests to run slow and found that version 2.8.50 is the first version that causes the slowdown (versions 2.8.48 and 2.8.49 run fine)

Our environment:
- Java 1.8.0_144
- Android Gradle Project (using Gradle 4.1 and Android Studio 3.0 Beta3)
- Kotlin 1.1.4-2
- JUnit 4.12

Other test dependencies
- Dagger 2.11
- threetenbp:1.3.6
- org.xerial:sqlite-jdbc:3.20.0

TimvdLippe:
@raphw Seems to be #1136 is the culprit. Any insight or immediate thoughts about it?

raphw:
@jeffdcamp Thanks for the description. Are you using the Android mock maker? Could you attach a profiler when running the tests to see what takes so much time?

raphw:
If you are using the final mock maker, I see how the resolution could be problematic using the method graph compiler. This is quite expensive and I did not think much about it when implementing the correct solution. We should definetly cache the result.

raphw:
@jeffdcamp Could you build Mockito from the branch `performance-fix` and try it out? I wonder if that fixes the regression. Thanks

jeffdcamp:
@raphw  We are not using "Android mock maker or the final mock maker".   Yes... 2.8.50 is the version that causes the slowdowns (we tried both  2.8.48 and 2.8.49, and those 2 versions work fine).

I'll try to perform a build of Mockito "performance-fix" branch

jeffdcamp:
@raphw I just built the "performance-fix" branch and changed the dependency on our project to "2.9.1" and that did not fix the slow down.

The following is using version 2.9.1 (Note the long run times):

![image](https://user-images.githubusercontent.com/1479095/29842041-d956de70-8cc4-11e7-9329-a7d1e88f9abc.png)

The following is using version 2.8.47:

![image](https://user-images.githubusercontent.com/1479095/29842193-58a65034-8cc5-11e7-8e13-9e023502340b.png)

raphw:
[...]

PR #1181, opened 2017-08-30
#1179: Fix performance regression caused by use of method graph compiler.

raphw:
In order to properly discover visibility bridges, we have to run a full analysis of the method graph. The compilation of this method graph is rather expensive and we require to cache a computed graph. The caching is however non-trivial as it references classes which might be unloaded. In order to avoid a leak, we cache classes using non-strong references.

This requires some real-world experience to discover if this caching mechanism suffices.

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodAdvice.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodAdvice.java
@@ -22,6 +22,7 @@ import java.io.ObjectInputStream;
 import java.io.Serializable;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
+import java.lang.ref.SoftReference;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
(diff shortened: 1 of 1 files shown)


src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodAdvice.java
```diff
@@ -119,10 +122,14 @@ public boolean isMocked(Object instance) {
...
-            .locate(new MethodDescription.ForLoadedMethod(origin).asSignatureToken());
-        return node.getSort().isResolved() && !node.getRepresentative().represents(origin);
+        SoftReference<MethodGraph> reference = graphs.get(instance.getClass());
+        MethodGraph methodGraph = reference == null ? null : reference.get();
+        if (methodGraph == null) {
+            methodGraph = compiler.compile(new TypeDescription.ForLoadedType(instance.getClass()));
+            graphs.put(instance.getClass(), new SoftReference<MethodGraph>(methodGraph));
+        }
+        MethodGraph.Node node = methodGraph.locate(new MethodDescription.ForLoadedMethod(origin).asSignatureToken());
+        return !node.getSort().isResolved() || !node.getRepresentative().represents(origin);
```
TimvdLippe: I think the condition is flipped now. Is that intentional?

raphw:
Yes, was incorrect before.

PR #1190, opened 2017-09-29
Improve interoperability with Java 8 by specializing parameterized types

ashleyfrieze:
It can be really hard to mock things like `Predicate<String>` easily as you cannot pass the specialization of type to the `mock` method.

This PR allows the user to write and they will avoid compiler warnings.

```java
Predicate<Foo> mockedPredicate = typed(mock(Predicate.class));
```

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -96,6 +96,7 @@ import org.mockito.verification.VerificationWithTimeout;
  *      <a href="#40">40. (*new*) Improved productivity and cleaner tests with "stricter" Mockito (Since 2.+)</a><br/>
  *      <a href="#41">41. (**new**) Advanced public API for framework integrations (Since 2.10.+)</a><br/>
  *      <a href="#42">42. (**new**) New API for integrations: listening on verification start events (Since 2.11.+)</a><br/>
+ *      <a href="#43">43. (**new**) Type casting Support (Since 2.11)</a><br/>
  * </b>
  *
  * <h3 id="0">0. <a class="meaningful_link" href="#mockito2" name="mockito2">Migrating to Mockito 2</a></h3>
@@ -1467,6 +1468,27 @@ import org.mockito.verification.VerificationWithTimeout;
  *     We found this method useful during the implementation.
  *     </li>
  * </ul>
+ *
+ * <h3 id="43">43. <a class="meaningful_link" href="#type_casting">Type casting Support</a> (Since 2.11)</h3>
+ *
+ * When mocking generic types you would get a compiler warning with code like this:
+ *
+ * <pre class="code"><code class="java">
+ *     Foo&lt;Bar&gt; mockFoo = mock(Foo.class);
+ * </pre>
+ *
+ * This is because there is no way to specify the generic type within the input to the <code>mock</code>
+ * method.
+ * <p>
+ *
+ * To reduce compiler warnings and to make Mockito more interoperable with inferred types in Java 8
+ * you can wrap your calls to <code>mock</code> with the function <code>typed</code>, which will apply
+ * the parameterized type.
+ *
+ * <pre class="code"><code class="java">
+ *     Foo&lt;Bar&gt; mockFoo = typed(mock(Foo.class));
+ * </pre>
+ *
  */
 @SuppressWarnings("unchecked")
 public class Mockito extends ArgumentMatchers {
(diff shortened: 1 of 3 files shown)


TimvdLippe:
I am not sure if we would want to add another dependency for this. What limits a current user of adding this `typed` function themselves when required? I see no usage of `typed` in the code changes of this PR.

ashleyfrieze:
@TimvdLippe well dang! Git stole some of my changes... I'll put them back in a short while. Watch this space.

ashleyfrieze:
Hey @TimvdLippe thanks for your patience. The PR is now in the shape I intended. Please offer feedback as you see fit.

src/main/java/org/mockito/Mockito.java
```diff
@@ -1772,6 +1794,31 @@ public static MockingDetails mockingDetails(Object toInspect) {
...
+     * get compiler warnings.
+     * <p>
+     *
+     * <pre class="code"><code class="java">
+     *   Foo&lt;Bar&gt; mock = typed(mock(Foo.class));
+     * </code></pre>
+     *
+     * <p>
+     * <em>Warning:</em> this method could be accidentally used to downcast to
+     * a subclass of the mock, resulting in a <code>ClasCastException</code>
```
TimvdLippe: Is there a test that shows this behavior? Or maybe include a direct example pointing out when this is an issue.

ashleyfrieze: I can add a test for it...

I will.

Essentially the constraint within the `<M, T extends M>` in the `typed` method means that the types must be related with the cast being to a more specific type than the mock's raw type from the class. However, if you tried to force a cast of `mock(List.class)` to `ArrayList`, it wouldn't work.

Give me a moment, and I'll add a test.

ashleyfrieze: There it is.

src/main/java/org/mockito/Mockito.java
```diff
@@ -1772,6 +1794,38 @@ public static MockingDetails mockingDetails(Object toInspect) {
...
+     *     // the mock is NOT an array list, so this will throw an exception
+     *     ArrayList&lt;String&gt; list = typed(mock(List.class));
+     * </code></pre>
+     *
+     * @param mock the return of a call to one of the <code>mock</code> methods
+     * @param <T> the return type (inferred by Java 8)
+     * @param <M> the actual type of the mock
+     * @return a downcasted version of the mock
+     */
+     public static <M, T extends M> T typed(M mock) {
```
ChristianSchwarz: I wonder if the same can be applied directly on `Mockito.mock(..)`?

ChristianSchwarz: Note: The type variable M can be replaced by Object. The current signature is equal to `<T> T typed(Object mock)`.
At the end of the day this PR safes a compile time warning in exchange of compiler error in case nonsens is passed.

ChristianSchwarz:
@TimvdLippe 

> I like that the solution is a single method and that we have more type-safety.

I am afraid we lose safety:

```
List<String> list = typed(mock(Runnable.class)); 
// ^^ hooray no compile time warning, we get an an CCE at runtime now
```

src/main/java/org/mockito/Mockito.java
ashleyfrieze: I couldn't make the above work directly on the `mock` method - I suspect it's an issue with Java 7's inferred type system, but some of the tests stopped compiling.

TimvdLippe:
Yes I played with this locally and it seems it will work for every single combination of types you insert. E.g. you can also call this with `Float` and `Predicate`. This means that we give the impression with `typed` that the compiler is happy with it, but in fact it is not.

As such, I think it will be hard to make this change land. Yes it works for the case where you pass a class and its corresponding generic subtype, but for all other cases it does not.

Let's see what the opinion of the others is.

ashleyfrieze:
@ChristianSchwarz - I just tried your counter example. You're right, that sucks!

It seems like the Java parameter constraints are finding a lowest-common type here... this is not what I expected it to do. Even the following code doesn't help:

```java
    public static <M, T extends M> T typedMock(Class<M> classToMock) {
        return (T)mock(classToMock, withSettings());
    }

...

List<String> list = typedMock(Runnable.class);
```

Very frustrating... it's as though the Java generics don't really care about types at all...

ashleyfrieze:
The issue comes down to type erasure. https://docs.oracle.com/javase/tutorial/java/generics/genTypes.html As @ChristianSchwarz said, the `M` in my code is just substituted with `Object` because Java doesn't need to care about it...

Unless someone says otherwise, I think I should drop this PR altogether. It would be more harm than good.

_Considers going back to C# with its stronger type system_ :)

TimvdLippe:
All right, I am closing this PR for now. Thanks for the investigation nonetheless! Now when we have similar feature requests, we can point to this discussion and explain why we are not going to adopt such a feature for the time being :smile:

Issue #1191, opened 2017-10-01
Ensure spring-boot can use Mockito public API


mockitoguy:
### Problem

Mockito framework integrators need robust public APIs from us. This way, they ship great Mockito integrations to their customers. We want to avoid the situation where framework integrators have to use Mockito internal APIs. This leads to overhead, relying on stale versions of Mockito, class conflicts in runtime.

spring-boot currently uses some internal API from Mockito. See https://github.com/spring-projects/spring-boot/issues/10352 for the problem statement, solution options and chosen design. 

### Solution

As discussed in spring-boot ticket, here's the API we would like to prototype:

- New "VerificationStartedListener"
- The instance of listener can be passed on mock creation
- Listener is notified when 'Mockito.verify' is invoked but before the verification is triggered
- It is possible to replace the mock object that is being verified (key feature needed by spring-boot)

### Example usage

```Java
mock(Foo.class, withSettings().verificationStartedListeners(new VerificationStartedListener() {
  public void onVerifyStarted(VerificationStartedEvent event) {
    event.getMock();
    event.setMock(arbitrayObject);
  }
}
```

szczepiq:
I put together a prototype in "mockito-spring-boot" branch. Opened PR #1192 for visibility. How to use it?

- clone mockito and check out "mockito-spring-boot" branch
- ./gradlew publishToMavenLocal
- grab newly published version of mockito from maven local and try it out!
- I don't have Javadocs yet. This test class documents the new API (hopefully :) -> https://github.com/mockito/mockito/pull/1192/commits/1d2cd207621c13251ef89b4d7e6ca9178f8115e8?diff=unified#diff-c7b40da00e36958d90e247ac5f759871R23

Suggested next steps:

- @wilkinsona, let me know if this is enough for you to try it out!
- we can use the PR that I have opened for review comments regarding the API

wilkinsona:
@szczepiq This looks great. I've built the `mockito-spring-boot` branch and prototyped the updates required for Spring Boot to use it: https://github.com/wilkinsona/spring-boot/tree/gh-10352. All our tests pass and, as you can hopefully see, Boot's code is now significantly simpler. Thanks very much.

szczepiq:
Fantastic! I will polish the code / documentation and get the release out soon. Thank you for testing this out and suppling sample project!

szczepiq:
@wilkinsona, can you try out one more time for us please?

[...]

PR #1192, opened 2017-10-01
Added experimental API for spring-boot use case

mockitoguy:
Not intended for merging yet, submitted for visibility.

- for design and rationale, see #1191
- for usage, see "VerificationStartedListenerTest" class

This is still experimental so refactorings, error handling, javadoc is missing.


src/main/java/org/mockito/mock/MockCreationSettings.java
```diff
@@ -94,4 +97,7 @@
      */
     @Incubating
     Object getOuterClassInstance();
+
+
+
```
TimvdLippe: nit: remove whitespace

szczepiq:
Hey @TimvdLippe, thank you for review! I will work on this today.

szczepiq:
>What happens if multiple verificationstartedlisteners set a different mock? Is it last-come wins? That could potentially break any framework that would be relying on the fact that the mock it set is the mock that is executed on.

The mock object is created only once so there is no practical possibility for 2 different frameworks to add different listeners.

We already have in our API similar listeners that are configured at mock creation. We did not encounter a problem with this API. The new "VerificationStartedListener" suggested in this PR follows an existing pattern that was already discussed in the past.

Hence, I think it's ok to allow multiple listeners, to keep the API consistent and simple.

Thank you for insightful feedback! Keep it coming because I'm finishing this feature off :)

TimvdLippe:
Sorry for being vague. I am okay with adding multiple listeners, but my comment was regarding calling `setMock`  multiple times. If franeworks A and B both call `setMock` and rely on its value, one of them will break (depending on which one is invoked latest). Therefore, I would like to block calling `setMock` twice. If another framework calls it, it should be an exception stating that mocks were overridden and there is an incompatibility between two dependencies. WDYT?

szczepiq:
>If franeworks A and B both call setMock and rely on its value, one of them will break (depending on which one is invoked latest)

How can 2 frameworks call setMock? Only 1 framework can create mock and pass a listener. Can you describe the scenario more clearly?

Happy to work out this scenario to design great API!

szczepiq:
FYI: I'm almost done with the code. I would love some feedback!

TimvdLippe:
I have some time tomorrow to write down the example

TimvdLippe:
Due to issues with Gradle from #1202 I am unable to compile or do any work. Have to wait till that is resolved until I can write the example :cry:

TimvdLippe:
Okay, so while trying to write down the example, I realized it is a bit more tricky. It would essentially require to add multiple listeners on the same `MockitoSettings` object:

```java
List mock2 = mock(List.class, Mockito.withSettings().verificationStartedListeners(new VerificationStartedListener() {
            public void onVerificationStarted(VerificationStartedEvent event) {
                //this is a hack to simulate desired behavior
                container.add(event.getMock());
                event.setMock(mock1);
            }
        }, new VerificationStartedListener() {
            @Override
            public void onVerificationStarted(VerificationStartedEvent event) {
                event.setMock(mock2)
            }
        }));
```

In this case, `mock2` will be the final mock. If the first listener was added based on framework 1 and the second one by a different one, there is a discrepancy between the mock that is actually in the event and the mock the framework expects. If framework 1 is relying on the mock being mock1 while at the same time framework 2 relies on the mock being mock2, one of them will break.

I would thus like to propose that if `setMock` is invoked twice, an exception is thrown. This is very explicit to the user indicating: hey you are using two framework listeners together, this can't work.

szczepiq:
Hey @TimvdLippe, I still don't understand the use case.

1. Only one framework can construct a mock object. If multiple frameworks integrate with Mockito, only one of those frameworks can construct given mock. The other framework can use the mock of course but it cannot create it (because it is already created). If we need to, let's discuss this further on slack.
2. I'd like the API to be consistent with other listeners, like invocation listeners. Using an exception makes verification started listeners work differently than invocation listeners because it makes the 1st listener more important than other listeners. Consistent API is less surprising to users and makes it easy to use, predictable.
3. Adding an exception adds some complexity that we need to maintain.
4. There is no use case to support 2 listeners, just like there is no use case to support only 1 listener. Supporting 2 listeners does not seem incorrect. If it is not incorrect, I'd rather keep things simple. Allowing 2 listeners is more open for experimentation and unexpected use cases. Using an exception is creating an arbitrary constraint without a use case.
5. If the use case presents itself, for example we have a user bug report, I'm happy to add code to address it. Perhaps then we add some validation.

FYI. When designing this, I had considered supporting just single listener but decided against it. That's why I have added "[listeners_are_executed_in_sequence](https://github.com/mockito/mockito/pull/1192/files#diff-c7b40da00e36958d90e247ac5f759871R72)" test.

Good discussion! Thank you for taking the time to describe what you meant. Do you have feedback about other parts of the API or the implementation? In the heat of the argument, let's not forget about the surrounding code :)

szczepiq:
The implementation is done. Would love some feedback!!!

szczepiq:
@szpak, can you review? There's a discussion about supporting one or many listeners but I don't think it should be blocking the PR.

Let's ship this so that Spring Boot can start using Mockito's public API!

TimvdLippe:
Its not about one or many listeners. It should be possible to add multiple
listeners. However, I want to limit `setMock` to be invoked only once.

mockitoguy:
Thank you! I'm pushing new version soon.

PR #1193, opened 2017-10-04
mockito#1151 - A reproducer where issue occurs on every invocation.

techpavan:
This is a PR on the issue branch to provide a reproducer which can demonstrate the issue with more clarity.

check list


techpavan:
The failure in the build is justified because this is a test case which fails with the current implementation of the library. Library needs to be fixed to make this test case pass.

mockitoguy:
Let's avoid merging this. See my comment: https://github.com/mockito/mockito/pull/1200#issuecomment-336717814

PR #1194, opened 2017-10-04
Issue 1151 fix for multi threaded environment

techpavan:
This provides a fix for the issue mockito#1151, but definitely requires a review as it might have unforeseen side effects. Code change is just one line to address the multiple threaded usage of mocks, but might require the static member to be accessed in a synchronized manner to avoid issues that might occur in a multi threaded environment. This works because most of the calls to the library invokes a call to validateState() method which internally uses stubbingInProgress member.

Note: The class SampleTest is a part of issue-1151 branch and unsure if this is required for the actual code branch.

check list

### DIFF
--- a/src/main/java/org/mockito/internal/progress/MockingProgressImpl.java
+++ b/src/main/java/org/mockito/internal/progress/MockingProgressImpl.java
@@ -31,7 +31,7 @@ public class MockingProgressImpl implements MockingProgress {
 
     private OngoingStubbing<?> ongoingStubbing;
     private Localized<VerificationMode> verificationMode;
-    private Location stubbingInProgress = null;
+    private static Location stubbingInProgress = null;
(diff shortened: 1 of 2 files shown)


src/main/java/org/mockito/internal/progress/MockingProgressImpl.java
```diff
@@ -31,7 +31,7 @@
 
     private OngoingStubbing<?> ongoingStubbing;
     private Localized<VerificationMode> verificationMode;
-    private Location stubbingInProgress = null;
+    private static Location stubbingInProgress = null;
```
ChristianSchwarz: This won't work in case subbings are made in multiple threads. e.g. when tests are executed parallel

techpavan: True, will check in detail when I find some time.

PR #1195, opened 2017-10-04
Fixes #1189: Add stable module name

### LINKED ISSUE #1189: JDK9: set a stable module name with `Automatic-Module-Name` entry in MANIFEST
JDK9 is out and introduces a module system.

Mockito does not currently use modules. When we use it in a JDK9 modular application, Java turns the JAR into a so-called automatic module, whose name is derived from the JAR archive name (*mockito*, in this case). However, the default name does not follow the recommended module naming conventions (reverse-dns style, module name derived from the main exported package).

It is possible to specify a stable automatic module name through a `Automatic-Module-Name` manifest entry, while still targeting JDK8:

```
Automatic-Module-Name: org.mockito
```

Selecting a stable module name is very important, because Java does not allow two modules to own the same package, and you may end up with a module hell. The naming issues are well-explained on Stephen Colebourne's blog:

 * http://blog.joda.org/2017/04/java-se-9-jpms-module-naming.html
 * http://blog.joda.org/2017/05/java-se-9-jpms-automatic-modules.html <- the explanation of module hell problem

And the comment of Mike Reinhold, chief Java architect:

> Strongly recommend that all modules be named according to the reverse Internet domain-name convention. A module's name should correspond to the name of its principal exported API package, which should also follow that convention. If a module does not have such a package, or if for legacy reasons it must have a name that does not correspond to one of its exported packages, then its name should at least start with the reversed form of an Internet domain with which the author is associated.

mureinik:
This patch adds a stable module name to the produced JAR in order to
be forwards compatible to Java 9.

### DIFF
--- a/gradle/mockito-core/osgi.gradle
+++ b/gradle/mockito-core/osgi.gradle
@@ -27,6 +27,8 @@ afterEvaluate {
                     'org.mockito.*'
 
             instruction '-removeheaders', 'Private-Package'
+
+            attributes 'Automatic-Module-Name': 'org.mockito'
         }
     }
 }


zyxist:
I started making a similar pull request, but it turns out that I don't have to :).

To make a dry test, just extract the JAR and make sure that the necessary entry is present in `MANIFEST.mf`: 

```
Manifest-Version: 1.0
Automatic-Module-Name: org.mockito
...
```

If the entry is present, I can confirm that Java 9 sees Mockito as `org.mockito` module.

zyxist:
So it will work. I'm working on a Gradle plugin for building Jigsaw modules, and Mockito is one of my use cases, so I can provide you the necessary feedback.

TimvdLippe:
All right, I trust you ;) Even if this would go wrong, its probably going to break no one (?).

mureinik:
And in the worst case - we can always revert it :-)


Thanks @zyxist @TimvdLippe

PR #1203, opened 2017-10-08
Fixed documentation issue, added unit tests

szczepiq:
- While working on new verification started listeners (#1191) I found an instance of incorrect documentation. I added unit tests and simplified the documentation.
- Currently, contrary to what the Javadoc said, we do guarantee the order in which invocation listeners get notified. Some of our users might have started to depend on this behavior. I added unit tests and fixed the Javadoc. In general, it is seems useful to guarantee the order. This way, the API is more predictable, easier to use.

### DIFF
--- a/src/test/java/org/mockitousage/debugging/InvocationListenerCallbackTest.java
+++ b/src/test/java/org/mockitousage/debugging/InvocationListenerCallbackTest.java
@@ -70,10 +70,11 @@ public class InvocationListenerCallbackTest {
         Foo foo = mock(Foo.class, withSettings().invocationListeners(listener1, listener1));
 
         // when
-        foo.giveMeSomeString("herb");
+        foo.giveMeSomeString("a");
+        foo.giveMeSomeString("b");
 
-        // then
-        assertThat(container).containsExactly(listener1, listener1);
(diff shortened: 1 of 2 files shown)


src/test/java/org/mockitousage/debugging/InvocationListenerCallbackTest.java
```diff
@@ -39,6 +46,37 @@ public void should_call_single_listener_when_mock_return_normally() throws Excep
...
+        List<InvocationListener> container = new ArrayList<InvocationListener>();
+        RememberingListener listener1 = new RememberingListener(container);
+
+        Foo foo = mock(Foo.class, withSettings().invocationListeners(listener1, listener1));
+
+        // when
+        foo.giveMeSomeString("herb");
+
+        // then
+        assertThat(container).containsExactly(listener1, listener1);
```
szpak: @szczepiq What do you thing about a test to document behavior how many times the listener is called if provided multiple times in `invocationListeners`?

szczepiq: I can add it. Good idea to document and preserve the current behavior of the public API.

szczepiq:
Thank you for review!

PR #1213, opened 2017-10-19
Standardize JUnit imports

mureinik:
This PR standardizes importing assertion methods, making the code base cleaner and easier to maintian.

### DIFF
--- a/src/test/java/org/mockito/internal/util/TimerTest.java
+++ b/src/test/java/org/mockito/internal/util/TimerTest.java
@@ -5,12 +5,15 @@
 package org.mockito.internal.util;
 
 import org.assertj.core.api.Assertions;
-import org.junit.Assert;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 import org.mockito.exceptions.misusing.FriendlyReminderException;
 import org.mockitoutil.TestBase;
 
 public class TimerTest extends TestBase {
+    @Rule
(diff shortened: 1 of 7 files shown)


src/test/java/org/mockito/internal/util/TimerTest.java
```diff
@@ -38,14 +37,9 @@ public void should_return_false_when_time_run_out() throws Exception {
         Assertions.assertThat(timer.isCounting()).isFalse();
     }
 
-    @Test
+    @Test(expected = FriendlyReminderException.class)
     public void should_throw_friendly_reminder_exception_when_duration_is_negative() {
-        try {
-            new Timer(-1);
-            Assert.fail("It is forbidden to create timer with negative value of timer's duration.");
```
ChristianSchwarz: We should not remove the check for the expected error message, ExpectedException-Rule to the rescue.

mureinik: @ChristianSchwarz nothing is removed here - `fail`'s argument is the error message that's printed once the statement is reached, it doesn't validate anything.

Regardless - using `ExpectedException` to verify the error message would make this a better test.
I'll revise the patch to include it.

mureinik:
@ChristianSchwarz as seen in the updated patch, the string in `fail` isn't the actual message in the thrown exception, but using the `ExpectedException` definitely makes the test more robust (and IMHO, easier to maintain). Thanks for the review!

Issue #1163, opened 2017-08-13
argThat fails with JUnit5


asarkar:
Mockito 2.8.47, JUnit 5.0.0-RC2, Maven 3.5.0, JDK 1.8.0_66-b17. Project [here](https://github.com/asarkar/java/tree/master/deploy-maven-plugin).

```
class DeployMojoTest {
    private DeployMojo deployMojo = new DeployMojo();
    private ArtifactDeployer artifactDeployer;

    static Stream<Arguments> gavProvider() {
        return Stream.of(
                Arguments.of("name-and-version.jar", "java.util", "name-and-version", "1.0.0")
        );
    }

    @BeforeEach
    void beforeEach() throws URISyntaxException, MalformedURLException {
        artifactDeployer = mock(ArtifactDeployer.class);
        deployMojo.artifactDeployer = artifactDeployer;
        deployMojo.lib = new File(getClass().getResource("/").toURI());

        Map<String, URL> repositories = new HashMap<>();
        repositories.put("test", new URL("file://url"));
        deployMojo.repositories = repositories;
    }

    @ParameterizedTest(name = "parse {0}")
    @MethodSource("gavProvider")
    void testExecute(String fileName, String groupId, String artifactId, String version)
            throws MojoFailureException, MojoExecutionException {
        deployMojo.execute();

        verify(artifactDeployer).deployArtifact(
                argThat(file -> file.getName().equals(fileName)),
                argThat(artifact -> artifact.getGroupId().equals(groupId)
                        && artifact.getArtifactId().equals(artifactId)
                        && artifact.getVersion().equals(version)),
                argThat(repoId -> repoId.equals("test")),
                argThat(repoUrl -> repoUrl.equals("file://url"))
        );
    }
}
```

Failure:

```
Argument(s) are different! Wanted:
artifactDeployer.deployArtifact(
    <Deploy mojo test$$ lambda$ 2 2 3/ 2 1 0 7 4 4 3 2 2 4>,
    <Deploy mojo test$$ lambda$ 2 2 4/ 1 0 4 0 7 7 6 9 9 6>,
    <Deploy mojo test$$ lambda$ 2 2 5/ 1 2 9 3 6 8 0 7 3 4>,
    <Deploy mojo test$$ lambda$ 2 2 6/ 1 3 8 8 1 7 3 2 9>
);
-> at org.abhijitsarkar.DeployMojoTest.testExecute(DeployMojoTest.java:52)
Actual invocation has different arguments:
[...]

PR #1221, opened 2017-10-24
MockitoExtension for JUnit5 

### LINKED ISSUE #445: Introduce MockitoExtension for JUnit Jupiter (a.k.a. JUnit 5)
## New Features in JUnit 5

The JUnit Jupiter extension model in JUnit 5 introduces support for constructor and method parameter resolution (i.e., dependency injection).

Specifically, third parties can implement the `ParameterResolver` extension API to inject dependencies, mocks, etc. into constructors and methods. In addition, the `TestInstancePostProcessor` extension API can be implemented to post-process a test instance (e.g., to perform field injection).

## Status Quo

Mockito supports _field injection_ for mocks via the `@Mock` annotation. In addition, #438 allows `@Mock` to be declared on parameters for constructors and methods which makes `@Mock` support an ideal candidate for both the `TestInstancePostProcessor` and `ParameterResolver` extension APIs in JUnit Jupiter. In fact, the JUnit Team has already developed a proof of concept: see the `MockitoExtension` in the [junit5-mockito-extension](https://github.com/junit-team/junit5-samples/tree/master/junit5-mockito-extension) sample project.

## Deliverables

- [ ] Introduce an official `MockitoExtension` for JUnit Jupiter to replace the proof of concept from the JUnit team.

ChristianSchwarz:
## ! WORK IN PROGRESS, DON'T MERGE !
Fixes #445

Discussion: https://groups.google.com/forum/#!topic/mockito/R1L1aG2RgBQ

First of all, thanks to the JUnit5-Team (@sbrannen) for that great API! The implementation was straight forward and the docu is top notch. Not comparable to a JUnitRule or Runner implementation. One can feel the hard work. 

----

The here provided `MockitoExtension` allows to mock/spy/validate like the well known Mockito-JUnitRule/Runner. This is still "work in progress", feel free to comment.

Usage:

```Java
class JUnit5Test {
 
  @Mock
  private Dependency mock ;

  @Test
  void checkMockCreation(){
    when(mock.foo()).thenReturn("Hello JUnit5");
    mock.foo();
    verify(mock).foo();
  }
}
```


ChristianSchwarz:
Possibly we need an annotation on class or method level to provide control of the "strictness". E.g.:

```
@MockitoStrictness(WARN)
class JUnit5Test {

    @MockitoStrictness(STRICT_STUBS)
    void testWithLocalDefinedStrictness(){
    }
}
```

subprojects/junit5/junit5.gradle
```diff
@@ -0,0 +1,14 @@
+description = "Mockito preconfigured inline mock maker (intermediate and to be superseeded by automatic usage in a future version)"
```
TimvdLippe: Just so we don't forget: we need to update this description

subprojects/junit5/junit5.gradle
```diff
@@ -0,0 +1,14 @@
...
+
+apply from: "$rootDir/gradle/java-library.gradle"
+
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
```
TimvdLippe: Needs to be 1.6

subprojects/junit5/src/main/java/org/mockito/junit5/MockitoExtension.java
```diff
@@ -0,0 +1,49 @@
...
+     * Callback that is invoked <em>before</em> each test is invoked.
+     *
+     * @param context the current extension context; never {@code null}
+     */
+    @Override
+    public void beforeEach(ExtensionContext context)  {
+        Optional<?> testInstance=context.getTestInstance();
+        if (!testInstance.isPresent()){
+            //for some reason junit don't give us the test instance
+            //there is nothing we can do in this case
```
TimvdLippe: This is weird. Can we construct a test for this?

subprojects/junit5/src/main/java/org/mockito/junit5/MockitoExtension.java
```diff
@@ -0,0 +1,49 @@
...
+            return;
+
+        }
+
+        session = mockitoSession()
+            .initMocks(testInstance.get())
+            .strictness(Strictness.WARN)
```
TimvdLippe: This should be an option then I guess

subprojects/junit5/junit5.gradle
ChristianSchwarz: I am afraid that don't work cause JUnit5 ExtensionContext.getTestInstance() returns Optional that was introduced in Java 8. IntelliJ refuses shows an error in this case.

subprojects/junit5/src/main/java/org/mockito/junit5/MockitoExtension.java
ChristianSchwarz: sure..

subprojects/junit5/junit5.gradle
ChristianSchwarz: Right, see https://github.com/mockito/mockito/pull/1221#issuecomment-339141751 for a proposal.

subprojects/junit5/src/main/java/org/mockito/junit5/MockitoExtension.java
ChristianSchwarz: Right, see https://github.com/mockito/mockito/pull/1221#issuecomment-339141751 for a proposal.

subprojects/junit5/junit5.gradle
```diff
@@ -0,0 +1,14 @@
+description = "Mockito JUnit 5 support"
+
+apply from: "$rootDir/gradle/java-library.gradle"
+
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
+
+dependencies {
+    compile project.rootProject
+    implementation libraries.junit5
```
epeee: Let's reduce/change this dependency to `org.junit.jupiter:junit-jupiter-api:5.0.1` since we should be fine with the api for implementing the extension.
Additionally let's add `testRuntime libraries.junit5`.

ChristianSchwarz: Thanks for these hints! 
Fixed!

ChristianSchwarz:
> Could we subclass the extension for the silent and nonsilent variant? Thats in line with the JUnit4 variant :)

You mean 3 extensions like `MockitoWarnExtension`, `MockitoLenientExtension`, `MockitoStrictStubExtension` ?

ChristianSchwarz:
Current state:

The Test class and nested test classes and methods can define the stricness by declaring `@Strictness`.
Note: The strictness level is inherited and can be overriden per context.


```
//default strictness WARN
class StricnessTest {

    @Mock
    Predicate mock;

    @Test
    void strictnessFromTestRoot_defaultIsWarn_shouldThrowUnnecessaryStubbingException() {
        when(mock.test(1)).thenReturn(true);
    }

    @Test
    @Strictness(LENIENT)
    void strictnessFromTestMethod_lenient_shouldIgnoreUnnecessaryStubbing() {
        when(mock.test(1)).thenReturn(true);
    }

    @Nested
    @Strictness(WARN)
    class NestedTest{

        @Mock
        Predicate mock;

        @Test
        void strictnessFromNestedTest_warn_shouldLogUnnecessaryStubbing() {
            when(mock.test(1)).thenReturn(true);
        }
    }
}

```

subprojects/junit5/src/main/java/org/mockito/junit5/MockitoExtension.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Optional;
+
+import static java.util.Optional.empty;
+import static java.util.Optional.ofNullable;
+import static org.mockito.Mockito.mockitoSession;
+import static org.mockito.quality.Strictness.WARN;
+
+public class MockitoExtension implements BeforeEachCallback, AfterEachCallback {
+
+    private MockitoSession session;
```
tmurakami: JUnit does not seem to guarantee how to handle Extension instances, so I think that it is better to use `ExtensionContext#getStore(Namespace)` rather than holding the MockitoSession instance in the field.

TimvdLippe:
@ChristianSchwarz Correct, it would have the 3 different versions again.

Also, Travis is unhappy because we are targeting Java 8 with this PR. Since we state Mockito 2 supports Java 6+, I am not sure if we can integrate into Mockito 2. Instead, we maybe should include it in master and publish a beta version of mockito 3?

ChristianSchwarz:
> Instead, we maybe should include it in master and publish a beta version of mockito 3?

That means stable JUnit5 support will only be available when mockito 3 final is released. I guess this will not happen before Q4 2018. 

An other option is to release junit5 support as seperate mockito 2 addon artefact which requires java 8, like junit 5 itself. It can be released much earlier. What do you think?

ChristianSchwarz:
>>You mean 3 extensions like MockitoWarnExtension, MockitoLenientExtension, MockitoStrictStubExtension ?

> Correct, it would have the 3 different versions again.

What is the benefit for the user here? What do you think of the annotation proposal above? It allows to define strictness at method and nested test level.

subprojects/junit5/junit5.gradle
epeee: hmm, the implementation of the extension should just depend on the api.
Therefore we could change it to 
```
implementation 'org.junit.jupiter:junit-jupiter-api:5.0.1'
testRuntime 'org.junit.vintage:junit-vintage-engine:5.0.1'
```
isn't it?

TimvdLippe:
> An other option is to release junit5 support as seperate mockito 2 addon artefact which requires java 8, like junit 5 itself. It can be released much earlier. What do you think?

If we can find a way to make that pass on the build, I am all for that :+1: :+1: 

> What is the benefit for the user here? What do you think of the annotation proposal above? It allows to define strictness at method and nested test level.

Yes definitely keep them, they are very useful to specify exceptions for a single method/nested class. However, to specify the default strictness on a whole test class, I would like to have that incorporated in the `@ExtendWith` annotation.

tmurakami:
JUnit 5 supports meta-annotations, so we could provide the following annotation:

```java
@Retention(RetentionPolicy.RUNTIME)
@Target({ElementType.ANNOTATION_TYPE, ElementType.TYPE, ElementType.METHOD})
public @interface WithMockito {
    Strictness value() default Strictness.STRICT_STUBS;
}
```

subprojects/junit5/src/main/java/org/mockito/junit5/MockitoExtension.java
ChristianSchwarz: fixed

ChristianSchwarz:
I am not sure if we should do that.  We would have two annotation concepts to define the strictness, that is maybe confusing and inconsequent and even conflicting.

Both annotations  `@Strictness` and `@WithMockito` are declared with target `ElementType.TYPE` to be applicable to classes. In JUnit5 we have 2 kind of test classes the root-test and nested-test. `@WithMockito`  will be ignored on nested classes, thats fine.   `@Strictness` can be placed on both, but this can lead to conflicts, like this:

```java
@WithMockito(WARN)
@Strictness(STRICT_STUBS)
class Test{
   ...
}
```

----
Anyway the meta-annotation `@WithMockito` is a good idea to avoid boilerplate code, but I would like to leave out the strictness.

subprojects/junit5/src/main/java/org/mockito/junit5/MockitoExtension.java
```diff
@@ -0,0 +1,96 @@
...
+
+        return strictness;
+    }
+
+    private static void store(ExtensionContext context, MockitoSession session) {
+        context.getStore(MOCKITO).put("session", session);
+    }
+
+    private static MockitoSession mockitoSession(ExtensionContext context) {
+        return (MockitoSession) context.getStore(MOCKITO).get(SESSION);
```
TimvdLippe: Can be `get(SESSION, MockitoSession.class)` per http://junit.org/junit5/docs/current/api/org/junit/jupiter/api/extension/ExtensionContext.Store.html#get-java.lang.Object-java.lang.Class-

TimvdLippe:
@ChristianSchwarz Could you also add a test for #1163 and verify we can fix that issue?

subprojects/junit5/src/main/java/org/mockito/junit5/MockitoExtension.java
tmurakami: Since a MockitoSession instance acquired here is finished on `afterEach()`, I think that you should call `Store#remove(Object, Class)` instead of `get()`.

subprojects/junit5/src/main/java/org/mockito/junit5/MockitoExtension.java
```diff
@@ -0,0 +1,96 @@
...
+            .strictness(strictness)
+            .startMocking();
+
+        store(context, session);
+    }
+
+
+    private static Strictness retrieveStrictness(ExtensionContext context) {
+        Optional<AnnotatedElement> annotatedElement = context.getElement();
+        if (!annotatedElement.isPresent()) {
```
tmurakami: I think that it is better to use `AnnotationSupport#findAnnotation(AnnotatedElement, Class)` which supports meta-annotations.

```java
context.getElement()
       .flatMap(e -> AnnotationSupport.findAnnotation(e, Strictness.class))
       .map(Strictness::value)
       .orElse(DEFAULT_STRICTNESS);
```

tmurakami:
@ChristianSchwarz Sorry I didn't explain it enough.
I thought that it would be better to provide an annotation (like @WithMockito) that was annotated with `@ExtendWith(MockitoExtension.class)` instead of providing `@Strictness`.
The reason is to reduce boilerplate code as you said.

> This violates the SRP , and may lead to problems down the road. What do you think?

I was lacking consideration.
As you say, my proposal may violate the SPR.

If MockitoExtension supports meta-annotation, I suppose that Mockito users would be able to create an annotation like `@WithMockito` as needed.
So, I think we don't have to provide `@WithMockito`.

tmurakami:
BTW, JUnit's MockitoExtension supports method/constructor injection.
http://junit.org/junit5/docs/current/user-guide/#writing-tests-dependency-injection

What do you think about supporting it?

subprojects/junit5/src/main/java/org/mockito/junit5/MockitoExtension.java
ChristianSchwarz: You are right, makes perfect sense!

ChristianSchwarz: see: https://github.com/mockito/mockito/pull/1221#discussion_r147455484

subprojects/junit5/src/main/java/org/mockito/junit5/MockitoExtension.java
ChristianSchwarz: True!

ChristianSchwarz:
> What do you think about supporting it?

It was discussed in #445 and definetly worth a look. For this PR I would like to provide only a MVP, more features should be added in the next steps.

ChristianSchwarz:
> So, I think we don't have to provide @WithMockito.

I like it and keep it for now but without stricness attribute.

ChristianSchwarz:
@TimvdLippe @tmurakami 
Do you guys have an idea how to get the build green? I must admit I'm a gradle noob :fearful:

TimvdLippe:
I think we should disable the animalSniffer in this specific subproject.
Can we disable the task in some way? If not, we should not import the full
java-library.gradle but only the logic we need. Then we dont import
animalSniffer straight away.

subprojects/junit5/junit5.gradle
```diff
@@ -0,0 +1,14 @@
+description = "Mockito JUnit 5 support"
+
+apply from: "$rootDir/gradle/java-library.gradle"
+
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
+
+dependencies {
+    compile project.rootProject
+    testRuntime libraries.assertj
```
tmurakami: I think you should use `testCompile` or `testImplementation` instead of `testRuntime`.
Otherwise, AssertJ classes would not be read at build time and you would not be able to compile test classes.

tmurakami:
@ChristianSchwarz Thank you. I'll read #445.

TimvdLippe:
@ChristianSchwarz I pushed fixes to this PR which should make the build pass now

ChristianSchwarz:
@TimvdLippe big thanks for fixing the build!

src/main/java/org/mockito/internal/framework/DefaultMockitoSession.java
```diff
@@ -43,7 +49,7 @@ public Throwable getFailure() {
                 return null;
             }
             public Object getTestClassInstance() {
-                return testClassInstance;
+                return testClassInstances.get(0);
```
TimvdLippe: This is unfortunate. We probably have to fix #1232 first to improve our internal API and then base this PR on that change.

src/main/java/org/mockito/internal/session/DefaultMockitoSessionBuilder.java
```diff
@@ -10,28 +10,48 @@
...
+        this.testInstances = testClassInstance;
+        return this;
+    }
 
+    @Deprecated
     @Override
     public MockitoSessionBuilder initMocks(Object testClassInstance) {
-        this.testClassInstance = testClassInstance;
+        if (testClassInstance == null) {
+            testClassInstance = new Object();
```
TimvdLippe: What is the purpose of this line?

src/main/java/org/mockito/internal/session/DefaultMockitoSessionBuilder.java
```diff
@@ -10,28 +10,48 @@
...
+        }
+
+        testInstances = singletonList(testClassInstance);
+
         return this;
     }
 
     @Override
     public MockitoSessionBuilder strictness(Strictness strictness) {
+        if (strictness==null){
```
TimvdLippe: Spaces around the `==`

subprojects/junit5/src/main/java/org/mockito/internal/junit5/AnnotationUtil.java
```diff
@@ -0,0 +1,43 @@
...
+import static org.mockito.quality.Strictness.WARN;
+
+public class AnnotationUtil {
+    /**
+     * FIXME we should use a more common place in mockito (MockitoCore?) to define the default strictness,
+     * it should  be uses by Runner & Rule too
+     */
+    private final static org.mockito.quality.Strictness DEFAULT_STRICTNESS = WARN;
+
+    public static org.mockito.quality.Strictness retrieveStrictness(ExtensionContext context) {
```
TimvdLippe: If I understand correctly, this method is recursively looking up the annotation. I think we should approach this issue differently:

1. Nested classes can specify a new runner which overrides the parent runner.
2. The `@ExtendWith` specifies the strictness as well and this strictness will be used for all tests. E.g. `@ExtendWith(MockitoExtension.Silent.class)` or `@ExtendWith(MockitoExtension.Strict.class)`. `@ExtendWith(MockitoExtension.class)` is equivalent to the strict variant.
3. The user can opt-in to a different strictness per `method`.

The algorithm to retrieve the strictness is then:

1. If Method has annotation, use that strictness
2. Else use strictness specified by runner

As every runner already defines it strictness level, we don't need to fallback to the `DEFAULT_STRICTNESS`. Moreover, we can remove this field as a whole, as we are now use runners instead that have it baked in.

subprojects/junit5/src/main/java/org/mockito/junit5/MockitoExtension.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    private void collectParentTestInstances(ExtensionContext context, List<Object> testInstances) {
+        Optional<ExtensionContext> parent = context.getParent();
+        while (parent.isPresent() && parent.get() != context.getRoot()){
+            ExtensionContext parentContext = parent.get();
+
+            Object testInstance = parentContext.getStore(MOCKITO).remove(TEST_INSTANCE);
+
+            if(testInstance!=null){
```
TimvdLippe: Spacing around `!=` and `if`

subprojects/junit5/src/main/java/org/mockito/junit5/MockitoExtension.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * Callback that is invoked <em>after</em> each test has been invoked.
+     *
+     * @param context the current extension context; never {@code null}
+     */
+    @Override
+    public void afterEach(ExtensionContext context) {
+        MockitoSession session;
```
TimvdLippe: This method can be a one-liner: `removeMockitoSession(context).finishMocking();`

subprojects/junit5/src/main/java/org/mockito/junit5/WithMockito.java
```diff
@@ -0,0 +1,28 @@
...
+ * Place this annotation on your root test class in order to initialize fields annotated with {@link Mock @Mock},
+ * {@link Spy @Spy} or {@link Captor @Captor}. To specify the stubbing strictness you can declare the
+ * annotation {@link Strictness @Stricness}
+ * <p>
+ * This is the Junit5 equivalent to {@link MockitoRule} and {@link MockitoJUnitRunner} for JUnit4.
+ */
+@Retention(RUNTIME)
+@Target({ANNOTATION_TYPE, TYPE})
+@ExtendWith(MockitoExtension.class)
+public @interface WithMockito {}
```
TimvdLippe: We have to bike-shed on this name. `WithMockito` is vague, but don't have a concrete proposal for rename yet

TimvdLippe:
@ChristianSchwarz Do you have some time in the near future to address the above feedback? Thanks!

ChristianSchwarz:
@TimvdLippe 
>Do you have some time in the near future to address the above feedback? Thanks!

I look into it next week.

subprojects/junit5/src/main/java/org/mockito/internal/junit5/AnnotationUtil.java
ChristianSchwarz: @TimvdLippe 

Defining a strictness-level is from my working experience a rarely used feature, the most teams tend to move fast forward and use the default strictness. Therefore I tend to provide an API that is as simple as possible to cover 80% of the uses cases. That means a plain annotation like `@WithMockito` instead of @ExtendWith(MockitoExtension.Silent.class) to avoid boilerplate code. Readability is a also to consider here. 
When users want to define the strictness-level one concept of defining it should be enough, different concepts may be a bit uncomfortable and harder to remember.

Lets take the discussion to the google group -> https://groups.google.com/forum/#!topic/mockito/R1L1aG2RgBQ

ChristianSchwarz:
@tmurakami In case you want to joint the API discussion, here it is -> https://groups.google.com/forum/#!topic/mockito/R1L1aG2RgBQ

tmurakami:
@ChristianSchwarz, thank you, I will check later.

TimvdLippe:
I have updated this PR with the refactoring. Really happy that the impact on our public API is minimal. The only sad part is the `MockitoSession` with `testClassInstances`, but the impact is limited. I have also updated the specification of the extension to be in line with other junit5 extension. This severely simplified the implementation of searching for strictness, while still allowing users to override per method.

wreulicke:
code coverage in this PR's diff is lower than 88.08%?
I'm waiting to merge this PR.

TimvdLippe:
This PR is work in progress. We will improve the code coverage later.

mockitoguy:
Hey @mockito/developers! Can you review #1272? It's a critical change because it's new public API. We are always very judicious about public API changes. It also simplifies this PR / JUnit5 integration because we don't need "@Strictness" annotation.

For this PR, how about we ship #1232 first?

marcphilipp:
Now that #1232 has been shipped, I think we should discuss how to continue with this PR.

@ChristianSchwarz Do you have time to continue with this effort? I'd be happy to help.

ChristianSchwarz:
Hi @marcphilipp !

> Do you have time to continue with this effort? I'd be happy to help.

This week I can invest some hours in this PR. You help is very welcome, thanks for your request! 
 * Do you know how to  set up gradle so the "junit" / mockito-extension module is build a independent artefact?

@mockitoguy @TimvdLippe Why you want to distribute the mockito-junit-extension as independent artefact and not as part of mockito-core? From a user perspective it is convenient to have only one dependency.

TimvdLippe:
Because mockito-core uses Java 6 and JUnit 5 and its extensions require
Java 8. Therefore we can not incorporate in core and have to publish a
separate artifact.

ChristianSchwarz:
@TimvdLippe @mockitoguy 
What is the current state of the strictness discussion? If you made no decisson yet, I would like leave the strictness toppic out of this first PR. It can be added later and use `Stricness.WARN` as default.

TimvdLippe:
@ChristianSchwarz I think our final decision was to remove `@Strictness` from test method level, but keep it in the extension. That leaves Strictness on runner, mock and soon invocation level. For test method level we will need concrete use-cases. (see https://github.com/mockito/mockito/pull/1272#issuecomment-361075606)

ChristianSchwarz:
@TimvdLippe 
Okay, what you mean by?

> remove @Strictness from test method level, but keep it in the extension.

Keep the `@Strictness` to be used on top-level-test-class and nested-test-classes? 

```
@Strictness(WARN)
public class NestedTests {
    @Nested
    @Strictness(SILENT)
    class NestedTest {
          ...
    }
}
```

TimvdLippe:
@ChristianSchwarz On test method level. E.g.

```java
@Strictness(WARN) // remove this one
public class NestedTests {
    @Nested
    @Strictness(SILENT) // remove this one
    class NestedTest {
          ...
         @Strictness(LENIENT) // remove this one
         public void test() {
         }
    }
}
```

ChristianSchwarz:
@TimvdLippe Should @ExtendWith(MockitoExtension.class) be supported on nested-test-classes?

```java
@ExtendWith(MockitoExtension.Lenient.class)
public class NestedTests {
    @Nested
    @ExtendWith(MockitoExtension.Warn.class)
    class NestedTest {
          ...
    }
}
```

TimvdLippe:
@ChristianSchwarz Yes please. In this case, all invocations in `NestedTest` should have Strictness `Warn`, while all others have `Lenient`.

marcphilipp:
I think using subclasses like `MockitoExtension.Warn` and `MockitoExtension.Lenient` is not a good idea because in that case both extensions would get registered and interfere with each other.

TimvdLippe:
@marcphilipp Hm good point. We wanted to keep our API in-line with how we deal with Strictness in JUnit4. Do we have any other options to use a similar interface but removing the duplication issue?

ChristianSchwarz:
Confirmed it doesn't work, `@Stricness`-Annotation to the rescue.

TimvdLippe:
That is really unfortunate :cry: Will keep on thinking for alternatives, but for now we have to resort to `@Strictness`

marcphilipp:
You could handle the duplication issue in the extensions by using lazy initialization and the same namespace to store the `MockitoSession` in `ExtensionContext`.

mockitoguy:
Thank you guys for picking this up!!!

Let's discuss the Strictness annotation (and potentially other settings of the extension). ```@Strictness``` annotation looks nice in the test code but it has 2 caveats:
 - it appears as general purpose API but it is only a setting for JUnit5 extension. This may be confusing for users: annotation used without MockitoExtension, user thinks it works but it has no effect.
 - it is not easy to discover

What do you think about following options:
 - ```@MockitoExtension.Settings(strictness=Strictness.LENIENT)```
 - ```@MockitoExtension.Settings(strictness=Strictness.LENIENT, logger=MyLogger.class)``` (example future setting)
 - ```@MockitoExtension.Strictness(Strictness.LENIENT)```

If we need to discuss the annotation more, we can always ship the initial version version of JUnit5 support without it. This way, we can unblock a feature for our users, while still improve the API incrementally.
 
It would be good to make the extension use STRICT_STUBS by default. It is consistent with MockitoSession and it is inline with our path to stricter stubs by default in future version of Mockito.
	
I agree with @TimvdLippe that we're not ready for strictness per method. Let's see some solid use cases first. To implement it cleanly, we should support it for all kinds of runners, not only JUnit5. Let's stay focused on JUnit5 extension and revisit strictness per method later. Also, when we merge #1272 we may not need strictness per method at all...

What's the best name of the new module: mockito-junit5, mockito-jupiter? What is the typical naming convention here?

ChristianSchwarz:
`@MockitoExtension.Settings(strictness=Strictness.LENIENT)` looks modern and more flexible than the JUnit4 look alike `@ExtendWith(MockitoExtension.Lenient.class)`. But I would like to shorten it to `@MockitoExtension(strictness=Strictness.LENIENT)`.

But I wonder how it behave with nested test classes?

ChristianSchwarz:
@TimvdLippe 
@mockitoguy 

How should the API finally look like? What should be/ shoud not be included in first version?

TimvdLippe:
@ChristianSchwarz The first version needs to:

1. Create mocks with `@Mock` and `@Spy` and such
1. Properly handle `MockitoSession` even with nested classes
1. Have strict stub support

ChristianSchwarz:
How should the extension look? 
 * `@ExtendWith(MockitoExtension.class)` vs  `@MockitoExtension`

So we leave the configurable stricness for a next step, right?

TimvdLippe:
`@ExtendWith(MockitoExtension.class)` such that users downstream can potentially introduce `@MockitoExtension` if they want to. That lets us stick to the mainstream JUnit5 notation with freedom for our users.

> So we leave the configurable stricness for a next step, right?

Yes please

mockitoguy:
+1 to @TimvdLippe's comments

@ChristianSchwarz, regarding your earlier suggestion with ```@MockitoExtension(strictness=Strictness.LENIENT)```

Can MockitoExtension be at the same time a) annotation for configuring our extension and b) a class referenced in Jupiter's ```@ExtendWith```?

```
@MockitoExtension(strictness=Strictness.LENIENT)
class JUnit5Test {}
```

sbrannen:
> Can MockitoExtension be at the same time a) annotation for configuring our extension and b) a class referenced in Jupiter's `@ExtendWith`?

No, that is not possible.

An annotation in Java is not allowed to implement any interface (other than `Annotation`). Any attempt to do so will result in a compiler warning similar to the following.

> Annotation type declaration cannot have explicit superinterfaces

marcphilipp:
You could meta-annotate the annotation with `@ExtendWith`, though.

For example

```java
@ExtendWithMockito(strictness=Strictness.LENIENT)
class JUnit5Test {}
```
```java
@Retention(RUNTIME)
public @interface ExtendWithMockito {
    Strictness strictness default Strictness.STRICT;
}
```

ChristianSchwarz:
@TimvdLippe 
@mockitoguy 

I would like to follow the proposal of @marcphilipp but leaving the strictness for now! The stricness can be added later.
 The annotation `@ExtendWithMockito`  reveal clearly its intention it is even better than `@MockitoExtension` .

TimvdLippe:
@ChristianSchwarz Yes that seems fine. Let's try to finish this PR and ship, as it has been lingering for quite a while now :smile:

TimvdLippe:
I am going to polish up this PR this weekend to hopefully get this close to merging. It has been waiting for quite a while now :smile:

TimvdLippe:
All right, I have polished up the implementation now (with help from @marcphilipp, much appreciated!).

The implementation now includes a `@ConfiguredWithMockito` which is an annotation that describes how Mockito should run. This includes the Strictness for now, but is extensible to future configuration we would like.

There are a couple of action points left:

- @mockitoguy could you review this PR in terms of documentation/implementation?
- `./gradlew build` does NOT include the junit5 subproject tests. I have no clue why yet :cry: 
- Running the tests directly works in IntelliJ, but if I run the module/package instead, I get the following stacktrace:
```
Mar 17, 2018 3:33:58 PM org.junit.platform.launcher.core.DefaultLauncher handleThrowable
WARNING: TestEngine with ID 'junit-jupiter' failed to discover tests
java.lang.ArrayStoreException: sun.reflect.annotation.TypeNotPresentExceptionProxy
	at sun.reflect.annotation.AnnotationParser.parseClassArray(AnnotationParser.java:724)
	at sun.reflect.annotation.AnnotationParser.parseArray(AnnotationParser.java:531)
	at sun.reflect.annotation.AnnotationParser.parseMemberValue(AnnotationParser.java:355)
	at sun.reflect.annotation.AnnotationParser.parseAnnotation2(AnnotationParser.java:286)
	at sun.reflect.annotation.AnnotationParser.parseAnnotations2(AnnotationParser.java:120)
	at sun.reflect.annotation.AnnotationParser.parseAnnotations(AnnotationParser.java:72)
	at java.lang.Class.createAnnotationData(Class.java:3521)
	at java.lang.Class.annotationData(Class.java:3510)
	at java.lang.Class.getDeclaredAnnotation(Class.java:3458)
	at org.junit.platform.commons.util.AnnotationUtils.findAnnotation(AnnotationUtils.java:135)
	at org.junit.platform.commons.util.AnnotationUtils.findAnnotation(AnnotationUtils.java:114)
	at org.junit.jupiter.engine.descriptor.JupiterTestDescriptor.determineDisplayName(JupiterTestDescriptor.java:86)
	at org.junit.jupiter.engine.descriptor.ClassTestDescriptor.<init>(ClassTestDescriptor.java:83)
	at org.junit.jupiter.engine.descriptor.ClassTestDescriptor.<init>(ClassTestDescriptor.java:77)
	at org.junit.jupiter.engine.discovery.TestContainerResolver.resolveClass(TestContainerResolver.java:100)
	at org.junit.jupiter.engine.discovery.TestContainerResolver.resolveElement(TestContainerResolver.java:45)
	at org.junit.jupiter.engine.discovery.JavaElementsResolver.tryToResolveWithResolver(JavaElementsResolver.java:208)
	at org.junit.jupiter.engine.discovery.JavaElementsResolver.lambda$resolve$9(JavaElementsResolver.java:195)
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
	at java.util.Iterator.forEachRemaining(Iterator.java:116)
	at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
	at org.junit.jupiter.engine.discovery.JavaElementsResolver.resolve(JavaElementsResolver.java:198)
	at org.junit.jupiter.engine.discovery.JavaElementsResolver.lambda$resolveForAllParents$5(JavaElementsResolver.java:164)
	at java.util.Collections$SingletonSet.forEach(Collections.java:4767)
	at org.junit.jupiter.engine.discovery.JavaElementsResolver.resolveForAllParents(JavaElementsResolver.java:164)
	at org.junit.jupiter.engine.discovery.JavaElementsResolver.resolveContainerWithParents(JavaElementsResolver.java:85)
	at org.junit.jupiter.engine.discovery.JavaElementsResolver.resolveClass(JavaElementsResolver.java:60)
	at java.util.ArrayList.forEach(ArrayList.java:1255)
	at java.util.Collections$UnmodifiableCollection.forEach(Collections.java:1080)
	at org.junit.jupiter.engine.discovery.DiscoverySelectorResolver.lambda$resolve$2(DiscoverySelectorResolver.java:66)
	at java.util.ArrayList.forEach(ArrayList.java:1255)
	at org.junit.jupiter.engine.discovery.DiscoverySelectorResolver.resolve(DiscoverySelectorResolver.java:65)
	at org.junit.jupiter.engine.discovery.DiscoverySelectorResolver.resolveSelectors(DiscoverySelectorResolver.java:50)
	at org.junit.jupiter.engine.JupiterTestEngine.discover(JupiterTestEngine.java:61)
	at org.junit.platform.launcher.core.DefaultLauncher.discoverEngineRoot(DefaultLauncher.java:130)
	at org.junit.platform.launcher.core.DefaultLauncher.discoverRoot(DefaultLauncher.java:117)
	at org.junit.platform.launcher.core.DefaultLauncher.discover(DefaultLauncher.java:82)
	at com.intellij.junit5.JUnit5IdeaTestRunner.startRunnerWithArgs(JUnit5IdeaTestRunner.java:48)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)

Mar 17, 2018 3:33:58 PM org.junit.platform.launcher.core.DefaultLauncher handleThrowable
WARNING: TestEngine with ID 'junit-jupiter' failed to discover tests
java.lang.ArrayStoreException: sun.reflect.annotation.TypeNotPresentExceptionProxy
	at sun.reflect.annotation.AnnotationParser.parseClassArray(AnnotationParser.java:724)
	at sun.reflect.annotation.AnnotationParser.parseArray(AnnotationParser.java:531)
	at sun.reflect.annotation.AnnotationParser.parseMemberValue(AnnotationParser.java:355)
	at sun.reflect.annotation.AnnotationParser.parseAnnotation2(AnnotationParser.java:286)
	at sun.reflect.annotation.AnnotationParser.parseAnnotations2(AnnotationParser.java:120)
	at sun.reflect.annotation.AnnotationParser.parseAnnotations(AnnotationParser.java:72)
	at java.lang.Class.createAnnotationData(Class.java:3521)
	at java.lang.Class.annotationData(Class.java:3510)
	at java.lang.Class.getDeclaredAnnotation(Class.java:3458)
	at org.junit.platform.commons.util.AnnotationUtils.findAnnotation(AnnotationUtils.java:135)
	at org.junit.platform.commons.util.AnnotationUtils.findAnnotation(AnnotationUtils.java:114)
	at org.junit.jupiter.engine.descriptor.JupiterTestDescriptor.determineDisplayName(JupiterTestDescriptor.java:86)
	at org.junit.jupiter.engine.descriptor.ClassTestDescriptor.<init>(ClassTestDescriptor.java:83)
	at org.junit.jupiter.engine.descriptor.ClassTestDescriptor.<init>(ClassTestDescriptor.java:77)
	at org.junit.jupiter.engine.discovery.TestContainerResolver.resolveClass(TestContainerResolver.java:100)
	at org.junit.jupiter.engine.discovery.TestContainerResolver.resolveElement(TestContainerResolver.java:45)
	at org.junit.jupiter.engine.discovery.JavaElementsResolver.tryToResolveWithResolver(JavaElementsResolver.java:208)
	at org.junit.jupiter.engine.discovery.JavaElementsResolver.lambda$resolve$9(JavaElementsResolver.java:195)
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
	at java.util.Iterator.forEachRemaining(Iterator.java:116)
	at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
	at org.junit.jupiter.engine.discovery.JavaElementsResolver.resolve(JavaElementsResolver.java:198)
	at org.junit.jupiter.engine.discovery.JavaElementsResolver.lambda$resolveForAllParents$5(JavaElementsResolver.java:164)
	at java.util.Collections$SingletonSet.forEach(Collections.java:4767)
	at org.junit.jupiter.engine.discovery.JavaElementsResolver.resolveForAllParents(JavaElementsResolver.java:164)
	at org.junit.jupiter.engine.discovery.JavaElementsResolver.resolveContainerWithParents(JavaElementsResolver.java:85)
	at org.junit.jupiter.engine.discovery.JavaElementsResolver.resolveClass(JavaElementsResolver.java:60)
	at java.util.ArrayList.forEach(ArrayList.java:1255)
	at java.util.Collections$UnmodifiableCollection.forEach(Collections.java:1080)
	at org.junit.jupiter.engine.discovery.DiscoverySelectorResolver.lambda$resolve$2(DiscoverySelectorResolver.java:66)
	at java.util.ArrayList.forEach(ArrayList.java:1255)
	at org.junit.jupiter.engine.discovery.DiscoverySelectorResolver.resolve(DiscoverySelectorResolver.java:65)
	at org.junit.jupiter.engine.discovery.DiscoverySelectorResolver.resolveSelectors(DiscoverySelectorResolver.java:50)
	at org.junit.jupiter.engine.JupiterTestEngine.discover(JupiterTestEngine.java:61)
	at org.junit.platform.launcher.core.DefaultLauncher.discoverEngineRoot(DefaultLauncher.java:130)
	at org.junit.platform.launcher.core.DefaultLauncher.discoverRoot(DefaultLauncher.java:117)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:90)
	at com.intellij.junit5.JUnit5IdeaTestRunner.startRunnerWithArgs(JUnit5IdeaTestRunner.java:65)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)


Process finished with exit code 0
Empty test suite.
```

You can run the individual tests though and see how it works. I would like to finish this PR asap and publish next week, to finally have Mockito + Junit5 working :tada:

TimvdLippe:
I was able to fix the Gradle integration after following https://junit.org/junit5/docs/current/user-guide/#running-tests-build-gradle IntelliJ still broken, but I got to go now.

subprojects/junit5/src/main/java/org/mockito/junit5/ConfiguredWithMockito.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright (c) 2018 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.junit5;
```
marcphilipp: Can we rename this package to `org.mockito.junit.jupiter`? The JUnit team has taken extra care not to put `5` into the package name to make major version updates possible.

TimvdLippe: Good call, done :+1:

build.gradle
```diff
@@ -81,7 +81,7 @@ dependencies {
 }
 
 task wrapper(type: Wrapper) {
-    gradleVersion = '3.5'
+    gradleVersion = '4.2.1'
```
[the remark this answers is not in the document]
TimvdLippe: Nice catch :+1:

gradle/root/java6-compatibility.gradle
```diff
@@ -3,10 +3,12 @@
...
-        plugins.withId('java') {
-            p.apply plugin: 'ru.vyarus.animalsniffer'
-            p.dependencies {
-                signature 'org.codehaus.mojo.signature:java16:1.0@signature'
+        if (p.name != 'junit5') {
+            plugins.withId('java') {
+                p.apply plugin: 'ru.vyarus.animalsniffer'
+                p.dependencies {
+                    signature 'org.codehaus.mojo.signature:java16:1.0@signature'
+                }
```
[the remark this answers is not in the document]
TimvdLippe: I am not really sure what you mean with baseline. We configure the Java 1.8 compatibility in the subproject:

```gradle
sourceCompatibility = 1.8
targetCompatibility = 1.8
```

Could you clarify your comment?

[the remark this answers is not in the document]
marcphilipp: Source and target compatibility make sure you only use Java 8 language features and compile to Java 8 byte code, respectively. The animalsniffer plugin checks that you don't use any APIs that are not present in whatever baseline you pick. For example, if you would use an API that was introduced in Java 9, and the build runs on Java 9, it would compile without any issues. Users on Java 8 would however get a runtime exception. For Java 8, there's `org.codehaus.mojo.signature:java18:1.0`.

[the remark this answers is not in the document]
TimvdLippe: Oh I see, yes good call. Done :+1:

.travis.yml
```diff
@@ -38,7 +38,7 @@ script:
   # We are using && below on purpose
   # ciPerformRelease must run only when the entire build has completed
   # This guarantees that no release steps are executed when the build or tests fail
-  - ./gradlew build idea -s -PcheckJava6Compatibility && ./gradlew ciPerformRelease
+  - ./gradlew build idea -s -PcheckJavaCompatibility && ./gradlew ciPerformRelease
```
mockitoguy: Nice cleanup!

subprojects/junit-jupiter/junit-jupiter.gradle
```diff
@@ -0,0 +1,19 @@
...
+    testCompile libraries.assertj
+    testCompile libraries.junitJupiterLauncher
+    implementation libraries.junitJupiterEngine
+}
+
+test {
+    useJUnitPlatform()
+}
+
+tasks.javadoc.enabled = true
```
mockitoguy: Most likely not needed - javadoc is enabled by default.

subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/ConfiguredWithMockito.java
```diff
@@ -0,0 +1,26 @@
...
+import java.lang.annotation.Retention;
+
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Annotation that can configure Mockito as invoked by the {@link MockitoExtension}.
+ */
+@ExtendWith(MockitoExtension.class)
+@Retention(RUNTIME)
+public @interface ConfiguredWithMockito {
```
mockitoguy: Can you update PR description with all the public API changes / additions so that the team can review them in one spot?

I'm curious why did you go with ```ConfiguredWithMockito```? Alternatives:
 - ```ExtendWithMockito``` - is syntactically consistent with ```ExtendWith```, default Jupiter annotation.
 - ```MockitoExtension.Settings``` - easier to discover, clearer intent (it's a setting for the extension, not general purpose Mockito configuration)
- ```MockitoExtensionSettings```

+1 to ```ExtendWithMockito```
-1 to ```ConfiguredWithMockito```

If I missed a discussion about it, please let me know so that I can read it and revise my vote.

subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/ConfiguredWithMockito.java
```diff
@@ -0,0 +1,26 @@
...
+
+    /**
+     * Configure the strictness used in this test.
+     * @return The strictness to configure, by default {@link Strictness#WARN}
+     */
+    Strictness strictness() default Strictness.WARN;
```
mockitoguy: New Mockito features (such as MockitoSession) uses Strictness.STRICT_STUBS by default because it helps us and the users:
 - cleaner tests that are easier to debug with STRICT_STUBS
 - it is a compatible way of introducing a new feature

How about we change the default for Jupiter to use STRICT_STUBS, too?

subprojects/junit-jupiter/src/test/java/org/mockitousage/StrictnessTest.java
```diff
@@ -0,0 +1,118 @@
...
+import org.mockito.junit.jupiter.ConfiguredWithMockito;
+import org.mockito.quality.Strictness;
+
+import java.util.function.Function;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.platform.engine.discovery.DiscoverySelectors.selectClass;
+
+@SuppressWarnings("ConstantConditions")
+class StrictnessTest {
```
mockitoguy: How come this test does not need ```MockitoExtension```?

subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/MockitoExtension.java
```diff
@@ -0,0 +1,165 @@
...
+
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Optional;
+
+import static org.junit.jupiter.api.extension.ExtensionContext.Namespace.create;
+import static org.junit.platform.commons.support.AnnotationSupport.findAnnotation;
+
+/**
+ * Extension that initializes mocks and handles strict stubbings. This extension is the JUnit Jupiter equivalent
```
mockitoguy: What do you think about a new entry in main mockito javadoc in ```Mockito.java```? It's a noteworthy feature!!!

subprojects/junit-jupiter/junit-jupiter.gradle
```diff
@@ -0,0 +1,19 @@
...
+
+dependencies {
+    compile project.rootProject
+    testCompile libraries.assertj
+    testCompile libraries.junitJupiterLauncher
+    implementation libraries.junitJupiterEngine
+}
+
+test {
+    useJUnitPlatform()
```
mockitoguy: Gradle team call Junit Jupiter "JUnit platform". I'm curious why? Should we follow this naming convention?

mockitoguy:
@sbrannen, what's the most consistent and popular naming convention? Gradle team calls new JUnit5 ```JUnitPlatform``` because in the Gradle file you write this:

```
test {
  useJUnitPlatform()
}
``` 

In this PR, we're shipping new module ```org.mockito:junit-jupiter``` (which I prefer). What is the best name?

subprojects/junit-jupiter/src/test/java/org/mockitousage/StrictnessTest.java
mockitoguy: Oh, meta-annotations. Cool.

subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/ConfiguredWithMockito.java
TimvdLippe: This was mostly after a discussion with @marcphilipp We concluded that this annotation is for "configuring" the test, as it takes configuration parameters of the extension. Therefore I would say this is more clear

subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/ConfiguredWithMockito.java
TimvdLippe: I was under the impression that the WARN behavior was still default, but I just checked https://github.com/mockito/mockito/blob/45a78a03a38fd2408f0ea7a3e1342fbffcfb9237/src/main/java/org/mockito/junit/MockitoJUnitRunner.java#L152-L155 and it is indeed Strict. Will update :+1:

subprojects/junit-jupiter/src/test/java/org/mockitousage/StrictnessTest.java
TimvdLippe: This test is intended to run a single JUnit test that then runs inner classes (which do have the extension)

TimvdLippe:
@mockitoguy The platform is something different than Jupiter. The platform is a separate organisation (https://mvnrepository.com/artifact/org.junit.platform) which handles integration with platform-oriented clients (like Gradle, IntelliJ, Maven, etc...). Instead, we are building an extension that is built on top of the JUnit API, and this is called Jupiter.

subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/MockitoExtension.java
TimvdLippe: Done :+1:

sbrannen:
BTW, I highly recommend my explanation of what "JUnit 5" is [here](https://junit.org/junit5/docs/current/user-guide/#overview-what-is-junit-5). 😉 

**_JUnit 5 = JUnit Platform + JUnit Jupiter + JUnit Vintage_**

blandger:
When can we expect releasing junit5 support by Mockito?

subprojects/junit-jupiter/src/test/java/org/mockitousage/StrictnessTest.java
```diff
@@ -0,0 +1,123 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.platform.engine.discovery.DiscoverySelectors.selectClass;
+
+/**
+ * Test that runs the inner test using a launcher {@see #invokeTestClassAndRetrieveMethodResult}.
+ * We then assert on the actual test run output, to see if test actually failed as a result
+ * of our extension.
+ */
+@SuppressWarnings("ConstantConditions")
+class StrictnessTest {
```
mockitoguy: Can you add a test that demonstrates that by default we're using STRICT_STUBS?

subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/ConfiguredWithMockito.java
mockitoguy: Are you attached to “ConfiguredWithMockito” annotation name? It looks odd in test. Isn’t “WithMockito” or “MockitoSettings” nicer? “Configured” is redundant, “MockitoSettings” is nice because we already have withSettings() method at mock creation. I'm not against "ConfiguredWithMockito" but I think we could do better. I'm looking for a best name because long term we will have similar annotation in mockito-core.

TimvdLippe:
@mockitoguy Feedback addressed! If no one else objects, I am merging this later today and publish a new Maven central release for Mockito Java 10/11 and Junit5 compatibility!

TimvdLippe:
All right, let's do this.

Thanks @ChristianSchwarz for the initial implementation, @marcphilipp @sbrannen for the JUnit feedback and help and the others for reviewing. Merging!

subprojects/junit-jupiter/junit-jupiter.gradle
```diff
@@ -0,0 +1,17 @@
...
+apply from: "$rootDir/gradle/java-library.gradle"
+
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
+
+dependencies {
+    compile project.rootProject
+    testCompile libraries.assertj
+    testCompile libraries.junitJupiterLauncher
+    implementation libraries.junitJupiterEngine
```
marcphilipp: @TimvdLippe I know it's a bit late now... but depending on junit-jupiter-api should be sufficient, shouldn't it?

subprojects/junit-jupiter/junit-jupiter.gradle
sormuras: **JUnit Platform** supports many engines, like "JUnit Jupiter (JUnit 5)", or "JUnit Vintage (JUnit 3 and 4)" or other `TestEngine` implementations, like those listed here: https://github.com/junit-team/junit5/wiki/Third-party-Extensions#junit-platform-test-engines

**JUnit Jupiter** is an implementation of the `TestEngine` interface ... which provides extension support. Here is where the Mockito Extension is living.

More details at: https://junit.org/junit5/docs/current/user-guide/#overview-what-is-junit-5

subprojects/junit-jupiter/junit-jupiter.gradle
sormuras: As types like `org.junit.platform.engine.discovery.DiscoverySelectors` and `org.junit.platform.engine.TestExecutionResult` are used within the tests at least a "test-related" configuration needs the `junit-jupiter-engine` dependency.

TimvdLippe: Um, I am not sure. There are many artifacts and I am not sure what is the minimum we must support. If api is sufficient, feel free to send in a PR :)

sbrannen: The current published [pom.xml](https://bintray.com/mockito/maven/download_file?file_path=org%2Fmockito%2Fmockito-junit-jupiter%2F2.17.0%2Fmockito-junit-jupiter-2.17.0.pom) contains the following mandatory transitive dependency.

```xml
    <dependency>
      <artifactId>junit-jupiter-engine</artifactId>
      <version>5.1.0</version>
      <scope>runtime</scope>
    </dependency>
```

I agree with @marcphilipp and @sormuras: you should likely have a mandatory `compile` dependency on `junit-jupiter-api` and an `optional` `runtime` dependency on `junit-jupiter-engine`.

sbrannen: Developers will of course need the dependency on the engine to actually execute tests, but I'm not so sure that the Mockito artifact should assume the developer wants that as a mandatory dependency.

Some users prefer to configure the dependency on the engine within the plug-in configuration for Maven Surefire as opposed to having it as a top-level dependency for the entire project.

sbrannen: Any volunteers for creating such a PR?

TimvdLippe: We have already updated the dependency in #1344

sbrannen: ... or perhaps just a `test` dependency on `junit-jupiter-engine`, since Mockito only depends on it within Mockito's own test suite.

sbrannen: Ahhh... I missed #1344.

Thanks for providing the link!

PR #1224, opened 2017-10-26
Fix issue #1222 and #584

### LINKED ISSUE #584: ArgumentCaptor can't capture varargs-arrays

### LINKED ISSUE #1222: Cannot verify varargs parameter as an array
Cannot verify the varargs by comparing with an array.
```
    @Test
    public void shouldVerifyVarargsAsArray() throws Exception {

        mock.mixedVarargs("1", "2", "3");

        verify(mock).mixedVarargs(any(), eq(new String[] {"2", "3"}));
    }
```

**Expected:** test runs successfully
**Actual:** the test fails with an argument mismatch error.

```
org.mockitousage.verification.BasicVerificationTest > shouldVerifyVarargsAsArray FAILED
    iMethods.mixedVarargs(<any>, ["2", "3"]);
    -> at org.mockitousage.verification.BasicVerificationTest.shouldVerifyVarargsAsArray(BasicVerificationTest.java:125)
    iMethods.mixedVarargs("1", "2", "3");
    -> at org.mockitousage.verification.BasicVerificationTest.shouldVerifyVarargsAsArray(BasicVerificationTest.java:123)
    ...
```

This was found while upgrading from 1.9.5. In that version it was possible to achieve something similar by creating a custom ArgumentMatcher that implemented VarargMatcher. In that case the ArgumentMatcher was given the varargs as an array. However, in the latest version the same matcher is passed each vararg separately which prevents it from matching.

It is not possible to inline the array as follows because in the actual code the contents of the array are passed as an array into the method that does the verification.
`
    verify(mock).mixedVarargs(any(), eq("1"), eq("2"));
`

I tried doing something like this:
```
    verify(mock).mixedVarargs(any(), inline(new String[] {"1", "2"}));
  }
  public static <T> T inline(T[] array) {
    for (Object o : array) {
      eq(o);
    }
    return null;
  }
```

But that failed in MatchersBinder because the number of matchers provided (3) did not match the number of arguments passed in the invocation (2).

I also tried to remove the VarargMatcher from the custom ArgumentMatcher but that failed too as the number of arguments did not match in MatcherApplicationStrategy.

paulduffin:
Adds a new method varargsAsArray(...) that indicates that the varargs
should be matched/captured as a single array rather than separate
values.

Fixes #1222 
Fixes #584

### DIFF
--- a/src/main/java/org/mockito/ArgumentMatchers.java
+++ b/src/main/java/org/mockito/ArgumentMatchers.java
@@ -6,12 +6,14 @@ package org.mockito;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.EmptyStackException;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
+import org.mockito.exceptions.misusing.InvalidUseOfMatchersException;
 import org.mockito.internal.matchers.Any;
 import org.mockito.internal.matchers.Contains;
 import org.mockito.internal.matchers.EndsWith;
@@ -22,6 +24,7 @@ import org.mockito.internal.matchers.NotNull;
 import org.mockito.internal.matchers.Null;
 import org.mockito.internal.matchers.Same;
 import org.mockito.internal.matchers.StartsWith;
+import org.mockito.internal.matchers.TreatVarargsAsArray;
 import org.mockito.internal.matchers.apachecommons.ReflectionEquals;
 import org.mockito.internal.util.Primitives;
 
@@ -1315,6 +1318,24 @@ public class ArgumentMatchers {
     public static double doubleThat(ArgumentMatcher<Double> matcher) {
         reportMatcher(matcher);
         return 0;
+    }
+
+    /**
+     * Indicates that the varargs should be matched as an array rather than individual values.
+     *
+     * The value must be specified as a call to a method for matching an argument, e.g.
+     * {@link #eq(Object)} or {@link ArgumentCaptor#capture()}. Failure to do so will result in
+     * either an {@link EmptyStackException} or an {@link InvalidUseOfMatchersException}.
+     *
+     * @param value expected to be a matcher such as {@code eq(...) or
+     * {@code captor.capture()}
+     * @param <T> the type of the vararg or array of varargs
+     * @return {@code null}
+     */
+    public static <T> T varargsAsArray(T value) {
+        ArgumentMatcher nestedMatcher = mockingProgress().getArgumentMatcherStorage().popMatcher();
+        reportMatcher(new TreatVarargsAsArray<Object>(nestedMatcher));
+        return null;
     }
 
     private static void reportMatcher(ArgumentMatcher<?> matcher) {
--- a/src/main/java/org/mockito/internal/invocation/MatcherApplicationStrategy.java
+++ b/src/main/java/org/mockito/internal/invocation/MatcherApplicationStrategy.java
@@ -6,13 +6,15 @@ package org.mockito.internal.invocation;
 
 import static org.mockito.internal.invocation.MatcherApplicationStrategy.MatcherApplicationType.ERROR_UNSUPPORTED_NUMBER_OF_MATCHERS;
 import static org.mockito.internal.invocation.MatcherApplicationStrategy.MatcherApplicationType.MATCH_EACH_VARARGS_WITH_LAST_MATCHER;
-import static org.mockito.internal.invocation.MatcherApplicationStrategy.MatcherApplicationType.ONE_MATCHER_PER_ARGUMENT;
+import static org.mockito.internal.invocation.MatcherApplicationStrategy.MatcherApplicationType.ONE_MATCHER_PER_EXPANDED_ARGUMENT;
+import static org.mockito.internal.invocation.MatcherApplicationStrategy.MatcherApplicationType.ONE_MATCHER_PER_RAW_ARGUMENT;
 
 import java.util.ArrayList;
 import java.util.List;
 
 import org.mockito.ArgumentMatcher;
 import org.mockito.internal.matchers.CapturingMatcher;
+import org.mockito.internal.matchers.TreatVarargsAsArray;
 import org.mockito.internal.matchers.VarargMatcher;
 import org.mockito.invocation.Invocation;
 
@@ -74,7 +76,7 @@ public class MatcherApplicationStrategy {
         if (matchingType == ERROR_UNSUPPORTED_NUMBER_OF_MATCHERS)
             return false;
 
-        Object[] arguments = invocation.getArguments();
+        Object[] arguments = matchingType.getArguments(invocation);
         for (int i = 0; i < arguments.length; i++) {
             ArgumentMatcher<?> matcher = matchers.get(i);
             Object argument = arguments[i];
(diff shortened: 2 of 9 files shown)


paulduffin:
I am working on fixing the code coverage problem and adding some more tests for TreatsVarargsAsArray.

TimvdLippe:
Code coverage seems mostly concerned regarding the boilerplate `toString` and friends. Not a huge dealbreaker, but adding more tests for to get full coverage is :100:

ChristianSchwarz:
I would like to see some test cases too, to get a feel for the API in corner cases

```
ArgumentCaptor<int[]>  captor;
when(mock).varargInts(varargsAsArray(captor.capture())) //varargInts(int...values)
```

```
when(mock).varargInts(varargsAsArray(eq(5))) //varargInts(int...values)
```

```
when(mock).stringVarargInts(eq("a"),varargsAsArray(123)) //stringVarargInts(String,int...values)
```

paulduffin:
I did add some more tests after the first upload. Do you want even more?

ChristianSchwarz:
In the end the mockito-core members or you decide! I am as you a contributor and can only give the advices to add more non-happy-path tests.

Last year I looked into the this issue and other related varargs, capturing and matching issues. Which leaded to some refactorings, regarding the whole capturing and matching process. The `MatcherApplicationStrategy` were the most of your changes were made are a result of it. I never found a good solution to handle / prevent API misuses. From that POV I would like to know 2 things:
* How does the API handle/prevent misuses and communicate them?
* Can we ensure this API solves the problem and introduces no new ones. 

Here is some other interesting test case:

```
  mock.varargsInt(1,2,3);
  verify(mock).varargInts(varargsAsArray(or(eq(null),eq(new int[]{1,2,3})))) 
  //what is the result: success, assertion failure or an exception indicating a bug
```

ChristianSchwarz:
Somewhere it should be checked that this new matcher is only allowed to be placed as last-arg of a vararg call.

mockitoguy:
Hey guys! This is a great discussion, thank you @TimvdLippe for bringing me in!

@paulduffin, this PR is conceivable! Before we merge it let's make an attempt to solve the use case more conveniently for our users.

@ChristianSchwarz, can you comment on #1235? I prototyped/hacked the implementation so that the use case is handled by Mockito out of the box. Can you point out gaps / add tests that demonstrate gaps?

PR #1235 makes both types of syntax work (and all existing Mockito tests are passing):

```Java
mock.mixedVarargs("1", "2", "3");

//traditional verify with varargs, same number of args
verify(mock).mixedVarargs(any(), eq("2"), eq("3"));

//new matching, using an array
verify(mock).mixedVarargs(any(), eq(new String[] {"2", "3"}));
verify(mock).mixedVarargs(any(), (String[]) any());
```

Thanks everyone for working on this use case!

ChristianSchwarz:
Some puzzelz, but they are solvable...

```java
    @Test
    public void shouldNoWorkOnNonVarargParameter_argOfTypeString()   {
        exception.expect(MockitoException.class);
        exception.expectMessage("varargsAsArray(..) is only allowed on vararg parameters");
        when(mock.simpleMethod(varargsAsArray(eq("")))).thenReturn("");
    }

    @Test
    public void shouldNoWorkOnNonVarargParameter_argOfTypeStringArray()   {
        exception.expect(MockitoException.class);
        exception.expectMessage("varargsAsArray(..) is only allowed on vararg parameters, actual parameter is of type: String");
        when(mock.simpleMethod("",varargsAsArray(eq(new String[0])))).thenReturn("");
    }

    @Test
    public void shouldNoWorkOnNonVarargParameter_argOfTypeStringArray2()   {
        exception.expect(MockitoException.class);
        exception.expectMessage("varargsAsArray(..) is only allowed on vararg parameters, actual parameter is of type: String[]");
        when(mock.simpleMethod("",varargsAsArray(new String[0]))).thenReturn("");
    }

    @Test
    public void shouldNotCompile()  {
        when(mock.intArgumentReturningInt(varargsAsArray(0))).thenReturn(1);
    }

    @Test
    public void shouldConvertRawValuesToEqualsMatcher()  {
        when(mock.varargsReturningString(varargsAsArray(0))).thenReturn("");
    }

```

paulduffin:
They could all cause a compile error as long as the code was compiled with Error Prone and a suitable Error Prone checker was written for it.

Otherwise, the only choice is to have them as runtime errors.

Preventing it being used with non-vararg and non-array parameters could be done by having it return T[] instead of T. Unfortunately, that will not work with primitive varargs. e.g. if the parameter was int... then varargsAsArray(eq(new int[] {1})) would have to return int[] but T[] cannot be assigned to int[] because T cannot be primitive. The closest it can be is Integer and Integer[] is not assignable to int[].

That would require primitive specializations of varargsAsArray, e.g. varargsAsIntArray, etc.

mockitoguy:
Thank you for explaining the PR here and in #1235. It is immensely helpful. My feedback:
 - +1 to adding explicit "varargs" matcher
 - what is the best name? What about shorter "vararg"?
 - with the current behavior, there are opportunities to improve usability (if we believe it is important enough, let's put together a ticket):
   - more context to the verification error message when vararg is involved and number of args don't match. We could hint to the new matcher.
   - improve behavior for the use cases we can solve unambiguously.

I'm doing the review of the code now.

src/main/java/org/mockito/internal/invocation/MatcherApplicationStrategy.java
```diff
@@ -91,21 +93,35 @@ private static MatcherApplicationType getMatcherApplicationType(Invocation invoc
         final int expandedArguments = invocation.getArguments().length;
         final int matcherCount = matchers.size();
 
+        boolean treatVarargsAsArray = false;
```
mockitoguy: Wow. I'm very impressed that you were able to inject the new algorithm here. The code is pretty complex around here. I'd love to see it refactored at some point in the future.

It is complex but you have added fantastic test coverage that I was not able to break down :)

src/test/java/org/mockitousage/verification/BasicVerificationTest.java
```diff
@@ -115,4 +115,49 @@ public void shouldDetectWhenOverloadedMethodCalled() throws Exception {
             fail();
         } catch(WantedButNotInvoked e) {}
     }
+
+    @Test
+    public void shouldVerifyVarargsAsArray_nullVarargsArray() throws Exception {
```
mockitoguy: Can you move those tests to a better place? "BasicVerification" is for basic operations. I suggest to move them to existing VarargTest or create a new one like VarargAsArrayTest.

BTW. thank you very much for adding truly comprehensive test coverage!!!

src/test/java/org/mockitousage/verification/BasicVerificationTest.java
```diff
@@ -115,4 +115,49 @@ public void shouldDetectWhenOverloadedMethodCalled() throws Exception {
...
+        IMethods mock = mock(IMethods.class);
+
+        mock.mixedVarargs(null, (String[]) null);
+
+        verify(mock).mixedVarargs(any(), varargsAsArray(eq((String[]) null)));
```
mockitoguy: Can you add coverage for negative cases, e.g. when verification error is thrown for use cases with "varargsAsArray"? This would also help us see the exception message emitted from Mockito, and if the message is handsome.

src/main/java/org/mockito/ArgumentMatchers.java
```diff
@@ -1317,6 +1320,24 @@ public static double doubleThat(ArgumentMatcher<Double> matcher) {
         return 0;
     }
 
+    /**
```
mockitoguy: Can you add a new point to the main Mockito Javadoc with a short description of the feature? This is a really cool feature and it deserves a note in main Javadoc. That doc acts as Mockito user guide and also shows notable features. Feel free to bump new minor version.

fransflippo:
@paulduffin @mockitoguy Would love to get this into a release version of Mockito! Can I help address some of the issues that @mockitoguy identified perhaps so we can finish it off?

ChristianSchwarz:
@mockitoguy 
@TimvdLippe 
Can you create a mockito branch of this PR, so other can work on it and fix the remaining issues?

TimvdLippe:
@ChristianSchwarz https://github.com/mockito/mockito/tree/capture-vararg-as-array published

paulduffin:
Apologies for not getting back sooner on this. Things are crazy at work at
the minute. If someone else wants to address the issues that would be
really helpful.

TungDoanDuong:
Any update on this issue?

TimvdLippe:

PR #1226, opened 2017-10-27
Change default strictness in deprecated runners/MockitoJUnitRunner to Silent

### LINKED ISSUE #1212: MockitoJUnitRunner too strict by default?
In one of the projects being migrated to Mockito 2 I've observed that deprecated `org.mockito.runners.MockitoJUnitRunner` (tested with 2.11.0) works in a strict mode by default (fails a test on unnecessary stubbing). I wonder if it is an expected behavior?

Even Mockito rule works in a "warn" mode by default. It makes it a little bit harder to migrate from Mockito 1.10.x to 2.x.

TimvdLippe:
This seemed to be an oversight incorrectly generating strict warnings for users migrating to Mockito 2. Locally I saw no test failures, so let's see what Travis says.

Fixes #1212

### DIFF
--- a/src/main/java/org/mockito/junit/MockitoJUnitRunner.java
+++ b/src/main/java/org/mockito/junit/MockitoJUnitRunner.java
@@ -152,7 +152,7 @@ public class MockitoJUnitRunner extends Runner implements Filterable {
         this(new StrictRunner(new RunnerFactory().createStrict(klass), klass));
     }
 
-    MockitoJUnitRunner(InternalRunner runner) throws InvocationTargetException {
+    protected MockitoJUnitRunner(InternalRunner runner) throws InvocationTargetException {
         this.runner = runner;
     }
 
--- a/src/main/java/org/mockito/runners/MockitoJUnitRunner.java
+++ b/src/main/java/org/mockito/runners/MockitoJUnitRunner.java
@@ -9,6 +9,8 @@ import org.junit.runner.Description;
 import org.junit.runner.manipulation.Filter;
 import org.junit.runner.manipulation.NoTestsRemainException;
 import org.junit.runner.notification.RunNotifier;
+import org.mockito.internal.runners.InternalRunner;
+import org.mockito.internal.runners.RunnerFactory;
 
 
 /**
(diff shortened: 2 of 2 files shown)


mockitoguy:
Looking into it!

src/main/java/org/mockito/junit/MockitoJUnitRunner.java
```diff
@@ -152,7 +152,7 @@ public MockitoJUnitRunner(Class<?> klass) throws InvocationTargetException {
         this(new StrictRunner(new RunnerFactory().createStrict(klass), klass));
     }
 
-    MockitoJUnitRunner(InternalRunner runner) throws InvocationTargetException {
+    protected MockitoJUnitRunner(InternalRunner runner) throws InvocationTargetException {
```
mockitoguy: Curious why changing the public modifier?

I suspect that this change may not be binary compatible in case somebody extends the runner.

src/main/java/org/mockito/runners/MockitoJUnitRunner.java
```diff
@@ -27,7 +29,7 @@
     @Deprecated
     public static class Silent extends MockitoJUnitRunner {
```
mockitoguy: The code works as designed (https://github.com/mockito/mockito/issues/1212#issuecomment-340266646).

If you want to keep the PR I suggest that we add a unit test that uses deprecated runner so that we preserve and document this behavior.

TimvdLippe:
@mockitoguy not sure how to proceed, as the linked issue was closed. I think @szpak is still encountering this issue in his project?

mockitoguy:
>@mockitoguy not sure how to proceed, as the linked issue was closed. I think @szpak is still encountering this issue in his project?

Sorry I was not clear. The code works as designed. @szpak encountered an issue with an intentional incompatible change we have done in Mockito 2.x. There is nothing to fix besides the making the behavior clearer in the documentation. You can close the PR as "won't fix" or update the docs. Hope that clears things!

TimvdLippe:
Closing this PR for now then. @szpak feel free to submit updates to the docs as you see fit :smile:

szpak:
@mockitoguy IMHO it's not consistent. Mockito JUnit rule and the new runner doesn't work that way. It makes the migration harder (more broken tests).

Nevertheless, it's quite easy to fix (by replacing that runner) and if you did it purposely I don't have strong arguments to revert it.

PR #1227, opened 2017-10-27
Update license plugin and add missing license headers

TimvdLippe:
The license plugin could be updated to later versions, as we dropped JDK6 on Travis. Moreover, quite some licenses were reported missing, so I ran `./gradlew licenseFormatMain licenseFormatTest` to add them.


mockitoguy:
Thank you!

PR #1228, opened 2017-10-27
Add @CheckReturnValue to stubbing/verification methods

### LINKED ISSUE #1130: Mockito should annotate when() methods with a custom annotation @CheckReturnValue
Static code checkers like error-Prone and FindBugs will help detecting invalid usage of Mockito in single-call situations if Mockito source code were annotated with @CheckReturnValue.

See:
* https://github.com/joel-costigliola/assertj-core/issues/157
* http://errorprone.info/bugpattern/CheckReturnValue
* https://github.com/joel-costigliola/assertj-core/pull/695
* https://github.com/joel-costigliola/assertj-core/blob/master/src/main/java/org/assertj/core/util/CheckReturnValue.java

Note like for AssertJ, any annotation with name should do, no need to depend on a specific library with  @CheckReturnValue.

TimvdLippe:
Introduce a new annotation `@CheckReturnValue` and add this to methods like `when` and `verify` such that static analysis can verify the value is actually used (http://errorprone.info/bugpattern/CheckReturnValue). This makes sure developers can get warnings on build-time instead of run-time (with our internal `validateState`)

Fixes #1130

### DIFF
--- a/src/main/java/org/mockito/CheckReturnValue.java
+++ b/src/main/java/org/mockito/CheckReturnValue.java
@@ -1,0 +1,23 @@
+package org.mockito;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+
+/**
+ * Findbugs handles any annotation with name "CheckReturnValue" in return value check.
+ *
+ * @see <a href="https://github.com/findbugsproject/findbugs/blob/264ae7baf890d2b347d91805c90057062b5dcb1e/findbugs/src/java/edu/umd/cs/findbugs/detect/BuildCheckReturnAnnotationDatabase.java#L120">Findbugs source code</a>
+ * @see <a href="http://errorprone.info/bugpattern/CheckReturnValue">ErrorProne check</a>
+ */
+@Target({
+    ElementType.CONSTRUCTOR,
+    ElementType.METHOD,
+    ElementType.PACKAGE,
+    ElementType.TYPE
+})
+@Retention(RetentionPolicy.CLASS)
+@interface CheckReturnValue {
+}
--- a/src/main/java/org/mockito/CheckReturnValue.java
+++ b/src/main/java/org/mockito/CheckReturnValue.java
@@ -7,10 +7,15 @@ import java.lang.annotation.Target;
 
 
 /**
- * Findbugs handles any annotation with name "CheckReturnValue" in return value check.
+ * This annotation is not supposed to be used by Mockito end-users. Instead, we
+ * use it to annotate methods for Static Analysis tools, including FindBugs and ErrorProne.
+ * These tools can check whether the return value of our Mockito methods are actually
+ * used. As such, Mockito State Validation can be performed at compile-time rather than run-time.
+ * This annotation is public, because we have to use it in multiple packages.
  *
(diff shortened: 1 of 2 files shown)


src/main/java/org/mockito/CheckReturnValue.java
```diff
@@ -0,0 +1,23 @@
...
+ * @see <a href="https://github.com/findbugsproject/findbugs/blob/264ae7baf890d2b347d91805c90057062b5dcb1e/findbugs/src/java/edu/umd/cs/findbugs/detect/BuildCheckReturnAnnotationDatabase.java#L120">Findbugs source code</a>
+ * @see <a href="http://errorprone.info/bugpattern/CheckReturnValue">ErrorProne check</a>
+ */
+@Target({
+    ElementType.CONSTRUCTOR,
+    ElementType.METHOD,
+    ElementType.PACKAGE,
+    ElementType.TYPE
+})
+@Retention(RetentionPolicy.CLASS)
```
mockitoguy: Shouldn't it be RetentionPolicy.RUNTIME? Annotations with RetentionPolicy.CLASS are not preserved for the vm runtime execution, which is what I assume Findbugs needs.

src/main/java/org/mockito/CheckReturnValue.java
```diff
@@ -0,0 +1,23 @@
+package org.mockito;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+
+/**
+ * Findbugs handles any annotation with name "CheckReturnValue" in return value check.
```
mockitoguy: Since this annotation is on our core public API methods, can you add a paragraph describing what this annotation means for the end users, written in a way that even a newbie would understand :)

src/main/java/org/mockito/CheckReturnValue.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @see <a href="https://github.com/findbugsproject/findbugs/blob/264ae7baf890d2b347d91805c90057062b5dcb1e/findbugs/src/java/edu/umd/cs/findbugs/detect/BuildCheckReturnAnnotationDatabase.java#L120">Findbugs source code</a>
+ * @see <a href="http://errorprone.info/bugpattern/CheckReturnValue">ErrorProne check</a>
```
mockitoguy: Let's add "since" Javadoc tag

src/main/java/org/mockito/CheckReturnValue.java
TimvdLippe: Actually, I think we should make this package-protected. It's not intended to be used by Mockito users

src/main/java/org/mockito/CheckReturnValue.java
TimvdLippe: I don't think so, it seems to work in AssertJ: https://github.com/karlicoss/assertj-core/blob/master/src/main/java/org/assertj/core/util/CheckReturnValue.java

src/main/java/org/mockito/CheckReturnValue.java
mockitoguy: If you make it package-protected we won't be able to use this annotation on our public types that reside is subpackages.

src/main/java/org/mockito/CheckReturnValue.java
mockitoguy: ok :)

mockitoguy:
Approved your changes, please consider my feedback and ship at will!

PR #1229, opened 2017-10-27
Enable jacoco coverage on subprojects

### LINKED ISSUE #1070: Fix coverage for subprojects
In #1032, the new coverage was incorrectly tracked by jacoco. A quick search showed this TODO (https://github.com/mockito/mockito/blob/release/2.x/gradle/root/coverage.gradle#L3) that points out it is actually not generating coverage for subprojects. Would like to see this fixed!

TimvdLippe:
Inspired by http://csiebler.github.io/blog/2014/02/09/multi-project-code-coverage-using-gradle-and-jacoco/, Gradle is now configured to incorporate coverage of subprojects. This should increase our current coverage, as we now also incorporate the coverage generated from `kotlinTest`.

Fixes #1070

### DIFF
--- a/gradle/root/coverage.gradle
+++ b/gradle/root/coverage.gradle
@@ -1,15 +1,18 @@
 apply plugin: 'jacoco'
-
-// TODO later include subprojects, possible ideas https://gist.github.com/aalmiray/e6f54aa4b3803be0bcac
+apply from: 'gradle/dependencies.gradle'
 
 jacoco {
-    toolVersion = "0.7.9"
+    toolVersion = libraries.jacocoVersion
 }
 
-task mockitoCoverage(type: JacocoReport, dependsOn: "test") {
-    executionData files("${buildDir}/jacoco/test.exec")
+task mockitoCoverage(type: JacocoReport, dependsOn: ["test", "kotlinTest:test", "extTest:test", "inline:test"]) {
+    executionData fileTree(project.rootDir.absolutePath).include("**/build/jacoco/*.exec")
 
     sourceSets sourceSets.main
+
+    subprojects {
+        sourceSets it.sourceSets.main
+    }
 
     reports {
         xml.enabled true
@@ -31,4 +34,4 @@ task mockitoCoverage(type: JacocoReport, dependsOn: "test") {
     }
 }
 
-task coverageReport(dependsOn: ["test", "mockitoCoverage"]) {}
+task coverageReport(dependsOn: ["mockitoCoverage"]) {}
--- a/gradle/root/coverage.gradle
+++ b/gradle/root/coverage.gradle
@@ -1,18 +1,27 @@
-apply plugin: 'jacoco'
-apply from: 'gradle/dependencies.gradle'
+//TODO: Standard JaCoCo coverage doesn't work in Android module
+task mockitoCoverage(type: JacocoReport) {
 
-jacoco {
-    toolVersion = libraries.jacocoVersion
-}
+    allprojects { currentProject ->
+        plugins.withId("java") {
+            mockitoCoverage.sourceSets currentProject.sourceSets.main
 
-task mockitoCoverage(type: JacocoReport, dependsOn: ["test", "kotlinTest:test", "extTest:test", "inline:test"]) {
-    executionData fileTree(project.rootDir.absolutePath).include("**/build/jacoco/*.exec")
+            apply plugin: "jacoco"
 
-    sourceSets sourceSets.main
+            jacoco {
+                toolVersion = '0.7.9' //Gradle 4.2.1 uses 0.7.8 by default
 
-    subprojects {
-        sourceSets it.sourceSets.main
+                if (currentProject != rootProject) {   //already automatically enhanced in mockito main project as "java" plugin was applied before
+                    applyTo test
+                }
+            }
+
+            currentProject.tasks.withType(Test) { testTask ->
+                mockitoCoverage.dependsOn testTask
+            }
+        }
     }
+
+    executionData(file("$buildDir/jacoco/test.exec"))
 
     reports {
         xml.enabled true
@@ -30,7 +39,8 @@ task mockitoCoverage(type: JacocoReport, dependsOn: ["test", "kotlinTest:test", 
     }
 
     doLast {
-        println "Jacoco report for client created: file://${reports.xml.destination.toURI().path}"
+        logger.lifecycle "Jacoco report for client created: file://${reports.xml.destination.toURI().path}"
+        logger.lifecycle "Jacoco HTML created: file://${new File(reports.html.destination, "index.html").toURI().path}"
     }
 }
 
--- a/gradle/root/coverage.gradle
+++ b/gradle/root/coverage.gradle
@@ -39,7 +39,6 @@ task mockitoCoverage(type: JacocoReport) {
     }
 
     doLast {
-        logger.lifecycle "Jacoco report for client created: file://${reports.xml.destination.toURI().path}"
         logger.lifecycle "Jacoco HTML created: file://${new File(reports.html.destination, "index.html").toURI().path}"
     }
 }
--- a/subprojects/extTest/extTest.gradle
+++ b/subprojects/extTest/extTest.gradle
@@ -25,7 +25,3 @@ configurations.all {
     //let's make those tests not use hamcrest
     //exclude group: 'org.hamcrest', module: 'hamcrest-core'
 }
-
-jacoco {
-    toolVersion = libraries.jacocoVersion
-}
--- a/subprojects/extTest/extTest.gradle
+++ b/subprojects/extTest/extTest.gradle
@@ -1,3 +1,5 @@
+apply from: "$rootDir/gradle/dependencies.gradle"
+
 apply plugin: 'java'
 description = "End-to-end tests for Mockito and its extensions."
 
@@ -22,4 +24,8 @@ configurations.all {
     //TODO SF enable when #154 is implemented
     //let's make those tests not use hamcrest
     //exclude group: 'org.hamcrest', module: 'hamcrest-core'
+}
+
+jacoco {
+    toolVersion = libraries.jacocoVersion
 }
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -18,5 +18,3 @@ libraries.bytebuddyandroid = "net.bytebuddy:byte-buddy-android:${versions.bytebu
 libraries.objenesis = 'org.objenesis:objenesis:2.6'
 
 libraries.asm = 'org.ow2.asm:asm:6.0'
-
-libraries.jacocoVersion = '0.7.9'
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -18,3 +18,5 @@ libraries.bytebuddyandroid = "net.bytebuddy:byte-buddy-android:${versions.bytebu
 libraries.objenesis = 'org.objenesis:objenesis:2.6'
 
 libraries.asm = 'org.ow2.asm:asm:6.0'
+
+libraries.jacocoVersion = '0.7.9'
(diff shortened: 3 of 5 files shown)


gradle/root/coverage.gradle
```diff
@@ -1,16 +1,19 @@
...
 
 jacoco {
-    toolVersion = "0.7.9"
+    toolVersion = libraries.jacocoVersion
 }
 
-task mockitoCoverage(type: JacocoReport, dependsOn: "test") {
-    executionData files("${buildDir}/jacoco/test.exec")
+task mockitoCoverage(type: JacocoReport, dependsOn: ["test", "kotlinTest:test", "extTest:test", "inline:test"]) {
+    executionData fileTree(project.rootDir.absolutePath).include("**/build/jacoco/*.exec")
```
szpak: Maybe it would be better to iterate over subprojects and depend on "test" task (by a type or by a name) and use that list to generate also files to include (istead of looking for `*.exec` file)?

It would be even less fragile to have an optional exclusions for specific modules if needed.

subprojects/extTest/extTest.gradle
```diff
@@ -23,3 +25,7 @@ configurations.all {
     //let's make those tests not use hamcrest
     //exclude group: 'org.hamcrest', module: 'hamcrest-core'
 }
+
+jacoco {
+    toolVersion = libraries.jacocoVersion
```
szpak: Do we really need it to be defined in every module separately? Maybe you could use the list of subprojects (I proposed above) to do it in a more generic way?

gradle/root/coverage.gradle
TimvdLippe: I tried to make that work, but ran into problems. First of all, not every subproject has this test task. Therefore, I had to somehow filter the correct project. However, if you do something like below Gradle will throw an exception:

```groovy
subprojects {
  if (it.getTaskByName('test')
    dependsOn "${it.path}:test"
}
```

My Groovy is lackluster and I just couldn't get this to work.

subprojects/extTest/extTest.gradle
TimvdLippe: Yes, because only this task will generate the correct `*.exec` files in `build/` which then have to be picked up by the root project. If you have a better solution, feel free to push to this PR!

szpak: I will try to tune that tomorrow. Stay tuned.

gradle/root/coverage.gradle
```diff
@@ -1,16 +1,19 @@
 apply plugin: 'jacoco'
-
-// TODO later include subprojects, possible ideas https://gist.github.com/aalmiray/e6f54aa4b3803be0bcac
+apply from: 'gradle/dependencies.gradle'
 
 jacoco {
-    toolVersion = "0.7.9"
+    toolVersion = libraries.jacocoVersion
```
mockitoguy: I'm curious why we need to specify the tool version. Can you get rid of the version and see if it works? If we bump Gradle, we generally want the best jacoco version that works with given Gradle version. Let's try if we can rely on the default.

Not blocking for this PR.

gradle/root/coverage.gradle
```diff
@@ -1,16 +1,19 @@
...
 
-task mockitoCoverage(type: JacocoReport, dependsOn: "test") {
-    executionData files("${buildDir}/jacoco/test.exec")
+task mockitoCoverage(type: JacocoReport, dependsOn: ["test", "kotlinTest:test", "extTest:test", "inline:test"]) {
+    executionData fileTree(project.rootDir.absolutePath).include("**/build/jacoco/*.exec")
 
     sourceSets sourceSets.main
 
+    subprojects {
+        sourceSets it.sourceSets.main
```
mockitoguy: Does this work? I suspect we can run into problems because we cannot guarantee that at this point subproject has any source sets (the sequence / order in which Gradle configures projects).

gradle/root/coverage.gradle
```diff
@@ -1,16 +1,19 @@
...
 }
 
-task mockitoCoverage(type: JacocoReport, dependsOn: "test") {
-    executionData files("${buildDir}/jacoco/test.exec")
+task mockitoCoverage(type: JacocoReport, dependsOn: ["test", "kotlinTest:test", "extTest:test", "inline:test"]) {
+    executionData fileTree(project.rootDir.absolutePath).include("**/build/jacoco/*.exec")
 
     sourceSets sourceSets.main
 
+    subprojects {
```
mockitoguy: I looked at the current "mockitoCoverage" implementation and I think that the change is a bit more involved. For example, we should also include ```classDirectories``` from the subprojects.

subprojects/extTest/extTest.gradle
mockitoguy: Let's avoid code duplication. I'm sure @szpak will figure this out.

gradle/root/coverage.gradle
mockitoguy: Let's avoid hardcoding the test task names and coupling with subprojects. 

Something like that should work (please note that I have not tested it :):

```Gradle
task mockitoCoverage(...)

allprojects {
  plugins.withId("java") {
    apply plugin: "jacoco"
    //configure jacoco version, etc.

    mockitoCoverage.dependsOn "test"
    //configure other parts of jacoco plugin or the mockitoCoverage task
  }
}
```

gradle/root/coverage.gradle
szpak: Gradle usually has outdated version of JaCoCo by default (see https://github.com/gradle/gradle/commit/987d86f36a05b9090ab498dfbe77fe3dbc0f8e39 - 6 months after the release of 0.7.9 - thanks to the PR from the JaCoCo committer). In addition Mockito tended (in the past) to do not use the newest Gradle version.
Therefore, it's redundant right now, but it can be needed once 0.7.10 is released. As a result, having it just one place would be probably enough to be able to uncomment one line in `coverage.gradle`.

subprojects/extTest/extTest.gradle
szpak: Sure, if not I will ask the Gradle 2.x core developer for help :).

WiP.

szpak: It was more tricky and took much more time that I expected... I did it in 3 somehow different ways and in the end I left the simplest version (with one file gathering all coverage data). Feel free to comment and to enhance.

gradle/root/coverage.gradle
```diff
@@ -1,15 +1,27 @@
...
-    toolVersion = "0.7.9"
-}
+            apply plugin: "jacoco"
+
+            jacoco {
+                toolVersion = '0.7.9' //Gradle 4.2.1 uses 0.7.8 by default
 
-task mockitoCoverage(type: JacocoReport, dependsOn: "test") {
-    executionData files("${buildDir}/jacoco/test.exec")
+                if (currentProject != rootProject) {   //already automatically enhanced in mockito main project as "java" plugin was applied before
```
szpak: @mockitoguy Do you have a good idea how to get rid of it without changing the main Gradle configuration in Mockito to much?

Subtasks doesn't have `Test` tasks enhances automatically due to the applying order.

szpak:
Btw, in the end the gain seems to be negligible. Just one 1 percent point higher coverage if submodules are taken into account. 
screenshot from 2017-10-31 01-24-18

However, the configuration itself seems to work ok - there is ~28% coverage with just submodules.
screenshot from 2017-10-31 01-24-02

gradle/root/coverage.gradle
mockitoguy: Not sure. I don't fully see the problem with ordering.

gradle/root/coverage.gradle
```diff
@@ -1,15 +1,27 @@
-apply plugin: 'jacoco'
+//TODO: Standard JaCoCo coverage doesn't work in Android module
+task mockitoCoverage(type: JacocoReport) {
```
mockitoguy: It looks somewhat confusing that "allprojects" section is within the body of the "mockitoCoverage" task configuration.

gradle/root/coverage.gradle
```diff
@@ -27,8 +39,9 @@ task mockitoCoverage(type: JacocoReport, dependsOn: "test") {
     }
 
     doLast {
-        println "Jacoco report for client created: file://${reports.xml.destination.toURI().path}"
+        logger.lifecycle "Jacoco report for client created: file://${reports.xml.destination.toURI().path}"
```
mockitoguy: Let's get rid of "for client" - I don't even know what it means :)

gradle/root/coverage.gradle
```diff
@@ -27,8 +39,9 @@ task mockitoCoverage(type: JacocoReport, dependsOn: "test") {
...
+        logger.lifecycle "Jacoco HTML created: file://${new File(reports.html.destination, "index.html").toURI().path}"
```
mockitoguy: Does it actually generate html and xml? In the past the task could only generate only one kind of report.

subprojects/extTest/extTest.gradle
```diff
@@ -1,3 +1,5 @@
+apply from: "$rootDir/gradle/dependencies.gradle"
```
mockitoguy: Is this line needed?

mockitoguy:
Nice. Please consider my feedback before merging!

gradle/root/coverage.gradle
szpak: It's caused by the fact that the root project in Mockito is not an empty parent project, but mockito-core in fact (which also generates coverage data).

gradle/root/coverage.gradle
szpak: According [documentation](https://docs.gradle.org/current/userguide/jacoco_plugin.html#sec:jacoco_specific_task_configuration):
> The JaCoCo plugin adds a JacocoTaskExtension extension to all tasks of type Test. This extension allows the configuration of the JaCoCo specific properties of the test task.

It doesn't happen for subprojects here. I needed to call "applyTo" manually. I didn't see it in multiprojects with "normal" empty parent/root project.

gradle/root/coverage.gradle
szpak: Yes, it is (it's enabled in the configuration)
.
For me the HTML link is enough. I don't know why the the link do XML was introduced. @TimvdLippe do you remember?

subprojects/extTest/extTest.gradle
szpak: Rather yes. There is:
```
dependencies {
    testCompile "org.mockito:mockito-core:$version"
    testCompile project(path: ':', configuration: 'testUtil')
    testCompile libraries.junit4
    testCompile libraries.assertj
}
```
in that file below.

gradle/root/coverage.gradle
TimvdLippe: That is used by CodeCov afaik

szpak: Yes, but do they use also a link? :)

TimvdLippe: Oh in that manner. I doubt it, but does it harm by reporting it?

gradle/root/coverage.gradle
```diff
@@ -1,15 +1,27 @@
-apply plugin: 'jacoco'
+//TODO: Standard JaCoCo coverage doesn't work in Android module
+task mockitoCoverage(type: JacocoReport) {
 
-// TODO later include subprojects, possible ideas https://gist.github.com/aalmiray/e6f54aa4b3803be0bcac
+    allprojects { currentProject ->
+        plugins.withId("java") {
```
TimvdLippe: Should we also take into account the Kotlin projects, as they are also generating coverage?

szpak: The groovy/kotlin plugins applies the `java` plugin under the hood.

TimvdLippe: Ah okay :+1:

TimvdLippe:
Merging this branch. Thanks for the help @szpak !

PR #1235, opened 2017-10-31
Experimental flexible matching of varargs

mockitoguy:
It is a prototype/hack. Not intended for check-in but demonstrates that we can match varargs flexibly.

See #1224

Can we follow down this path and fix problem #1222 and #584 in the way the API is most convenient to use?

Interested in helping out? Can you pull down the "varargs-experiment" branch and add tests to FlexibleVarargsTest class so that we can see the gaps?

### DIFF
--- a/src/main/java/org/mockito/internal/invocation/MatcherApplicationStrategy.java
+++ b/src/main/java/org/mockito/internal/invocation/MatcherApplicationStrategy.java
@@ -4,25 +4,24 @@
  */
 package org.mockito.internal.invocation;
 
-import static org.mockito.internal.invocation.MatcherApplicationStrategy.MatcherApplicationType.ERROR_UNSUPPORTED_NUMBER_OF_MATCHERS;
-import static org.mockito.internal.invocation.MatcherApplicationStrategy.MatcherApplicationType.MATCH_EACH_VARARGS_WITH_LAST_MATCHER;
-import static org.mockito.internal.invocation.MatcherApplicationStrategy.MatcherApplicationType.ONE_MATCHER_PER_ARGUMENT;
-
-import java.util.ArrayList;
-import java.util.List;
-
 import org.mockito.ArgumentMatcher;
 import org.mockito.internal.matchers.CapturingMatcher;
 import org.mockito.internal.matchers.VarargMatcher;
 import org.mockito.invocation.Invocation;
+
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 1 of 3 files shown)


paulduffin:
The problem with this approach (which I did try) is that you cannot differentiate how to match a single argument properly. e.g. given the following two cases in the first one you want to match against the single expanded argument but in the second case you want to match against the varargs array containing a single item.

```
mock.mixedVarargs(any(), eq("1"))
mock.mixedVarargs(any(), eq(new String[] {"1"}))
```

In this case you could conceivably check to see what type the matcher was expecting but that would not work if say the varargs was Object.... In that case one call could pass a String and another could pass a String[] and if you based how you matched it on the type of the matcher then you could end up with false positives or negatives.

The only way that I found which could work in all cases is for the developer to have the ability to explicitly state "I want to match this as an array of all the varargs" at the time where they are defining their match.

mockitoguy:
>In this case you could conceivably check to see what type the matcher was expecting but that would not work if say the varargs was Object

Thank you for explaining!

>The problem with this approach (which I did try) is that you cannot differentiate how to match a single argument properly.

Brainstorming idea: what if we implement a double match? First, we attempt to match with expanded arguments. If that fails, we try matching against an array. Offering an explicit “varargsAsArray” is totally conceivable, yet it’s not the ideal user experience. Ideally, Mockito works intuitively out of the box.

Thoughts?

ChristianSchwarz:
> Brainstorming idea: what if we implement a double match? 

Won't work in null/non-null checks

`verify(mock).vararg(isNonNull()); `

This is ambiguous. Do we check if the vararg-array is non-null or the first element of the vararg-array?

mockitoguy:
> This is ambiguous. Do we check if the vararg-array is non-null or the first element of the vararg-array?

Good point. The most intuitive expectation is:

```java
mock.varargs("x");

//pass
verify(mock).varargs(ArgumentMatchers.isNotNull());
verify(mock).varargs((String) ArgumentMatchers.isNotNull());
verify(mock).varargs((String[]) ArgumentMatchers.isNotNull());

//fail
verify(mock).varargs(ArgumentMatchers.isNull());
verify(mock).varargs((String) ArgumentMatchers.isNull());
verify(mock).varargs((String[]) ArgumentMatchers.isNull());
```

I will pass my feedback to the #1224 ticket. I will close this PR for now.

PR #1236, opened 2017-10-31
Tests for:  #1235 Experimental flexible matching of varargs 

ChristianSchwarz:
Experimental flexible matching of varargs #1235

added tests that are failing


mockitoguy:
Hey Christian! Thank you for adding all those test methods. This really helps understand the problem. What do you think about https://github.com/mockito/mockito/pull/1235#issuecomment-341333168

mockitoguy:
I used those tests to understand the problem. They were very useful. Thank you for submitting this!!!

PR #1242, opened 2017-11-03
Enable Checkstyle for checking license headers

### LINKED ISSUE #1239: Enable Checkstyle rule for checking existence of license headers
Whenever we add new files to the repository, we usually forget to add our license headers. This results in PRs such as #1227 where retroactively we add the license headers.

CheckStyle has a rule called [`Header`](http://checkstyle.sourceforge.net/config_header.html) which can enforce that we add our license header. I would propose to start enabling this rule to make sure all our source files comply to our licensing.

@mockito/core yay/nay?

TimvdLippe:
Apparently we weren't shipping license headers on the Android subproject as well :cry: 

Fixes #1239

### DIFF
--- a/gradle/license.gradle
+++ b/gradle/license.gradle
@@ -10,5 +10,4 @@ license {
         groovy = 'SLASHSTAR_STYLE'
     }
     ext.year = Calendar.getInstance().get(Calendar.YEAR)
-    ext.name = 'Mockito contributors'
 }
--- a/gradle/license.gradle
+++ b/gradle/license.gradle
@@ -1,13 +1,16 @@
-apply plugin: 'com.github.hierynomus.license'
+def licenseHeaderFile = rootProject.file('doc/licenses/HEADER.txt')
 
-license {
-    header = rootProject.file('doc/licenses/HEADER.txt')
-    strictCheck = true
-    ignoreFailures = true
-    skipExistingHeaders = true
-    mapping {
-        java = 'SLASHSTAR_STYLE'
-        groovy = 'SLASHSTAR_STYLE'
+allprojects {
+    apply plugin: 'com.github.hierynomus.license'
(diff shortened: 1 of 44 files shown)


subprojects/testng/testng.gradle
```diff
@@ -1,5 +1,6 @@
 apply plugin: 'java'
 apply plugin: 'maven'
+apply from: "$rootDir/gradle/license.gradle"
```
szpak: Coudn't we apply it automatically to all projects that apply Java plugin? Similar way it is done with JaCoCo - to do not miss another module once added.

TimvdLippe: Uh, yeah maybe. Not a Gradle expert, but I can give it a shot tomorrow!

TimvdLippe: Done. Please take a look :+1:

gradle/license.gradle
```diff
@@ -0,0 +1,16 @@
+def licenseHeaderFile = rootProject.file('doc/licenses/HEADER.txt')
+
+allprojects {
+    apply plugin: 'com.github.hierynomus.license'
```
szpak: Do we use that plugin to handle any other type of files than `.java` and `.groovy` (`.kt`?)? If not, maybe it would be good to narrow the scope to only submodules with code? It can be achieved the same way as with the coverage: `plugins.withId("java") { ... }`.

TimvdLippe: Yes also any configuration or extra files. We can limit to Java projects, but I would rather check that all our files comply. Apparently we were missing quite some in total and the more we can catch, the better. WDYT?

szpak: If it's also for non source code files I'm ok with `allprojects` unconditionally.

PR #1245, opened 2017-11-03
Fixes #1175: thenCallRealMethod() supports Kotlin interface function with body

### LINKED ISSUE #1175: [Kotlin] thenCallRealMethod() does not work for default functions
Mockito has supported applying `thenCallRealMethod()` for Java 8 default methods.
However, the feature does not work in Kotlin.

The Kotlin compiler generates separate classes for an interface with default functions.

In Kotlin, the file

```kotlin
interface MyInterface {
    fun foo(s: String): String = "Foo $s"
}

class MyClass : MyInterface
```

is compiled like the following code:

```java
public interface MyInterface {

    String foo(String s);

    final class DefaultImpls {
        public static String foo(MyInterface thisObj, String s) {
            return "Foo " + s;
        }
    }

}

public final class MyClass implements MyInterface {
    public String foo(String s) {
        return DefaultImpls.foo(this, s);
    }
}
```

Kotlin 1.1.4-2 (current stable version) generates such bytecode even if the `-jvm-target` option is specified as `1.8`.

tmurakami:
In Kotlin, the default implementation of an interface function is defined in `DefaultImpls` class member of that interface.
This PR fixes so that `CallsRealMethods` calls methods on that class.

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodInterceptor.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodInterceptor.java
@@ -19,6 +19,7 @@ import org.mockito.internal.invocation.MockitoMethod;
 import org.mockito.internal.invocation.RealMethod;
 import org.mockito.internal.invocation.SerializableMethod;
 import org.mockito.internal.progress.SequenceNumber;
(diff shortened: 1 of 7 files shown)


TimvdLippe:

PR #1247, opened 2017-11-04
Remove several container classes and inline code

### LINKED ISSUE #904: Remove unnecessary classes

TimvdLippe:
Remove several classes that were containers for code that is easier to be inlined. They were only referenced once, or never at all. This makes it easier to reason about the code, while also maintaining less code and classes.

Closes #904


TimvdLippe:
Since no one reviewed this PR for a while, I assume it is safe to be merged.

mockitoguy:
Sorry nobody picked up the work! You can try to ping us for feedback. I suspect that because it was a safely looking internal refactoring we put a low priority on this. In the future, let's try to solicit approval for PRs :)

PR #1263, opened 2017-11-18
Update Kotlin to latest version

TimvdLippe:
Bump Kotlin to the latest version. Also extract the version to our dependencies so we can more easily keep track of it.

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -18,3 +18,8 @@ libraries.bytebuddyandroid = "net.bytebuddy:byte-buddy-android:${versions.bytebu
 libraries.objenesis = 'org.objenesis:objenesis:2.6'
 
 libraries.asm = 'org.ow2.asm:asm:6.0'
+
+libraries.kotlin = [:]
+libraries.kotlin.version = '1.1.60'
+libraries.kotlin.stdlib = "org.jetbrains.kotlin:kotlin-stdlib:${libraries.kotlin.version}"
+libraries.kotlin.coroutines = 'org.jetbrains.kotlinx:kotlinx-coroutines-core:0.19.3'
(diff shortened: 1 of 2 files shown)


gradle/dependencies.gradle
```diff
@@ -18,3 +18,8 @@ libraries.bytebuddyandroid = "net.bytebuddy:byte-buddy-android:${versions.bytebu
 libraries.objenesis = 'org.objenesis:objenesis:2.6'
 
 libraries.asm = 'org.ow2.asm:asm:6.0'
+
+libraries.kotlin = [:]
```
mockitoguy: To reduce the duplication, you just do:

```
libraries.kotlin = [
  stdlib: "...",
  ...
]
```

subprojects/kotlinTest/kotlinTest.gradle
```diff
@@ -1,5 +1,6 @@
 plugins {
-    id 'org.jetbrains.kotlin.jvm' version '1.1.4-3'
+    // Can not use libraries.kotlin.version here per https://stackoverflow.com/questions/37555196/in-gradle-how-to-use-a-variable-for-a-plugin-version
+    id 'org.jetbrains.kotlin.jvm' version '1.1.60'
```
mockitoguy: We could probably wrap the duplication if we used traditional way of declaring plugins, via "buildscript" section. See the main "build.gradle" file if you are interested. I think it is a good idea to avoid duplication but I'm not too concerned.

subprojects/kotlinTest/kotlinTest.gradle
```diff
@@ -15,6 +16,6 @@ dependencies {
     testCompile project(":")
     testCompile libraries.junit4
 
-    testCompile 'org.jetbrains.kotlin:kotlin-stdlib:1.1.4-3'
-    testCompile 'org.jetbrains.kotlinx:kotlinx-coroutines-core:0.18'
+    testCompile libraries.kotlin.stdlib
```
mockitoguy: Reads super nice! Thank you!

gradle/dependencies.gradle
TimvdLippe: Hm, I edited online on GitHub, but it seems to be broken now 😢

Issue #1255, opened 2017-11-10
ArgumentMatchers support for @NonNull / Kotlin non-null


olivierg13:
Hello folks,

There is an issue when using ArgumentsMatcher.eq() method with either:
- A Java @NonNull annotated param 
- A Kotlin non-null param

Example (in kotlin):
- Let's say you have a method:
`doSomething(test: String) {...}`
- When trying to do:
`Mockito.verify(mock).doSomething(ArgumentMatchers.eq("toto"))`

At runtime the test is going to fail with message:
`java.lang.IllegalStateException: ArgumentMatchers.eq("toto") must not be null`

Is there a way to return NonNull objects from the eq method?

Thank you in advance :)

TimvdLippe:
I am going to publish a prototype soon that can probably address this issue.

lemartin:
The `eq` problem can be addressed quite easily. A different name was only chose to not get into import hell.
```kotlin
fun <T : Any> safeEq(value: T): T = eq(value) ?: value
```

ChristianSchwarz:
This can be fixed in mockito by changing

```java
public static <T> T eq(T value) {
        reportMatcher(new Equals(value));
        if (value == null)
            return null;
        return (T) Primitives.defaultValue(value.getClass());
}
```

to

```java
public static <T> T eq(T value) {
        reportMatcher(new Equals(value));
        return value;
}
```

mockitoguy:
I really like @ChristianSchwarz's idea because it simplifies the code. Problem is that it is a breaking change because there could be client code depending on the value returned from ```eq```. When working with test spies, the real behavior is invoked during stubbing with ```when``` syntax. That real behavior may depend on parameters. Example:

```Java
Foo spy = spy(Foo.class);
when(spy.foo(eq("blah")).thenReturn(...);
```

I like the idea of including @ChristianSchwarz's change in next major version of Mockito.

giovannicandido:
`any` and `isA` have the same problem

ChristianSchwarz:
You might want to take a look at https://github.com/nhaarman/mockito-kotlin, they solved the null-return problem and other's too, like the "when"-keyword issue.

giovannicandido:
Thanks @ChristianSchwarz,

I'm using mockito-kotlin to work around this issue.

MRezaNasirloo:
Same for `Mockito.argThat`

capezzbr:
How is still an outstanding issue?

[...]

Issue #1237, opened 2017-11-02
Re-stubbing from exceptional to non exceptional doesn't work


baztian:
Hi,

I want to re-stub a method that already has a default behaviour in the @Before method. The default behaviour is to throw an exception (CustomerNotFound) to simulate the behaviour on a newly setup instance of my system for any customer number. Of course I need tests describing the behaviour for existing customers. Unfortunately if I override the stubbing for some (or any) customer numbers the exception gets throws during stubbing phase.

I've written a minimal test case to show the problem. The second test `testOverrideWithNonExceptionalCase` fails.

    package mockito;
    
    import static org.hamcrest.core.Is.is;
    import static org.junit.Assert.assertThat;
    import static org.mockito.Mockito.when;
    
    import java.io.IOException;
    import java.io.InputStream;
    import org.junit.Rule;
    import org.junit.Test;
    import org.junit.rules.ExpectedException;
    import org.junit.runner.RunWith;
    import org.mockito.Mock;
    import org.mockito.junit.MockitoJUnitRunner;
    
    @RunWith(MockitoJUnitRunner.class)
    public class StubbingOverrideTest {
    
      @Mock
      private InputStream is;
    
      @Rule
      public final ExpectedException exception = ExpectedException.none();
    
      @Test
      public void testOverrideWithExceptionalCase() throws Exception {
        when(is.read()).thenReturn(42);
        when(is.read()).thenThrow(new IOException());
    
        exception.expect(IOException.class);
        is.read();
      }
    
      @Test
      public void testOverrideWithNonExceptionalCase() throws Exception {
        when(is.read()).thenThrow(new IOException());
        when(is.read()).thenReturn(42);
    
        assertThat(is.read(), is(42));
      }
    }

When running the test I get the following exception:
    
    java.io.IOException
    	at mockito.StubbingOverrideTest.testOverrideWithNonExceptionalCase(StubbingOverrideTest.java:37)
    	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    	at java.lang.reflect.Method.invoke(Method.java:498)
    	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
[...]

Issue #1174, opened 2017-08-25
java.lang.ClassCastException with generic types


shark300:
Hi,

I've tried this:

```
@Test
public void test() {
    given(testCollectionSourceProvider.getCollection(new ArrayList<Integer>())).willReturn(new ArrayList<>());
}

static class TestCollectionSourceProvider {
    <T extends Collection<E>, E> T getCollection(T collection) {
        return collection;
    }
}
```

But I've got an Exception with a very confusing message in Idea 2017.1.5 and Eclipse Neon.3:

```
java.lang.ClassCastException: java.util.LinkedList cannot be cast to java.util.ArrayList
	at com.example.demo.FailingTestWithCollectionsTest.test(FailingTestWithCollectionsTest.java:24)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```
But it works from Maven.

If I cast it manually, it works from IDEs:

```
given((List<Integer>) testCollectionSourceProvider.getCollection(new ArrayList<Integer>())).willReturn(new ArrayList<>());
```

Java versions:

1. machine
> java version "1.8.0_144"
> Java(TM) SE Runtime Environment (build 1.8.0_144-b01)
> Java HotSpot(TM) 64-Bit Server VM (build 25.144-b01, mixed mode)

> IntelliJ IDEA 2017.1.5
> Build #IC-171.4694.70, built on July 4, 2017
> JRE: 1.8.0_112-release-736-b21 amd64
> JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
> Windows 10 10.0

> Apache Maven 3.5.0 (ff8f5e7444045639af65f6095c62210b5713f426; 2017-04-03T21:39:06+02:00)
> Maven home: C:\Maven
> Java version: 1.8.0_131, vendor: Oracle Corporation
> Java home: C:\Program Files\Java\jdk1.8\jre
> Default locale: en_US, platform encoding: Cp1252
> OS name: "windows 10", version: "10.0", arch: "amd64", family: "windows"

2. machine
> java version "1.8.0_141"
> Java(TM) SE Runtime Environment (build 1.8.0_141-b01)
> Java HotSpot(TM) 64-Bit Server VM (build 25.141-b01, mixed mode)

> Platform: Eclipse Neon.3 (4.6.3)
> Version: 3.9.0.RELEASE
> Build Id: 201707061903
> Windows 7

Please check it: https://github.com/shark300/ClassNotFoundException-in-Mockito

[...]

Issue #899, opened 2017-01-22
Matcher Support for Kotlin


ihanjo:
I've recently started using Kotlin and am loving it so far. I've however had to use some workarounds for Mockito matchers, since Kotlin doesn't like null -- a great thing IMO. See http://stackoverflow.com/q/30305217 

There is a library out there called [MockitoKotlin](https://github.com/nhaarman/mockito-kotlin) that provides matchers that do work with Kotlin, but I'd rather use Mockito directly since that way I can have access to the latest features and patches by official Mockito.

What do you guys think? Is this a reasonable ask? I know Kotlin isn't Java, but it is gaining popularity. :)

TimvdLippe:
I think this usecase can be safely supported with the new lambda when syntax and implementation. Will try it out!

ihanjo:
@TimvdLippe Do you think it's a lot of work to support it? What do you envision this snippet would look like with the lambda when syntax?

```java
when(mockGreeter.welcome(eq("John"))).thenReturn("Welcome, John!");
```

TimvdLippe:
@ihanjo There is an open proposal PR at #742 . The syntax I imagine will be

```java
when(mockGreeter::welcome).invokedWith(eq("John")).thenReturn("Welcome, John!");
```

Note this is typesafe and Kotlin-friendly because passing the matcher is decoupled from the method invocation.

bric3:
I'm not sure this should be part of mockito yet. As matchers will create new mocks for types without default values.

PR #1264, opened 2017-11-18
Working prototype of Mockito lambda

TimvdLippe:
This PR contains a working prototype of Mockito lambda.

# What is this lambda thing about?

This prototype shows a design of our API based on lambdas. Currently, the invocation of a method is controlled by the user. E.g. in `when(mock.method())`, the user controls the point of invocation. In turn, Mockito tries it best to deduce which method it was and what it has to do when `when` (or `verify`) is handled.

# Why this change?

Historically, we have gathered a wide variety of (fundamental) issues that we are not able to solve. This is mostly in relation the exact problem of user-controlled vs Mockito-controlled invocation. In some cases, the user has to invoke the method, but this results in unwanted side-effects.

Previously, it was not possible to change this behavior, it was simply the only option. However, with the introduction of lambdas with JDK8, we do have an alternative.

# How does it work?

Instead of the user invoking the method directly, Mockito will invoke the method. To do so, Mockito requires the lambda that references the method. This is based on the principle of [section 15.13](https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.13) of the Java specification which explains how `Method Reference Expressions` work.

This prototype gathers all necessary data (e.g. matchers, returnvalue, answer) and only then performs the invocation.

# Consequences of this approach

There are several (positive) consequences to this approach:

1. We no  longer have static state across interactions with the Mockito API. E.g. `when(mock.method())` will no longer leave behind corrupt state.
2. We can handle any kind of lambda, which means the user has more freedom on how the invocation is performed (opens opportunity for static method mocking if we desire so)
3. Creating custom answers is easier and type-safe. This means we no longer need wrapping with `AdditionalAnswers` of http://static.javadoc.io/org.mockito/mockito-core/2.12.0/org/mockito/Mockito.html#37
4. Creation of Matchers is simplified, because we no longer need to supply our own (best guess) value. E.g. right now, when we construct a matcher, we store it statically and then `return` our best guess. This results in problems with generics, auto-boxing and non-nullness languages like Kotlin
5. Per above point, this prototype has first-class Kotlin support (check [this test suite](https://github.com/mockito/mockito/blob/mockito-lambda/subprojects/lambda/src/test/java/org/mockitousage/MockitoLambdaKotlinTest.kt)) and users would no longer have to resort to https://github.com/nhaarman/mockito-kotlin/ The solution to this problem is based on the approach that @nhaarman took in that artifact.

# Full list of fundamental issues
The following issues can be solved with this approach (sorted on date opened):

#308: This is more related to the problematic creation of matchers. This means that we no longer need to force ship Matchers for specific types (like [`anyCollection`](http://static.javadoc.io/org.mockito/mockito-core/2.12.0/org/mockito/ArgumentMatchers.html#anyCollection--)). Shown in [this test](https://github.com/mockito/mockito/blob/a7839ddbb2746c4f528bf6203dca05ece32b7bda/subprojects/lambda/src/test/java/org/mockitousage/MockitoLambdaStubTest.java#L171-L181)

#357: Matchers can not deal with Generics, meaning that our best guess violates check casts. This is problematic on for example JDK9. Matchers can now control the `returnValue`, which means users can control it themselves. However, for this case, no custom class is necessary as [shown in these tests](https://github.com/mockito/mockito/blob/a7839ddbb2746c4f528bf6203dca05ece32b7bda/subprojects/lambda/src/test/java/org/mockitousage/MockitoLambdaStubTest.java#L147-L169)

#500: Because the user controls the invocation, we don't know for a spy that the invocation was not meant to be tracked. This approach requires no extra mechanics, it happens naturally, see [this test](https://github.com/mockito/mockito/blob/a7839ddbb2746c4f528bf6203dca05ece32b7bda/subprojects/lambda/src/test/java/org/mockitousage/MockitoLambdaSpyTest.java#L36-L49)

#899: Kotlin can be first-class supported and an extra artifact is no longer necessary. Check out [this test class](https://github.com/mockito/mockito/blob/mockito-lambda/subprojects/lambda/src/test/java/org/mockitousage/MockitoLambdaKotlinTest.kt)

#1174: Similar to #357, this can be solved by simply specifying the types. See [this test](https://github.com/mockito/mockito/blob/a7839ddbb2746c4f528bf6203dca05ece32b7bda/subprojects/lambda/src/test/java/org/mockitousage/MockitoLambdaStubTest.java#L129-L136)

#1237: This is a perfect showcase of the user-controlled vs Mockito-controlled approach. This issue is naturally resolved without any additional machinery: see [this test](https://github.com/mockito/mockito/blob/a7839ddbb2746c4f528bf6203dca05ece32b7bda/subprojects/lambda/src/test/java/org/mockitousage/MockitoLambdaStubTest.java#L77-L87)

#1255: As shown, Kotlin is supported and also works with non-nullness parameters, see [this test](https://github.com/mockito/mockito/blob/a7839ddbb2746c4f528bf6203dca05ece32b7bda/subprojects/lambda/src/test/java/org/mockitousage/MockitoLambdaVerificationTest.java#L39-L47) as well

#1279: Since we control the invocation, the static state issue with calls to stubs no longer exists. Therefore calls to stubs in answers are no longer an issue.

#1515: Expose Matchers rather than internally `reporting`. This greatly reduces the need of users to rely on our internal implementations.

# What now?

This PR is not meant to be merged; this is a prototype and far from final. It has no documentation, is not feature complete, but does have a lot of tests to show how it all works together. Instead, this PR is meant for starting the discussion about this approach.

Most of the implementation is boiler-plate, as our API is quite extensive. If you would like to review the implementation, I would suggest to check out [`MockitoLambdaHandlerImpl`](https://github.com/mockito/mockito/blob/mockito-lambda/subprojects/lambda/src/main/java/org/mockito/MockitoLambdaHandlerImpl.java) first. All other classes simply define methods that are only 1-2 lines long.

I would like to emphasize that the syntax of this API is by all means not final. I have already heard developers disliking this specific syntax. As this is a prototype, I have not focused at all on final API design. Instead, I would like to ask everyone to first consider the concept of using lambdas as means of conveying stubs/verifications, later we can do that full API design.


mockitoguy:
Thank you for filing this PR with amazing amount of detail. I will read it and give feedback!

TimvdLippe:
@mockito/core Friendly ping. Would like to hear your thoughts on the lambda-based stubbing approach 😄

mockitoguy:
Quick feedback:
 - very interesting! solves handful of Mockito caveats!
 - feels too different from the ubiquitous Mockito
 - not as readable (for some use cases, a lot less readable)
 - I'm pondering about: a) replace existing syntax with it b) offer it as an option c) do nothing / separate framework. My current stance in order of preference: c, b, a.
 - I couldn't get this to work (I'm getting compiler errors):
   - stubbing void methods
   - stubbing overloaded methods
 - it would be great to see example API in the PR description

I think we should push forward to see it fully working. We will learn how / if we can address all the issues. On the way we will be getting to a decision.

I'm not yet convinced this is the future of Mockito but this is an amazing work nonetheless. It's first time since the inception of Mockito in 2007 that we are innovating on the side of syntax. Thank you!!!

TimvdLippe:
@mockitoguy Thanks for your feedback! As I pointed out earlier, I wanted to avoid the syntax discussion and focus on the usage of lambdas, but alas.

I would approach this similarly to `mockito-inline`. It's incubating and advised in specific cases, for hard/next-to-impossible issues.

You can see examples of this API in https://github.com/mockito/mockito/blob/mockito-lambda/subprojects/lambda/src/test/java/org/mockitousage/MockitoLambdaStubTest.java

RE void methods: Consumers should be supported, see https://github.com/mockito/mockito/blob/mockito-lambda/subprojects/lambda/src/test/java/org/mockitousage/MockitoLambdaStubTest.java#L183-L190
RE overloaded methods: Yes Java method reference is unfriendly with overloaded methods. See for an example https://github.com/mockito/mockito/blob/mockito-lambda/subprojects/lambda/src/test/java/org/mockitousage/MockitoLambdaStubTest.java#L166 I am not sure how many times users actually run into this problem.

For now, only `Supplier`, `Function` and `Consumer` are supported: https://github.com/mockito/mockito/blob/mockito-lambda/subprojects/lambda/src/main/java/org/mockito/MockitoLambda.java#L43-L53 I do not want to put a lot of effort in supporting all possible cases to then get it shot down. Therefore this prototype should give a fairly decent viewpoint on the feasibility.

Otacon:
RESOLVE THE CONFLICTS!!!

mikehc:
Any clue if this is going to get merged any time soon?

fduminy:
This PR sounds interesting. 

Any news since 2018 ?

TimvdLippe:
This approach has a fundamental problem, in that it relies on overload choice based on lambda types. This is typically brittle and discouraged. Therefore, I am closing this PR for now 😢

Marcono1234:
Would another option be to use a similar approach to `MockedStatic#when`, that is, let the users specify a lambda expression which calls the method to stub. For example:
```java
when(() -> myMock.doSomething(anyString())).thenReturn("result);
```

Then Mockito knows that whatever method call is happening inside that lambda is only performing stubbing and is not a real method call.

The API could look like this:
```java
@FunctionalInterface
interface StubbingRecorder<T> {
    T recordStubbing() throws Throwable;
}

class Mockito {
    public static <T> OngoingStubbing<T> when(StubbingRecorder<T> s) {
        ...
    }

    ...
}
```
(the API names might not be that ideal; this is just a rough draft)

Though maybe it should be a differently named method (e.g. `whenL`) to not cause the compiler to pick the wrong overload.

<details>

<summary>Details about overload issues (click to expand)</summary>

Let's imaging this simplified Mockito API:
```java
interface StubbingRecorder<T> {
    T recordStubbing() throws Throwable;
}

class Mockito {
    // New method
    public static <T> void when(StubbingRecorder<T> s) {
    }
    
    // Existing `Mockito.when`
    public static <T> void when(T t) {
    }
}
```

And this test code:
```java
<T> T getValue() {
    return (T) (Object) "test";
}

// Fails with 'ClassCastException: class java.lang.String cannot be cast to class StubbingRecorder'
Mockito.when(getValue());
```

This is maybe a bit contrived, but it shows that the compiler might pick the wrong overload for existing code using Mockito.

</details>

----

Another approach could be to have a `<T> T mocked(T mock)` method or similar which wraps access to the mock and indicates that the next call is stubbing a method, for example:
```java
when(mocked(myMock).doSomething(anyString()).thenReturn("result);
```
Though this might be a bit verbose.

I am also not sure if that would solve all of the above mentioned issues.

----

What do you think? Sorry if these approaches have been suggested before, or if they would not solve the majority of the issues.

PR #1268, opened 2017-11-27
Fixes #1262: update doc for Answers.CALLS_REAL_METHODS

### LINKED ISSUE #1262: Documentation mistake
I think the documentation here is wrong:
https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/Mockito.java#L1646 

I believe the doReturn() syntax should be used there instead. At least when I do that it works but when I do the when(myFoo.myMethod()).thenReturn() that line itself runs the real impl of myMethod.

SeriyBg:
Update the documentation for the Answers.CALLS_REAL_METHODS according to #1262 and also included a note with recommendation not to use when().thenReturn() for partial mock.

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -1643,11 +1643,15 @@ public class Mockito extends ArgumentMatchers {
      * // this calls the real implementation of Foo.getSomething()
      * value = mock.getSomething();
      *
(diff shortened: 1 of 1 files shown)


mockitoguy:
Hey @SeriyBg, this is a nice change! Sorry we have not reviewed earlier. It must have slipped our radar. Continually improving documentation is absolutely important. Thank you for contributing!

PR #1272, opened 2017-12-11
Strictness configurable per mock / stubbing

### LINKED ISSUE #792: New lenient() strictness setting available at mock and at stubbing level
## Problems

Overview of strictness: #769

1. It is not possible to have any common stubbing when using Strictness.STRICT_STUBS. Strict stubs is very useful and it is most likely the future default for Mockito 3. Common stubbing typically is a part of "before" method or a test utility class that generates mocks.

```java
public MockitoRule rule = MockitoJUnit.rule().strictness(Strictness.STRICT_STUBS);

@Before public void before() {
  //a) common stubbing, needed for most test methods
  when(mock.foo()).thenReturn("bar");  
}
```

2. API caveats:

```java
@Test public void demo() {
  //b) not possible to stub the same method using 'when', with different args:
  when(mock.foo(1)).thenReturn(1);
  when(mock.foo(2)).thenReturn(2); // <- will throw PotentialStubbingProblem

  //c) code under test cannot use stubbed call with different argument (sometimes we need it)
  when(mock.foo(1)).thenReturn(1);
  mock.foo(2); // <- will throw PotentialStubbingProblem whether we need it or not
}
```

## Suggested solution

New public API overview:

```java
@Test public void demo() {
  //1. New method on MockSettings interface:
  Foo mock = Mockito.mock(Foo.class, withSettings().lenient());

  //2. New method on Mockito class:
  Mockito.lenient().when(mock.foo(1)).thenReturn(1);
  Mockito.lenient().doReturn(1).when(mock).foo(1);
}
```

Details:
 - Why 2 new public methods? Sometimes common stubbing has a form of one or few stubbings in "before" method - it's best to configure leniency per stubbing. Sometimes a mock object has many common stubbings (described in detail in the ticket thread below) - it's best to configure leniency per mock.

## Examples

1. Common stubbing

```java
public MockitoRule rule = MockitoJUnit.rule().strictness(Strictness.STRICT_STUBS);
@Mock(lenient = true) Foo mock;

@Before public void before() {
  //a) common stubbing, needed for most test methods
  when(mock.foo()).thenReturn("bar");  
}
```

2. API caveats:

```java
@Test public void demo() {
  //b) not possible to stub the same method using 'when', with different args:
  lenient().when(mock.foo(1)).thenReturn(1);
  when(mock.foo(2)).thenReturn(2); // <- works!

  //c) code under test cannot use stubbed call with different argument (sometimes we need it)
  lenient().when(mock.foo(1)).thenReturn(1);
  mock.foo(2); // <- works!
}
```

## Original report

The new UnnecessaryStubbingException logic is great. However, it is sometimes useful to disable this detection for specific methods. Could it be possible to add an annotation that says "unnecessary stubs should not be detected in this method"?

This would also make it easier to migrate from Mockito 1.* to Mockito 2. This is the case in the project I am currently working at, where we have created some utility methods that creates a mock and configures it to fit most of our use cases. As we use this method at many places, where different configuration is needed, it will cause a lot of unnecessary stubbing. Thus, we would like to keep this method out of unneccessary stubbing-check, while doing this check on the rest of the code base.

mockitoguy:
Fixes #792

### Problem

Today we have strict stubs feature, a key enabler for cleaner tests and improved debuggability. Strict stubs don't work well with "common stubbing" pattern. They also don't like when we call stubbed methods with unexpected arguments - even though sometimes it is legit. For more, see #792 

### Solution

```java
@Test public void demo() {
  //1. "lenient()" method on MockSettings interface:
  Foo mock = Mockito.mock(Foo.class, withSettings().lenient());

  //2. "lenient()" method on Mockito class:
  Mockito.lenient().when(mock.foo(1)).thenReturn(1);
  Mockito.lenient().doReturn(1).when(mock).foo(1);
}
```

For more, see #792

### Public API changes

- New methods:
   - Mockito#lenient()
   - MockSettings#lenient()
   - MockCreationSettings#isLenient()
   - Stubbing#getStrictness()  
- New interfaces:
   - BaseStubber - extracted out from existing "Stubber" interface, so that we can reuse API with "LenientStubber"
   - LenientStubber - instance returned by "Mockito.lenient()"
- Other changes:
   - Made @org.mockito.NotExtensible annotation public so that we can use it in subpackages


mockitoguy:
>As a cross-point with JUnit5: I would actually say that we still want method level lenient stubbing, thus these two features can co-exist in my opinion.

Can you provide a use case (test case) that demonstrates the value? You can also reply to the long email thread we had about it on Mockito mailing list :)

mockitoguy:
>Moreover, I see null used a lot of times, which is unclear to me what the usefulness is.

Can you be more specific what nulls /  are you not clear about? The code changes are light on the javadoc and comments, apologies! Thank you for review!!!

mockitoguy:
>For example, the Strictness.LENIENT is scattered throughout the code. This make it very hard to expand later on. Rather, I would add logic to strictness with methods that based on the type determine what they should do. This makes the code a lot more maintainable.

Strictness is a public type and we want to be very judicious adding methods to it. I'm very curious about your idea, though! Perhaps we can put those methods on some new internal Strictness object? Let me try to work on this and get back. Great feedback!

mockitoguy:
Hey @mockito/developers, any feedback?

@TimvdLippe, I've addressed portion of your feedback. Can you clarify the other portion (see my questions to your questions :).

src/main/java/org/mockito/internal/stubbing/InvocationContainerImpl.java
```diff
@@ -17,24 +12,32 @@
...
-    private final List<Answer<?>> answersForStubbing = new ArrayList<Answer<?>>();
+    private final DoAnswerStyleStubbing doAnswerStyleStubbing;
     private final RegisteredInvocations registeredInvocations;
+    private final Strictness mockStrictness;
 
     private MatchableInvocation invocationForStubbing;
 
     public InvocationContainerImpl(MockCreationSettings mockSettings) {
         this.registeredInvocations = createRegisteredInvocations(mockSettings);
+        this.mockStrictness = mockSettings.isLenient() ? Strictness.LENIENT : null;
```
TimvdLippe: We assign `null` here and use it throughout the code. I still think this is brittle, as it is really easy to miss. Moreover, in `addConsecutiveAnswer` `null` is passed in and in `addAnswer` again `null` is propagated. Rather, we should define a default somewhere and use this instead. That way, this field is never `null` and we always know what "not specifying a strictness for this method" actually means (e.g. fallback to `mockStrictness`)

mockitoguy: Thank you for feedback! I absolutely agree with you on the nulls.

When implementing this feature, I tried to have a default Strictness. This would require us to add new public value to the enum, for example: ```Strictness.UNSPECIFIED```. I decided against it because it's adding a new awkward value to the public API.

Do you think we should add ```Strictness.UNSPECIFIED```? What would that mean to the user? (e.g. how would we document it?).

TimvdLippe: I would add a constant static attribute in `Strictness` like `public static final Strictness DEFAULT = Strictness.LENIENT;` (or whatever value we decide on).

mockitoguy: For mock objects and stubbings, the default Strictness is "unspecified", e.g. it will be inherited from MockitoSession or JUnit rule. The question is whether we want to add ```Strictness.UNSPECIFIED``` to the public API?

TimvdLippe: I think we should extract this discussion into an issue, as it seems that this is a more fundamental issue with `Strictness`. Since we are using it more and more and it is configurable in multiple places, maybe we should extract all this logic into a separate class? In this way, you would make a chain of configurations, where we take the "most local" definition. Anyways, let's create an issue with the requirements so that we can tinker.

mockitoguy: There is already an issue: #792 - please write comment there with your idea! What do you mean about extracting it into a separate class?

TimvdLippe: @mockitoguy Filed #1305 of what I meant and to hopefully address this issue.

TimvdLippe: I have been trying to incorporate #1305 in this PR, but sadly `Strictness` is used in many places in the codebase. I feel like the Strictness feature as a whole has had more impact on our overall architecture than I would have liked. For now I am okay with these changes, but I think that we should revisit the whole Strictness story in a later stadium to see if we can reduce duplication and boilerplate code (such as the `LenientStubber`)

mockitoguy: I agree. Thank you for giving it a shot!

I'll fix the conflicts so that we're ready to merge.

src/main/java/org/mockito/internal/junit/UnusedStubbings.java
```diff
@@ -47,16 +47,19 @@ public String toString() {
...
-            });
-
+    void reportUnused() {
+        if (unused.isEmpty()) {
+            return;
+        }
 
-            Reporter.unncessaryStubbingException(invocations);
+        List<Invocation> invocations = new LinkedList<Invocation>();
+        for (Stubbing stubbing : unused) {
```
TimvdLippe: Java 8 will make this so much nicer, but this is already an improvement :+1:

src/main/java/org/mockito/internal/stubbing/DefaultLenientStubber.java
```diff
@@ -0,0 +1,68 @@
...
+package org.mockito.internal.stubbing;
+
+import org.mockito.internal.MockitoCore;
+import org.mockito.quality.Strictness;
+import org.mockito.stubbing.Answer;
+import org.mockito.stubbing.LenientStubber;
+import org.mockito.stubbing.OngoingStubbing;
+import org.mockito.stubbing.Stubber;
+
+public class DefaultLenientStubber implements LenientStubber {
```
TimvdLippe: Unfortunate that we require all this boilerplate here. Would be great to have a future improvement to get rid of that. Not blocking though.

src/main/java/org/mockito/mock/MockCreationSettings.java
```diff
@@ -109,4 +109,7 @@
      */
     @Incubating
     Object getOuterClassInstance();
+
+    @Incubating//TODO x javadoc
```
TimvdLippe: Please fix this TODO

src/main/java/org/mockito/stubbing/Stubbing.java
```diff
@@ -50,4 +52,7 @@
      * @since 2.2.3
      */
     boolean wasUsed();
+
+    @Incubating//TODO x javadoc
```
TimvdLippe: Please fix this TODO

src/test/java/org/mockito/internal/stubbing/InvocationContainerImplStubbingTest.java
```diff
@@ -46,7 +46,7 @@ public void setup() {
     public void should_finish_stubbing_when_wrong_throwable_is_set() throws Exception {
         state.stubbingStarted();
         try {
-            invocationContainerImpl.addAnswer(new ThrowsException(new Exception()));
+            invocationContainerImpl.addAnswer(new ThrowsException(new Exception()), null);
```
TimvdLippe: Would be great to overload `addAnswer`, to not having to modify all tests.

src/test/java/org/mockitousage/strictness/StrictnessPerMockTest.java
```diff
@@ -0,0 +1,117 @@
...
+import static junit.framework.TestCase.assertFalse;
+import static junit.framework.TestCase.assertTrue;
+import static org.junit.Assert.assertNull;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.mockingDetails;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.mockito.Mockito.withSettings;
+
+//TODO 792 also move other Strictness tests to this package (unless they already have good package)
```
TimvdLippe: I am not really sure what you  mean with this TODO. Is this something we need to address before or after merging this PR?

mockitoguy:
>Please address the left-over TODO's and add a new section in the Mockito.java javadoc. Then you can merge this PR 👍

Absolutely! Thank you for review!

mockitoguy:
@mpkempson, thank you for upvoting - that's what I needed! I'll work on this shortly.

src/main/java/org/mockito/internal/stubbing/DoAnswerStyleStubbing.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private final List<Answer<?>> answers = new ArrayList<Answer<?>>();
+    private Strictness stubbingStrictness;
+
+    void setAnswers(List<Answer<?>> answers, Strictness stubbingStrictness) {
+        this.stubbingStrictness = stubbingStrictness;
+        this.answers.addAll(answers);
+    }
+
+    boolean isSet() {
```
marchpig: Is there any special meaning of this method name?
I think `isEmpty()` is clear.

Issue #1279, opened 2017-12-21
Cannot override stubbed method that calls a stubbed method


r-smirnov:
Mockito 2.13.0

Overriding stubbed method that calls a stubbed method in its answer may result in a number of different failure modes:

1) WrongTypeOfReturnValue is thrown.

```java
import static org.junit.Assert.assertEquals;
import static org.mockito.Mockito.mock;
import static org.mockito.Mockito.when;

import org.junit.Test;

public class TestStubbing {
    @Test
    public void test() throws Exception {
        Foo foo = mock(Foo.class);
        Bar bar = mock(Bar.class);

        when(foo.doFoo()).thenAnswer(invocation -> String.valueOf(bar.doBar()));
        assertEquals("0", foo.doFoo());

        when(foo.doFoo()).thenReturn("");
        assertEquals("", foo.doFoo());
    }

    interface Foo {
        String doFoo();
    }

    interface Bar {
        int doBar();
    }
}
```

2) Wrong method is stubbed.

```java
import static org.junit.Assert.assertEquals;
import static org.mockito.Mockito.mock;
import static org.mockito.Mockito.when;

import org.junit.Test;

public class TestStubbing {
    @Test
    public void test() throws Exception {
        Foo foo = mock(Foo.class);
        Bar bar = mock(Bar.class);

        when(foo.doFoo()).thenAnswer(invocation -> bar.doBar());
        assertEquals(0, foo.doFoo());
        assertEquals(0, bar.doBar());

        when(foo.doFoo()).thenReturn(1);
        assertEquals(1, foo.doFoo());
        assertEquals(0, bar.doBar());
    }

    interface Foo {
        int doFoo();
    }

    interface Bar {
        int doBar();
    }
}
```

3) StackOverflowError is thrown.

```java
import static org.junit.Assert.assertEquals;
import static org.mockito.Mockito.mock;
import static org.mockito.Mockito.when;

import org.junit.Test;

public class TestStubbing {
    @Test
    public void test() throws Exception {
        Foo foo = mock(Foo.class);
        Bar bar = mock(Bar.class);

        when(foo.doFoo()).thenAnswer(invocation -> bar.doBar());
        assertEquals(0, foo.doFoo());

        when(foo.doFoo()).thenAnswer(invocation -> bar.doBar() + 1);
        assertEquals(1, foo.doFoo());
    }

    interface Foo {
        int doFoo();
    }

    interface Bar {
        int doBar();
    }
}
```

[...]

PR #1296, opened 2018-01-22
Add files via upload

r-smirnov:
Applied patch from issue #1279


r-smirnov:
Added test

mockitoguy:
@bric3 - great to see you online!!! Hope things are settling for you and you're getting more sleep :)

I'm reviewing and trying out this change now.

mockitoguy:
@r-smirnov, this is an amazing bugfix for the problem that I initially thought was not fixable. Thank you very much!!! I've opened a separate PR with your commits at #1310 (I tried to push to your repo but it was not possible because your branch got out of sync). I wanted to address the feedback from @bric3 and try out this change to make sure the solution is what we want.

@mockito/developers, please take a look at #1310 - this bugfix carries incompatible behavior change that I think is justified given circumstances. Please weight on this.

PR #1299, opened 2018-01-27
Fix link of Maven Central badge

arend-von-reinersdorff:
Link to list of mockito-core versions instead of linking to version 2.8.47

Even better would be linking to the latest version in Maven Central. But I'm not aware of a way to do this.


TimvdLippe:
Thanks!

PR #1301, opened 2018-01-28
Extend MockitoSession(Builder) API to enable usage by testing frameworks

### LINKED ISSUE #1232: [JUnit5] MockitoSession#initMocks should support multiple test instances
In order to support nested tests in JUnit5 (see #1221)  MockitoSession#initMocks should allow to pass more than one test instance or allow multiple calls on initMocks with different test instances.

In other words/code:

```
 Mockito.mockitoSession()
            .initMocks(rootTestInstance)
            .initMock(nestedTestInstance)
```

or 

```
 Mockito.mockitoSession()
            .initMocks(rootTestInstance,nestedTestInstance)
```

marcphilipp:
The first commit in this PR fixes #1232. In the additional commits, I've taken the liberty to implement my proposals from https://github.com/mockito/mockito/issues/445#issuecomment-358407562:

- Add `MockitoSessionBuilder.name(String)` to pass a name from the testing framework to the `MockitoSession`
- Add `MockitoSessionBuilder.logger(MockitoSessionLogger)` to customize the logger used for hints/warnings produced when finishing mocking (useful for testing and to connect reporting capabilities provided by testing frameworks such as JUnit Jupiter)
- Add `MockitoSession.finishMocking(Throwable)` to avoid potentially confusing warnings for failing tests
- Add `MockitoSession.setStrictness(Strictness)` to change the strictness of the current session (there were two test cases requiring that)

Each of these changes is in a separate commit. Thus, it should be relatively easy to take only those changes that you think good, or just the first one to fix #1232.

To showcase that this new API is sufficient to implement an extension of a testing framework, I've changed the `JUnitRule` implementation to use it.


src/main/java/org/mockito/internal/junit/JUnitRule.java
```diff
@@ -18,44 +19,32 @@
...
-                    testFailure = evaluateSafely(base);
-                } finally {
-                    Mockito.framework().removeListener(listener);
-                }
-
-                //If the 'testFinished' fails below, we don't see the original failure, thrown later
-                DefaultTestFinishedEvent event = new DefaultTestFinishedEvent(target, method.getName(), testFailure);
-                listener.testFinished(event);
-
+                session = Mockito.mockitoSession()
```
mockitoguy: This is amazing cleanup. Thank you so much!!!

src/main/java/org/mockito/internal/junit/JUnitRule.java
```diff
@@ -70,11 +59,14 @@ private Throwable evaluateSafely(Statement base) {
...
 
     public MockitoRule silent() {
-        return new JUnitRule(logger, Strictness.LENIENT);
+        return strictness(Strictness.LENIENT);
     }
 
     public MockitoRule strictness(Strictness strictness) {
-        this.listener.setStrictness(strictness);
+        this.strictness = strictness;
+        if (session != null) {
```
mockitoguy: Can you write a comment explaining when session is null? Useful for folks who don't know this code very well and future maintainers!

src/main/java/org/mockito/internal/session/DefaultMockitoSessionBuilder.java
```diff
@@ -1,23 +1,51 @@
...
+            this.testClassInstances.add(testClassInstance);
+        }
+        return this;
+    }
+
+    @Override
+    public MockitoSessionBuilder initMocks(Object... testClassInstances) {
+        if (testClassInstances != null) {
+            for (Object instance : testClassInstances) {
+                initMocks(instance);
```
mockitoguy: Few questions:
- why this method initializes the mocks but the non-vararg method just appends to the list?
- why do we need 2 methods: the vararg and non-vararg initMocks?

src/main/java/org/mockito/internal/session/DefaultMockitoSessionBuilder.java
```diff
@@ -27,11 +55,27 @@ public MockitoSessionBuilder strictness(Strictness strictness) {
...
+
     @Override
     public MockitoSession startMocking() {
         //Configure default values
-        Object effectiveTest = this.testClassInstance == null ? new Object() : this.testClassInstance;
+        List<Object> effectiveTestClassInstances;
+        String effectiveName;
+        if (testClassInstances.isEmpty()) {
+            effectiveTestClassInstances = emptyList();
+            effectiveName = this.name == null ? UUID.randomUUID().toString() : this.name;
```
mockitoguy: Random String may not provide great user experience. Can we make it more self-describing in case our users see that name? Can we just have empty String or "test". With the current functionality of Mockito, are there any circumstances where random UUID will be visible to the user?

src/main/java/org/mockito/session/MockitoSessionBuilder.java
```diff
@@ -30,17 +30,47 @@
...
+     * like {@link org.mockito.Mock}.
+     *
+     * @param testClassInstances test class instances that contains fields with Mockito annotations to be initialized.
+     *  Passing {@code null} or an empty array is permitted but will be ignored.
+     * @return the same builder instance for fluent configuration of {@code MockitoSession}.
+     * @see #initMocks(Object)
+     * @since 2.13.4
+     */
+    @Incubating
+    MockitoSessionBuilder initMocks(Object... testClassInstances);
```
mockitoguy: I suggest to kill the other 'initMocks' method and just leave the vararg. It is a compatible change.

src/main/java/org/mockito/session/MockitoSessionBuilder.java
```diff
@@ -55,6 +85,21 @@
     @Incubating
     MockitoSessionBuilder strictness(Strictness strictness);
 
+    /**
+     * Configures logger used by {@code MockitoSession} for emitting hints when finishing
```
mockitoguy: Can you add javadoc link to MockitoHint class? This will help getting a big picture what is the deal with hints.

src/main/java/org/mockito/session/MockitoSessionBuilder.java
```diff
@@ -55,6 +85,21 @@
...
+     * the session.
+     * See examples in {@link MockitoSession}.
+     *
+     * @param logger for hints emitted when finishing {@code MockitoSession}.
```
mockitoguy: How about we explain to the user when to use this method? We can say that we recommend to use strict stubs rather than warnings, and that this method is used by framework integrations to keep backwards compatibility with early versions of Mockito 2.x

src/main/java/org/mockito/session/MockitoSessionBuilder.java
```diff
@@ -30,17 +30,47 @@
...
+     *  Passing {@code null} or an empty array is permitted but will be ignored.
+     * @return the same builder instance for fluent configuration of {@code MockitoSession}.
+     * @see #initMocks(Object)
+     * @since 2.13.4
+     */
+    @Incubating
+    MockitoSessionBuilder initMocks(Object... testClassInstances);
+
+    /**
+     * Configures the name of the {@code MockitoSession} instance.
```
mockitoguy: It would be great to explain when the user is interested using this method. We can say that it is for framework integrations, example JUnit5, and that we don't expect the users calling this method.

src/main/java/org/mockito/session/MockitoSessionLogger.java
```diff
@@ -0,0 +1,26 @@
...
+ * Copyright (c) 2018 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.session;
+
+import org.mockito.Incubating;
+import org.mockito.MockitoSession;
+
+/**
+ * Logger for hints emitted when finishing mocking for a {@link MockitoSession}.
```
mockitoguy: Let's link to MockitoHint class. I'd also explain that this API is used by framework integrations, example JUnit5.

src/main/java/org/mockito/MockitoSession.java
```diff
@@ -48,7 +48,9 @@
  *         //initialize session to start mocking
  *         mockito = Mockito.mockitoSession()
  *            .initMocks(this)
+ *            .name("Example")
```
mockitoguy: 'name' and 'logger' are targeted to framework integrations rather than regular users. I'd rather we don't include them here in this example. This example is for regular users who want to use mockito session in their test. They won't know what name or logger to configure :)

I suggest to remove those 2 methods from this example completely. I don't think they need an example at the top level.

src/main/java/org/mockito/internal/session/MockitoLoggerAdapter.java
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) 2018 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.internal.session;
+
+import org.mockito.internal.util.MockitoLogger;
+import org.mockito.session.MockitoSessionLogger;
+
+class MockitoLoggerAdapter implements MockitoLogger {
```
mockitoguy: This and other changes related to the logger are great. THANKS!

src/main/java/org/mockito/session/MockitoSessionBuilder.java
mockitoguy: Can you explain in what circumstances we pass multiple test instances? Can we recommend to pass single instance and explain that multiple instances are for advanced framework integrations, for example...

src/main/java/org/mockito/session/MockitoSessionLogger.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * @since 2.13.4
+ */
+@Incubating
+public interface MockitoSessionLogger {
```
mockitoguy: Nice! Thank you for making this type public and placing it in the right package.

src/main/java/org/mockito/session/MockitoSessionBuilder.java
marcphilipp: AFAIK it's a source compatible change, but breaks binary compatibility because old code is compiled against a method with signature `initMocks(Object)` which would no longer exists, because there's now only `initMocks(Object[])`.

src/main/java/org/mockito/MockitoSession.java
```diff
@@ -89,6 +91,16 @@
 @Incubating
 public interface MockitoSession {
 
+    /**
+     * Changes the strictness of this {@code MockitoSession}.
```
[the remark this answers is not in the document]
marcphilipp: Will do!

src/main/java/org/mockito/MockitoSession.java
```diff
@@ -105,8 +117,22 @@
...
      * For example, see javadoc for {@link MockitoSession}.
      *
+     * @see #finishMocking(Throwable)
      * @since 2.7.0
      */
     @Incubating
     void finishMocking();
+
+    /**
+     * Must be invoked when the user is done with mocking for given session (test method).
```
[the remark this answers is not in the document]
marcphilipp: Sure, will do!

src/main/java/org/mockito/internal/junit/JUnitRule.java
marcphilipp: Yep, will do!

src/main/java/org/mockito/internal/session/DefaultMockitoSessionBuilder.java
marcphilipp: It doesn't initialize the mocks, it calls the `initMocks(Object)` method on the builder to not have to duplicate the null check.

src/main/java/org/mockito/internal/session/DefaultMockitoSessionBuilder.java
marcphilipp: It is visible in the emitted hints, e.g. in https://github.com/mockito/mockito/blob/c8f07c70a2e2be4714b4fc6a1a7f6e8b473b3bac/src/main/java/org/mockito/internal/junit/StubbingHint.java#L14-L17

How about using sth. like `"<unnamed session>"`?

src/main/java/org/mockito/session/MockitoSessionBuilder.java
marcphilipp: Will do!

src/main/java/org/mockito/session/MockitoSessionBuilder.java
marcphilipp: Sure!

src/main/java/org/mockito/session/MockitoSessionBuilder.java
marcphilipp: 👍

src/main/java/org/mockito/session/MockitoSessionBuilder.java
marcphilipp: Sure!

src/main/java/org/mockito/session/MockitoSessionLogger.java
marcphilipp: Will do!

TimvdLippe:
@marcphilipp Thank you very much for your PR! It looks like this also includes several changes I introduced in https://github.com/mockito/mockito/pull/1221/commits/68b71db63a6a90d5f7bbfd8ddead295be6972c86 for #1221. We will need to look how this merges.

Sadly I do not have a lot of time for review, but I will try to find some time as soon as possible to provide feedback. Initial look is promising :smile:

marcphilipp:
@mockitoguy I've improved the documentation in 360fb5d.

@TimvdLippe Sorry, I wasn't aware of your changes in #1221. There will definitely be conflicts, but the general direction seems to be the same.

mockitoguy:
>@mockitoguy I've improved the documentation in 360fb5d.

Great! Taking a look now.

mockitoguy:
@marcphilipp, can you configure your PR to allow our pushes? I have minor tweaks I want to do before merging and would like to push to your repo for visibility first. I've tried to push to your repo branch but I got this:

```
~/mockito/src$ git push https://github.com/marcphilipp/mockito.git
...
To https://github.com/marcphilipp/mockito.git
 ! [remote rejected]     marcphilipp-issues/1232-session-additions -> marcphilipp-issues/1232-session-additions (permission denied)
error: failed to push some refs to 'https://github.com/marcphilipp/mockito.git'
```

mockitoguy:
@marcphilipp, this PR is great and I'm almost ready to merge. The updates to the docs are good. I'm making tweaks to this PR in this branch: https://github.com/mockito/mockito/commits/marcphilipp-issues/1232-session-additions. Making tweaks helps me understand the code and keeps the docs consistent. If you allow upstream pushes from this PR, it would be useful!

We are very happy to take this contribution. Thank you!

marcphilipp:
Hmm, "allow edits from maintainers" is enabled, so it should work:

> If checked, users with write access to mockito/mockito can add new commits to your `issues/1232-session-additions` branch. You can always change this setting later.

Anyway, I've sent you an invite as a collaborator to my fork. That should work in any case. Please let me know if it did.

mockitoguy:
@mockito/developers, can we get your input on method naming?

This PR adds:

```
session.setStrictness(strictness);
```

Since in our API we already use "strictness" method, I suggest we keep it consistent:

```java
//existing:
mockitoRule.strictness(strictness);
mockitoSession().strictness(strictness);

//suggested change for this PR:
session.strictness(strictness);
```

Thoughts?

TimvdLippe:
@mockitoguy I am okay with that :+1:

mockitoguy:
Thank you @TimvdLippe for feedback! @marcphilipp, thoughts on 'strictness' vs 'setStrictness' method? In the next day or two, I'll finish tweaking this PR, push changes upstream to your fork, and we should be ready to merge.

This is very nice improvement and I'm waiting eagerly to merge!!!

marcphilipp:
I went with the setter to emphasize that the old value might already have been used for previously created mocks but didn‘t want to break the API of the Rule implementation. However, I‘d be fine with `strictness()`, too!

src/main/java/org/mockito/session/MockitoSessionBuilder.java
ChristianSchwarz: Q: Should we throw an IAE in case:
 * an empty vararg/array is passed
 * an array with `null` elements is passed
 * `(Object[])null` is passed

mockitoguy:
@marcphilipp, thank you for feedback. We should be ready to merge soon. Today's Mockito time budget I used on different PR but I have this one on mind every day :)

src/main/java/org/mockito/internal/session/DefaultMockitoSessionBuilder.java
mockitoguy: Got it. Thank you for clarifying. I misread the code.

src/main/java/org/mockito/session/MockitoSessionBuilder.java
mockitoguy: Good idea.

mockitoguy:
@marcphilipp, before I merge I have 2 requests:
 - can you update the version in "version.properties" to next minor number (and also the ```@since``` tags)? I believe it will e 2.15.0
 - can you add a paragraph to our main Javadoc with concise summary of the public API changes. Check out how we have done it in the past: https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/Mockito.java#L1451

This is a really, really nice improvement of the public API for framework integrations. Big thanks!

marcphilipp:
@mockitoguy Done! Feel free to polish what I wrote.

I merged master into the PR branch to resolve the conflict in version.properties. I'll be on vacation the next few days but please go ahead and rebase, squash, or merge the commits in this branch.

mockitoguy:
Great, thank you! Making a final review and merging...

src/main/java/org/mockito/Mockito.java
```diff
@@ -1398,7 +1398,7 @@
  * <a href="https://github.com/mockito/mockito/issues/769">issue 769</a>.
  *
  * <h3 id="41">41. <a class="meaningful_link" href="#framework_integrations_api" name="framework_integrations_api">
- *      (**new**) Advanced public API for framework integrations (Since 2.10.+)</h3>
+ *      (**new**) Advanced public API for framework integrations (Since 2.10.+)</a></h3>
```
mockitoguy: Nice catch, thank you!

PR #1304, opened 2018-02-05
Make InstantiatorProvider a public API

### LINKED ISSUE #1303: Make  org.mockito.internal.creation.instance.Instantiator a public API
The public API org.mockito.plugins.InstantiatorProvider#getInstantiator returns an Instantiator object. The Instantiator is no public API though.

moltmann:
The public API org.mockito.plugins.InstantiatorProvider#getInstantiator returns an Instantiator object that is an internal type. This PR exposes public API for InstantiatorProvider plugin.

```
public interface InstantiatorProvider2 {
    Instantiator getInstantiator(MockCreationSettings<?> settings);
}
```

In order to preserve backwards compatibility, we need to introduce new ```InstantiatorProvider2``` public interface. Previous ```InstantiatorProvider``` is still supported but the new type takes precedence in case both plugins are found on the classpath. For more information, see the Javadoc.

Fixes #1303


src/main/java/org/mockito/creation/instance/InstantiationException.java
```diff
@@ -2,10 +2,15 @@
...
  */
-package org.mockito.internal.creation.instance;
+package org.mockito.creation.instance;
 
 import org.mockito.exceptions.base.MockitoException;
 
+/**
+ * Exception generated when {@link Instantiator#newInstance(Class)} failed.
+ *
+ * @since 5.2.18
```
TimvdLippe: Should be `2.14.0`

src/main/java/org/mockito/plugins/InstantiatorProvider.java
```diff
@@ -4,7 +4,7 @@
  */
 package org.mockito.plugins;
 
-import org.mockito.internal.creation.instance.Instantiator;
+import org.mockito.creation.instance.Instantiator;
```
mockitoguy: This is a backwards incompatible change.

TimvdLippe: But we regard `internal` as an exception to our backwards compatibility requirement right?

moltmann: @mockitoguy I see it like this:

Having a class marked as internal means that using this class might break if mockito is updated. Hence anybody using Instantiator should be aware that this might break. Hence I think breaking people is just the expected behavior.

Of course InstantiatorProvider is a public API and upgrading mockito should not break this API. I think there is only one use case of InstantiatorProvider without using internal APIs. It is calling getInstantiator and compare it to null or an existing object. This still works, even after my move.

In summary: Yes I am braking everybody using of the internal class. I don't find a use case where I break anyone using the public API.

src/main/java/org/mockito/creation/instance/InstantiationException.java
moltmann: done

src/main/java/org/mockito/plugins/InstantiatorProvider.java
mockitoguy: >But we regard internal as an exception to our backwards compatibility requirement right?

@TimvdLippe, yes. We have few exceptions from our standard rule, where internal API leaked and is considered public. The other exception is [VerificationData](https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/internal/verification/api/VerificationData.java) type that lives in "internal" package but is considered public (see the Javadoc).

mockitoguy: >In summary: Yes I am braking everybody using of the internal class. I don't find a use case where I break anyone using the public API.

@moltmann, you are right that it is not a breaking change provided that InstantiatorProvider is only used to null-check result from ```getInstantiator()``` method. Problem is that our "plugins" API contract tells users to provide implementations of plugin interfaces, such as ```InstantiatorProvider```. So anyone who uses ```InstantiatorProvider``` would actually provide his own implementation, and his code would break if we ship this PR as-is.

If you are concerned about using "internal" class that we consider public, what do you think about option b) (create new type, make the change compatible)?

moltmann: If there is precedent for a internal-quite-not-internal API, I can mark it with the same statement as VerificationData.

The statement in the JavaDoc of VerificationData states that the class will be moved though. Will this actually happen?

ChristianSchwarz:
Option d)

Keep and deprecate the internal `org.mockito.internal.creation.instance.Instantiator` and let it extend the _"new"_ public `org.mockito.*.Instantiator`. This way Mockito remains compatible.

mockitoguy:
@ChristianSchwarz, thank you for suggestion!

How would that work? If I deprecate existing ```Instantiator``` what is the action the existing users (who provide own implementation of InstantiatorProvider) should do to avoid using deprecate type? The user cannot change the signature of his ```MyInstantiatorProvider#getInstantiator()``` method because compiler would not like that :)

Plus, there's also InstantiationException that is currently internal and would have to be made public. @ChristianSchwarz, can you put together an experimental PR to show us how this idea would work?

moltmann:
I think my last update f6d1081e95b9c28be4c1a71b82f2f9d2305c00fd is what @ChristianSchwarz meant.

The other way round (make the internal type a subclass of the external type) does not work as this would change the return type of InstantiatorProvider#getInstantiator in an incompatible way.

moltmann:
@ChristianSchwarz I tried both: making the internal Instantiator subclass of the external one and the other way round. In both cases I ended up with non-compatible code.

The problem is the return type of InstantiatorProvider.

ChristianSchwarz:
@moltmann 
@mockitoguy 

Yes the problem is the return type of InstantiatorProvider, but if i am not wrong it is only a source compability issue, the binary should be fine ? 

Here is the requested PR #1309 illustarting the problem @moltmann described.

It can be solved with proposal **b)**

mockitoguy:
@ChristianSchwarz, thanks!

I left the comment on your PR with suggestions how to test the change. If you make it work in a compatible way, I will be happy to merge ;)

mockitoguy:
@ChristianSchwarz's approach is not compatible (details in the PR)

@moltmann, what do you think about option b)? Do you want to take a stab at it?

moltmann:
I uploaded version b) as an additional change on top of version c)

ChristianSchwarz:
@mockitoguy 
@moltmann 

Some questions about the compability:
 * Doesn't the leaked internal API of `*.internal.*.Instantiator` render the public API ( `InstantiatorProvider` ) uses les. It is per contract forbidden to reference it, read: "use on you own risk".
 * Whould this allow it to change the return type? 
 * How many clients implement this interface, do we know one?
 * What efforts will the implementers have to take, if we change the return type?

mockitoguy:
@ChristianSchwarz,

Leaked internal API sucks. As library owners with large user base, we don't know how many projects depend on specific API portion. We play on the safe side because staying compatible is extremely useful for adoption, trust and happiness of our customers. In the long term, compatibility often decides about the success of the software.

mockitoguy:
@moltmann, reviewing...

src/main/java/org/mockito/internal/configuration/plugins/PluginRegistry.java
```diff
@@ -50,10 +59,36 @@ MockMaker getMockMaker() {
...
      *
      * <p>Returns {@link org.mockito.internal.creation.instance.DefaultInstantiatorProvider} if no
-     * {@link org.mockito.plugins.InstantiatorProvider} extension exists or is visible in the current classpath.</p>
+     * {@link org.mockito.plugins.InstantiatorProvider2} extension exists or is visible in the
+     * current classpath.</p>
      */
-    InstantiatorProvider getInstantiatorProvider() {
-      return instantiatorProvider;
+    InstantiatorProvider2 getInstantiatorProvider() {
+        // Lazily init defaultInstantiatorProvider as all dependencies of the default instantiator
```
mockitoguy: To avoid problems with concurrency, we prefer to instantiate all plugins statically vs. lazily. Can you explain why you suggest doing it lazily here?

src/main/java/org/mockito/internal/configuration/plugins/PluginRegistry.java
```diff
@@ -50,10 +59,36 @@ MockMaker getMockMaker() {
...
+        // provider might not be available when this class is initialized
+        if (defaultInstantiatorProvider == null) {
+            defaultInstantiatorProvider = (new DefaultMockitoPlugins()).getDefaultPlugin(InstantiatorProvider.class);
+        }
+
+        if (instantiatorProvider.getClass() != defaultInstantiatorProvider.getClass()) {
```
mockitoguy: Let's keep the complexity of loading plugins in the ```PluginLoader``` type. Otherwise it is hard to understand what's going on here. Example implementation:

```
Object plugin = new PluginLoader(pluginSwitch).loadPlugin(InstantiatorProvider.class, InstantiatorProvider2.class);
//returned plugin is either of those 2 types, we can use 'instanceof' to find out
```

then we can create a new type that adapts to the ```InstantiatorProvider2``` interface so that PluginRegistry could look like this:

```
private final InstantiatorProvider2 instantiatorProvider2 = 
        new InstantiatorProviderWrapper(
            new PluginLoader(pluginSwitch).loadPlugin(InstantiatorProvider.class, InstantiatorProvider2.class));
```

I think that implementing above will remove the need for ```LegacyDefaultInstantiatorProvider``` class included in this PR.

src/main/java/org/mockito/creation/instance/InstantiationException.java
```diff
@@ -0,0 +1,19 @@
...
+package org.mockito.creation.instance;
+
+import org.mockito.exceptions.base.MockitoException;
+
+/**
+ * Exception generated when {@link Instantiator#newInstance(Class)} failed.
+ *
+ * @since 2.14.0
+ */
+public class InstantiationException extends MockitoException {
```
mockitoguy: Nice!

src/main/java/org/mockito/creation/instance/Instantiator.java
```diff
@@ -0,0 +1,19 @@
...
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.creation.instance;
+
+/**
+ * Provides instances of classes.
+ *
+ * @since 2.14.0
+ */
+public interface Instantiator {
```
mockitoguy: Nice!

src/main/java/org/mockito/plugins/InstantiatorProvider.java
```diff
@@ -45,6 +45,7 @@
  *     <code>mockito-extensions/org.mockito.plugins.InstantiatorProvider</code>.
  * </p>
  *
+ * @deprecated Use {@link InstantiatorProvider2} instead
  * @since 21.10.15
```
mockitoguy: Bonus: if you can help us identify what version this was released in the past, that would be cool. It looks like this version is a date. It should be a number :)

src/main/java/org/mockito/plugins/InstantiatorProvider2.java
```diff
@@ -0,0 +1,56 @@
...
+package org.mockito.plugins;
+
+import org.mockito.creation.instance.Instantiator;
+import org.mockito.mock.MockCreationSettings;
+
+/**
+ * <p>
+ *     Mockito will invoke this interface in order to fetch an instance instantiator provider.
+ * </p>
+ *
```
mockitoguy: Can you explain somewhere here why deprecated previous InstantiatorProvider and created this class? It's a good dev experience when a concise justification of deprecation/API change is offered in Javadoc (plus link to github ticket).

Please also explain that the old InstantiatorProvider is supported but the new one takes precedence during plugin loading.

Finally, we'd love to have a new paragraph in the main Mockito class explaining the addition of new public API (see example: https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/Mockito.java#L100)

mockitoguy:
During review I was making tweaks to the documentation and I pushed it directly to your branch. I hope it's ok!

src/main/java/org/mockito/internal/configuration/plugins/PluginRegistry.java
moltmann: Good suggestion, done.

I still have to keep legacy support in DefaultMockitoPlugins#getDefaultPlugin though.

I factored the compatibility wrappers InstantiatorProvider2 <-> InstantiatorProvider into separate classes.

src/main/java/org/mockito/internal/configuration/plugins/PluginRegistry.java
moltmann: By using the loadPlugin with two parameters, this can be avoided.

src/main/java/org/mockito/plugins/InstantiatorProvider.java
moltmann: At the time the file was created version.properties showed 2.0.31-beta, so 2.0.31 ?

src/main/java/org/mockito/plugins/InstantiatorProvider2.java
moltmann: - I think you already added a comment why this was deprecated
- Added link to #1303
- Added comment about plugin preference if both a InstantiatorProvider and InstantiatorProvider2 are configured
- Created entry 44 in Mockito.java

moltmann:
I uploaded a new version that addresses the comment including adding the tests.

"./gradlew build" succeeds, but I get

Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.

from using the InstantiatorProvider in the tests.

mockitoguy:
I haven't had the time to work on this today. I plan on reviewing tomorrow morning. Thank you!

src/main/java/org/mockito/internal/creation/instance/InstantiatorProvider2Wrapper.java
```diff
@@ -0,0 +1,34 @@
...
+package org.mockito.internal.creation.instance;
+
+import org.mockito.mock.MockCreationSettings;
+import org.mockito.plugins.InstantiatorProvider;
+import org.mockito.plugins.InstantiatorProvider2;
+
+/**
+ * Map a {@link InstantiatorProvider2} onto a {@link InstantiatorProvider}
+ */
+public class InstantiatorProvider2Wrapper implements InstantiatorProvider {
```
mockitoguy: I suggest to call this type ```InstantiatorProvider2Adapter``` to clearly indicate the intent. Same applies to the other adapter: ```InstantiatorProviderWrapper``` -> ```InstantiatorProviderAdapter```

src/main/java/org/mockito/plugins/InstantiatorProvider.java
```diff
@@ -44,9 +48,17 @@
...
+ * <p>
+ *     This class is deprecated and was replaced by
+ *     {@link org.mockito.plugins.InstantiatorProvider2}. Hence if there is both a
+ *     <code>mockito-extensions/org.mockito.plugins.InstantiatorProvider</code> and
+ *     <code>mockito-extensions/org.mockito.plugins.InstantiatorProvider2</code> the second one
+ *     takes preference.
+ * </p>
  *
- * @since 21.10.15
+ * @since 2.0.31
```
mockitoguy: Thank you for fixing the version!

src/main/java/org/mockito/plugins/InstantiatorProvider2.java
mockitoguy: Great!

version.properties
```diff
@@ -1,8 +1,8 @@
 #Currently building Mockito version
-version=2.15.2
```
mockitoguy: hmmm, this PR is picking up commits from "release/2.x" branch. Please prefer rebase vs. merge. Unless I caused it by pushing to your branch :)

src/main/java/org/mockito/internal/configuration/plugins/PluginRegistry.java
mockitoguy: Thank you! I like the new types a lot.

version.properties
moltmann: This seems to be fixed now

moltmann:
Removed the TODOs and replaced them by 2.15.4

mockitoguy:
We're ready. Thanks a lot @moltmann!!! I will wait for final feedback by tomorrow and merge it.

mockitoguy:
I merged it from command line, not sure why GitHub did not pick it up.

Thank you for contribution! We really like to keep the public APIs clean (and compatible)

PR #1307, opened 2018-02-06
Update public API of InvocationFactory needed for Android static mocking

### LINKED ISSUE #1306: InvocationFactory.createInvocation's realmethod cannot throw a Throwable
InvocationFactory.createInvocation takes a Callable that is used to call the real method of the Invocation. If the Invocation throws an Exception the Exception is forwarded from the Callable. Unfortunately Callable can only throw an Exception, not a Throwable.

Hence if the real method throws a Throwable that is neither an Error, RuntimeException or Exception it cannot be forwarded.

I recommend for InvocationFactory.createInvocation to take realMethod argument that can throw a full Throwable.

This is required for fixing: linkedin/dexmaker#64

moltmann:
Fixes #1306

A real method might throw a Throwable, but the invocations produced by
InvocationFactory call the real method via a Callable that can only
throw a Exception.

Hence add a new method to InvocationFactory that allows to create
invocations with a RealMethodBehavior == a Callable that throws a
Throwable.

### DIFF
--- a/src/main/java/org/mockito/internal/invocation/DefaultInvocationFactory.java
+++ b/src/main/java/org/mockito/internal/invocation/DefaultInvocationFactory.java
@@ -18,4 +18,9 @@ public class DefaultInvocationFactory implements InvocationFactory {
         RealMethod.FromCallable superMethod = new RealMethod.FromCallable(realMethod);
         return MockMethodInterceptor.createInvocation(target, method, args, superMethod, settings);
     }
+
+    public Invocation createInvocation(Object target, MockCreationSettings settings, Method method, RealMethodBehavior realMethod, Object... args) {
+        RealMethod.FromBehavior superMethod = new RealMethod.FromBehavior(realMethod);
+        return MockMethodInterceptor.createInvocation(target, method, args, superMethod, settings);
+    }
 }
--- a/src/main/java/org/mockito/internal/invocation/DefaultInvocationFactory.java
+++ b/src/main/java/org/mockito/internal/invocation/DefaultInvocationFactory.java
@@ -15,17 +15,16 @@ import java.util.concurrent.Callable;
 public class DefaultInvocationFactory implements InvocationFactory {
 
     public Invocation createInvocation(Object target, MockCreationSettings settings, Method method, final Callable realMethod, Object... args) {
-        RealMethod superMethod = new RealMethod.FromBehavior(new RealMethodBehavior<Object>() {
-            @Override
-            public Object call() throws Throwable {
-                return realMethod.call();
-            }
-        });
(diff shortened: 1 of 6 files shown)


src/main/java/org/mockito/internal/invocation/DefaultInvocationFactory.java
```diff
@@ -14,8 +14,13 @@
 
 public class DefaultInvocationFactory implements InvocationFactory {
 
-    public Invocation createInvocation(Object target, MockCreationSettings settings, Method method, Callable realMethod, Object... args) {
+    public Invocation createInvocation(Object target, MockCreationSettings settings, Method method, final Callable realMethod, Object... args) {
```
TimvdLippe: Nit: no need to add `final` here

moltmann: done

moltmann:
- "R call() throws Throwable" would conflict with ""R call() throws Exception" so I cannot use subclassing
- When just an anonoymous inner RealMethodBehavior that wrapps the Callable org.mockitousage.basicapi.MocksSerializationTest fails.

I added a test case that tests the case where the invocation is handled without a Throwable.

mockitoguy:
Reviewing...

mockitoguy:
>Else you should be able to update our internal FromCallable to use a RealMethodBehavior and let both createInvocation call this method. That should remove the need for a duplicate inner class (since these classes are in internal anyways)

+1. I like @TimvdLippe's idea to reduce duplication. I applied that idea to the changes I made on top @moltmann's branch (https://github.com/mockito/mockito/commits/moltmann-ThrowCallable)

moltmann:
@mockitoguy
- You change https://github.com/mockito/mockito/commit/017cf577910c2156ed785a18914396bcba3d51f9 seem to break org.mockitousage.basicapi.MocksSerializationForAnnotationTest , the same issue I saw
- How do I apply the change changes? Manually?

mockitoguy:
>You change 017cf57 seem to break org.mockitousage.basicapi.MocksSerializationForAnnotationTest , the same issue I saw

Ouch, I missed that use case. We need to make the new interface Serializable. Otherwise, non-android use cases like Powermockito, will use the new public API and break when mocks are being serialized (unless they remember to keep the implementation classes Serializable). For android use cases we don't need serialization, but this new public API will be used in other context, too.

Our strategy is that public interfaces that we expect our users to extend are marked Serializable if they are needed for Mockito's serialization feature.

>How do I apply your changes? Manually via "git am"?

Just pull my changes on top of your branch. Alternatively, configure your PR to allow pushes from Mockito devs (did you do it already?)

mockitoguy:
@TimvdLippe, once the build passes, I'm merging. Thanks for review!

PR #1310, opened 2018-02-11
Ensure isolation of stubbings

mockitoguy:
### Bugfix compatibility

For certain corner cases, this bugfix may be incompatible. We believe that the exposure is minimal (corner cases only, possibly incorrect tests), and the bugfix important. If your test fails after upgrading to Mockito version that contains this fix, please scrutinize the test - it is very likely that the test is incorrect. We apologize for any compatibility problems arising from this change. It is a hard decision to make when a bugfix changes the behavior in a way it may not be compatible with every Mockito test in the world.

### Problem

When stubbing with answers that call different mocked methods:
```java
when(mock.foo()).thenAnswer(() -> { otherMock.bar() }
```
we need to ensure that ```mock.foo()``` and ```otherMock.bar()``` don't interfere. Prior to this bugfix, this scenario resulted in various problems like: stack overflow error, unexpected stubbings declared on different mock, unexpected invalid return type errors.

For more details, see #1279

### Solution

The solution is based on great PR #1296 by @r-smirnov. **Huge thanks** to @r-smirnov for reporting the problem diligently, then digging into the bowels of Mockito to find a clean solution, providing excellent test cases that demonstrate the problem. When I initially saw the issue report, I thought that it's not fixable and we would close it as "documented Mockito syntax caveat". Thank you @r-smirnov for persistence and getting this fixed!

### DIFF
--- a/src/main/java/org/mockito/internal/handler/MockHandlerImpl.java
+++ b/src/main/java/org/mockito/internal/handler/MockHandlerImpl.java
@@ -8,7 +8,6 @@ import org.mockito.internal.creation.settings.CreationSettings;
 import org.mockito.internal.invocation.InvocationMatcher;
 import org.mockito.internal.invocation.MatchersBinder;
 import org.mockito.internal.listeners.StubbingLookupListener;
-import org.mockito.invocation.InvocationContainer;
 import org.mockito.internal.stubbing.InvocationContainerImpl;
 import org.mockito.internal.stubbing.OngoingStubbingImpl;
 import org.mockito.internal.stubbing.StubbedInvocationMatcher;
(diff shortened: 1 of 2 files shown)


TimvdLippe:
Can we give an example of a test that breaks with this "new" behavior?

mockitoguy:
> Can we give an example of a test that breaks with this "new" behavior?

Take a look (or run) the test class attached to this PR. There are test methods that demonstrate invalid behavior such as SOE. There are also test methods that demonstrate incorrect behavior of stubbing.

mockitoguy:
@TimvdLippe, the ticket #1279, linked from this PR, contains examples.

TimvdLippe:
@mockitoguy I understand that the supplied test cases are now working as intended. What I meant was: are there any other tests that previously passed but now fail and what we consider the "breaking change"?

TimvdLippe:
In that case I wouldn't consider this a breaking change, simply a major bugfix 😄

mockitoguy:
>In that case I wouldn't consider this a breaking change, simply a major bugfix 😄

Sounds good. I updated the description. Can you review the code / approve?

mockitoguy:
Thank you again @r-smirnov!!! Merging.

PR #1315, opened 2018-02-17
Update Kotlin version to be compatible with the IntelliJ runtime version

TimvdLippe:
The version of the IntelliJ Kotlin runtime has been updated to 1.2.10, therefore bump our version as well.


mockitoguy:
Thank you!

PR #1316, opened 2018-02-17
Fix several static analysis warnings as reported by IntelliJ

TimvdLippe:
- Public access can be stronger (package-)private
- Redundant throws Exception/Throwable in tests
- Unused fields

All warnings were auto-fixed with the IntelliJ quickfixes.

Especially the access changes should limit users from implicitly relying on internal code that we should not expose. Other than that, the throws Exception is mostly an artifact of really old (5+ years) code.


mockitoguy:
Thank you! Do you mind if we merge 2 outstanding PRs first (#1304, #1272)?

TimvdLippe:
Yeah, we should have merged this one before the others. The merge conflicts are a mess now. Closing 😢

Issue #1112, opened 2017-06-07
mockito-android has extremely large method count


drewhannay:
I've been investigating why my test apk has hit the 65k "multidex" method limit. It seems like `mockito-android` is by far the dominant contributor to method count.

To help get cleaner data, I created a brand new empty project in Android Studio 3.0-alpha3 and removed all dependencies except the following:
```
dependencies {
    androidTestImplementation 'org.mockito:mockito-android:2.8.9'
    androidTestImplementation('com.android.support.test.espresso:espresso-core:2.2.2', {
        exclude group: 'com.android.support', module: 'support-annotations'
    })
}
```

After this, I built the test apk using `./gradlew assembleDebugAndroidTest` and used the APK Analyzer tool in Android Studio to examine the APK. Here's what it reports:

`org.mockito` --> 2402 referenced methods
`net.bytebuddy` --> 13512 referenced methods
`com.android.dx` --> 4661 referenced methods

This totals to 20575 referenced methods, which is roughly 1/3 of the total allowable referenced method count for an Android test apk. Since test apks don't support legacy multidex, if a test apk goes over the 65k method limit, they can no longer run their tests on versions of android older than API 21 (Lollipop).

For extra info, here's the (relevant) output of `./gradlew :app:dependencies` for my sample project:

```
debugAndroidTestCompileClasspath - Resolved configuration for compilation for variant: debugAndroidTest
+--- org.mockito:mockito-android:2.8.9
|    +--- org.mockito:mockito-core:2.8.9
|    |    +--- net.bytebuddy:byte-buddy:1.6.14
|    |    +--- net.bytebuddy:byte-buddy-agent:1.6.14
|    |    \--- org.objenesis:objenesis:2.5
|    \--- net.bytebuddy:byte-buddy-android:1.6.5
|         +--- net.bytebuddy:byte-buddy:1.6.5 -> 1.6.14
|         \--- com.jakewharton.android.repackaged:dalvik-dx:1
|              \--- com.jakewharton.android.repackaged:libcore-dex:2
```

In my real app, we are already using Proguard on our test APK, and that only strips out ~3k of the 20k methods referenced by mockito-android (so this probably isn't a viable solution).

Is there anything we can do to reduce the method count pulled in by `mockito-android`?

szczepiq:
Thank you for reporting!

[...]

PR #1320, opened 2018-02-21
Avoid bytebuddy import Issue to enable repackaging without bytebuddy

moltmann:
This allows libs that use alternate mockmakers do cleanly exclude the bytebuddy subdirectory.

Test: grep -r -e "import org\.mockito\.internal\.creation\.bytebuddy\.[^\.]*$" src/main does not return anything

Offers workaround for #1112

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodInterceptor.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodInterceptor.java
@@ -13,12 +13,8 @@ import net.bytebuddy.implementation.bind.annotation.RuntimeType;
 import net.bytebuddy.implementation.bind.annotation.StubValue;
 import net.bytebuddy.implementation.bind.annotation.SuperCall;
 import net.bytebuddy.implementation.bind.annotation.This;
-import org.mockito.internal.creation.DelegatingMethod;
 import org.mockito.internal.debugging.LocationImpl;
-import org.mockito.internal.invocation.MockitoMethod;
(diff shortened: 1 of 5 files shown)


mockitoguy:
Hey @moltmann - thank you for another contribution! Can you elaborate the goals for this PR?

This PR changes only internal classes. How does it help alternate mockmakers? What does "cleanly exclude the bytebuddy subdirectory" mean?

moltmann:
InterceptedInvocation is used by all mockmakers, hence if you want to order it correctly it should not be in a bytebuddy specific directory.

There are currently two ways to build mockito without bytebuddy.
1. Just load mockito-core and exclude net.bytebuddy. This leaves the bytebuddy-mockmaker in place. Hence there will be unresolved references into net-bytebuddy. This still works as none of these references will be called and (at least on Android) loading something with unresolved references just throws some log messages but nothing worse.
2. You can also take mockito-core, unpack it and remove the dependency on net.bytebuddy and remove the whole src/org/mockito/internal/creation/bytebuddy folder. Then you have no references into net-bytebuddy. To be able to do this you need this change to clean up the directory structure of mockito as described above. Also if you even wanted to produce a byte-buddy-less version of mockito, this change would be the first you would make.

moltmann:
Regarding my last comment: During a regular android-gradle build we run a lint checker which finds the issue (1) in my last comment. You can suppress this and you are good though.

mockitoguy:
Thank you for explanation. I updated the title and I'm absolutely merging!

PR #1321, opened 2018-02-21
Prevent memory leaks with inline MockMakers by using weak refs to mocks

### LINKED ISSUE #1313: InlineByteBuddyMockMaker does not clean up stale mocks
**Repro:**

1. Mockito 2.15.2 using mockito-inline
2. With -Xmx64m run
```
@RunWith(JUnit4.class)
public class Stress {
    private static final int NUM_RUNS = 20000;

    public static class TestClass {
        public String returnA() {
            return "A";
        }
    }

    /**
     * Test that mocks that are not referenced anymore are properly cleaned up.
     */
    @Test
    public void callALotOfMocks() {
        for (int i = 0; i < NUM_RUNS; i++) {
            TestClass mock = mock(TestClass.class);
            when(mock.returnA()).thenReturn("B");

            assertEquals("B", mock.returnA());

            if (i % 1024 == 0) {
                System.out.println("Mocked " + i + "/" + NUM_RUNS);
            }
        }
    }
}
```

**Expected:**

The test eventually finishes

**Observed:**

> Mocked 0/20000
> Mocked 1024/20000
> Mocked 2048/20000
> Mocked 3072/20000
> Mocked 4096/20000
> Mocked 5120/20000
> Mocked 6144/20000
> Mocked 7168/20000
> Mocked 8192/20000
> Mocked 9216/20000
> Mocked 10240/20000
> Mocked 11264/20000
> Exception in thread "main" java.lang.OutOfMemoryError: GC overhead limit exceeded
> 	at java.lang.StringCoding$StringDecoder.decode(StringCoding.java:149)
> 	at java.lang.StringCoding.decode(StringCoding.java:193)
> 	at java.lang.String.<init>(String.java:426)
> 	at java.util.jar.Attributes.read(Attributes.java:418)
> 	at java.util.jar.Manifest.read(Manifest.java:199)
> 
> Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "main"
> 
> Process finished with exit code 1

**Analysis:**

The `InlineByteBuddyMockMaker` uses a Map<WeakRef\<Mock\>, Handler> in the `mocks` field to store the mocks and their handlers. It relies that the mocks get garbage collected to remove the entry in the map. Unfortunately if there is an invocation on the mock, the handlers keep strong references to the mock in `InterceptedInvocation.mock`, `RealMethodCall.instance`, and `SerializableRealMethodCall.instance`. These are (strong) references by the handler. Hence the mock never gets cleaned up and the map grows indefinitely.

This is not a problem if the mocks get `Mockito.reset` properly or if using a mock maker that does not store the handlers.

moltmann:
Fixes #1313

This is needed for mock makers that store a map<weakref<mock>, invocation>. If the invocation holds a strong reference to the mock, the key never gets gc-ed and the map keeps on growing.

As weak references cannot be serialized, temporarily create a strong reference during serialization.

As the tests use invocations without keeping references to the mock we need to keep strong references in these cases.

### DIFF
--- a/gradle/java-library.gradle
+++ b/gradle/java-library.gradle
@@ -10,9 +10,6 @@ targetCompatibility = 1.6
 test {
     include "**/*Test.class"
 
-    // To make sure we do not leak memory during tests
-    maxHeapSize = "256m"
-
     testLogging {
         exceptionFormat 'full'
         showCauses true
--- a/gradle/java-library.gradle
+++ b/gradle/java-library.gradle
@@ -9,6 +9,10 @@ targetCompatibility = 1.6
 
 test {
     include "**/*Test.class"
+
+    // To make sure we do not leak memory during tests
+    maxHeapSize = "256m"
+
     testLogging {
         exceptionFormat 'full'
         showCauses true
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodAdvice.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodAdvice.java
@@ -26,7 +26,6 @@ import java.io.Serializable;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.ref.SoftReference;
-import java.lang.ref.WeakReference;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodAdvice.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodAdvice.java
@@ -24,6 +24,7 @@ import java.io.Serializable;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.ref.SoftReference;
+import java.lang.ref.WeakReference;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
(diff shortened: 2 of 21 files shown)


mockitoguy:
I'll review this weekend. Thank you!

mockitoguy:
My understanding is that we need to update the code and use weak reference wrapper everywhere where we currently keep the mock objects. @moltmann diligently identified following spots: InterceptedInvocation.mock, RealMethodCall.instance, and SerializableRealMethodCall.instance. I reviewed the remaining code and found other instances safe (implementations of MockCreationListener)

@moltmann, I like your idea with SerializableReference! I'm assuming we need to use it also in RealMethodCall, right? I'd love to see a very descriptive exception message thrown when serializableReference.get() method returns null (rather than just returning null as it is the case for WeakReference.get().

Thank you for contribution!

moltmann:
@mockitoguy
- I added an error message to be thrown when SerializableReference returns null. This should never be seen for users of the public API. Internal users might run into this though.
- I think RealMethodCall does not need SerializableReference as it never gets serialized. I am not sure why I added it there.

mockitoguy:
Hey @moltmann, can you rebase on the current release/2.x? It contains your previous PR and it created some conflicts?

In the meantime, I'll complete the review of this so that we know what we need to do before we merge it.

mockitoguy:
@moltmann - great contribution, BIG THANKS!

I made tweaks on top of your changes. We're ready to merge once the build is green.

@moltmann, @raphw any final remarks before we merge it?

raphw:
LGTM!

moltmann:
lgtm2

PR #1324, opened 2018-02-23
Improved stubbing internals and test coverage

dmitry-timofeev:
Some improvements in exception-stubbing code:

- Clarify intentions in BaseStubbing:
  - Replace an obsolete TODO with a clarifying comment and add some tests covering related test cases.
  - Extract the code aborting the ongoing stubbing in a separate method that always throws. Use return `abortNullExceptionType` so that static-analysis tools (e.g., IDEA) yield no false-positives.

- Clarify that ThrowsException always returns the _same_ throwable.
The previous `#answer` implementation might give an illusion that `Throwable#fillInStackTrace` produces a new throwable, but that is not correct — it always returns `this`.
- Also, replace a possible NPE with a descriptive IllegalStateException.

### DIFF
--- a/src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
+++ b/src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
@@ -66,7 +66,7 @@ public abstract class BaseStubbing<T> implements OngoingStubbing<T> {
     @Override
     public OngoingStubbing<T> thenThrow(Class<? extends Throwable> toBeThrown, Class<? extends Throwable>... nextToBeThrown) {
         if (nextToBeThrown == null) {
-            thenThrow((Class<Throwable>) null);
+            return thenThrow((Class<Throwable>) null);
         }
         OngoingStubbing<T> stubbing = thenThrow(toBeThrown);
         for (Class<? extends Throwable> t : nextToBeThrown) {
--- a/src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
+++ b/src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
@@ -24,7 +24,7 @@ public abstract class BaseStubbing<T> implements OngoingStubbing<T> {
     public OngoingStubbing<T> thenReturn(T value, T... values) {
         OngoingStubbing<T> stubbing = thenReturn(value);
         if (values == null) {
-            // TODO below does not seem right
+            // Return null in the 2nd and subsequent invocations.
             return stubbing.thenReturn(null);
         }
         for (T v : values) {
@@ -56,16 +56,16 @@ public abstract class BaseStubbing<T> implements OngoingStubbing<T> {
     @Override
     public OngoingStubbing<T> thenThrow(Class<? extends Throwable> throwableType) {
         if (throwableType == null) {
-            mockingProgress().reset();
-            throw notAnException();
+            return abortNullExceptionType();
         }
         return thenThrow(newInstance(throwableType));
     }
 
     @Override
-    public OngoingStubbing<T> thenThrow(Class<? extends Throwable> toBeThrown, Class<? extends Throwable>... nextToBeThrown) {
+    public OngoingStubbing<T> thenThrow(Class<? extends Throwable> toBeThrown,
+                                        Class<? extends Throwable>... nextToBeThrown) {
         if (nextToBeThrown == null) {
-            thenThrow((Class<Throwable>) null);
+            return abortNullExceptionType();
         }
         OngoingStubbing<T> stubbing = thenThrow(toBeThrown);
         for (Class<? extends Throwable> t : nextToBeThrown) {
@@ -77,6 +77,11 @@ public abstract class BaseStubbing<T> implements OngoingStubbing<T> {
     @Override
     public OngoingStubbing<T> thenCallRealMethod() {
         return thenAnswer(new CallsRealMethods());
+    }
+
+    private OngoingStubbing<T> abortNullExceptionType() {
+        mockingProgress().reset();
+        throw notAnException();
     }
 }
 
--- a/src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
+++ b/src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
@@ -24,7 +24,8 @@ public abstract class BaseStubbing<T> implements OngoingStubbing<T> {
     public OngoingStubbing<T> thenReturn(T value, T... values) {
         OngoingStubbing<T> stubbing = thenReturn(value);
         if (values == null) {
-            // Return null in the 2nd and subsequent invocations.
+            // For no good reason we're configuring null answer here
+            // This has been like that since forever, so let's keep it for compatibility (unless users complain)
             return stubbing.thenReturn(null);
         }
         for (T v : values) {
@@ -56,16 +57,16 @@ public abstract class BaseStubbing<T> implements OngoingStubbing<T> {
     @Override
     public OngoingStubbing<T> thenThrow(Class<? extends Throwable> throwableType) {
         if (throwableType == null) {
-            return abortNullExceptionType();
+            mockingProgress().reset();
+            throw notAnException();
         }
         return thenThrow(newInstance(throwableType));
     }
 
     @Override
-    public OngoingStubbing<T> thenThrow(Class<? extends Throwable> toBeThrown,
-                                        Class<? extends Throwable>... nextToBeThrown) {
+    public OngoingStubbing<T> thenThrow(Class<? extends Throwable> toBeThrown, Class<? extends Throwable>... nextToBeThrown) {
         if (nextToBeThrown == null) {
-            return abortNullExceptionType();
+            thenThrow((Class<Throwable>) null);
         }
         OngoingStubbing<T> stubbing = thenThrow(toBeThrown);
         for (Class<? extends Throwable> t : nextToBeThrown) {
@@ -77,11 +78,6 @@ public abstract class BaseStubbing<T> implements OngoingStubbing<T> {
     @Override
     public OngoingStubbing<T> thenCallRealMethod() {
         return thenAnswer(new CallsRealMethods());
-    }
-
-    private OngoingStubbing<T> abortNullExceptionType() {
-        mockingProgress().reset();
-        throw notAnException();
     }
 }
 
--- a/src/main/java/org/mockito/internal/stubbing/StubberImpl.java
+++ b/src/main/java/org/mockito/internal/stubbing/StubberImpl.java
@@ -83,9 +83,9 @@ public class StubberImpl implements Stubber {
         Throwable e;
         try {
             e = newInstance(toBeThrown);
-        } catch (RuntimeException instanciationError) {
+        } catch (RuntimeException instantiationError) {
             mockingProgress().reset();
-            throw instanciationError;
+            throw instantiationError;
         }
         return doThrow(e);
     }
--- a/src/main/java/org/mockito/internal/stubbing/answers/ThrowsException.java
+++ b/src/main/java/org/mockito/internal/stubbing/answers/ThrowsException.java
@@ -14,27 +14,35 @@ import org.mockito.stubbing.ValidableAnswer;
 import static org.mockito.internal.exceptions.Reporter.cannotStubWithNullThrowable;
 import static org.mockito.internal.exceptions.Reporter.checkedExceptionInvalid;
 
+/**
+ * An answer that always throws the same throwable.
+ */
 public class ThrowsException implements Answer<Object>, ValidableAnswer, Serializable {
 
     private static final long serialVersionUID = 1128820328555183980L;
     private final Throwable throwable;
     private final ConditionalStackTraceFilter filter = new ConditionalStackTraceFilter();
 
+    /**
+     * Creates a new answer always throwing the given throwable. If it is null,
+     * {@linkplain ValidableAnswer#validateFor(InvocationOnMock) answer validation}
+     * will fail.
+     */
     public ThrowsException(Throwable throwable) {
         this.throwable = throwable;
     }
 
     public Object answer(InvocationOnMock invocation) throws Throwable {
+        if (throwable == null) {
+            throw new IllegalStateException("throwable is null: " +
+                "you shall not call #answer if #validateFor fails!");
+        }
         if (MockUtil.isMock(throwable)) {
             throw throwable;
         }
-        Throwable t = throwable.fillInStackTrace();
-
-        if (t == null) {
-            throw throwable;
-        }
-        filter.filter(t);
-        throw t;
+        throwable.fillInStackTrace();
+        filter.filter(throwable);
+        throw throwable;
     }
 
     @Override
--- a/src/main/java/org/mockito/internal/stubbing/answers/ThrowsException.java
+++ b/src/main/java/org/mockito/internal/stubbing/answers/ThrowsException.java
@@ -40,9 +40,14 @@ public class ThrowsException implements Answer<Object>, ValidableAnswer, Seriali
         if (MockUtil.isMock(throwable)) {
             throw throwable;
         }
-        throwable.fillInStackTrace();
-        filter.filter(throwable);
-        throw throwable;
+        Throwable t = throwable.fillInStackTrace();
+
+        if (t == null) {
+            //Custom exceptions sometimes return null, see #866
+            throw throwable;
+        }
+        filter.filter(t);
+        throw t;
     }
 
     @Override
--- a/src/test/java/org/mockito/internal/stubbing/answers/ThrowsExceptionTest.java
+++ b/src/test/java/org/mockito/internal/stubbing/answers/ThrowsExceptionTest.java
@@ -28,62 +28,6 @@ public class ThrowsExceptionTest {
         }
     }
 
-    /**
-     * Tests that ThrowsException raises the same instance as passed in the constructor
-     * in case of a well-behaved Throwable implementation.
-     *
-     * @see #should_raise_same_wanted_throwable_even_if_it_returns_null_in_fill_in_stack_trace()
-     */
-    @Test
-    public void should_raise_same_wanted_throwable_on_subsequent_invocations() {
-        Throwable expectedThrowable = new IllegalStateException("An expected throwable");
-        ThrowsException throwingAnswer = new ThrowsException(expectedThrowable);
-        int numInvocations = 2;
-        for (int i = 0; i < numInvocations; i++) {
-            try {
-                throwingAnswer.answer(createMethodInvocation());
-                Assertions.fail("Should have raised wanted exception");
-            } catch (Throwable throwable) {
-                // All invocations must result in the same throwable as passed to the constructor.
-                assertThat(expectedThrowable).isSameAs(expectedThrowable);
-            }
-        }
-    }
-
-    /**
-     * Tests that a custom exception that breaks the Throwable contract returning null
-     * in fillInStackTrace does not break ThrowsException#answer.
-     *
-     * Any Throwable implementation must always return a reference to this
-     * from #fillInStackTrace.
-     *
-     * @see Throwable#fillInStackTrace()
-     * @see <a href="https://github.com/mockito/mockito/issues/866">#866</a>
-     */
-    @Test
-    public void should_raise_same_wanted_throwable_even_if_it_returns_null_in_fill_in_stack_trace() {
-        /*
-         * An evil throwable implementation that breaks Throwable#fillInStackTrace contract.
-         *
-         * We can't use a mock here because ThrowsException#answer handles mocks of exceptions
-         * differently.
-         */
-        class EvilThrowable extends Throwable {
-            @Override public synchronized Throwable fillInStackTrace() {
-                return null;
-            }
-        }
-
-        Throwable expectedThrowable = new EvilThrowable();
-        ThrowsException throwingAnswer = new ThrowsException(expectedThrowable);
-        try {
-            throwingAnswer.answer(createMethodInvocation());
-            Assertions.fail("should have raised wanted exception");
-        } catch (Throwable throwable) {
-            assertThat(throwable).isSameAs(expectedThrowable);
-        }
-    }
-
     @Test
     public void should_throw_mock_exception_without_stacktrace() throws Exception {
         try {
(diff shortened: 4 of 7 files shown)


src/test/java/org/mockitousage/stubbing/StubbingConsecutiveAnswersTest.java
```diff
@@ -41,6 +41,34 @@ public void should_return_consecutive_values_for_two_nulls() throws Exception {
         assertNull(mock.simpleMethod());
     }
 
+    @Test
+    public void should_return_consecutive_values_first_var_arg_null() throws Exception {
```
mockitoguy: This is the current behavior, correct? E.g. your PR did not change it?

src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
```diff
@@ -24,7 +24,7 @@
     public OngoingStubbing<T> thenReturn(T value, T... values) {
         OngoingStubbing<T> stubbing = thenReturn(value);
         if (values == null) {
-            // TODO below does not seem right
+            // Return null in the 2nd and subsequent invocations.
```
mockitoguy: This comment explains what the code does but does not explain why (which is the point of code comments). Unfortunately, I don't know either :) I think it was a judgement call years ago and we will stick to it to keep compatibility. Thank you for cleaning up the TODO and adding test coverage.

src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
```diff
@@ -56,16 +56,16 @@
...
         return thenThrow(newInstance(throwableType));
     }
 
     @Override
-    public OngoingStubbing<T> thenThrow(Class<? extends Throwable> toBeThrown, Class<? extends Throwable>... nextToBeThrown) {
+    public OngoingStubbing<T> thenThrow(Class<? extends Throwable> toBeThrown,
+                                        Class<? extends Throwable>... nextToBeThrown) {
         if (nextToBeThrown == null) {
-            thenThrow((Class<Throwable>) null);
+            return abortNullExceptionType();
```
mockitoguy: This looks like a change in behavior. What is driving this? We are very careful when making changes in behavior that can cause issues when migrating to newer version of Mockito.

src/main/java/org/mockito/internal/stubbing/StubberImpl.java
```diff
@@ -83,9 +83,9 @@ public Stubber doThrow(Class<? extends Throwable> toBeThrown) {
         Throwable e;
         try {
             e = newInstance(toBeThrown);
-        } catch (RuntimeException instanciationError) {
+        } catch (RuntimeException instantiationError) {
```
mockitoguy: Thanks!

src/main/java/org/mockito/internal/stubbing/answers/ThrowsException.java
```diff
@@ -14,27 +14,35 @@
...
+     */
     public ThrowsException(Throwable throwable) {
         this.throwable = throwable;
     }
 
     public Object answer(InvocationOnMock invocation) throws Throwable {
-        if (MockUtil.isMock(throwable)) {
-            throw throwable;
+        if (throwable == null) {
+            throw new IllegalStateException("throwable is null: " +
```
mockitoguy: This exception message is not consistent with how we handle exceptions. Take a look at a Reporter class. Typically we throw an exception that is or derives from MockitoException. The message needs to be very clean, actionable to the end user. I suggest to create a high level unit test that demonstrates this use case from the standpoint of the end user. Run this test, observe the exception, would a newbie Mockito understand the exception message and was able to resolve the problem quickly?

src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
dmitry-timofeev: I see, thank you! Now I think the TODO was added because it's strange when a _user_ passes a null varargs argument:
```java
thenReturn("foo", (String []) null);
// instead of 
thenReturn("foo", (String) null);
```
If that null array comes from somewhere else (e.g., a method `nextInvocationsArguments`), that almost always a programming error that must be fixed (we can't say if a user really means `{ null }`, or {}).

Will Mockito stick with this decision in 2.x only, or in 3+ too? Will you consider changing that in the next releases? In the latter case I'd keep a TODO, but expand it with explanation _and_ a reference to an issue.

src/main/java/org/mockito/internal/stubbing/answers/ThrowsException.java
dmitry-timofeev: This one is tricky, as it must not occur because of an _external_ precondition that _Mockito_ first validates this answer. If this precondition is met (i.e., Mockito never calls `answer` if `validateFor` fails), it will never fire. Therefore, I can't think of how to yield this exception through _public_ API.

I've checked on 2.x with the following test, `answer` throws an NPE:

```java
    @Test
    public void should_not_answer_null_throwable() throws Throwable {
        try {
            // Shall #validateFor first!
            new ThrowsException(null).answer(new InvocationBuilder().toInvocation());
            Assertions.fail("Should have raised NPE");
        } catch (NullPointerException expected) {}
    }
```

I see such alternatives:
1. Keep an implicit null check. I believe it's not OK because this check is performed on the object state, and not on a user-supplied argument.
2. Add an explicit null check, throw NPE with description.
3. Add an explicit null check, throw IllegalStateException with description.
4. Add an explicit null check, report in a more user-friendly format.
5. Add a `validateFor` check in `answer`. It will _definetely_ catch when validateFor isn't called, but is redundant in case when Mockito always validates this answer (= adheres to the Answer+ValidableAnswer contract).

I'd say (5) is the safest one, but you make the same check twice (externally in the framework code + internally in the class itself). I can't say how expensive that is, or shall we trust the rest of the framework here, what do you think?

It seems `Reporter` is used to supply user-facing exceptions, but this one shall never occur in correct Mockito implementation. How do you report _internal_ errors?

src/test/java/org/mockitousage/stubbing/StubbingConsecutiveAnswersTest.java
dmitry-timofeev: Exactly, I've added to confirm that.

src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
dmitry-timofeev: I considered how to clarify that in the case when any argument is null, we abort the stubbing.

It doesn't change _observable_ behavior, because `thenThrow(Class<? extends Throwable> throwableType)` does the same if null is passed (that was extracted in `abortNullExceptionType`).

If any code is added in `thenThrow(Class<? extends Throwable> throwableType)` _[before](https://github.com/mockito/mockito/pull/1324/files/84332358ad08f225c9bb293e5200792046d57ed7#diff-01345510e64bc217eb06fbf16e77739dR58)_ a null check, the behavior will be different.

dmitry-timofeev: If you think it's worth it, I'd do the same to  [`thenThrow(Throwable.. throwables)`](https://github.com/mockito/mockito/pull/1324/files/84332358ad08f225c9bb293e5200792046d57ed7#diff-01345510e64bc217eb06fbf16e77739dR43).

src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
mockitoguy: Unless there is a strong reason not to, I plan to keep the code as is. Nobody complained about it. There is risk that the change will be incompatible. It's better to focus Mockito 3 on more interesting features!

Thank you for pointing this out and attempting to clear up the TODO!

src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
mockitoguy: Oh I see now, thank you for clarification!

I don't see the value of this change. We are adding one extra method and creating small code duplication.

src/main/java/org/mockito/internal/stubbing/answers/ThrowsException.java
mockitoguy: Oh, you're right, the user will never see it when using the public API. I missed that.

If this is only internal, IllegalStateException is perfect.

src/main/java/org/mockito/internal/stubbing/answers/ThrowsException.java
```diff
@@ -14,27 +14,35 @@
...
         }
-        Throwable t = throwable.fillInStackTrace();
-
-        if (t == null) {
+        if (MockUtil.isMock(throwable)) {
             throw throwable;
         }
-        filter.filter(t);
-        throw t;
+        throwable.fillInStackTrace();
```
mockitoguy: I'd rather keep the assignment for backwards compatibility. I made this change and will push it back to your pull request. Can you contribute a test that preserves this behavior?

src/test/java/org/mockito/internal/stubbing/answers/ThrowsExceptionTest.java
```diff
@@ -59,24 +60,42 @@ public void should_fill_in_exception_stacktrace() throws Exception {
...
+            new ThrowsException(null).validateFor(invocation);
             Assertions.fail("should have raised a MockitoException");
         } catch (MockitoException expected) {}
     }
 
+    @Test
+    public void should_throw_illegal_state_exception_if_null_answer() throws Throwable {
+        Invocation invocation = createMethodInvocation();
+        try {
+            new ThrowsException(null)
```
mockitoguy: Can you contribute a fix that makes it fail early when null passed via constructor? This would be cleaner (earlier feedback).

src/main/java/org/mockito/internal/stubbing/answers/ThrowsException.java
dmitry-timofeev: Oh, that's unexpected, thank you for adding a clarifying comment!

I think I'll add a couple of tests:
- The one that prevents the change I did.
- The one that verifies we get the throwable returned from `fillInStackTrace` in subsequent invocations.

dmitry-timofeev: What's interesting is that the custom exception from #866 breaks the [Throwable#fillInStackTrace contract](https://docs.oracle.com/javase/9/docs/api/java/lang/Throwable.html#fillInStackTrace--): this method shall always return `this`, never `null` or a reference to another throwable.

src/test/java/org/mockito/internal/stubbing/answers/ThrowsExceptionTest.java
dmitry-timofeev: I actually did it locally earlier, and got failures in some integration tests. It seems the constructor has to allow null throwable so that we don't get UnfinishedStubbingException (I certainly can't say if that shall or can be fixed). Here is an example of failing test output:
```
org.mockitousage.stubbing.StubbingWithThrowablesTest > shouldNotAllowSettingNullThrowable FAILED
    java.lang.AssertionError: 
    Expected: (an instance of org.mockito.exceptions.base.MockitoException and exception with message a string containing "Cannot stub with null throwable")
         but: an instance of org.mockito.exceptions.base.MockitoException <org.junit.internal.runners.model.MultipleFailureException: There were 2 errors:
      org.mockito.exceptions.base.MockitoException(
    Cannot stub with null throwable!)
      org.mockito.exceptions.misusing.UnfinishedStubbingException(
    Unfinished stubbing detected here:
    -> at org.mockitousage.stubbing.StubbingWithThrowablesTest.shouldNotAllowSettingNullThrowable(StubbingWithThrowablesTest.java:154)

    E.g. thenReturn() may be missing.
    Examples of correct stubbing:
        when(mock.isOk()).thenReturn(true);
        when(mock.isOk()).thenThrow(exception);
        doThrow(exception).when(mock).someVoidMethod();
    Hints:
     1. missing thenReturn()
     2. you are trying to stub a final method, which is not supported
     3: you are stubbing the behaviour of another mock inside before 'thenReturn' instruction if completed
    )> is a org.junit.internal.runners.model.MultipleFailureException
    Stacktrace was: org.junit.internal.runners.model.MultipleFailureException: There were 2 errors:
      org.mockito.exceptions.base.MockitoException(
    Cannot stub with null throwable!)
      org.mockito.exceptions.misusing.UnfinishedStubbingException(
    Unfinished stubbing detected here:
    -> at org.mockitousage.stubbing.StubbingWithThrowablesTest.shouldNotAllowSettingNullThrowable(StubbingWithThrowablesTest.java:154)

    E.g. thenReturn() may be missing.
    Examples of correct stubbing:
        when(mock.isOk()).thenReturn(true);
        when(mock.isOk()).thenThrow(exception);
        doThrow(exception).when(mock).someVoidMethod();
    Hints:
     1. missing thenReturn()
     2. you are trying to stub a final method, which is not supported
     3: you are stubbing the behaviour of another mock inside before 'thenReturn' instruction if completed
    )
        at org.junit.runners.model.MultipleFailureException.assertEmpty(MultipleFailureException.java:67)
        at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:39)
        at org.junit.rules.ExpectedException$ExpectedExceptionStatement.evaluate(ExpectedException.java:239)
        at org.junit.rules.RunRules.evaluate(RunRules.java:20)
        at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
        at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
        at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
        at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
        at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
        at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
        at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
        at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecuter.runTestClass(JUnitTestClassExecuter.java:114)
        at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecuter.execute(JUnitTestClassExecuter.java:57)
        at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassProcessor.processTestClass(JUnitTestClassProcessor.java:66)
        at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.processTestClass(SuiteTestClassProcessor.java:51)
        at sun.reflect.GeneratedMethodAccessor60.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:35)
        at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
        at org.gradle.internal.dispatch.ContextClassLoaderDispatch.dispatch(ContextClassLoaderDispatch.java:32)
        at org.gradle.internal.dispatch.ProxyDispatchAdapter$DispatchingInvocationHandler.invoke(ProxyDispatchAdapter.java:93)
        at com.sun.proxy.$Proxy1.processTestClass(Unknown Source)
        at org.gradle.api.internal.tasks.testing.worker.TestWorker.processTestClass(TestWorker.java:108)
        at sun.reflect.GeneratedMethodAccessor59.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:35)
        at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
        at org.gradle.internal.remote.internal.hub.MessageHubBackedObjectConnection$DispatchWrapper.dispatch(MessageHubBackedObjectConnection.java:146)
        at org.gradle.internal.remote.internal.hub.MessageHubBackedObjectConnection$DispatchWrapper.dispatch(MessageHubBackedObjectConnection.java:128)
        at org.gradle.internal.remote.internal.hub.MessageHub$Handler.run(MessageHub.java:404)
        at org.gradle.internal.concurrent.ExecutorPolicy$CatchAndRecordFailures.onExecute(ExecutorPolicy.java:63)
        at org.gradle.internal.concurrent.ManagedExecutorImpl$1.run(ManagedExecutorImpl.java:46)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at org.gradle.internal.concurrent.ThreadFactoryImpl$ManagedThreadRunnable.run(ThreadFactoryImpl.java:55)
        at java.lang.Thread.run(Thread.java:748)

```

dmitry-timofeev:
@mockitoguy , would you please consider this patch
```
Index: src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/main/java/org/mockito/internal/stubbing/BaseStubbing.java	(revision f44a4eed8dbfc06bb832be98ab169ab4237b1b47)
+++ src/main/java/org/mockito/internal/stubbing/BaseStubbing.java	(date 1520970673000)
@@ -66,7 +66,7 @@
     @Override
     public OngoingStubbing<T> thenThrow(Class<? extends Throwable> toBeThrown, Class<? extends Throwable>... nextToBeThrown) {
         if (nextToBeThrown == null) {
-            thenThrow((Class<Throwable>) null);
+            return thenThrow((Class<Throwable>) null);
         }
         OngoingStubbing<T> stubbing = thenThrow(toBeThrown);
         for (Class<? extends Throwable> t : nextToBeThrown) {
```
so that readers do not get 'nextToBeThrown can be null' warning from their IDEs?

src/test/java/org/mockito/internal/stubbing/answers/ThrowsExceptionTest.java
```diff
@@ -28,6 +28,62 @@ public void should_raise_wanted_throwable() throws Throwable {
...
+     * in fillInStackTrace does not break ThrowsException#answer.
+     *
+     * Any Throwable implementation must always return a reference to this
+     * from #fillInStackTrace.
+     *
+     * @see Throwable#fillInStackTrace()
+     * @see <a href="https://github.com/mockito/mockito/issues/866">#866</a>
+     */
+    @Test
+    public void should_raise_same_wanted_throwable_even_if_it_returns_null_in_fill_in_stack_trace() {
```
mockitoguy: Isn't it already tested in NPEWhenCustomExceptionStackTraceReturnNullTest?

src/test/java/org/mockito/internal/stubbing/answers/ThrowsExceptionTest.java
```diff
@@ -28,6 +28,62 @@ public void should_raise_wanted_throwable() throws Throwable {
...
     }
 
+    /**
+     * Tests that ThrowsException raises the same instance as passed in the constructor
+     * in case of a well-behaved Throwable implementation.
+     *
+     * @see #should_raise_same_wanted_throwable_even_if_it_returns_null_in_fill_in_stack_trace()
+     */
+    @Test
+    public void should_raise_same_wanted_throwable_on_subsequent_invocations() {
```
mockitoguy: I like this test. Can you check if we have top level integration tests for those use cases? I did a quick search and I think we have some gaps.

I'm wondering if top level tests under 'org.mockitousage.*' packages aren't better to verify and document this behavior. ThrowsExceptionTest is becoming big and complicated and it does seem to repeat the coverage of the other tests. Thoughts?

mockitoguy:
>so that readers do not get 'nextToBeThrown can be null' warning from their IDEs?

Good idea! Can you make this change?

src/test/java/org/mockito/internal/stubbing/answers/ThrowsExceptionTest.java
dmitry-timofeev: It is, shall I remove this unit test or expand the IT in `NPEWhenCustomExceptionStackTraceReturnNullTest`?

src/test/java/org/mockito/internal/stubbing/answers/ThrowsExceptionTest.java
dmitry-timofeev: There is a similar integration test: `org.mockitousage.stubbing.StubbingConsecutiveAnswersTest#should_throw_consecutively`

Shall I remove this one?

src/test/java/org/mockito/internal/stubbing/answers/ThrowsExceptionTest.java
dmitry-timofeev: As this test passed, the code 
```
throwable.fillInStackTrace();
filter.filter(throwable);
throw throwable;
```
didn't break #866 , did it?

src/test/java/org/mockitousage/bugs/FillInStackTraceScenariosTest.java
```diff
@@ -12,25 +12,40 @@
...
+    private class SomeException extends RuntimeException {}
 
     class NullStackTraceException extends RuntimeException {
-        @Override
         public Exception fillInStackTrace() {
             return null;
         }
     }
 
+    class NewStackTraceException extends RuntimeException {
```
dmitry-timofeev: Shan't we add a comment that these implementations of `fillInStackTrace` are totally incorrect -- they violate the method contract? Any Throwable implementation must always return a reference to `this` from `fillInStackTrace`.

mockitoguy: Go ahead! I wanted to preserve current behavior in a test.

src/main/java/org/mockito/internal/stubbing/answers/ThrowsException.java
```diff
@@ -14,23 +14,36 @@
...
+ */
 public class ThrowsException implements Answer<Object>, ValidableAnswer, Serializable {
 
     private static final long serialVersionUID = 1128820328555183980L;
     private final Throwable throwable;
     private final ConditionalStackTraceFilter filter = new ConditionalStackTraceFilter();
 
+    /**
+     * Creates a new answer always throwing the given throwable. If it is null,
+     * {@linkplain ValidableAnswer#validateFor(InvocationOnMock) answer validation}
```
TimvdLippe: With `linkplain` here instead of `link`?

PR #1327, opened 2018-03-01
Fixes #1326 : Reduce the allocation rate for the typical use of Locat…

### LINKED ISSUE #1326: Reduce memory consumption of the typical LocationImpl
I recently noticed very heavy memory usage on our testsuite (which is using mockito) so I started profiling.

I eventually found #1313 but that's already being worked on. Also spotted a low hanging fruit: our testsuite is allocating about 6 *million* instances of `LocationImpl`.
It looks like I can't avoid that easily, however each `LocationImpl` also allocates a new `StackTraceFilter`, which can be easily avoided when the default is being applied; sending a PR shortly.

This should save us about 6 million empty and pointless `StackTraceFilter` :)

Sanne:
…ionImpl

A trivial one, but should be very useful as I noticed this class being instantiated quite a lot.

### DIFF
--- a/src/main/java/org/mockito/internal/debugging/LocationImpl.java
+++ b/src/main/java/org/mockito/internal/debugging/LocationImpl.java
@@ -11,11 +11,14 @@ import org.mockito.invocation.Location;
 public class LocationImpl implements Location, Serializable {
 
     private static final long serialVersionUID = -9054861157390980624L;
+    //Limit the amount of objects being created, as this class is heavily instantiated:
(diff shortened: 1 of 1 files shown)


Sanne:
Will do, thanks! Another strong memory consumer is the `Throwable` stored in this same class as in most cases it seems empty, yet has an array of 10 elements prepared for filling in the stack.

I didn't change that as I'm not sure when this would be safe to do.

PR #1339, opened 2018-03-22
Enable errorprone build

moltmann:
- fix up issues found
- placate error prone warning where we know we do something funky
- Ignore two tests as it is not immediately clear what is broken
  - org.mockitousage.bugs.varargs.VarargsNotPlayingWithAnyObjectTest#shouldNotAllowUsingAnyObjectForVarArgs
  - org.mockito.internal.util.PrimitivesTest#should_check_that_value_type_is_assignable_to_wrapper_reference

NOTE: Please carefully review the changes made to the code. I fixed them up as good as I could but a real mockito expert need to verify them.

### DIFF
--- a/src/main/java/org/mockito/internal/util/Primitives.java
+++ b/src/main/java/org/mockito/internal/util/Primitives.java
@@ -44,7 +44,7 @@ public class Primitives {
 
     public static boolean isAssignableFromWrapper(Class<?> valueClass, Class<?> referenceType) {
         if(isPrimitiveOrWrapper(valueClass) && isPrimitiveOrWrapper(referenceType)) {
-            return Primitives.primitiveTypeOf(valueClass).isAssignableFrom(referenceType);
+            return Primitives.primitiveTypeOf(valueClass).isAssignableFrom(Primitives.primitiveTypeOf(referenceType));
         }
         return false;
     }
--- a/src/test/java/org/mockito/internal/util/PrimitivesTest.java
+++ b/src/test/java/org/mockito/internal/util/PrimitivesTest.java
@@ -4,6 +4,7 @@
  */
 package org.mockito.internal.util;
 
+import org.junit.Ignore;
 import org.junit.Test;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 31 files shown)


src/main/java/org/mockito/internal/util/Primitives.java
```diff
@@ -44,7 +44,7 @@ public static boolean isPrimitiveOrWrapper(Class<?> type) {
 
     public static boolean isAssignableFromWrapper(Class<?> valueClass, Class<?> referenceType) {
         if(isPrimitiveOrWrapper(valueClass) && isPrimitiveOrWrapper(referenceType)) {
-            return Primitives.primitiveTypeOf(valueClass).isAssignableFrom(referenceType);
+            return Primitives.primitiveTypeOf(valueClass).isAssignableFrom(Primitives.primitiveTypeOf(referenceType));
```
TimvdLippe: This seemed to be the most sensible fix. It did not break any other test :thinking:

mockitoguy: If it does not break tests, we're good ;) This strategy applies to 99% of our codebase. We have few weak spots that are not covered with tests (like OSGi bundle configuration), but primitives handling should be good.

Thanks for checking!

src/test/java/org/mockito/internal/util/PrimitivesTest.java
```diff
@@ -69,21 +69,21 @@ public void should_return_null_for_everything_else() throws Exception {
...
-        assertThat(Primitives.isAssignableFromWrapper(Float.class, Float.class));
-        assertThat(Primitives.isAssignableFromWrapper(char.class, Character.class));
-        assertThat(Primitives.isAssignableFromWrapper(Character.class, Character.class));
-        assertThat(Primitives.isAssignableFromWrapper(short.class, Short.class));
-        assertThat(Primitives.isAssignableFromWrapper(Short.class, Short.class));
-        assertThat(Primitives.isAssignableFromWrapper(byte.class, Byte.class));
-        assertThat(Primitives.isAssignableFromWrapper(Byte.class, Byte.class));
-        assertThat(Primitives.isAssignableFromWrapper(boolean.class, Boolean.class));
-        assertThat(Primitives.isAssignableFromWrapper(Boolean.class, Boolean.class));
+        assertThat(Primitives.isAssignableFromWrapper(int.class, Integer.class)).isTrue();
```
mockitoguy: So previously we didn't have any assertions here, right? Nice :)

GREAT CATCH, thanks!

src/test/java/org/mockitousage/bugs/varargs/VarargsNotPlayingWithAnyObjectTest.java
```diff
@@ -35,18 +37,6 @@ public void shouldMatchAnyVararg() {
         verify(mock, never()).run(anyString(), eq("f"));
     }
 
-    //we cannot use anyObject() for entire varargs because it makes the verification pick up extra invocations
```
mockitoguy: Are you removing the test because errorprone does not like empty catch statement? Hmm, it is not a good (even though I love removing code :). If you propose to remove it, please validate carefully and ensure we're not reducing coverage.

From my initial look, the test is valid, even though it covers a corner case. I suggest to either suppress the warning or use AssertJ's assertThrownBy

src/test/java/org/mockitousage/matchers/MoreMatchersTest.java
```diff
@@ -48,12 +48,14 @@ public void any_class_should_be_actual_alias_to_isA() {
         mock.simpleMethod((String) null);
         try {
             verify(mock).simpleMethod(isA(String.class));
-            fail();
-        } catch (AssertionError ignored) { }
-        try {
-            verify(mock).simpleMethod(any(String.class));
-            fail();
-        } catch (AssertionError ignored) { }
+        } catch (AssertionError ignored) {
```
mockitoguy: The resulting code is less readable. I suggest AssertJ or suppression.

src/test/java/org/mockitousage/verification/BasicVerificationTest.java
```diff
@@ -47,8 +47,11 @@ public void shouldFailVerificationOnMethodArgument() throws Exception {
         verify(mock).clear();
         try {
             verify(mock).add("bar");
-            fail();
-        } catch (AssertionError expected) {}
+        } catch (AssertionError expected) {
```
mockitoguy: Empty catch again? :) AssertJ or suppress will work better for us I think.

src/test/java/org/mockitousage/bugs/varargs/VarargsNotPlayingWithAnyObjectTest.java
TimvdLippe: It was because we were actually not asserting anything in this case. The reason is that the `fail()` statement was always executed, because `verify(mock).run((String[]) anyObject());` is completely valid and works. This would then throw an `AssertionError`, which is consequently caught in the `try-catch` in the test case. In other words, whether the `fail` is reached or not, we would always catch the error and the test would never fail (even if there is a bug). Errorprone is correct in identifying that this is an issue.

src/test/java/org/mockitousage/matchers/MoreMatchersTest.java
moltmann: done. Will look nicer once we can use lambdas

src/test/java/org/mockitousage/verification/BasicVerificationTest.java
moltmann: done. Will look nicer once we can use lambdas

src/test/java/org/mockitousage/bugs/varargs/VarargsNotPlayingWithAnyObjectTest.java
TimvdLippe: @mockitoguy Any update regarding this comment?

mockitoguy: Oh I see! Thank you for clarification!

In that case we should absolutely clean up the test method. I'll push a commit with a desired fix and merge the PR.

Thank you for digging deep into this one. I missed the flaw in this test. Clean tests for the win!!!

mockitoguy:
We're ready. I changed and kept the test that you identified as flawed (thanks for that!).

When Travis is green I'll merge.

PR #1340, opened 2018-03-23
Update to Byte Buddy 1.8.0

PascalSchumacher:


raphw:
I was just on it but this PR is all that is necessary. Good to merge. And yes, ASM 6.1 is required, Byte Buddy relies on its API.

PR #1341, opened 2018-03-23
Update to ASM 6.1

PascalSchumacher:


raphw:
This is also good to go.

PR #1344, opened 2018-03-25
junit jupiter extension dependencies

epeee:
- we should be fine with the api for implementing the extension, therefore let's only depend on junitJupiterApi
 - change junitJupiterEngine dependency to testRuntime
 - rename junitJupiterLauncher to junitPlatformLauncher


epeee:
based on feedback in #1221 (e.g. [this](https://github.com/mockito/mockito/pull/1221#discussion_r146949185) and [this](https://github.com/mockito/mockito/pull/1221#discussion_r176917572) discussion).

PR #1349, opened 2018-03-26
Fix #1346 - Different mocks are used for @Mock and @InjectMock in the same test class with JUnit 5 extension

### LINKED ISSUE #1346: Different mocks are used for @Mock and @InjectMock in the same test class with JUnit 5 extension
I've been playing with the new JUnit 5 support in Mockito 2.17.0. I have found that with `@Mock` and `@InjectMocks` used in the same test class different mocks are injected. I would expect to have the same mock reused (as it works with JUnit 4 or manual `@Mock` initialization).

Sample code:
```
public class SpaceShip5BugMiniTest {

    @InjectMocks
    private SpaceShip spaceShip;

    @Mock
    private TacticalStation tacticalStation;

    @Test
    public void shouldInjectMocks() {
        assertThat(tacticalStation).isNotNull();
        assertThat(spaceShip.getTacticalStation()).isEqualTo(tacticalStation);  
        //Bug ↑ - different mocks are use in @Mock and @InjectMock
    }
}
```

Error message:
```
org.opentest4j.AssertionFailedError: 
Expecting:
 <"tacticalStation (TacticalStation$MockitoMock$991606095@61d6015a)">
to be equal to:
 <"tacticalStation (TacticalStation$MockitoMock$991606095@b62d79)">
but was not.
Expected :tacticalStation
Actual   :tacticalStation
```

The minimal (non-)working example is available [here](https://github.com/szpak/code-examples-and-poc/tree/mockito2junit5-bug/mockito2-junit5).

@TimvdLippe Do you have an idea what could be wrong?

Mockito 2.17.0 (2.17.1 is not being released due to (https://github.com/mockito/shipkit/issues/679), junit-jupiter 5.1.0, OpenJDK 1.8.0_161, Linux.

### Checklist:

SeriyBg:
Fix #1346 - fix `MockitoExtension` which has init mocks with two identical test instances which cause the mock to be initialized and injected two times

### DIFF
--- a/subprojects/junit-jupiter/src/test/java/org/mockitousage/JunitJupiterTest.java
+++ b/subprojects/junit-jupiter/src/test/java/org/mockitousage/JunitJupiterTest.java
@@ -4,14 +4,14 @@
  */
 package org.mockitousage;
 
+import java.util.function.Function;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
 import org.mockito.Mock;
 import org.mockito.Mockito;
 import org.mockito.junit.jupiter.MockitoExtension;
-
-import java.util.function.Function;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
@@ -20,6 +20,9 @@ class JunitJupiterTest {
 
     @Mock
     private Function<Integer, String> rootMock;
+
+    @InjectMocks
+    private ClassWithDependency classWithDependency;
 
     @Test
     void ensureMockCreationWorks() {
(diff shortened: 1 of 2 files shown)


subprojects/junit-jupiter/src/test/java/org/mockitousage/JunitJupiterTest.java
```diff
@@ -70,4 +73,17 @@ void shouldWeCreateMocksInTheParentContext() {
             assertThat(rootMock).isNotNull();
         }
     }
+
+    @Test
+    void should_be_injected_correct_instance_of_mock() {
+        assertThat(classWithDependency.dependency).isEqualTo(rootMock);
```
TimvdLippe: Lets use `isSame` here, as we literally want this object to be there.

SeriyBg: Good catch, have changed it. Thanks!

SeriyBg: I have amended the previous commit as it's a simple and small change

davinkevin:
Sadly, this doesn't cover the case where we are using `@TestInstance(TestInstance.Lifecycle.PER_CLASS)`

> https://github.com/mockito/mockito/issues/1437

PR #1350, opened 2018-03-26
Support local parameters in test methods with JUnit Jupiter

### LINKED ISSUE #1348: Support @Mock injection in JUnit 5 method parameters
JUnit 5 has the ability to "inject" parameters into test methods via an extensible ``ParameterResolver`` interface.  The JUnit 5 users guide provides an example that shows how such a ``ParameterResolver`` can supply mock objects as test parameters as follows (shamelessly copied from https://junit.org/junit5/docs/current/user-guide/#writing-tests-dependency-injection):

```
class MyMockitoTest {

    @BeforeEach
    void init(@Mock Person person) {
        when(person.getName()).thenReturn("Dilbert");
    }

    @Test
    void simpleTestWithInjectedMock(@Mock Person person) {
        assertEquals("Dilbert", person.getName());
    }

}
```

The prototype ``MockitoExtension`` provided in the JUnit 5 samples project shows a simple implementation of the required ``supportsParameter()`` and ``resolveParameter()`` methods. (See: https://github.com/junit-team/junit5-samples/blob/r5.1.0/junit5-mockito-extension/src/main/java/com/example/mockito/MockitoExtension.java)

[ ]   Add the ability to inject mock objects into test method parameters to the official ``MockitoExtension``.  References #445

TimvdLippe:
Note that this implementation differs from the prototype implementation of the JUnitTeam (https://github.com/junit-team/junit5-samples/blob/7bf40178345d5ca837579c8ddb8c025401a98788/junit5-mockito-extension/src/main/java/com/example/mockito/MockitoExtension.java#L41-L73). Instead, it will do not do any parameter resolution between test methods.

Fixes #1348


sbrannen:
> Instead, it will do not do any parameter resolution between test methods.

I always thought that was a _cool_ feature of the original prototype, but I can understand your rationale for omitting that feature; however, I'd make sure it's very well documented in the user guide, etc. for Mockito.

Otherwise, users will likely get confused quite quickly.

sbrannen:
Regarding that latter, in the prototype we added explicit support for determining the mock name here:

https://github.com/junit-team/junit5-samples/blob/7bf40178345d5ca837579c8ddb8c025401a98788/junit5-mockito-extension/src/main/java/com/example/mockito/MockitoExtension.java#L64

sbrannen:
> I think the javadoc of MockitoExtension is quite explicit about this, do you have any suggestions to extend it?

Well, as the code is currently implemented, `supportsParameter()` will return `true` for any parameter annotated with `@Mock`, whether such a parameter is in a _constructor_, `@BeforeAll`/@AfterAll`, `@BeforeEach`/`@AfterEach`, `@Test` method etc.

And it _sounds_ like you wish to restrict usage to `@Test`, `@TestFactory`, `@ParameterizedTest`, and `@RepeatedTest` methods.

Or maybe you don't want to restrict it?

But it's unclear what is supported and what isn't.

I think your goal is restrict sharing of mock instances between setup and test methods. So if you might want to restrict mock injection for lifecycle callback methods.

On the other hand, it would be a perfectly valid use case to have a mock injected into a constructor and then store it in a `private final` field (i.e., immutability).

In summary, you'll need to think through the possible use cases and document which ones you actually support, and... the implementation should actually only return `true` from `supportsParameter()` if Mockito actually supports the use case.

TimvdLippe:
Oh I was not aware of all these cases, I thought it was only for test methods. Have to think about what we do want to support. (Maybe it all magically works already with this implementation)

sbrannen:
Sure.... I'd recommend you _play around_ with it a bit to get a feel for what's possible.

Then you'll be better equipped to make an informed decision. 😉

TWiStErRob:
+1 for sharing, I think `@BeforeEach void f(@Mock x)` `@Test void f(@Mock x)` could be the same object. @TimvdLippe as I read you're clearing the session after each test, so sharing between methods would affect a single test method execution.

TimvdLippe:
@TWiStErRob Please see https://github.com/mockito/mockito/issues/1348#issuecomment-376295133 for the full explanation

sbrannen:
@TimvdLippe,

> Maybe it all magically works already with this implementation

Yes, that should be the case.

But I still think the Mockito Team should provide guidance for end users demonstrating best practices when using the `MockitoExtension` for JUnit Jupiter.

For example, if it were my project, I'd make sure that there are tests in place and documented examples demonstrating parameter injection for constructors, lifecycle methods, testable methods (e.g., `@Test`, `@TestFactory`, etc.), _and_ I'd cover the use of `@TestInstance(PER_CLASS)` semantics as well.

I'm sure that either I or someone else from the JUnit team would be happy to review your work before you publish it. 😉

TimvdLippe:
For reference https://github.com/junit-team/junit5/issues/1345 is the issue that is blocking this PR for now, but the upstream issue is actively being worked on atm.

mockitoguy:
Thank you guys for working on this!

sbrannen:
@TimvdLippe, I would't hold off on working on this PR just to wait on JUnit Jupiter 5.2 to be released.

That will literally take months. Plus it would prevent Mockito users from using the feature with JUnit Jupiter 5.1.x (unless we backport the convenience methods to 5.1.x).

For example, in Spring Framework 5.0.5 (likely being released today), I already added [custom support](https://github.com/spring-projects/spring-framework/commit/9244090ba04011902568aba4b854668312c025e1) for the JDK bug so that Spring users aren't affected by the issue even on JUnit Jupiter 5.0.x.

TimvdLippe:
> unless we backport the convenience methods to 5.1.x

As a user, I would strongly appreciate a backport. Have to think whether we would like to maintain this code ourselves (like Spring does) or whether we simply note the limitation that this will be addressed in the next release of JUnit.

TimvdLippe:
Per https://github.com/junit-team/junit5/issues/1345#issuecomment-378632027 this will be backported to JUnit 5.1.1. We just have to wait till that release is out and then I can finish up this PR :tada:

TimvdLippe:
Awesome, constructor parameter resolution has been fixed! @mockito/core PTAL

sbrannen:
👍

ammerritt:
Will this be getting released soon?  Thank you!

TimvdLippe:
This is still awaiting review of @mockito/core

mockitoguy:
I finished reviewing and I'm happy to merge! There's some issue with our Travis build at the moment but we should merge soon!

mockitoguy:
I rebased and wait for Travis feedback before merging.

PR #1351, opened 2018-03-28
Update Byte Buddy and ASM for bug fix releases

raphw:
This updates a few minor regressions that came with the updates of ASM and Byte Buddy.

In particular, this fixes a bug with the processing of parameter annotations caused by an API change in ASM. This also introduces a new handling for stack map frames in advice code that is used by the inline mock maker which can now translate inconsistent frames, typically caused by non-javac compilers such as Scala and Kotlin and by code obfuscators.

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -4,7 +4,7 @@ ext {
 
 def versions = [:]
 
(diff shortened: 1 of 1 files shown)


mockitoguy:
Thank you!

PR #1355, opened 2018-03-31
Use java.lang.MethodHandles.Lookup::defineClass 

raphw:
Use java.lang.MethodHandles.Lookup::defineClass if this mechanism is available on the current JVM. This might not work when interoperating with foreign modules, if those modules do not open and export their types, most prominently, the Java core modules where there are introduced issues with serialization.

This introduces a change in behavior for Mockito when using Java 9+. Unfortunately, I think that this is necessary for Mockito to continue to function after introduction of the JPMS. For now, it seems like this limitation only affects working with serializability in combination with platform types what is not too bad but there is a chance that we will discover issues once this new mechanism is picked up by the community.

### DIFF
--- a/src/main/java/org/mockito/codegen/InjectionBase.java
+++ b/src/main/java/org/mockito/codegen/InjectionBase.java
@@ -4,6 +4,10 @@
  */
 package org.mockito.codegen;
 
+/**
+ * This class is required to resolve a method handle lookup for the {@code org.mockito.codegen} package what requires a preexisting class for the package.
+ * By defining this class, the JVM (starting from Java 9) assures that this package is a part of the Mockito module such that we gain full access rights.
+ */
 public class InjectionBase {
 
     private InjectionBase() {
--- a/src/main/java/org/mockito/codegen/InjectionBase.java
+++ b/src/main/java/org/mockito/codegen/InjectionBase.java
@@ -1,0 +1,12 @@
+/*
+ * Copyright (c) 2007 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.codegen;
+
+public class InjectionBase {
+
+    private InjectionBase() {
+        throw new UnsupportedOperationException();
+    }
+}
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassInjectionLoader.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassInjectionLoader.java
@@ -20,7 +20,7 @@ class SubclassInjectionLoader implements SubclassLoader {
     private final SubclassLoader loader;
 
     SubclassInjectionLoader() {
-        if (!Boolean.getBoolean("org.mockito.reflection.disabled") && ClassInjector.UsingReflection.isAvailable()) {
+        if (!Boolean.getBoolean("org.mockito.internal.simulateJava11") && ClassInjector.UsingReflection.isAvailable()) {
             this.loader = new WithReflection();
         } else if (ClassInjector.UsingLookup.isAvailable()) {
             this.loader = tryLookup();
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassInjectionLoader.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassInjectionLoader.java
@@ -4,12 +4,84 @@
  */
 package org.mockito.internal.creation.bytebuddy;
 
+import net.bytebuddy.dynamic.loading.ClassInjector;
 import net.bytebuddy.dynamic.loading.ClassLoadingStrategy;
+import org.mockito.codegen.InjectionBase;
+import org.mockito.exceptions.base.MockitoException;
+import org.mockito.mock.SerializableMode;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+import static org.mockito.internal.util.StringUtil.join;
 
 class SubclassInjectionLoader implements SubclassLoader {
 
+    private final boolean disableReflection;
+
+    private final Object lookup, codegenLookup;
+
+    private final Method privateLookupIn;
+
+    public SubclassInjectionLoader() {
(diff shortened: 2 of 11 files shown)


mockitoguy:
Will review shortly. Thank you!

src/test/java/org/mockitousage/serialization/DeepStubsSerializableTest.java
```diff
@@ -33,6 +35,8 @@ public void should_serialize_and_deserialize_mock_created_with_deep_stubs() thro
 
     @Test
     public void should_serialize_and_deserialize_parameterized_class_mocked_with_deep_stubs() throws Exception {
+        assumeTrue(ClassFileVersion.ofThisVm().isLessThan(ClassFileVersion.JAVA_V9));
```
mockitoguy: When assumption fails it emits an exception that is ignored by the runner (effectively, later part of this test method is not executed)?

src/test/java/org/mockitousage/serialization/DeepStubsSerializableTest.java
```diff
@@ -47,6 +51,8 @@ public void should_serialize_and_deserialize_parameterized_class_mocked_with_dee
 
     @Test(expected = ClassCastException.class)
     public void should_discard_generics_metadata_when_serialized_then_disabling_deep_stubs_with_generics() throws Exception {
```
mockitoguy: Let's add a reverse test that shows developer experience when we're running on Java9 and try to use serializable deep stubs. This will cover the exception handling logic.

raphw:
This change is driven by Java 11's removal of Unsafe::defineClass. Byte Buddy's injection strategy relies on this method what means that it is no longer available in Java 11. Other codegeneration libraries will stop working completely, such as cglib.

Instead, Mockito must now rely on MethodHandles.Lookup::defineClass. This requires a change as the method MethodHandles::lookup is call site sensitive. This means that it must be called by Mockito and be provided to Byte Buddy using the Lookup-injection strategy.

Unsafe::defineClass was - as the name suggests - unsafe and allowed about anything. With lookups, there are a few constraints. For example, we cannot just inject classes into the system class loader but this needs to happen in an existing package. 

As for that, I found a solution to again allow for the serialization of system classes. Instead of using a synthetic package `codegen` we now use a package `org.mockito.codgen` that already defines a type. Doing so, we can inject classes into this package such that the serialization support works as it did before.

I agree with your argument that we should use the more powerful injection if it is available. From Java 11, Mockito does however require opening all packages that contain types that are mocked to function as it does today.

src/test/java/org/mockitousage/serialization/DeepStubsSerializableTest.java
raphw: This is obsolete as serialization works again as it did before.

src/main/java/org/mockito/internal/creation/bytebuddy/SubclassInjectionLoader.java
```diff
@@ -4,12 +4,64 @@
...
+import static org.mockito.internal.util.StringUtil.join;
 
 class SubclassInjectionLoader implements SubclassLoader {
 
+    private final Object lookup;
+
+    private final Method privateLookupIn;
+
+    public SubclassInjectionLoader() {
+        Object lookup;
```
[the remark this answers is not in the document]
raphw: Added that.

raphw:
To bad we cannot run this on Java 11 due to Gradle. To try this, just use Java 9 and comment out the injection strategy part such that is always uses method handles.

src/main/java/org/mockito/internal/creation/bytebuddy/SubclassInjectionLoader.java
```diff
@@ -4,12 +4,100 @@
...
+
+    private static SubclassLoader tryLookup() {
+        try {
+            Class<?> methodHandles = Class.forName("java.lang.invoke.MethodHandles");
+            Object lookup = methodHandles.getMethod("lookup").invoke(null);
+            Method privateLookupIn = methodHandles.getMethod("privateLookupIn", Class.class, Class.forName("java.lang.invoke.MethodHandles$Lookup"));
+            Object codegenLookup = privateLookupIn.invoke(null, InjectionBase.class, lookup);
+            return new WithLookup(lookup, codegenLookup, privateLookupIn);
+        } catch (Exception ignored) {
+            throw mockingImpossible();
```
TimvdLippe: Lets add the exception in the trace here. That makes sure we can understand the root cause

raphw:
@mockitoguy Changes are looking good! Could you merge and release (to Maven Central)?

src/main/java/org/mockito/codegen/InjectionBase.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (c) 2007 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.codegen;
+
+public class InjectionBase {
```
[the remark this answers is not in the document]
raphw: Added doc.

src/main/java/org/mockito/internal/creation/bytebuddy/SubclassLoader.java
```diff
@@ -5,9 +5,12 @@
 package org.mockito.internal.creation.bytebuddy;
 
 import net.bytebuddy.dynamic.loading.ClassLoadingStrategy;
+import org.mockito.mock.SerializableMode;
 
 public interface SubclassLoader {
 
-    ClassLoadingStrategy<ClassLoader> getStrategy(Class<?> mockedType);
-
+    ClassLoadingStrategy<ClassLoader> resolveStrategy(Class<?> mockedType,
```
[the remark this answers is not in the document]
raphw: Added doc and removed the serialization strategy as a parameter which is no longer required.

src/main/java/org/mockito/internal/creation/bytebuddy/SubclassInjectionLoader.java
raphw: Good idea, added the cause and resolve the platform to aid debugging.

PR #1356, opened 2018-04-01
Add JDK 10 and 11 to the build matrix

TimvdLippe:
Work in progress PR to try to make Mockito build on Java 10 and 11. Let's see how this goes...


TimvdLippe:
JDK 11 support is blocked on a new Gradle release, as they fixed JDK 11 support in https://github.com/gradle/gradle/pull/4759

TimvdLippe:
Oh, Gradle appears to have nightly builds (https://gradle.org/nightly/) so let's try that instead.

TimvdLippe:
We are now hitting https://github.com/gradle/gradle/issues/4860 so have to revert to Gradle 4.6 and not build on JDK 11 for now.

.travis.yml
```diff
@@ -22,12 +22,26 @@ matrix:
...
+    env: SKIP_RELEASE=true JDK='Oracle JDK 10'
+  - install: . ./install-jdk.sh -F 10 -L GPL
+    env: SKIP_RELEASE=true MOCK_MAKER=mock-maker-inline JDK='Oracle JDK 10'
+  - install: . ./install-jdk.sh -F 11 -L BCL
+    env: SKIP_RELEASE=true JDK='Oracle JDK 11'
+  - install: . ./install-jdk.sh -F 11 -L BCL
+    env: SKIP_RELEASE=true MOCK_MAKER=mock-maker-inline JDK='Oracle JDK 11'
+  allow_failures:
+    - env: SKIP_RELEASE=true JDK='Oracle JDK 11'
+    - env: SKIP_RELEASE=true MOCK_MAKER=mock-maker-inline JDK='Oracle JDK 11'
```
szpak: +1 for preferring `allow_failures` over disabling the Java 11 build completely

.travis.yml
```diff
@@ -22,12 +22,26 @@ matrix:
...
 
 branches:
   #Don't build tags
   except:
   - /^v\d/
 
+before_install:
+  - wget https://raw.githubusercontent.com/sormuras/bach/master/install-jdk.sh
```
szpak: I personally don't like to just `wget` files from GitHub in my (release) builds. The file could be downloaded (we lose automatic updates, though). However, I might be exaggerating (and maybe we already "wget" something in the build ;) ).

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.6-bin.zip
```
szpak: When have we lost the distribution checksum verification? ;)

szpak:
I hit https://github.com/sormuras/bach/pull/12/ in my project, but Mockito uses other infrastructure, so it may be not needed.

szpak:
And @TimvdLippe, please squash your commits :)

.travis.yml
sormuras: I don't like it either.

I managed a [list of copies](https://sormuras.github.io/blog/2017-12-08-install-jdk-on-travis.html#used-by) this file in various OS projects, but whenever a new feature version of the JDK drops or Oracle changes the path to the binaries, I had to spam pull requests. Some of them are not merged, yet and the projects will be hit by an outdated version of `install-jdk.sh` soon: 
- https://github.com/google/google-java-format/pull/261 
- https://github.com/square/javapoet/pull/629

sormuras: I think, the preferred way to refer to raw files on GitHub is using following URI:

`https://github.com/sormuras/bach/raw/master/install-jdk.sh`

TimvdLippe: @sormuras I took the url stated in https://github.com/junit-team/junit5/blob/13de8c51701b69769328b55b1857631e75bc5cdc/.travis.yml#L13 which is the same as https://github.com/raphw/byte-buddy/blob/a576a7af2e4c9d3621526f40f7faa15fda16eafb/.travis.yml#L8 so I think this is okay?

gradle/wrapper/gradle-wrapper.properties
TimvdLippe: That seems to be a regression in https://github.com/mockito/mockito/commit/70a7cf8c17bcbf0ffb5316d6dd2457d9708bbe5f#diff-fd9d61e37d939c941e7a62077f38d62d Fixing :+1:

TimvdLippe:
@szpak I will squash at the moment of merging :smile:

.travis.yml
sormuras: That's the current target URI where you are redirected to when browsing `https://github.com/sormuras/bach/raw/master/install-jdk.sh` -- so, if GitHub changes this to some other address someday, it might stop to work. Both should be okay.

But perhaps Travis CI will solve this by calling `install-jdk.sh` for us: https://github.com/travis-ci/travis-cookbooks/issues/976 ;)

PR #1358, opened 2018-04-02
Apply errorprone to all java projects

epeee:
I thought it might be a good idea to apply the errorprone plugin to all projects using the java plugin.
Furthermore, I think it is a good idea to specify the version of errorprone in the gradle file instead of always using the latest one (since this might break the build e.g. due to new checks).

@TimvdLippe, @mockitoguy what do you think?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ buildscript {
     dependencies {
         classpath 'gradle.plugin.nl.javadude.gradle.plugins:license-gradle-plugin:0.14.0'
         classpath 'ru.vyarus:gradle-animalsniffer-plugin:1.4.3' //for 'java-compatibility-check.gradle'
-        classpath "net.ltgt.gradle:gradle-errorprone-plugin:0.0.13"
+        classpath 'net.ltgt.gradle:gradle-errorprone-plugin:0.0.13'
 
         //Using buildscript.classpath so that we can resolve shipkit from maven local, during local testing
         classpath 'org.shipkit:shipkit:2.0.13'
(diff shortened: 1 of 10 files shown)


build.gradle
```diff
@@ -16,7 +16,7 @@ buildscript {
 
 plugins {
     id 'com.gradle.build-scan' version '1.12.1'
-    id 'net.ltgt.errorprone' version '0.0.13'
+    id 'net.ltgt.errorprone' version '0.0.13' apply false
```
TimvdLippe: Why not remove this line and put it on line 53 instead?

build.gradle
```diff
@@ -45,10 +45,16 @@ apply from: 'gradle/dependencies.gradle'
...
 
-allprojects {
+allprojects { proj ->
     repositories {
         jcenter()
     }
+    plugins.withId('java') {
+        proj.apply plugin: 'net.ltgt.errorprone'
+        proj.dependencies {
+            errorprone 'com.google.errorprone:error_prone_core:2.2.0'
```
TimvdLippe: Let's extract the version number into the `dependencies.gradle` file

szpak:
There seems to be a problem with errorprone and Java 10 - Travis in master [fails](https://travis-ci.org/mockito/mockito/jobs/361255753#L853-L865).

TimvdLippe:
Errorprone on JDK 10 works on `release/2.x` here though 🤔 https://travis-ci.org/mockito/mockito/builds/361158556

epeee:
thx, I will have a look at this one.

Issue #1333, opened 2018-03-12
Add mockito-scala


bbonanno:
It would be nice to have a separate module (?) with Scala specific features and syntax sugars
I've been working on this for a while now and I think I could improve the experience of using Mockito with Scala

I'm not sure what the community thinks about a separate module for it, I think it's right to keep it clean and separated as they will be some .scala files

Things I'd initially put there would be:

- Wrapping Mockito API on traits: This is so it can either be used as now (calling the static methods on Mockito or ArgumentMatchers) or just mixing in those traits to avoid having to import each thing we want to use.
- Bridge between Java and Scala: Said traits will also fix annoying issues like Scala not playing well with Java overloaded methods that use varargs (like doReturn) or renaming the eq ArgumentMatcher given that eq in Scala is a special operator to do identity equality.
- Adding support for default arguments: Currently when you stub or verify a method with default arguments the mock gets null as the value for that param instead of the default value, this is easy to fix with some default stubbings.

And more to come

Thoughts, ideas?

Thanks

Bruno

TimvdLippe:
I do not have a lot of Scala experience nor am actively developing Scala. Could you publish a very small PR with 1 example API to show: 1. what the original problem/annoyance was and 2. how this prototype would fix it? That way we can make a judgement call whether we want a full-blown module :smile:

bbonanno:
@TimvdLippe sure

TimvdLippe:
@szpak Oh yes I forgot to point that out. @bbonanno for the first PR, you can create a new subproject just like https://github.com/mockito/mockito/tree/release/2.x/subprojects/kotlinTest to have this as a separate module.

bbonanno:
@szpak @TimvdLippe Yeah, I was thinking in doing that as adding the Scala dependencies to mockito-core looks bad from the beginning.
Keep tuned, as soon as I have something decent to show the problem and solution I'll raise a PR so you guys can check it out

mockitoguy:
Hey @bbonanno! Thank you for suggestion! Given that the core team has no expertise in Scala, most likely we won't be publishing mockito-scala from our project. Please get the prototype working in a separate project. Make it great, then for sure you will attract community!

Thanks for suggestion! I'll close this ticket to keep our issue tracker clean.

bbonanno:
That's a problem then, I only have permission to contribute to an existing project, not to start a new one...

TimvdLippe:
[...]

PR #1362, opened 2018-04-06
Draft of the proposed solution for #1333

bbonanno:
[Please do not merge]

This PR is just to show the proposed solution to #1333, it needs more work before it's in a state that can be published


bbonanno:
@TimvdLippe Thanks for taking the time to review it.
Your summary is correct, do you want me to add more comments in the code or just here in the PR?
I'll also take a look on how to simplify the `ArgumentMatchers` calls

TimvdLippe:
Would be best to add comments in the code, for us to understand what it is doing. Mostly because we have little experience with Scala and do not know what were the problems. The code submitted in this PR seems rather straightforward, but the reason for existence is a bit unclear for me. If we can clear that up, future Scala users will understand why they need to use these methods as well: 2 in 1! :smile:

bbonanno:
@TimvdLippe I've split some files and add a ton of documentation and tests, I hope it's easier to get the reason behind the changes now.
Please let me know your thoughts.

TimvdLippe:
@bbonanno Thank you very much for the update. The intent of most of the methods is clear to me now. For me to summarize: we have 2 categories of methods. The first method is methods that are required to play nicely with either the Scala compiler or with Scala SDK implementation. The other methods are simply to complement the problematic methods to cover our full API.

To me, this change provides real value, as we are now more nicely integrating with Scala users. However, the one concern we still have is the maintenance of the Scala code, as we are no Scala developers. I think the level of complexity is still limited in these cases, so I am positive in that regard, but I would like to know what the others are thinking.

@mockitoguy I would like to hear your opinion (per our internal discussion) about these changes are.

bbonanno:
@TimvdLippe For what is worth, my plan is to keep a close eye on it and keep adding features, so if there is an issue with it in the future I'll be able to help

szpak:
I'm not a Scala developer and I will not comment the implementation. However, one thing that I spotted is a lack of methods from [`BDDMockito`](https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/BDDMockito.java), including stubbing with `given..will` and a verification method `then()`. Probably it would be good to add it them in some future iteration to catch up the Java API (possibly once this PR is accepted and merged).

bbonanno:
@szpak You're right, I barely use that one so I totally forgot about it, I added it to my TODO list for when this piece gets merged.
Thanks for the feedback

TimvdLippe:
Hm, I am not sure if that is a good idea. After thinking about this more I would actually say that shipping the "filler"-methods is not a good idea. Instead, we should ship only the methods that are actually problematic in Scala. We could always add our other API surface later, but I would like to use a conservative approach here.

bbonanno:
@TimvdLippe The filler methods are there because it would be really weird to have to remember which ones you have to call through MockitoSugar ans which ones directly through the Java Mockito class.
It also eases the pain of imports and auto-complete as importing the all object members or mixin in the trait would make the API available effortlessly.
Maybe we don't need that for BDDMockito, but if at least one of the methods there need to be modified to play nicer with Scala, then the whole API should be exposed through the same place.

subprojects/mockito-scala/src/main/scala/org/mockito/MockitoSugar.scala
```diff
@@ -0,0 +1,278 @@
...
+      * but fixes the following compiler issue that happens because the overloaded vararg on the Java side
+      *
+      * Error:(33, 25) ambiguous reference to overloaded definition,
+      * both method doReturn in class Mockito of type (x$1: Any, x$2: Object*)org.mockito.stubbing.Stubber
+      * and  method doReturn in class Mockito of type (x$1: Any)org.mockito.stubbing.Stubber
+      * match argument types (`Type`)
+      *
+      */
+    def doReturn[T](toBeReturned: T, toBeReturnedNext: T*): Stubber =
+        Mockito.doReturn(toBeReturned, toBeReturnedNext.map(_.asInstanceOf[Object]): _*)
```
tpalanga: Check if it works by default without mapping to Object

subprojects/mockito-scala/src/main/scala/org/mockito/MockitoSugar.scala
```diff
@@ -0,0 +1,278 @@
...
+
+    /**
+      * Delegates to <code>Mockito.doThrow</code>, it's only here so we expose all the `ArgumentMatchers`
```
tpalanga: Replace "ArgumentMatchers" with Mockito API

subprojects/mockito-scala/src/test/scala/org/mockito/MockitoSugarTest.scala
```diff
@@ -0,0 +1,66 @@
...
+
+        aMock.iHaveSomeDefaultArguments("I'm not gonna pass the second argument")
+
+        verify(aMock).iHaveSomeDefaultArguments("I'm not gonna pass the second argument", "default value")
+    }
+
+    "verifyNoMoreInteractions" should "ignore the calls to the methods that provide default arguments" in {
+        val aMock = mock[Foo]
+
+        reset(aMock)
```
tpalanga: No need to reset here

subprojects/mockito-scala/src/main/scala/org/mockito/MockitoSugar.scala
bbonanno: Sadly, it doesn't

bbonanno:
@TimvdLippe do you think anything else needs to be changed to get this code merged?

TimvdLippe:
@bbonanno This PR is pending other core members to chime in and give their opinion, most prominently @mockitoguy per our internal discussion earlier.

bbonanno:
@TimvdLippe fair enough, thanks!

bbonanno:
@TimvdLippe @mockitoguy Hi Guys, have you reached a decision regarding this PR?

Cheers

Bruno

mockitoguy:
Hey guys!

I'm not in favor of adding Scala code to the Mockito project. Core developers don't work with Scala. How can we review PRs / evolve this feature in a meaningful way?

@bbonanno, how can we help you? I can create a subproject under "Mockito" organization on GitHub. This way you get Mockito official brand and the software ownership is cleaner: you work with Scala, you are the maintainer, you decide on evolution/vision for the library. This means that the versioning is separate, too.

@bbonanno, we really appreciate the contribution. I hope we can arrive at a solution that works for everybody!

bbonanno:
@mockitoguy I think that would be a good solution, please let me know when it's ready so I can set-up the project and move the code over.
I guess I'd also need your help to get that published into the maven repositories once everything is done

Many thanks!

mockitoguy:
@mockito/developers, any feedback before I create "mockito-scala" project on GitHub and give @bbonanno commit rights?

bbonanno:
@mockitoguy Sorry to insist, but I think this week I'll have time to tackle this, do you think the repo could be ready any time soon?
Thanks!

mockitoguy:
@bbonanno, thank you for patience!

I've created the repo and set you as a committer/maintainer:

https://github.com/mockito/mockito-scala

Please sent an RB to that repo. We'll help you set up automatic publications once you're ready with the code.

bbonanno:
@mockitoguy Thanks! I just got access to it, I'll set up the project and I'll let you know when I have something that can be released so we do the publications

mockitoguy:
@bbonanno, sounds good. I'll keep an eye on this PR but if you're not getting support from me hit me directly via linkedin/twitter/email

bbonanno:
@mockitoguy I've committed the code for a first release to a branch called "initial-code" as I don't wanna have it in master until we get the artefact published in the maven repositories, how can we do that?
BTW, this project uses SBT instead of Gradle as it's the standard Scala build tool, not sure if your CI server has it installed

Anyway, please let me know what are the next steps

mockitoguy:
@bbonanno, thank you for neat progress!!!

Please submit a PR to mockito-scala project and let's merge your work to the master. Set the version to 0.0.1 to indicate that the project is in progress and there is no guarantee of compatibility (yet :).

Can you take a look at "play-parseq" project, and how the SBT build is integrated with Gradle? https://github.com/linkedin/play-parseq/blob/master/build.gradle

In a nutshell we need:
 - the sbt build needs to be self-contained, e.g. I run "./gradlew build" and the right version of sbt needs to be provisioned and used to run the build.
 - sbt build needs to build the version as indicated in "version.properties" file.
 - in general, we want to replicate what "play-parseq" project does. This will give us continuous delivery and publications to repos. When we are ready, we change the version from 0.0.x -> 1.0.0.

Let me know how I can help further. Awaiting PRs to mockito-scala :)

mockitoguy:
We are working on this in a separate project: https://github.com/mockito/mockito-scala. I'm closing this PR.

PR #1402, opened 2018-06-06
Fixes #1401 : Correctly handle VarargsMatcher check for HamcrestMatchers

### LINKED ISSUE #1401: VarargsMatcher doesn't work correctly when using HamcrestMatchers in Mockito 2
```
class A {
    private B b;
    public A(B b) {
        this.b = b;
    }
    public Integer perform() {
        return b.run("abcd", 1L, 2L);
    }
}

class B {
    public Integer run(String str, Long... price) {
        return 10;
    }
}

class HamcrestVarArgsMatcher extends TypeSafeMatcher<Long> implements VarargMatcher {
    protected boolean matchesSafely(Long item) {
        return item < 3;
    }
    public void describeTo(Description description) {
    }
}

class Test {
   @Test
   public void varArgsMatcherUsingHamcrestMatcherWorks() {
        B b = Mockito.mock(B.class);
        A a = new A(b);

        when(b.run(eq("abcd"), MockitoHamcrest.argThat(new HamcrestVarArgsMatcher()))).thenReturn(15);

        assertThat(a.perform(), is(15)); // Actually returns 0
    }
}
```

aj-jaswanth:


aj-jaswanth:
Hi @TimvdLippe, this change is meant to handle the situation in which we use a custom hamcrest vararg matcher. As, https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/hamcrest/MockitoHamcrest.java#L178 wraps the actual matcher it is not possible to check if it implements varargmatcher interface.

Please see https://github.com/mockito/mockito/issues/1401 for an example.

PR #1409, opened 2018-06-15
Fixes #1408: Remove redundant thenReturn

### LINKED ISSUE #1408: remove redundant thenReturn from OngoingStubbing
There are two methods in  `OngoingStubbing`.
```
OngoingStubbing<T> thenReturn(T value);
OngoingStubbing<T> thenReturn(T value, T... values);

```
The one, with a single argument is redundant, because based on both documentation and implementation - it is covered by the one with varargs.

On top of that, such overloading might cause interoperability issues (at least in Scala it causes compilation error due ambiguous reference), where one needs to pass empty List as the second argument explicitly. (you may look in one of the answers here https://stackoverflow.com/questions/13358705/force-single-argument-in-scala-varargs )

GodPlaysChess:
Motivation is here: https://github.com/mockito/mockito/issues/1408

### DIFF
--- a/src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
+++ b/src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
@@ -15,8 +15,7 @@ import org.mockito.stubbing.OngoingStubbing;
 
 public abstract class BaseStubbing<T> implements OngoingStubbing<T> {
 
-    @Override
-    public OngoingStubbing<T> thenReturn(T value) {
+    private OngoingStubbing<T> thenReturn(T value) {
         return thenAnswer(new Returns(value));
     }
 
--- a/src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
+++ b/src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
@@ -15,13 +15,9 @@ import org.mockito.stubbing.OngoingStubbing;
 
 public abstract class BaseStubbing<T> implements OngoingStubbing<T> {
 
-    private OngoingStubbing<T> thenReturn(T value) {
-        return thenAnswer(new Returns(value));
-    }
-
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
```diff
@@ -15,8 +15,7 @@
 
 public abstract class BaseStubbing<T> implements OngoingStubbing<T> {
 
-    @Override
-    public OngoingStubbing<T> thenReturn(T value) {
+    private OngoingStubbing<T> thenReturn(T value) {
```
TimvdLippe: We are only calling this on line 24, so let's just inline this and remove the method as a whole.

GodPlaysChess: Yes, will do that.

bric3:
I don't want these change merged.

My comment in #1408 

> I'm opposed to this change in 2.x, because it may break exernal libraries that require binary compatibility.
> 
> I'm OK for 3.x though. Note however that requires to allocate de facto an array for each thenReturn / willReturn.
>
> Regarding scala stuff, what about mockito sugar ?

Also this diff is not mirroring the variations of this method, in BDD aliases, in the doReturn style methods.

TimvdLippe:
We should probably add a way to test for binary compatibility if we want to enforce that. I wonder if we can generate an artifact of our test folder and use that binary to then test the current source code with. But that's besides this PR.

Given that we then don't want to do this in 2.x, this PR has to target 3.x and also fix the other `thenReturn` versions in BDD. However, let's settle the work for now if there is no sign of merging in the near future.

bric3:
I'm closing the PR. I'd like to thank you for contributing to this project @GodPlaysChess. It's relly nice to take some time and contribute. Thank you !

Issue #1417, opened 2018-06-29
Allow @InjectMock into static/final fields


HonoluluHenk:
For some of my projects it would be nice if @InjectMock would work with spies on final/static fields.
Unfortunately I cannot refactor the code to use constructor/setter injection since this would alter the API (package-private is not sufficient).

I can see that injecting into static final is problematic (unexpected test behavior on subsequent tests)...
but injecting into just static *or* final fields should pose no problems - at least none I can see (care to enlighten if your opinion differs?).

TimvdLippe:
I am inclined to not support this feature, as it encourages bad mocking behavior. Moreover, I think `PowerMock` already handles this case. In general, relying on static state during tests is bad, especially if mocks are mixed in. Writing adapters in that case improves readability.

HonoluluHenk:
Actual use case:
my instance does have an internal cache of some sort:
`private final Cache cache = new Cache();`

I do not ever want to expose this cache in the API of the containing class (not event package-private) since this is a very private implementation detail no one else should know/care of.

> In general, relying on static state during tests is bad, especially if mocks are mixed in.
Fully agree, but not being able to easily test the classes containing static state is (imho) event worse.
Workaround is to set these class-fields via reflection - which is bad, too.

> PowerMock already handles this case
This requires yet another framework everyone on the team has to learn.

>  as it encourages bad mocking behavior
Not being able to mock easily encourages skipping testing altogether :(

PR #1418, opened 2018-06-29
Allow injection of static/final fields

HonoluluHenk:
Possible implementation for #1417


src/main/java/org/mockito/InjectUnsafe.java
```diff
@@ -0,0 +1,71 @@
...
+ * And then there's always legacy/library code that cannot be changed easily :(</p>
+ *
+ * <p>InjectUnsafe to the rescue:<br/>
+ * Modifies the behavior of InjectMocks in a way that allows injection into static and final fields. </p>
+ *
+ * <p> Please note: not all mock builders are supported equally:
+ * bytebuddy (the default) will throw {@link IllegalAccessError}
+ * if you try to use <code>staticFields = {@link OverrideStaticFields#STATIC_FINAL}</code>. </p>
+ */
+@Documented
```
bric3: Could you add example code blocks like those in `@InjectMocks`

src/main/java/org/mockito/InjectUnsafe.java
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * <p>However, sometimes you don't want to/cannot expose fields via constructor as this might change your api
+ * (even if the constructor is package-private).<br/>
+ * Even worse with static fields: you'd have to make setters available via method call.<br/>
+ * And then there's always legacy/library code that cannot be changed easily :(</p>
+ *
+ * <p>InjectUnsafe to the rescue:<br/>
+ * Modifies the behavior of InjectMocks in a way that allows injection into static and final fields. </p>
+ *
+ * <p> Please note: not all mock builders are supported equally:
```
bric3: mock ~builder~ maker ?

src/main/java/org/mockito/InjectUnsafe.java
```diff
@@ -0,0 +1,71 @@
...
+ * Change the behavior of {@link InjectMocks}: allow injection of static/final fields that are normally skipped.
+ *
+ * <p>Typically, injection into instances is done via the constructor.</p>
+ *
+ * <p>In most cases you should refactor your code to support these best practices!</p>
+ *
+ * <p>However, sometimes you don't want to/cannot expose fields via constructor as this might change your api
+ * (even if the constructor is package-private).<br/>
+ * Even worse with static fields: you'd have to make setters available via method call.<br/>
+ * And then there's always legacy/library code that cannot be changed easily :(</p>
```
bric3: Please add another warning about static, e.g. changing them might have repercussion on other part of the code because static fields can be used anywhere.

src/main/java/org/mockito/InjectUnsafe.java
```diff
@@ -0,0 +1,71 @@
...
+ * Modifies the behavior of InjectMocks in a way that allows injection into static and final fields. </p>
+ *
+ * <p> Please note: not all mock builders are supported equally:
+ * bytebuddy (the default) will throw {@link IllegalAccessError}
+ * if you try to use <code>staticFields = {@link OverrideStaticFields#STATIC_FINAL}</code>. </p>
+ */
+@Documented
+@Target({ElementType.FIELD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface InjectUnsafe {
```
bric3: Annotate with `@Incubating`.

src/main/java/org/mockito/InjectUnsafe.java
```diff
@@ -0,0 +1,71 @@
...
+        /**
+         * Allow mock injection into final instance fields.
+         */
+        FINAL
+    }
+
+    /**
+     * Allow mock-injection into instance fields that get skipped during the normal injection cycle.
+     */
+    OverrideInstanceFields instanceFields() default OverrideInstanceFields.NONE;
```
bric3: Instead of _override something_ I propose this naming :

* `NONE` to `NON_FINAL`
* `FINAL` to  `ALLOW_FINAL` 
* `InstanceFields instanceFields() default InstanceFields.NON_FINAL`

Thoughts ?

src/main/java/org/mockito/InjectUnsafe.java
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * Only use this if you <strong>absolute know what you're doing!</strong>.
+     *
+     * This will change the value of static fields value for <strong>all subsequent usages of the static field</strong>
+     * as long as the current class loader lives.
+     * This might lead to very unexpected and hard-to-debug behavior of other tests!
+     */
+    OverrideStaticFields staticFields() default OverrideStaticFields.NONE;
```
bric3: Same here 

* `NONE` to `NON_STATIC`
* `STATIC` to  `ALLOW_STATIC` 
* `InstanceFields staticFields() default InstanceFields.NON_STATIC`

bric3: Thinking about it more, why not simplify the annotation further ? In pseudo code :

```java
enum UnsafeFieldModifier { NONE, FINAL, STATIC, STATIC_FINAL }
UnsafeModifier[] allow() default NONE;
```

src/main/java/org/mockito/internal/util/collections/ListUtil.java
```diff
@@ -34,6 +34,20 @@
         return converted;
     }
 
+    public static <T> Filter<T> combineOr(final Filter<T> a, final Filter<T> b) {
```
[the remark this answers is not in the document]
HonoluluHenk: Inside, the anonymous class inside accesses the parameters directly and thus needs final.
I'll assign the params to local final variables and let the anon class access these locals.

src/main/java/org/mockito/InjectUnsafe.java
HonoluluHenk: I like this!

With this modification, the "NONE" modifier is kinda redundant since the user has to explicitly add the annotation and (imho) could already expect some changed behavior.

I'd risk `default FINAL;` so the 99% use-case of this feature is as simple to use as possible.

Resulting in:
```java
enum UnsafeFieldModifier { FINAL, STATIC, STATIC_FINAL }
UnsafeModifier[] allow() default FINAL;
```

HonoluluHenk:
I noticed that PropertyAndSetterInjection::orderedInstanceFieldsFrom only handles fields in the "awaitingInjectionClazz" but not its superclasses.
Is this expected behavior?

HonoluluHenk:
@bric3 I think I implemented all your change requests. Care to review again?

src/main/java/org/mockito/internal/util/collections/ListUtil.java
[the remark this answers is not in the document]
bric3: Ah yes I forgot that part now that working with effectively final variable is possible with Java 8.

👍

src/main/java/org/mockito/internal/util/reflection/StaticFinalOverrider.java
```diff
@@ -0,0 +1,47 @@
...
+    public void enableWrite(Field field) {
+        if ((field.getModifiers() & (Modifier.FINAL + Modifier.STATIC)) == 0) {
+            return;
+        }
+
+        try {
+            modifiersField = Field.class.getDeclaredField("modifiers");
+            modifiersField.setAccessible(true);
+            modifiersField.setInt(field, field.getModifiers() & ~Modifier.FINAL);
+        } catch (NoSuchFieldException e) {
```
bric3: Thinking about this, I'm not sure this always work. I know this trick, but it may not work under all circumstaces e.g. when there's a security manager.

Which means doing it in `AccessController.doPrivileged(...)`. However I suspects that a lot of mockito code base isn't complient with secuirty manager anyway.

src/main/java/org/mockito/internal/util/reflection/FieldSetter.java
```diff
@@ -6,21 +6,56 @@
...
+public final class FieldSetter {
+
+    private FieldSetter() {}
+
+    /**
+     * Set value of fields, except static final (use {@link #setAnyField(Object, Field, Object)} for that).
+     */
+    public static void setField(Object target, Field field, Object value) {
+        // keep synchronization in sync with setAnyField method!
+        synchronized (field) {
```
bric3: I'm not sure this synchronization needed since this call is wrapped by `setAnyField` which is supposed to handle static fields as well.

src/main/java/org/mockito/internal/util/reflection/StaticFinalOverrider.java
```diff
@@ -0,0 +1,47 @@
...
+public class StaticFinalOverrider {
+    private Field modifiersField = null;
+
+    public void enableWrite(Field field) {
+        if ((field.getModifiers() & (Modifier.FINAL + Modifier.STATIC)) == 0) {
+            return;
+        }
+
+        try {
+            modifiersField = Field.class.getDeclaredField("modifiers");
```
bric3: I think this one should be initialized statically.

src/main/java/org/mockito/internal/util/reflection/FieldSetter.java
HonoluluHenk: TIL: Foo.class.getDeclaredField("bar") returns a new Field instance for every call
=> synchronizing is probably not worth it.
I'll remove this

src/main/java/org/mockito/internal/util/reflection/StaticFinalOverrider.java
HonoluluHenk: TIL (same as above ;)): Foo.class.getDeclaredField("bar") returns a new Field instance for every call.
So I'm not sure if a static variable is The Right Thing (tm)

src/main/java/org/mockito/internal/util/reflection/StaticFinalOverrider.java
HonoluluHenk: Hmm... setAccessible itself is guarded by the SecurityManager as per JavaDoc.
Calling setInt only throws IllegalAccessException (from Field javadoc) "if this Field object is enforcing Java language access control and the underlying field is either inaccessible or final."
Since we're setting the Field accessible, the IllegalAccessException *should* not occur.

Unfortunately I don't have any experience with SecurityManagers but do share your assumption: 
"However I suspects that a lot of mockito code base isn't complient with secuirty manager anyway." :)

Risk it to leave it as-is?

src/main/java/org/mockito/internal/util/reflection/StaticFinalOverrider.java
bric3: So let's try with a local variable then.

HonoluluHenk:
> If you want I can take over the little tweaks and then merge ?
Awesome, thx!

src/main/java/org/mockito/internal/util/reflection/FieldSetter.java
```diff
@@ -6,21 +6,51 @@
...
+    }
+
+    /**
+     * Compared to {@link #setField(Object, Field, Object)}, this method also allows setting static final fields.
+     *
+     * Setting static final fields needs some mangling on the field modifiers of the {@link Field} class and thus should
+     * only be used when absolutely necessary. See {@link StaticFinalOverrider} for implementation details.
+     */
+    public static void setAnyField(Object target, Field field, Object value) {
+        synchronized (field) {
```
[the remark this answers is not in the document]
HonoluluHenk: prolly not needed, see my TIL  comment about Field instances above

KoichiSenada:
> This branch has conflicts that must be resolved
Any progress on this?

HonoluluHenk:
> Any progress on this?
Done

HonoluluHenk:
Deprecated in favor of #2700

PR #1429, opened 2018-07-13
Updated test coverage for async verification

mockitoguy:
Updated tests with after() verification mode
- Some existing tests for the feature did not work due to flaws in ReplyRule. Those tests (now ignored) consistently failed but succeeded only when used with reply rule.
- Some existing tests had weak assertions that did not fully cover the feature.
- Suggested simpler and cleaner way of implementing delayed execution via new 'AsyncTesting' class, instead of 'DelayedExecution'. Dropping the reply rule may introduce instabilities to the CI builds, let's observe and measure before making an action.

Although we are considering deprecating 'after()', we need good coverage for an existing feature even if it will be removed in the long run.


mockitoguy:
@ChristianSchwarz, any feedback about this change?

ChristianSchwarz:
I will take a look.

mockitoguy:
I'll merge this if that's ok. It mostly test code change with a compatible change to exceptions. Happy to take review feedback after it's merged.

PR #1430, opened 2018-07-15
Allow mockito-logger to be configurable as Mockito extension

bedla:
Hi,
we have many of legacy tests that do over-mocking code base - lots of unused mocking there (we are cleaning it slowly). We have user `@MockitoSettings(strictness = Strictness.WARN)` switch to log and not fail in this situactions. By default all logs are printed out to console using `System.out.println` and this make us trouble because we have "silent" profile at CI server which is configured to silent Loggers / as you expect sith `.println` we are not trully silent in this case :]
Thats why I have created this PR proposal and want to know what do you think.
Thx,
Ivos


mgajdos:
+1

bedla:
rebased with current upstream

src/main/java/org/mockito/plugins/MockitoLogger.java
```diff
@@ -2,7 +2,7 @@
  * Copyright (c) 2007 Mockito contributors
  * This program is made available under the terms of the MIT License.
  */
-package org.mockito.internal.util;
+package org.mockito.plugins;
 
 public interface MockitoLogger {
```
TimvdLippe: We are going to need some additional documentation here. https://static.javadoc.io/org.mockito/mockito-core/2.23.4/org/mockito/plugins/MockMaker.html is probably a good starting point of what we can add as documentation.

PR #1432, opened 2018-07-18
Update gradle plugins

epeee:
- gradle-animalsniffer-plugin (v1.4.4)
 - gradle-errorprone-plugin (v0.0.15)
 - build-scan-plugin (v1.15.1)

I will provide another pr about updating shipkit to the latest version soon (which we'd need in order to be able to use gradle 4.8/4.9).


mockitoguy:
I've restarted the build and if it OK, I'll merge!

PR #1433, opened 2018-07-18
Fixed Travis build

mockitoguy:
We don't need private constructor on an 'internal' class. Internal classes can be changed at any time based on our compatibility policy.


bric3:
I think we should align this on other no-arg matchers like `Null`, and `NotNull`

PR #1434, opened 2018-07-19
Javadoc: Update broken code example

marchpig:
It updates a broken code example in Javadoc.
I just copied the example from [Answer.java](https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/stubbing/Answer.java#L16) to [Mockito.java](https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/Mockito.java#L525).


mockitoguy:
Thank you!

PR #1440, opened 2018-07-22
Javadoc: Correct wrong links and code examples

marchpig:
Some wrong links and code examples are corrected.


mockitoguy:
Thank you @marchpig!

Issue #1398, opened 2018-05-31
Allow partial mixing of concrete values with matchers


eyalroth:
I'm pretty sure we're all familiar with the dreadful message "Invalid use of argument matchers" which is caused (most of the time) due to mixing of concrete values and matchers. After using Mockito for several years, I gotta say that this is one of the most annoying drawbacks of the framework.

Recently I've been getting more and more into the code of the framework, and started wondering why this restriction is there in the first place. I looked it up, and found [this thread](https://groups.google.com/forum/#!topic/mockito/4W-PFmG5jTM) on the Google mailing list, in which a guy named Brice (@bric3 ?) explains the reasoning behind this in a fairly simple manner.

In short, the matchers cannot be passed to the mock invocation, since they have a different type than that of the arguments. The way Mockito is handling this is by "registering" the matchers prior to the invocation on a local-thread stack, which the mock-handler later inspects upon the method invocation. When only matchers are used, it is easy to match each matcher to a parameter of the method--by simply matching the stack to the parameter list--but when concrete values are mixed in, this becomes ambiguous.

I would like to suggest a middle-ground solution that will probably cover (at least) 80% of the cases - How about assuming that each `null` argument a method is invoked with is supposed to correspond to a matcher, while the rest will be treated as concrete values? This way, only cases where `null` is passed "concretely" alongside matchers will be invalid. For example:
```java
when(myMock.foo(any(), null)) // invalid
when(myMock.foo(any(), "1")) // valid
```

There would also be a problem with primitive types, as matchers cannot be passed to the method as `null`s, but in this case they could be passed as default values and then an invocation will be considered invalid if the number of default primitive values passed to the method exceeds the number of matchers of said primitive type on the stack.

Thoughts?

TimvdLippe:
Yes, this is indeed one of the gotchas I consider most negative of Mockito. That said, I think solving this with the current implementation is near impossible. Matchers currentlly return default values, to make sure method invocations compile, while in the background we register the matcher. There is technically no way we can differentiate between a regular value and the default value of a matcher.

In #1264 I lifted this restriction to the type-level instead: https://github.com/mockito/mockito/blob/c7a326c85d7eb446f29b2e772d9ec218ab0d91cb/subprojects/lambda/src/main/java/org/mockito/OngoingStubbingFunction.java#L18-L26

That approach does allow us to differentiate and make this requirement more explicit.

ChristianSchwarz:
[...]

PR #1441, opened 2018-07-23
[Prototype] allow mixing of int matchers and raw int values

ChristianSchwarz:
see #1398

This is a prototype implementation to demonstrate that mixing raw-values and matchers is partially possible. 

how it works: 
 * 0 is treated as special value, it is used to identify a int-matcher 
 * when a non 0 argument value is found an `Equals` matcher will be used at the current argument position
 * when 0 is found the first matcher from the matcher-stack is used to match at the current position

possible improvements:
 * 0 is a suboptimal value to indicate a matcher, cause it is a often used int-value. To reduce the chance for clashes e.g. `Integer.MIN+3` should be used.
  * the applicable type of the matcher can be taken into account (e.g. anyInt() fit's only to an int argument )

how to deal with non primitive types:
 * the same rule as for int-values can be applied
 * for every object type a matcher-indicator must be returned, e.g. `null` or to avoid clashes a mock (requires inline mocks)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -58,10 +58,10 @@ allprojects { proj ->
         options.warnings = false
         options.encoding = 'UTF-8'
     }
-    apply plugin: 'checkstyle'
-    checkstyle {
-       configFile = rootProject.file('config/checkstyle/checkstyle.xml')
-    }
+//    apply plugin: 'checkstyle'
+//    checkstyle {
+//       configFile = rootProject.file('config/checkstyle/checkstyle.xml')
+//    }
 }
 
 configurations {
--- a/src/main/java/org/mockito/internal/invocation/ArgumentsProcessor.java
+++ b/src/main/java/org/mockito/internal/invocation/ArgumentsProcessor.java
@@ -36,7 +36,7 @@ public class ArgumentsProcessor {
         Object[] varArgs;
         if (args[nonVarArgsCount] == null) {
             // in case someone deliberately passed null varArg array
-            varArgs = new Object[] { null };
+            varArgs = new Object[]{null};
         } else {
             varArgs = ArrayEquals.createObjectArray(args[nonVarArgsCount]);
         }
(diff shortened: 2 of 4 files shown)


mockitoguy:
This is nice! Thank you for putting together a prototype. Here're my suggestions:

1. Can you explore the non-happy paths? How can this break / don't work correctly and produce misleading / surprising test result? What about some odd combinations of matchers.
2. Can you write a design spec with details on how would all types of primitives and Objects be disambiguated. Please make it very crisp what is the user experience - when things work and when don't.
3. For Objects, I suggest we start off with 'null' instead of creating a mock. It is simpler and safer.

Let's make it work!

ChristianSchwarz:
* Yes I try to write a little bit about the toppic.
* Using `null` as indicator value has some implications that can't be rolled back without breaking compability.  Also it will not be possible to provide this feature for Kotlin which would throws in the majority of cases when null is passed.  On the other hand an Kotlin-Plugin could do the indicator detection job, this way the plugin could provide its own non-null indicator values for a given type.

ChristianSchwarz:
I found some test of  @bric3 and  @mockitoguy you that block this implementation. These test assume that
anyString(), anyList(), anyCollection() don't return null, which makes currently makes no sense. Also these matchers are used outside the context of when/verify, which seems strange.

 * [StubbingSpiesDoesNotYieldNPETest](https://github.com/mockito/mockito/blob/e9516f42b1182ae675022c101954e721e20b2c72/src/test/java/org/mockitousage/spies/StubbingSpiesDoesNotYieldNPETest.java)
 * [DeepStubbingTest](https://github.com/mockito/mockito/blob/d25b75c124917e57217fa2d21d5dc2724dfa50e2/src/test/java/org/mockitousage/stubbing/DeepStubbingTest.java#L248)

Is there a uses case I am missing or are the test outdated?

mockitoguy:
We have to stay compatible if we want to ship this feature in 2.x. The use case exists but it is not a strong one:

```java
List spy = spy(List.class);
//in below, real method 'add' is called
//if that method does not accept nulls and anyString() returns null, the user gets a confusing exception
when(spy.add(anyString())).thenReturn(true);

//below works even if anyString() returns null
doReturn(true).when(spy).add(anyString());
```

> Using null as indicator value has some implications that can't be rolled back without breaking compatibility.

Can you suggest alternatives and list pros & cons? This will help us make a decision.

ChristianSchwarz:
@mockitoguy 
Seem like your last message slipped throu my fingers, sorry for the delayed answer.

>> Using null as indicator value has some implications that can't be rolled back without breaking compatibility.
>
>Can you suggest alternatives and list pros & cons? This will help us make a decision.

I looked it up again, I can see no implications anymore. I guess I was wrong here.

TimvdLippe:
Closing this PR as stale. I don't think we are likely to accept changes related to mixing matchers and raw values, unless we revamp the overall API.

Issue #793, opened 2016-11-29
Expose StubbingLookupListener publicly and create listener distinction


mockitoguy:
Make StubbingLookupListener public. Originating ticket: #770

As a rule of a thumb we try to expose the internal APIs that the top level features are built upon. Since StubbingLookupListener is useful for us to implement strictness, let's get the former exposed as public API. This way we build stronger framework with the concept of "onion skin API".

bric3:
If exposed publicly the API should use an event object.

PR #1458, opened 2018-07-28
Add StubbingLookupEvent and Notifier (a part of #793)

marchpig:
This PR adds `StubbingLookupEvent` and `StubbingLookupNotifier` as a part of the issue #793.

### DIFF
--- a/src/main/java/org/mockito/internal/listeners/StubbingLookupNotifier.java
+++ b/src/main/java/org/mockito/internal/listeners/StubbingLookupNotifier.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2018 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.internal.listeners;
+
+import org.mockito.internal.creation.settings.CreationSettings;
+import org.mockito.invocation.Invocation;
+import org.mockito.mock.MockCreationSettings;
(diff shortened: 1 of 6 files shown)


src/main/java/org/mockito/internal/listeners/StubbingLookupNotifier.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class StubbingLookupNotifier {
+
+    public static void notifyStubbedAnswerLookup(Invocation invocation, Stubbing stubbingFound,
+                                                 Collection<Stubbing> allStubbings, CreationSettings creationSettings) {
+        List<StubbingLookupListener> listeners = creationSettings.getStubbingLookupListeners();
+        if (listeners.isEmpty()) {
+            return;
+        }
+        // TODO #793 - onStubbingLookup() should not be called if stubbingFound is null.
```
mockitoguy: > onStubbingLookup() should not be called if stubbingFound is null.

Why? We actually have code that needs to act when stubbing is not found (e.g. strictness - detecting argument mismatch).

marchpig: Oh, I misunderstood that this listener should be called only if the stubbed answer is found.
I will remove this comment!

PR #1459, opened 2018-07-29
stubOnly() must prevent verifyNoMoreInvocations()

marchpig:
stubOnly() must prevent verifyNoMoreInvocations() otherwise the tests are not correct:

Prior behavior:
```java
Foo stub = mock(Foo.class, withSettings().stubOnly());
stub.foo();
verifyNoMoreInvocations(stub); //succeeds even though there is an invocation on stub!
```

Desired behavior:
```java
Foo stub = mock(Foo.class, withSettings().stubOnly());
stub.foo();
verifyNoMoreInvocations(stub); //fails explaining that stubs should not be verified
```


mockitoguy:
This looks good! It is nicer to fail as early as possible.

It is an incompatible change for certain edge cases but the new early failure is absolutely desired - users should not be verifying anything on the 'stubOnly' mocks because those mocks don't preserve any invocations. Hence, running verifyNoMoreInvocations() on 'stubOnly' mock will always succeed.

Can you resolve conflicts so I can merge? The code looks great, I have no constructive corrective feedback. Thank you!

marchpig:
@mockitoguy,
I'm sorry for making you confused but this PR is not for #1460, but only for `verify()`.
Anyway, I can create another PR for `verifyZeroInteractions()` after work today!
Thanks!

PR #1461, opened 2018-07-29
Fixes issue #1222

### LINKED ISSUE #1222: Cannot verify varargs parameter as an array

Alias-m:
Simple vararg argument matcher

Fixes #1222

### DIFF
--- a/src/main/java/org/mockito/internal/matchers/VarargMatcherImpl.java
+++ b/src/main/java/org/mockito/internal/matchers/VarargMatcherImpl.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) 2007 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.internal.matchers;
+
+import org.mockito.ArgumentMatcher;
+
+/**
+ * This class is used to verify with a vararg matcher
+ */
+public class VarargMatcherImpl<T> implements VarargMatcher, ArgumentMatcher<T> {
+    private T[] values;
+    private int iterator;
+    public VarargMatcherImpl(T... value) {
+        this.values = value;
+    }
+
+    @Override
+    public boolean matches(T argument){
+        iterator %= values.length;
+        return values[iterator++].equals(argument);
+    }
+
+    public String toString() {
+        String name = values.getClass().getComponentType().getSimpleName();
+        String str = "vararg<"+name+">(";
+        for(int i = 0 ; i < values.length; i++)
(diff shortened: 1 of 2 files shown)


mockitoguy:
Thank you for your contribution. The ticket suggests to add a new matcher method in the public API. In your PR, you added an internal class that is not a public API.

Alias-m:
I could have add this method in the ArgumentMatchers class (which would have been a better solution), but I didn't want to change it because it has not been updated since last year

```
    public static <T> T vararg(T... value) {
        reportMatcher(new VarargMatcherImpl(value));
        if (value == null)
            return null;
        return (T) Primitives.defaultValue(value.getClass());
    }
```

TimvdLippe:

PR #1462, opened 2018-07-30
Added Gradle-based retry logic for tests

mockitoguy:
- removed JUnit Rule that was doing retries because it was too dangerous. It was hiding tests that did not work properly. Due to how Mockito works and maintains state, it is hard to safely build a retry test rule that works at the level of JUnit rules.
- we still need retry logic for some of our concurrent tests. I wrote a Gradle script plugin for it. It is a pragmatic solution to the problem - having convincing unit tests for concurrent API inherently can have timing issues.

This work is preparing us for @ChristianSchwarz's within implementation.


mockitoguy:
Thank you @TimvdLippe for taking a look!

PR #1464, opened 2018-08-01
stripAnnotations fix

victormakarov:
It was unable to use MockSettings.withoutAnnotations().
stripAnnotation setting had no effect in call to Mockito.mock(class, MockSettings)
because of line CreationSettings<T> settings = new CreationSettings<T>(source);


marchpig:
LGTM!

But it would be better to add a test case to the `MockSettingsTest.java`.
```java
@Test
public void test_without_annotations() {
    MockCreationSettings<List> settings = Mockito.withSettings()
        .withoutAnnotations()
        .build(List.class);

    assertEquals(List.class, settings.getTypeToMock());
    assertTrue(settings.isStripAnnotations());
}
```

victormakarov:
I've just report a bug... And its seems to me that your @Test doesn't cover problem I'm addressed.
I migrating our project's tests from Mockito 1.19 to 2.21 and can't mock any interface annotated with @MessagingGateway (from Spring Integration). Method GatewayHeader[] defaultHeaders() default {}; causes Mockito.mock() to throw IllegalArgumentException: Not a constant annotation type.
So, I've try to use Mockito.withSettings().withoutAnnotations() and found this annoying bug...

mockitoguy:
Thank you for submitting! We definitely want to get this fixed. We will need a test before the code is merged.

@victormakarov, can you contribute a test, too?

victormakarov:
Test added

mockitoguy:
I've pulled those changes (the history was messy, I had to squash), added top-level user test and I'm about to push.

mockitoguy:
Merged. Given that we had this PR, I did not bother creating a new PR with squashed changes. Thank you for contribution!

victormakarov:
Sorry for messy history. It's my first steps on github...
Thank you for very usefull product!

Issue #962, opened 2017-02-23
MockitoTestNGListener not released yet


cedrik23:
Hello,

This is more of an inquiry rather than an issue. I apologize up front if this is the wrong place to post this. 

In our company we use TestNG together with Mockito for unit testing. As such we experience the known issues which are resolved by using MockitoTestNGListener found at: https://github.com/mockito/mockito/tree/release/2.x/subprojects/testng/src/main/java/org/mockito/testng

Perhaps i am mistaken and it is released, yet i am unable to use it through adding dependency of mockito-all either version 1.+ or 2.+

Is there an expected date for releasing this as part of Mockito? I know i can just copy the code to our code base but i'm sure that's not the best way to do this.

Thanks in advance for the assistance.

mockitoguy:
Update: we will work on this soon.

mockitoguy:
For Mockito core team it will be hard to find time to maintain TestNG integration. We will publish the TestNG support via a separate repository similar to how we have done it for "mockito-scala". If you want to use TestNG, check out the repo at https://github.com/mockito/mockito-testng

We're looking for owners/maintainers for this library. The version of TestNG we built the integration with is somewhat old. I don't know if the integration works - I literally copied the code from main Mockito repo. The tests still pass :)

Pending:
 - maven central releases
 - Travis CI and automated releases
 - delete TestNG from main Mockito repo
 - announce

Let us know if you want to help out!!!

skabashnyuk:
@mockitoguy I've checked mockito/mockito-testng 0.0.2 against Eclipse Che codebase and it works just fine. Could you please release mockito-testng to Maven central.

mockitoguy:
I submitted a request to get mockito-testng include in JCenter. This is a first step of getting this in Maven Central. Once the request is approved, I'll configure the project to automatically publish to Maven Central.

We are getting close. Thank you everybody for patience with this ticket!!!

skabashnyuk:
https://bintray.com/mockito/maven/mockito-testng looks like approved

TimvdLippe:
Closing this per the above comment.

PR #1465, opened 2018-08-02
Moved TestNG from the main repo into separate mockito/mockito-testng repo

mockitoguy:
New home for TestNG integration is: https://github.com/mockito/mockito-testng

For details and discussion, see: https://github.com/mockito/mockito/issues/962


mockitoguy:
Thank you for review and for support on strategy to minimize Mockito repo!

PR #1475, opened 2018-08-13
MockitoSession API does not clean up listener when initMocks fails

mockitoguy:
When initMocks() fails we did not clean up the session listener. This led to a confusing error message on the next attempt to create a session (next test method run). I stumbled upon this problem when attempting to configure TestNG to use strict stubbing (https://github.com/mockito/mockito-testng/issues/1).

In order to fix this cleanly I need to keep the existing functionality that protects the user from forgetting to use 'finishMocking()'. Hence, I still need to throw an exception if the user adds the same listener multiple times. However, in the event that the listener is dirty, I clean it up automatically instead of failing.

### DIFF
--- a/src/main/java/org/mockito/exceptions/misusing/InjectMocksException.java
+++ b/src/main/java/org/mockito/exceptions/misusing/InjectMocksException.java
@@ -1,0 +1,16 @@
+/*
+ * Copyright (c) 2018 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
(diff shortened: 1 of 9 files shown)


mockitoguy:
I'm merging this because it blocks https://github.com/mockito/mockito-testng/issues/1. Happy to address code review feedback after this is merged. The change does not involved public API changes, it fixes a bug. All existing tests pass, and the new tests pass, too. Let's ship it!

Issue #1500, opened 2018-09-27
Mockito 2.x.x does not support Kotlin release coroutines


dzharkov:
In https://github.com/mockito/mockito/pull/1032 support for Kotlin suspend function has been introduced.

But the change contains a hard-coded class name of `Continuation` (https://github.com/elizarov/mockito/blob/bf092378fc42f31efa054b4e5e555c10c33f5994/src/main/java/org/mockito/internal/creation/SuspendMethod.java#L10) that is going to be renamed eventually.

Kotlin compiler with languge-version set to 1.3 generates suspend function with continuation parameter of type `kotlin.coroutines.Continuation` and with the current state of mockito it becomes impossible to mock them properly

PR #1501, opened 2018-09-27
Support mocking kotlin suspend functions compiled by Kotlin 1.3 (#1500)

dzharkov:
This is an update for https://github.com/mockito/mockito/pull/1032

Coroutines support is not experimental anymore and a type of continuation-parameter is renamed to `kotlin.coroutines.Continuation`

This PR is assumed to fix https://github.com/mockito/mockito/issues/1500


subprojects/kotlinReleaseCoroutinesTest/kotlinReleaseCoroutinesTest.gradle
```diff
@@ -0,0 +1,26 @@
...
+    mavenCentral()
+    maven { url 'https://dl.bintray.com/kotlin/kotlin-eap' }
+}
+
+dependencies {
+    testCompile project(":")
+    testCompile libraries.junit4
+
+    testCompile  "org.jetbrains.kotlin:kotlin-stdlib:1.3.0-rc-57"
+    testCompile 'org.jetbrains.kotlinx:kotlinx-coroutines-core:0.26.1-eap13'
```
TimvdLippe: We are using `dependencies.gradle` to define our dependencies (https://github.com/mockito/mockito/blob/release/2.x/gradle/dependencies.gradle). Could you use the same format as https://github.com/mockito/mockito/blob/release/2.x/subprojects/kotlinTest/kotlinTest.gradle and add the extra dependency in `dependencies.gradle`?

subprojects/kotlinReleaseCoroutinesTest/kotlinReleaseCoroutinesTest.gradle
```diff
@@ -0,0 +1,26 @@
...
+    }
+}
+
+apply plugin: "org.jetbrains.kotlin.jvm"
+
+description = "Kotlin tests for Mockito."
+
+repositories {
+    mavenCentral()
+    maven { url 'https://dl.bintray.com/kotlin/kotlin-eap' }
```
TimvdLippe: While I do understand we are breaking soon, I would rather wait for an official release before making any changes. (I do want to incorporate change, just when the package has been released).

TimvdLippe: This also leaves the question whether we should maintain 2 versions of our Kotlin tests. Maybe we should just have one and state that we only support the latest version? WDYT @mockito/core ?

subprojects/kotlinReleaseCoroutinesTest/kotlinReleaseCoroutinesTest.gradle
dzharkov: Yes of course, I can. 
I haven't done that just to avoid introducing the version used only within a single test module

subprojects/kotlinReleaseCoroutinesTest/kotlinReleaseCoroutinesTest.gradle
dzharkov: @TimvdLippe Just for clarification, this dependency is test-only and won't be visible for library users.
The relevant change is at https://github.com/mockito/mockito/pull/1501/files#diff-67883a6458cc322a1cf66cbaa8b50b1eL14
And it would be very cool if in the moment of release we had a version of mockito that would work both with old and new kotlin compiler versions.

Thanks!

TimvdLippe: Yes that is true. It is mostly a question regarding compatibility. How long do we want to maintain 2 projects to ensure we remain compatible. I think that is a question the core team should answer to resolve this situation.

jossiwolf: @TimvdLippe This really is a pressing issue - no reply from the core team for 7 days is quite bad. Can you check again? :)

TimvdLippe: I have requested their opinion again. Is there a timeline for release of Kotlin 1.3?

dzharkov: @TimvdLippe We expect it in a few weeks. And since people are already struggling after updating to 1.3-rc getting their test failing for unknown reasons it would be cool to merge it sooner than later

raphw: This looks good to me. 

On a side note, we always take the time we need. This is a free time project for us after all and you cannot expect a reply within five work days. We just got Java 11 support out of the door and if Kotlin decides to break its APIs this is on them and not on us, they must expect such problems when evolving the language that way.

subprojects/kotlinReleaseCoroutinesTest/kotlinReleaseCoroutinesTest.gradle
TimvdLippe: Could you resolve this comment? Then I think this is ready for a merge.

dzharkov: Done

raphw:
@TimvdLippe Could you just also release to MC? Otherwise I will do it tonight.

TimvdLippe:
@raphw Yup doing that right now :smile:

dzharkov:
@TimvdLippe Cool, thanks!

asarkar:
Mocked suspending function returns null with v3.3.3, and this PR is referenced from [this SO thread](https://stackoverflow.com/q/53098975/839733) as a solution. Shall I open a new ticket?

PR #1502, opened 2018-09-27
Try to mock all constructor dependencies if possible.

esarbanis:
Although I couldn't find any relevant issue for this change, I think it will help a lot of people that need to test legacy or flat out bad code.

What I have come across very regularly is test classes that have a lot or `@Mock` fields that are not used for testing, but just to wire the component under test. If the `@InjectMocks` tried to mock needed constructor dependencies, most of the `@Mock` unused field could be eliminated and thus have fewer distractions in the test classes.

I also spotted a [ `TODO`](https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/internal/configuration/injection/ConstructorInjection.java#L30) in `ConstructorInjection` with the same idea. So, I thought I would give it a try.

### DIFF
--- a/src/main/java/org/mockito/internal/configuration/injection/ConstructorInjection.java
+++ b/src/main/java/org/mockito/internal/configuration/injection/ConstructorInjection.java
@@ -5,7 +5,9 @@
 
 package org.mockito.internal.configuration.injection;
 
+import org.mockito.Mockito;
 import org.mockito.exceptions.base.MockitoException;
+import org.mockito.internal.util.MockUtil;
 import org.mockito.internal.util.reflection.FieldInitializationReport;
 import org.mockito.internal.util.reflection.FieldInitializer;
 import org.mockito.internal.util.reflection.FieldInitializer.ConstructorArgumentResolver;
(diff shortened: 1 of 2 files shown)


src/main/java/org/mockito/internal/configuration/injection/ConstructorInjection.java
```diff
@@ -26,13 +28,8 @@
...
  *
- * <blockquote>
- * TODO on missing mock type, shall it abandon or create "noname" mocks.
- * TODO and what if the arg type is not mockable.
- * </blockquote>
- *
  * <p>
- * For now the algorithm tries to create anonymous mocks if an argument type is missing.
+ * The algorithm tries to create anonymous mocks if an argument type is missing.
  * If not possible the algorithm abandon resolution.
```
freakomonk: Very minor typo, it should be algorithm abandons* resolution.
```suggestion
 * If not possible the algorithm abandons resolution.
```

esarbanis: Nice catch! Thank you!

mockitoguy:
Hey @esarbanis! Thank you for reporting and the PR! Can you write pros/cons of this change and also more clearly what is the old behavior vs. new behavior? (was Mockito previously throwing an exception, and now it is "silently" attempting to mock all constructor arguments?).

TimvdLippe:
Per above, could you explain what the original issue was?

src/main/java/org/mockito/internal/configuration/injection/ConstructorInjection.java
```diff
@@ -58,7 +55,7 @@ public boolean processInjection(Field field, Object fieldOwner, Set<Object> mock
     }
 
     /**
-     * Returns mocks that match the argument type, if not possible assigns null.
+     * Returns mocks that match the argument type, if not possible tries to assign an anonymous mock or defaults to null.
```
freakomonk: Can you pls explain on assigning an anonymous mock would help , especially when there is no mock for that argument type ?

TimvdLippe:
Closing per the above comments. I don't think we want to introduce more magic behavior in our injection implementation.

PR #1523, opened 2018-10-25
Added 'lenient' annotation toggle

mockitoguy:
This way we can conveniently configure an existing 'lenient' setting via an annotation. See the unit test that describes the feature.

```Java
@Mock(lenient = true) SomeClass mock;
```

### DIFF
--- a/src/main/java/org/mockito/Mock.java
+++ b/src/main/java/org/mockito/Mock.java
@@ -76,4 +76,6 @@ public @interface Mock {
     Class<?>[] extraInterfaces() default {};
 
     boolean serializable() default false;
+
+    boolean lenient() default false;
 }
--- a/src/main/java/org/mockito/Mock.java
+++ b/src/main/java/org/mockito/Mock.java
@@ -9,6 +9,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
 import org.mockito.junit.MockitoJUnitRunner;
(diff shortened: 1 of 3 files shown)


src/main/java/org/mockito/Mock.java
```diff
@@ -76,4 +76,6 @@
     Class<?>[] extraInterfaces() default {};
 
     boolean serializable() default false;
+
+    boolean lenient() default false;
```
TimvdLippe: This is a more a general point, but we probably should add `mockSettings` here and not have all arguments duplicated in the annotation.

mockitoguy: Why do you think so? It is very nice and clean to configure the parameters directly?

mockitoguy:
Thank you for feedback. I'll add documentation and merge it.

src/main/java/org/mockito/Mock.java
TimvdLippe: Because we duplicate all values from `mockSettings` into this annotation. That feels brittle and easy-to-forget.

mockitoguy: That's fair. 

I don't see what's brittle here - this is public API that we don't change. We can forget to add annotation but... so what... our users will remind us and they can always submit a PR.

It's a rather small trade-off for having a good looking API :)

PR #1535, opened 2018-11-19
Allow delegating to non-public methods for AdditionalAnswers#delegatesTo

li-wjohnson:
Fixes an issue where doing "delegatesTo(new AnonymousClass() {})" would cause an IllegalAccessException, and other access-related issues.

### DIFF
--- a/src/main/java/org/mockito/internal/stubbing/defaultanswers/ForwardsInvocations.java
+++ b/src/main/java/org/mockito/internal/stubbing/defaultanswers/ForwardsInvocations.java
@@ -40,6 +40,7 @@ public class ForwardsInvocations implements Answer<Object>, Serializable {
             }
 
             Object[] rawArguments = ((Invocation) invocation).getRawArguments();
+            delegateMethod.setAccessible(true);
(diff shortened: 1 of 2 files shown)


src/main/java/org/mockito/internal/stubbing/defaultanswers/ForwardsInvocations.java
```diff
@@ -40,6 +40,7 @@ public Object answer(InvocationOnMock invocation) throws Throwable {
             }
 
             Object[] rawArguments = ((Invocation) invocation).getRawArguments();
+            delegateMethod.setAccessible(true);
```
TimvdLippe: We should probably `try-catch` this one, in case we are unable to set it accessible. However, in that case I think we should still continue with calling `delegateMethod.invoke`. E.g. the try-catch-clause should be only around this statement, but not the other statements.

li-wjohnson:
@TimvdLippe: Actually, the existing tests would fail, if they were moved into a different package. Should I create a package "foopackage" within org.mockito.internal.stubbing.defaultanswers, to add my new test?

PR #1536, opened 2018-11-20
Allow delegating to non-public methods for AdditionalAnswers#delegatesTo

li-wjohnson:
Fixes an issue where doing "delegatesTo(new AnonymousClass() {})" would cause an IllegalAccessException, and other access-related issues.

(See https://github.com/mockito/mockito/pull/1535 - Sorry, wasn't aware that you can't change the PR source branch.)


src/test/java/org/mockito/internal/stubbing/defaultanswers/foopackage/ForwardsInvocationsTest.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package org.mockito.internal.stubbing.defaultanswers.foopackage;
+
+import org.junit.Test;
+import org.mockito.internal.stubbing.defaultanswers.ForwardsInvocations;
+import org.mockitoutil.TestBase;
+
+import static org.junit.Assert.assertEquals;
+
+public class ForwardsInvocationsTest extends TestBase {
```
mockitoguy: Do we need this test? Does this test improves coverage on top of the existing ForwardsInvocationsTest?

src/main/java/org/mockito/internal/stubbing/defaultanswers/ForwardsInvocations.java
```diff
@@ -40,6 +40,11 @@ public Object answer(InvocationOnMock invocation) throws Throwable {
             }
 
             Object[] rawArguments = ((Invocation) invocation).getRawArguments();
+            try {
+                delegateMethod.setAccessible(true);
```
mockitoguy: Can you add an e2e user test? Can you put together a test class that shows that without this change Mockito indeed throws access exceptions?

mockitoguy:
Please ignore my earlier comment, I can see the new test is doing what we need to do!

src/test/java/org/mockito/internal/stubbing/defaultanswers/foopackage/ForwardsInvocationsTest.java
TimvdLippe: The original problem was the inability to mock anonymous classes. In line 21 of this test, an anonymous class is created. I think this is therefore testing the appropriate codepath?

mockitoguy:
I reworked the test a little bit and will merge once the build is happy. Thank you @li-wjohnson!

src/test/java/org/mockito/internal/stubbing/defaultanswers/foopackage/ForwardsInvocationsTest.java
mockitoguy: It is. Please disregard my comment.

TimvdLippe: Yeah, our comments were cross-posted. I saw your reply the second I posted my comment 😛

PR #1466, opened 2018-08-03
Expose StubbingLookupListener publicly (#793)

marchpig:
This PR makes `StubbingLookupListener` public and adds `MockObjectListener` as parent of `InvocationListener` and `StubbingLookupListener`. (#793)

PR #1543, opened 2018-11-24
Exposed new API - StubbingLookupListener

mockitoguy:
Exposed new public API based on an internal API - StubbingLookupListener - #793.

As a rule of a thumb we try to expose the internal APIs that the top level features are built upon. Since StubbingLookupListener is useful for us to implement strictness, let's get the former exposed as public API. This way we build stronger framework with the concept of "onion skin API".

This PR replaces #1466 - very nice PR from @marchpig. Due to number of changes I needed to do, I decided to reimplement the feature, rather than use the original PR. @marchpig, thank you for understanding!


marchpig:
It looks good to me!

mockitoguy:
I'll update the version / since tags and will merge shortly (most likely tomorrow).

PR #1544, opened 2018-11-24
Fixes #1541: Prevent premature garbage collection of mock objects

### LINKED ISSUE #1541: Mock object premature garbage collected when using 'One-liner stubs'
### Problem:
'One-liner stubs' may throw an exception because the actual mock object is already garbage collected when trying to return the mock.

### Reproduce:
1. Mockito 2.23.5 using mockito-inline
2. -Xmx256m

```
public class OneLinerStubStressTest {

    public class TestClass {
        public String getStuff() {
            return "A";
        }
    }

    private static String generateString() {
        final int length = 10000;
        final StringBuilder stringBuilder = new StringBuilder(length);
        for (int i = 0; i <= length; i++) {
            stringBuilder.append("B");
        }
        return stringBuilder.toString();
    }

    @Test
    public void call_a_lot_of_mocks() {
        final String returnValue = generateString();
        for (int i = 0; i < 40000; i++) {
            final TestClass mock = when(mock(TestClass.class).getStuff())
                                   .thenReturn(returnValue).getMock();
            assertEquals(returnValue, mock.getStuff());
        }
    }
}
```
#### Expected:
Test finishes successfully

#### Actual:
```
java.lang.IllegalStateException: The mock object was garbage collected. This should not happen in normal circumstances when using public API. Typically, the test class keeps strong reference to the mock object and it prevents getting the mock collected. Mockito internally needs to keep weak references to mock objects to avoid memory leaks for certain types of MockMaker implementations. If you see this exception using Mockito public API, please file a bug. For more information see issue #1313.
        at org.mockito.internal.invocation.mockref.MockWeakReference.get(MockWeakReference.java:32)
        at org.mockito.internal.invocation.InterceptedInvocation.getMock(InterceptedInvocation.java:103)
        at org.mockito.internal.stubbing.InvocationContainerImpl.invokedMock(InvocationContainerImpl.java:157)
        at org.mockito.internal.stubbing.ConsecutiveStubbing.getMock(ConsecutiveStubbing.java:28)
```

### Analysis:
As part of #1313 `DefaultInvocationFactory` and `InterceptedInvocation` were refactored to keep week reference to the mock. This was necessary in order to get rid of strong references by the handlers which caused memory leaks. On the other hand this requires to have a strong reference to the mock in the test class in order to prevent premature garbage collection. Unfortunately if using 'One-liner stubs' (https://static.javadoc.io/org.mockito/mockito-core/2.23.4/org/mockito/Mockito.html#one_liner_stub) the mock object may be premature cleaned up. This occurs because we gain access to the `InterceptedInvocation` when calling `getMock()` (which tries to get the mock from the invocation), without having any strong reference to the mock itself. At this point the mock may already be cleaned up.

maxgrabenhorst:
Fixes #1541.

If using 'One-liner stubs' (https://static.javadoc.io/org.mockito/mockito-core/2.23.4/org/mockito/Mockito.html#one_liner_stub) the mock object may be premature cleaned up and returning the mock fails with an exception. This occurs because there is no strong reference to the mock itself.

Therefore we need to maintain a strong reference to the mock until we've returned it, while making sure that the GC can still cleanup the mock correctly when needed.

Cause implementations of `BaseStubbing` are intended to be cleaned up after usage, we can store a strong ref to the mock there. `getMock` then uses the strong ref (instead of the weak ref provided by the invocation). After cleaning up the implementations of  `BaseStubbing` there is no strong ref to the mock itself anymore (except the one in the test class).

I tried some alternative solutions but I believe this is the only way it works.

----

Bind last mock creation to mockingProgress does not work because of this:
```
when(mock(TestClass2.class).getTestClass()).thenReturn(mock(TestClass.class)).getMock();
```

Bind mock of last invocation for stubbing to mockingProgress does not work because of this:
```
when(mock(TestClass.class).getStuff()).thenReturn("X").thenReturn(
    when(mock(TestClass.class).getStuff()).thenReturn("XXX").<TestClass>getMock().getStuff()
).getMock();
```

### DIFF
--- a/src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
+++ b/src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
@@ -4,15 +4,15 @@
  */
 package org.mockito.internal.stubbing;
 
-import static org.mockito.internal.exceptions.Reporter.notAnException;
-import static org.mockito.internal.progress.ThreadSafeMockingProgress.mockingProgress;
-import static org.objenesis.ObjenesisHelper.newInstance;
-
 import org.mockito.internal.stubbing.answers.CallsRealMethods;
 import org.mockito.internal.stubbing.answers.Returns;
 import org.mockito.internal.stubbing.answers.ThrowsException;
 import org.mockito.stubbing.Answer;
 import org.mockito.stubbing.OngoingStubbing;
+
+import static org.mockito.internal.exceptions.Reporter.notAnException;
+import static org.mockito.internal.progress.ThreadSafeMockingProgress.mockingProgress;
+import static org.objenesis.ObjenesisHelper.newInstance;
 
 public abstract class BaseStubbing<T> implements OngoingStubbing<T> {
 
@@ -20,8 +20,8 @@ public abstract class BaseStubbing<T> implements OngoingStubbing<T> {
     // Keep strong ref to mock preventing premature garbage collection when using 'One-liner stubs'. See #1541.
     private final Object strongMockRef;
 
-    public BaseStubbing(InvocationContainerImpl invocationContainer) {
-        this.strongMockRef = invocationContainer.invokedMock();
+    BaseStubbing(Object mock) {
+        this.strongMockRef = mock;
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


mockitoguy:
This is great! I pushed a couple of commits on top of this PR. Can you take a look? If you're ok with those changes, I'm merging!

maxgrabenhorst:
LGTM. Thanks for your improvements.
Only passing the mock was my first take but I thought passing the `InvocationContainerImpl` is more type safe than having just `Object` as parameter. Also the "mock retrieval logic" would be located in one central place. But I'm very fine with the refactoring.

mockitoguy:
Thank you!

PR #1546, opened 2018-11-26
Enable gradle build cache

### LINKED ISSUE #1545: Enable Gradle Build cache for faster local builds
Let's enable Gradle build cache for faster local builds.

https://docs.gradle.org/current/userguide/build_cache.html

epeee:
This pr fixes #1545.

I was not entirely sure about the settings to choose for the local build cache.
Let's discuss if we should go for a local folder and if keeping cache entries for 30 days is ok. 

check list

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -20,9 +20,7 @@ rootProject.children.each { project ->
 enableFeaturePreview('STABLE_PUBLISHING')
 
 buildCache {
-    local(DirectoryBuildCache) {
+    local {
         enabled = !System.getenv().containsKey("CI")
-        directory = new File(rootDir, 'build-cache')
-        removeUnusedEntriesAfterDays = 30
     }
 }
--- a/settings.gradle
+++ b/settings.gradle
@@ -21,6 +21,7 @@ enableFeaturePreview('STABLE_PUBLISHING')
 
 buildCache {
     local(DirectoryBuildCache) {
+        enabled = !System.getenv().containsKey("CI")
         directory = new File(rootDir, 'build-cache')
         removeUnusedEntriesAfterDays = 30
     }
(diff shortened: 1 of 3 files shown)


szpak:
Thanks for the PR. Have you check the performance gain with your change locally? Do you propose to enable it also with Travis?

epeee:
Ad Travis, good point, let's disable it in this case.
I will update the pr accordingly.

szpak:
How much faster does it build on your machine in a subsequent build with no changes?

epeee:
currently, e.g. the way we have configured jacoco prevents us from cache hits for test tasks.
I've already a fix for this problem locally but it looks like we have more challenges in order to fully benefit from build cache.
With the local jacoco fix build durations are reduced to 
`BUILD SUCCESSFUL in 44s`
compared to 
`BUILD SUCCESSFUL in 2m 56s`

I will dig into the remaining problems (always executed tasks) tomorrow.

settings.gradle
```diff
@@ -18,3 +18,11 @@ rootProject.children.each { project ->
 }
 
 enableFeaturePreview('STABLE_PUBLISHING')
+
+buildCache {
+    local(DirectoryBuildCache) {
+        enabled = !System.getenv().containsKey("CI")
```
mockitoguy: Great! We don't want local cache on CI.

settings.gradle
```diff
@@ -18,3 +18,11 @@ rootProject.children.each { project ->
...
+        directory = new File(rootDir, 'build-cache')
```
mockitoguy: Why do we need this setting or the next one? Can we rely on Gradle's defaults?

mockitoguy:
Great! Thanks a lot for taking a stab at it!

settings.gradle
epeee: good question, let's rely on defaults for now and adjust them if needed later on.

mockitoguy:
Welcome faster builds :) Let's merge it!

PR #1549, opened 2018-11-27
Fixed issue with leftover state when stubbing with bad throwables

### LINKED ISSUE #1514: Mockito remains in unfinished stubbing state if a stubbing fails
If one of `doThrow` instructions failed due to InstantiationError, Mockito remains in "unfinished stubbing" state and can no longer mock anything.

I had this issue in a single JUnit 4 test, but the fact that Mockito remained in "unfinished stubbing" state used to result in _all_ tests being marked as failed (I couldn't reproduce the last effect with the latest Mockito neither in JUnit 4 nor with JUnit 5 now, so the issue is applicable to a single test invocation only).

```java
import java.util.List;
import org.junit.jupiter.api.RepeatedTest;
import org.junit.jupiter.api.RepetitionInfo;
import org.mockito.Mockito;

class UninstantiableTest {

  @RepeatedTest(2)
  void tryThrow01(RepetitionInfo i) {
    if (i.getCurrentRepetition() == 1) {
      // Stub using uninstantiable exception.
      try {
        List mock = Mockito.mock(List.class);

        Mockito.doThrow(UninstantiableException.class).when(mock).clear();
      } catch (InstantiationError expected) {
        // Get a j.l.InstantiationError.
      }
    }

    // Try to stub another mock, this time using a valid exception type.
    {
      List mock = Mockito.mock(List.class);

      // The following operation results in "UnfinishedStubbing"
      Mockito.doThrow(RuntimeException.class).when(mock).clear();
    }
  }

  abstract static class UninstantiableException extends RuntimeException {}
}
```

Mockito version 2.23.0

check that

mockitoguy:
Fixed issue with leftover state when stubbing with bad throwables. Fixes #1514

### DIFF
--- a/src/main/java/org/mockito/internal/stubbing/StubberImpl.java
+++ b/src/main/java/org/mockito/internal/stubbing/StubberImpl.java
@@ -87,12 +87,16 @@ public class StubberImpl implements Stubber {
             mockingProgress().reset();
             throw notAnException();
         }
-        Throwable e;
+        Throwable e = null;
         try {
             e = newInstance(toBeThrown);
-        } catch (RuntimeException instantiationError) {
-            mockingProgress().reset();
-            throw instantiationError;
+        } finally {
+            if (e == null) {
+                //this means that an exception or error was thrown when trying to create new instance
+                //we don't want 'catch' statement here because we want the exception to be thrown to the user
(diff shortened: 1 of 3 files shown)


bric3:
Good catch

Issue #1551, opened 2018-11-28
ClassCastException with generics and smart nulls


isae:
Hello. 

I am trying to migrate my JUnit4 tests (Java version is 1.8.0_74) from Mockito v1.10.19 to v2.0.10 and I have the following exception: 

```java
java.lang.ClassCastException: codegen.java.lang.Object$MockitoMock$220764719 cannot be cast to com.solanteq.solar.fp.dao.scheme.MockitoTest$Entity
```

when I mock my objects using `Answers.RETURNS_SMART_NULLS`.

However, when I change the answer to default, or change the stubbing to do..when approach, the exception disappears. Before version upgrade, everything worked fine.

Test case to reproduce the issue: 
 
```java
import org.junit.Before;
import org.junit.Test;
import org.junit.runner.RunWith;
import org.mockito.Answers;
import org.mockito.Mock;
import org.mockito.Mockito;
import org.mockito.junit.MockitoJUnitRunner;

import static org.junit.Assert.assertEquals;

@RunWith(MockitoJUnitRunner.class)
public class MockitoTest {

    @Mock(answer = Answers.RETURNS_SMART_NULLS)
    private ConcreteDao concreteDao;

    @Before
    public void setUp() {
        Mockito.when(concreteDao.findById(1L))
                .thenReturn(new Entity(1L));
    }

    @Test
    public void testMockito() {
        assertEquals(1L, concreteDao.findById(1L).id);
    }

    public static class AbstractDao<T> {
        T findById(final long id) {
            throw new UnsupportedOperationException("Not implemented");
        }
    }

    public static class Entity {
        private final long id;

        public Entity(long id) {
            this.id = id;
        }
    }

    public static class ConcreteDao extends AbstractDao<Entity> {
    }
}
```

Thanks in advance.

mockitoguy:
Thank you for reporting! Based on your test I created a minimal test that reproduces the issue: #1552. I marked this ticket with "please contribute" hoping you or someone from the community can help. The issue is tricky and it is caused by Java generics combined with smart nulls.

Patouche:
Hello !

If you're agree with that, I can work on this. I've start watching how it works and how it can be solve. I  also think that it can be pretty tricky but I can even try ;) !

PR #1552, opened 2018-11-28
Minimal test to repro generic issue with smart nulls

mockitoguy:
Created a minimal test that reproduces bug with smart nulls and generics, see issue #1551


mockitoguy:
Since this is a new test, I'll just merge it without review. Happy to fix / change things later if someone reviews it.

PR #1554, opened 2018-11-28
Fixes #1553: Fix error when JUnit 3 is in the classpath.

### LINKED ISSUE #1553: VerifyError thrown when JUnit 3 is in the classpath
With JUnit 3 in the classpath, VerifyError will be thrown during JUnit detection in ExceptionFactory. This is because JUnit's ComparisonFailure class only starts extending AssertionError in JUnit 4.

**Expected Behavior**

When JUnit 3 is in the classpath, Mockito should behave as if JUnit was not there at all.

A real example where handling this here instead of updating JUnit in the consumer would be beneficial is related to my company's usage of ZooKeeper. ZooKeeper 3.4 leaks a JUnit 3.8.1 dependency that is fixed in 3.5, but 3.5 is still in beta and has been in the works for several years. Projects that use TestNG (most of ours) but which use ZooKeeper (usually transitively) run into this issue.

**Example**

src/test/java/com/example/ExampleTest.java
```
package com.example;

import org.testng.annotations.Test;

@Test
public class ExampleTest {
  public void testMockitoWithJUnit3() {
    Client mockClient = mock(Client.class);
    mockClient.call("one");
    verify(mockClient).call("two");
  }

  public interface Client {
    void call(String argument);
  }
}
```
build.gradle
```
plugins {
  id 'java'
  id 'idea'
}

repositories {
  mavenCentral()
}

dependencies {
  testImplementation group: 'junit', name: 'junit', version: '3.8.2'
  testImplementation group: 'org.mockito', name: 'mockito-core', version: '2.23.4'
  testImplementation group: 'org.testng', name: 'testng', version: '6.14.3'
}

test {
  useTestNG()
}
```

**Error**
```
java.lang.VerifyError: Bad return type
Exception Details:
  Location:
    org/mockito/internal/junit/ExceptionFactory$JUnitArgsAreDifferent.create(Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;)Ljava/lang/AssertionError; @10: areturn
  Reason:
    Type 'org/mockito/exceptions/verification/junit/ArgumentsAreDifferent' (current frame, stack[0]) is not assignable to 'java/lang/AssertionError' (from method signature)
  Current Frame:
    bci: @10
    flags: { }
    locals: { 'java/lang/String', 'java/lang/String', 'java/lang/String' }
    stack: { 'org/mockito/exceptions/verification/junit/ArgumentsAreDifferent' }
  Bytecode:
    0000000: bb00 0259 2a2b 2cb7 0003 b0            

	at org.mockito.internal.junit.ExceptionFactory.canLoadJunitClass(ExceptionFactory.java:33)
	at org.mockito.internal.junit.ExceptionFactory.<clinit>(ExceptionFactory.java:12)
	at org.mockito.internal.exceptions.Reporter.argumentsAreDifferent(Reporter.java:313)
        ...
```

tjni:
Mockito will throw VerifyError during its JUnit detection when JUnit 3
is in the classpath, because the code only supports JUnit 4 and later.
This change will catch this error and pretend that JUnit is not there.

### DIFF
--- a/src/main/java/org/mockito/internal/junit/ExceptionFactory.java
+++ b/src/main/java/org/mockito/internal/junit/ExceptionFactory.java
@@ -33,6 +33,8 @@ public class ExceptionFactory {
             JUnitArgsAreDifferent.create("message", "wanted", "actual");
         } catch (NoClassDefFoundError onlyIfJUnitIsNotAvailable) {
             return false;
+        } catch (VerifyError onlyIfJUnitIsIncompatible) {
+            return false;
         }
         return true;
     }
--- a/src/main/java/org/mockito/internal/junit/ExceptionFactory.java
+++ b/src/main/java/org/mockito/internal/junit/ExceptionFactory.java
@@ -31,9 +31,7 @@ public class ExceptionFactory {
     private static boolean canLoadJunitClass() {
         try {
             JUnitArgsAreDifferent.create("message", "wanted", "actual");
-        } catch (NoClassDefFoundError onlyIfJUnitIsNotAvailable) {
-            return false;
-        } catch (VerifyError onlyIfJUnitIsIncompatible) {
+        } catch (Throwable onlyIfJUnitIsNotAvailable) {
             return false;
         }
         return true;


tjni:
I would like some opinions on testing, because I was not able to figure out a good way to test this. I tried changing the class hierarchy of ComparisonFailure using Byte Buddy and a custom classloader but was not able to get that to work. I also thought it would be possible (though heavy) to use a Gradle property to change the version of the JUnit dependency to run a test and add that to the test matrix, but I don't expect that to work without other changes because Mockito itself uses JUnit 4 to compile and run its own tests.

mockitoguy:
I'm making a couple of tweaks and I'll merge it.

Don't worry about tests. I don't want to keep tests that use JUnit3 because that version is too old. Maintaining it is more cost than value. Can you verify that this fix resolved the problem you were seeing?

Thank you for the contribution!

mockitoguy:
Please confirm that the fix works for you and I'll merge it. Thank you again!

tjni:
Verified. I tested this in a small project and in the larger work project I encountered this in as well. Thanks again!

PR #1555, opened 2018-11-30
Fixed JaCoCo build cache issue

mockitoguy:
Jacoco Gradle plugin breaks cacheability, hence we only configure the plugin on demand.

Bumped Build Scans plugin so that it is easier to debug build cache issues.

This PR is needed for #1546

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,5 +58,5 @@ script:
 
 after_success:
   #Generates coverage report:
-  - ./gradlew --stacktrace coverageReport && cp build/reports/jacoco/mockitoCoverage/mockitoCoverage.xml jacoco.xml || echo "Code coverage failed"
+  - ./gradlew coverageReport -Pjacoco -s && cp build/reports/jacoco/mockitoCoverage/mockitoCoverage.xml jacoco.xml || echo "Code coverage failed"
   - bash <(curl -s https://codecov.io/bash) || echo "Codecov did not collect coverage reports"
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,5 +58,5 @@ script:
 
 after_success:
   #Generates coverage report:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -58,5 +58,5 @@ script:
 
 after_success:
   #Generates coverage report:
-  - ./gradlew --stacktrace coverageReport && cp build/reports/jacoco/mockitoCoverage/mockitoCoverage.xml jacoco.xml || echo "Code coverage failed"
```
TimvdLippe: I think having `--stacktrace` here makes it easier to debug on Travis, so let's revert that change?

epeee: What about enable build scans in order to use build scans to debug problems like this?

gradle/root/coverage.gradle
```diff
@@ -1,3 +1,11 @@
+if (!project.hasProperty("jacoco")) {
+    //JaCoCo plugin prevents using Build Cache. We will enable jacoco configuration only when specific property is supplied
+    logger.info "JaCoCo code coverage will not be configured because 'jacoco' project property is not present."
+    return //don't evaluate this Gradle file any further
+} else {
+    logger.lifecycle "Configuring JaCoCo code coverage because 'jacoco' project property is present."
+}
```
epeee: I'm ok with this change. 
The other option would be to set `tasks.test.jacoco.enabled = false`.
In this case we don't have to add another parameter to the command line.

.travis.yml
mockitoguy: @TimvdLippe, I agree. I just shorthanded this option to "-s".

@epeee, good idea, I'll add it.

gradle/root/coverage.gradle
mockitoguy: Great! The workaround in this PR is cheap so I'll merge it. Happy to remove it when we migrate to Gradle 5.0. I'll add a note.

PR #1558, opened 2018-12-03
Use errorprone gradle plugin v0.6 and extract errorprone config

epeee:
let's use latest errorprone plugin and extract errorprone configuration.

This should also make the build java11 ready.


szpak:
Thanks.

PR #1560, opened 2018-12-04
 Simplify travis.yml (openjdk10 and openjdk11)

epeee:
... and remove no longer needed wget of install-jdk.sh

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,9 +32,6 @@ matrix:
     env: SKIP_RELEASE=true JDK='Oracle JDK 11'
   - jdk: openjdk11
     env: SKIP_RELEASE=true MOCK_MAKER=mock-maker-inline JDK='Oracle JDK 11'
-  allow_failures:
-    - env: SKIP_RELEASE=true JDK='Oracle JDK 11'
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -41,9 +41,6 @@ branches:
   except:
   - /^v\d/
 
-before_install:
-  - wget https://raw.githubusercontent.com/sormuras/bach/master/install-jdk.sh
```
TimvdLippe: Do we maybe need this for future jdk versions?

szpak: We can always bring it back.

szpak: Travis under the hood uses that script (in some version).

epeee:
@TimvdLippe we should pass on java11 once #1558 is merged.

.travis.yml
epeee: We can add openjdk-ea in order to use java12 early access builds.

szpak: Good idea with Java 12 EA

szpak:
I merged #1558. @TimvdLippe Feel free to merge it once you are convinced :).

epeee:
@TimvdLippe I just rebased and got rid of `allow_failures`.

Issue #1469, opened 2018-08-07
Recorded invocation arguments can be modified after invocation


joeybic:
When an argument passed to a mock method is modified following the mock method invocation, the recorded invocation contains the modified argument, __NOT__ the argument it was originally called with. This causes some confusing, deceptive behavior.

Consider the following test and set of objects:

```java
import lombok.AllArgsConstructor;
import lombok.Data;
import lombok.extern.slf4j.Slf4j;
import org.junit.Test;
import org.mockito.MockingDetails;
import org.mockito.Mockito;
import org.mockito.invocation.Invocation;

import java.util.List;

import static org.mockito.ArgumentMatchers.any;
import static org.mockito.Mockito.*;

@Slf4j
public class MockitoIssueTest {
    @Test
    public void doStuffShouldCallMethodToVerifyWithAMessage() {
        final String expectedMessage = "A message.";
        final ArgumentClass argument = new ArgumentClass(expectedMessage);
        final ArgumentClass expectedArgument = new ArgumentClass(expectedMessage);

        final MockClass mockClass = mock(MockClass.class);
        final ServiceClass service = new ServiceClass(mockClass);

        doAnswer(invocation -> {
            log.info("argument (during mock invocation):          {}", argument);
            log.info("vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv SEE ACTUAL ARGUMENT HERE vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv");
            log.info("invocation.getArgument(0):                  {}",
                    (Object) invocation.getArgument(0));
            log.info("^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ SEE ACTUAL ARGUMENT HERE ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^");
            log.info("argument.equals(invocation.getArgument(0)): {}",
                    argument.equals(invocation.getArgument(0)));
            return null;
        }).when(mockClass).methodToVerify(any(ArgumentClass.class));

        log.info("argument (before call):                     {}", argument);
        service.doStuff(argument);
        log.info("argument (after call):                      {}", argument);

        final MockingDetails mockingDetails = Mockito.mockingDetails(mockClass);
        final Invocation invocation = ((List<Invocation>) mockingDetails.getInvocations()).get(0);
[...]

Issue #1153, opened 2017-07-26
Modifying invocation arguments results in incorrect verification


m35:
### Environment
mockito-core-2.8.9.jar
Windows 10
Java 1.8

### Issue

It's hard to summarize this issue. Modifying the invocation arguments also modifies what the verification uses. That may not be clear, so here is an example.

```
import org.junit.Assert;
import org.junit.Test;
import org.mockito.AdditionalMatchers;
import org.mockito.Mockito;
import org.mockito.invocation.InvocationOnMock;
import org.mockito.stubbing.Answer;

public class ModifyArgTest {

    public interface ModifyMyArgs {
        void modifyMe(byte[] buffer);
    }

    @Test
    public void testBug() {

        ModifyMyArgs mock = Mockito.mock(ModifyMyArgs.class);

        Mockito.doAnswer(new Answer<Void>() {
            @Override
            public Void answer(InvocationOnMock invocation) throws Throwable {
                byte[] buffer = invocation.getArgument(0);
                buffer[0] = 1;
                buffer[1] = 2;
                buffer[2] = 3;
                return null;
            }
        }).when(mock).modifyMe(AdditionalMatchers.aryEq(new byte[3]));

        byte[] actual = new byte[3];
        mock.modifyMe(actual);

        Assert.assertArrayEquals(new byte[] {1, 2, 3}, actual);

        // this next line should pass, but fails
        Mockito.verify(mock).modifyMe(AdditionalMatchers.aryEq(new byte[3]));
    }
}
```

The output is
```
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.306 sec <<< FAILURE! - in ModifyArgTest
testBug on testBug(ModifyArgTest)(ModifyArgTest)  Time elapsed: 0.766 sec  <<< FAILURE!
org.mockito.exceptions.verification.junit.ArgumentsAreDifferent: 
Argument(s) are different! Wanted:
modifyMyArgs.modifyMe(
    [(byte) 0x00, (byte) 0x00, (byte) 0x00]
);
-> at ModifyArgTest.testBug(ModifyArgTest.java:38)
Actual invocation has different arguments:
modifyMyArgs.modifyMe(
    [(byte) 0x01, (byte) 0x02, (byte) 0x03]
);
-> at ModifyArgTest.testBug(ModifyArgTest.java:34)

	at ModifyArgTest.testBug(ModifyArgTest.java:38)


Results :

Failed tests: 
[...]

PR #1563, opened 2018-12-07
Fix ClonesArguments for null and arrays

tombrown52:
The ClonesArguments class works for many cases, but fails when trying to clone an argument that is an array, or is null. This PR fixes those edge cases. This does not directly resolve any open issue, but it is related to #1153, #1469

check list


tombrown52:
@TimvdLippe  I have modified the existing test to include the new code

TimvdLippe:
This PR was made to the wrong branch. Sorry for missing your reply :cry: Do you mind opening the same PR with a new regression test on `release/2.x` instead?

PR #1566, opened 2018-12-11
Support lambda

### LINKED ISSUE #1068: feature request: spying on lambdas
Here's some Java8 code that works today (even with Mockito 1.x):

```
Function<Integer, Integer> incrementer = x -> x + 1;
Function<Integer, Integer> spyIncrementer = 
    mock(Function.class, AdditionalAnswers.delegatesTo(incrementer));
```

I then use `spyIncrementer` as an argument to the code that I really want to test and write a series of `verify` calls to check how `spyIncrementer` was used. In particular, I'm testing a memoization system that should only ever call call this lambda once with any given argument, so there are a series of Mockito calls along the lines of `verify(spyIncrementer, atMost(1)).apply(1);`

It would be preferable to be able to write something more like:
```
Function<Integer, Integer> spyIncrementer = spy(x -> x + 1);
```

With `org.mockito:mockito-inline:2.7.22`, this yields a `NoClassDefFoundError`. If you instead use `org.mockito:mockito-core:2.7.22`, you get _Mockito cannot mock/spy because : - final class_.

brachi-wernick:
fixes #1068 

need help how to make this work on java 6, because Mockito-core is still on Java-6

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -1805,6 +1805,15 @@ public class Mockito extends ArgumentMatchers {
         return mock(classToMock, withSettings());
     }
 
+    public static <T, R> java.util.function.Function<T,R> mock(java.util.function.Function<T,R> function){
+        java.util.function.Function<T,R> mock = Mockito.mock(java.util.function.Function.class);
+                when(mock.apply(any())).thenAnswer(invocation -> {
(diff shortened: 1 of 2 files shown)


TimvdLippe:
Closing this per https://github.com/mockito/mockito/issues/1068#issuecomment-446266057

PR #1567, opened 2018-12-13
Add warning to inline mock maker when running on a JRE

HoldYourWaffle:
I made a topic about this in the [mailing list](https://groups.google.com/forum/#!topic/mockito/vKhx3R4_doQ).

Long story short: the inline mock maker doesn't work when running on just a JRE, so I added a warning in the error message if that's that case. This would've saved me a lot of headache, so I figured I'd maybe help a poor soul that encounters this in the future.


TimvdLippe:
Thanks again!

PR #1569, opened 2018-12-13
Fix #1398 : Allow partial mixing of concrete values with matchers

### LINKED ISSUE #1398: Allow partial mixing of concrete values with matchers

eyalroth:
This PR intends to fix / implement #1398 by allowing mixing of concrete values with matchers in stubbing and / or verification, while also changing most of the default / marker values for the matchers of primitive types. Other matcher "marker" values remain the same; i.e, most of them still return `null`.

A few notes:
- This is my first contribution to this repository, so please double check me.
- I wasn't sure if I should / how should I use the `@since` documentation (I just replaced the old documentation with a newer one).
- As I mentioned a few months back (in the issue), a big part of the tests did not pass on my environment upon cloning the project. It seems to me though that all the tests relevant to this change have passed and are still passing after the change, but I may very well be missing something. 

@TimvdLippe @ChristianSchwarz @mockitoguy - you guys have been a part of the original discussion, so I would love your review here :)


eyalroth:
@TimvdLippe Fixed that. I wasn't sure what to do with the "regexp header" for the new file, so I just added a generic license message copied from one of the currently opened PRs.

`mock.method(1398)` should work as if `1398` is just a concrete value - this is tested by `testStubUsingConcreteValuesEqualToMatchersMarkersAndNoOtherMatchers` in  `VerificationAndStubbingUsingMatchersTest`.

`1398` (`short`) will fail to work only if mixed with other matchers, which didn't work before anyhow, so that shouldn't break anything.

ChristianSchwarz:
Just for reference, there is an other PR open dealing with the same issue: #1441

eyalroth:
@ChristianSchwarz Oh god, I didn't see the PR, i'm so sorry.

Do you want to continue the previous one or is it OK to take this one? I feel this one is a bit more complete in a few senses:
- It supports all of the primitive types.
- It uses a marker values for the primitives which is (in most cases) not `0`, and that allows for mixing of this value alongside matchers (say something like `verify(calculator, never()).divide(anyInt(), 0)`) , which I believe is quite a common case (definitely not rare).
- It has updated documentation and error reporting.

Despite that, I would like to take the approach of your PR which places the entire logic behind the binding of matchers in the `MatchersBinder` and leaves out the `InvocationMatcher` alone. It would make it simpler to test this specific functionality in a unit-test (which is currently only covered in cross-units usage tests), and will keep the original unit-test for the `InvocationMatcher` simpler. Also, it would make more sense to do the actual binding and the usage-verification in the same place.

eyalroth:
@ChristianSchwarz I have added changes to this PR to reflect the separation of the matchers' binding logic from the `InvocationMatcher` to the `MatchersBinder`.

I'm not sure why, but the code-coverage report insists that some lines in `MatchersBinder` are not covered, where in fact the new `MatchersBinderTest` covers them for certain (I've debugged it). I don't have a lot of experience with this coverage tool, perhaps it's buggy? I would've asserted that the new test was actually executed, but I can't figure how to see which tests were executed in the build.

Anyhow, @TimvdLippe, @mockitoguy and perhaps even @bric3   - I would love to have your review :)

ChristianSchwarz:
@eyalroth 
Thank you for looking into my PR and taking over some concepts from it! 

> Do you want to continue the previous one or is it OK to take this one?

My strategy was to wait for a definite "Go!" from the mockito reviewers. I didn't get "Go!" so I put the prototype on hold. Mockito PR's with a bigger change set get hardly a review, if they were not written by one of the core members. So tried to get a review for that prototype first. 
Since your PR is more complete I would say it make sense to continue here. On the other hand I see room for improvements in your implementation for the matcher binding process. In that particular case I would like follow my implementation. 

Once the mockito-core team give you a "Go!" for the feature in general, I would like to give you some review feedback. So we can be sure we don't waste time here.

eyalroth:
@ChristianSchwarz Hmm, good to know.

Honestly I just have some time off right now, so I didn't mind putting up a complete PR instead of a prototype. Since I have the time now I would also love it if you'd provide a full review, but I understand that reviewing is also quite a time-consuming task, so no hard feelings if you wish to wait :)

Regarding the binding process, I didn't see much of a difference between the two implementations other than coding style which I have no problem changing (especially since I've gotten used to Scala).

The major difference I guess is the explicit check of the argument type, which in my PR is "implicitly" being done in the `isMarker` method by getting the argument's class. The other difference is that in my PR, in case the matcher-count and the argument-count are the same, the code still asserts that all the arguments are markers (otherwise it throws an exception).

ChristianSchwarz:
@eyalroth 

I think it is a good idea to take the argument type into account when the marker vaule is checked. Otherwise multiple marker values a allowed per type:
 - If the argument is of type `Long` the values `null` and `MARKER_LONG` are recognized as marker, but `null` is not required to be a marker.
- If the argument is of type 'Object' the values `null` and any `MARKER_xxx` will be recognized.

Why is `null` not the marker value for all Object types? In other words why have List's, Set's, Iterable's there own marker? I mean they can have ther own marker value but then null should not be a marker too.

eyalroth:
@ChristianSchwarz 

> If the argument is of type `Long` the values `null` and `MARKER_LONG` are recognized as marker, but `null` is not required to be a marker.

This is exactly why the documentation on `ArgumentMatchers` now state that `null` is not allowed to be mixed with other matchers:

> Since `null` is used as a "marker" value by many argument-matchers, it is impossible to mix concrete `null`s with other argument matchers; use `isNull` instead.

And obviously if no matchers are used then the `null` value on the `Long` parameter will not be evaluated and will be treated as a concrete value (much like `MARKER_LONG` would).

> If the argument is of type 'Object' the values null and any `MARKER_xxx` will be recognized.

This is actually intended. Why should these behave differently?

```java
public void intArgs(int a, int b);
public void objectArgs(Object a, Object b);

when(mock.intArgs(gt(1), 13981398)); // throws exception because 13981398 is the marker value for int
when(mock.objectArgs(eq("lala"), 13981398)); // should also throw exception, since this is a mixture of matchers with concrete values that happen to be marker values
```

After all, stubbing / verifying a method with `Object` arguments but with a primitive matcher (such as `gt(1)`) will be invoking the method with the primitive marker, not `null`, therefore this protection is still very much needed.

> Why is `null` not the marker value for all Object types? In other words why have List's, Set's, Iterable's there own marker? I mean they can have ther own marker value but then null should not be a marker too.

I just kept it as it is now, as I wasn't sure exactly what mechanism might rely on these values being collections and not `null`s.

TimvdLippe:
> This is exactly why the documentation on ArgumentMatchers now state that null is not allowed to be mixed with other matchers:

This is my primary concern with this feature. Having to rely on documentation rather than code will mean that users will run into difficult and hard to understand situations, until they find out this one subtlety. I prefer to not ship features with subtleties unless we can catch them in our code and report it in a nice manner to our users.

eyalroth:
@TimvdLippe

> This is my primary concern with this feature. Having to rely on documentation rather than code will mean that users will run into difficult and hard to understand situations, until they find out this one subtlety. I prefer to not ship features with subtleties unless we can catch them in our code and report it in a nice manner to our users.

Well, that is already the case now; i.e, there is not possibility to detect this at compile time, so Mockito throws an exception which directs the user to the documentation. This is why I've also updated the `Reporter.java`:


> This exception may occur if matchers are combined with "marker" values:
> ```java
>     //incorrect:
>     someMethod(anyObject(), null);
> ```
> When mixing matchers with concrete values, concrete values should not contain any "marker" value. For example:
> ```java
>     //correct,
>     someMethod(anyObject(), \"Just a string\");
> ```
> This exception might also occur if matchers are casted to a similar type:
> ```java
>     //incorrect:
>     someMethod(eq(100)); // actual method accepts long, not int
>     //correct:
>     someMethod(eq(100L));
> ```
> For more info see javadoc for Matchers class.

And the javadocs for the deprecated `Matchers` class leads to the javadoc for the `ArgumentMatchers` class, which explains these subtleties in a fairly simple manner IMHO.

eyalroth:
@TimvdLippe @mockitoguy Hey, so what's basically holding this up?

TimvdLippe:
I think we need to discuss this feature whether we actually want to ship this. I am still sceptical, as I dislike the magic values and having to rely on documentation. We have existing debt in these spaces and I would like us to see move away from that instead of to it. So I would need some time to really dig into this and figure out what the edge-cases are. Sadly I am quite busy atm (and so are the others), so we have little time reviewing Mockito PRs atm. 😢

TimvdLippe:
Closing per https://github.com/mockito/mockito/pull/1441#issuecomment-488753561

eyalroth:
> Closing per [#1441 (comment)](https://github.com/mockito/mockito/pull/1441#issuecomment-488753561)

Disappointing, but understandable 👍

Issue #1444, opened 2018-07-25
Artifact ID changes


ming13:
Since `2.20.0` Mockito Maven artifact IDs [have changed](http://search.maven.org/#search%7Cga%7C1%7Cg%3A%22org.mockito%22).

* `mockito-android` → `android`
* `mockito-inline` → `inline`
* `mockito-junit-jupiter` → `junit-jupiter`

Not sure if it is an intended change or not, so decided to clarify things a bit. So... Is it an intended change? It isn’t mentioned in the changelog.

mockitoguy:
It's a bug. Thank you for reporting!

mockitoguy:
@epeee, can you take a look at this?

mockitoguy:
Upstream bug in Shipkit: https://github.com/mockito/shipkit/issues/741

epeee:
@mockitoguy sorry, I was busy yesterday. 
I can have a look at the upstream bug in shipkit tomorrow.

mockitoguy:
There is some delay. Bintray log shows that the sync was successful. Let's wait until the end of the day.

Thanks for checking!

mockitoguy:
OSS repo has the right artifacts so I think it's just a delay before it's searchable in central. Proof - [pom.xml](https://oss.sonatype.org/service/local/artifact/maven/redirect?r=releases&g=org.mockito&a=mockito-inline&v=2.20.1&e=pom) in the Sonatype OSS 'release' repo.

Let's keep the ticket open until we add tests for the artifacts. @ming13, do you want to contribute a test that would verify artifact IDs? This should be a form of a Gradle task.

mockitoguy:
The issue is fixed in 2.20.1. I've added tests to prevent the issue in the future.

Thanks again for reporting!!!

PR #1575, opened 2018-12-21
Remove Workaround for #1444

nowakpawel:
Removed workaround for #1444.


TimvdLippe:
Hm, Travis fails. Must we also update Shipkit? @mockitoguy do you know more about this?

TimvdLippe:
This still requires an update to Shipkit. Closing this for now and hopefully it is fixed soon upstream.

PR #1585, opened 2019-01-09
Fix #1584 where MultipleParentClassLoader lacks of classloaders of superclasses

### LINKED ISSUE #1584: ClassLoader built by mockito is not able to load declared types in the hierarchy in OSGi
We develop in OSGi environment and we are using latest version of mockito available on maven central 2.23.4

We got an issue when we mock classes given to guice (sisu) for injection. Sisu/Guice analyse all declared fields and methods for the input type and its super classes to built a sort of index of known types. In OSGi it fails with ClassNotFoundException on mocked classes. The reason is that the mocked class built with `SubclassBytecodeGenerator` has a classloader not able to load types reachable only by supertypes. In OSGi this does not happen because each class has associated the classloader from which is loaded. This guarantee that the class is able to reach/load each type it declares (by its methods, fields or annotations).

Example:

- Bundle A
`foo.Class1` has a protected method that returns foo.impl.OtherClass
`foo.impl.OtherClass`

- Bundle B (imports only `foo` package from bundle A)
`acme.Class2 extends Class1`

`Mockito.mock(foo.Class1).getDeclaredMethods()` will fails. To fix this issue `SubclassBytecodeGenerator` should compose a classloader (`MultipleParentClassLoader`) with also the classloader of superclass the mocked type (`features.mockedType`).

nfalco79:
The MultipleParentClassLoader is now composed also with classloaders of supertypes of the mockedType so in OSGi is able to load all declared types in the hierarchy.

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
@@ -28,7 +28,9 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.Random;
 
 import static java.lang.Thread.currentThread;
@@ -79,7 +81,7 @@ class SubclassBytecodeGenerator implements BytecodeGenerator {
     @Override
     public <T> Class<? extends T> mockClass(MockFeatures<T> features) {
         ClassLoader classLoader = new MultipleParentClassLoader.Builder()
-            .appendMostSpecific(features.mockedType)
+            .appendMostSpecific(getAllTypes(features.mockedType))
             .appendMostSpecific(features.interfaces)
             .appendMostSpecific(currentThread().getContextClassLoader())
             .appendMostSpecific(MockAccess.class)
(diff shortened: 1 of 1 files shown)


nfalco79:
The problem of test case is not possible reflect an real OSGi environment. If for you is the same i create test case to test if the multiparentclassloader contains more classloaders

nfalco79:
A job fails to install jdk11, other jobs are ok

raphw:
I think that this can cause issues with class loader hierarchies where we would now generate a new class loader even if all classes were visible from the top-most type. This would effectively prevent mocking package-private classes and methods for such types which worked until now.

I just extended the multiple parent class loader in Byte Buddy to allow for adding

nfalco79:
> I think that this can cause issues with class loader hierarchies where we would now generate a new class loader even if all classes were visible from the top-most type.

it's already like that, the actual classloader is built with a composition of classloader of actual type and the classloader from all interfaces class. My PR add classloader from superclasses. I had a look to the implementation of `MultipleParentClassloader` and it delegate to parents in the exact order they are added. This means that if the classloader of the actual type is able to load all classes other classloaders are not visited.

TimvdLippe:
This needs a rebase now.

nfalco79:
rebase done

TimvdLippe:
We encountered an issue with Java 11 which should now be fixed. Could you rebase again? Sorry for the troubles.

Awaiting @raphw review and then we can merge.

raphw:
This works for me as we now apply a filter to only add the most specific class loader in a hirarchy. Before, we would have added several class loaders in a hierarchy what would had ruined injection and therewith mocking of package-private types.

Looks good for me but I am a bit excited to see if this breaks things in a way we do not anticipate. However, I do not see an obvious problem which is why I suggest to merge.

nfalco79:
Please also consider a release on maven central

nfalco79:
I had attached an OSGi test case but I have no idea how to integrate in gradle.
It is a simple maven project that run a test case using pax-exam that cover the scenario described i the issue. If in the pom.xml you change the version of mockito <= 2.23.17 the test will fail  with NoClassDefFound. With mockito version >= 2.23.18 it works.

[osgi-test.zip](https://github.com/mockito/mockito/files/2808558/osgi-test.zip)

PR #1590, opened 2019-01-15
Add read edge

raphw:
If Mockito is executed as a named module, the dispatcher class needs to be made accessible to Mockito's module as the dispatcher is always loaded on the bootstrap loaders unnamed module.


raphw:
Yes we can and doing so I found out that this is actually not even necessary. I will push the tests on another branch.

PR #1592, opened 2019-01-16
Module fix

raphw:
Allows to run Mockito as an automated module and adds additional tests.


chanseokoh:
Making a note that this is to fix https://github.com/mockito/mockito/issues/1483#issuecomment-454087291.

PR #1600, opened 2019-01-26
Automatic dependency update with Dependabot

szpak:
Configuration for [Dependabot](http://dependabot.com/) to evaluate automatic dependency update (via PRs).

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,9 @@
+version: 1
+
+update_configs:
+  - package_manager: "java:gradle"
+    directory: "/"
+    update_schedule: "daily"
+    # Redundant - default repository branch by default
+    target_branch: "release/2.x"
+


TimvdLippe:
Travis was unhappy, but that should be fixed on `release/2.x` and would need a simple rebase.

It seems like the parser is specified in https://github.com/dependabot/dependabot-core/blob/master/gradle/lib/dependabot/gradle/file_parser/property_value_finder.rb and that would need to be updated to support our pattern (or we change our pattern to something that is compatible with dependabot).

szpak:
Thanks @TimvdLippe for looking on it. I created an [issue](https://github.com/dependabot/feedback/issues/345). It would be good to work out some easy to match with regex good practices to re-use also in other Gradle projects.
I propose to merge it as is (after Travis is happy), activate via GitHub Market to see the current results and see changes after potential improvement on their side (Dependabot seems to use Continuous Delivery).

szpak:
@TimvdLippe https://github.com/dependabot/feedback/issues/345 is fixed. Can you accept that PR?

szpak:
I had problem reaching out @mockitoguy to enable [Dependabot](https://github.com/marketplace/dependabot/), so in the end I tested it with my fork. Seem quite ok: https://github.com/szpak/mockito/pulls?q=is%3Apr+is%3Aopen+label%3Adependencies.

PR #1613, opened 2019-02-08
Add test that mocks an interface with a finalize method

TimvdLippe:
This is a regression compared to the CGLibMockmaker, for which an interface with a method named "finalize" can not be mock. The call here to `finalized.finalize()` throws `java.lang.IllegalAccessError: org.mockitousage.stubbing.StubbingFinalizeTest$FinalizeInterface$MockitoMock$809502244.finalize()V`

I think this is a bug with ByteBuddy or a bug in how we define the type in `SubclassBytecodeGenerator`?

### DIFF
--- a/src/test/java/org/mockitousage/stubbing/StubbingFinalizeTest.java
+++ b/src/test/java/org/mockitousage/stubbing/StubbingFinalizeTest.java
@@ -1,0 +1,20 @@
+package org.mockitousage.stubbing;
+
+import static org.mockito.Mockito.mock;
+
+import org.junit.Test;
+
+public class StubbingFinalizeTest {
+
+    interface FinalizeInterface {
+        void finalize();
+    }
+
+    @Test
+    public void should_not_throw_illegal_access_error_when_stubbing_finalize() {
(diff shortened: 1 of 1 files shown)


TimvdLippe:
ByteBuddy assumes that an interface is a `TypeDescription.Generic.OBJECT` in https://github.com/raphw/byte-buddy/blob/ad7aa3e96f9b6a97a019525ca5c3ce42a02b1fd9/byte-buddy-dep/src/main/java/net/bytebuddy/ByteBuddy.java#L407-L409 The `isDefaultFinalizer` checks for `OBJECT` as well: https://github.com/raphw/byte-buddy/blob/ad7aa3e96f9b6a97a019525ca5c3ce42a02b1fd9/byte-buddy-dep/src/main/java/net/bytebuddy/matcher/ElementMatchers.java#L1587-L1589

Do we need a special case where we can say "this is an object from an interface", so that `isDefaultFinalizer` does not inadvertently skip the definition of the `finalize()` method from the interface?

TimvdLippe:
Closing this for now.

PR #1619, opened 2019-02-12
New API to clean up all inline mocks after test

### LINKED ISSUE #1532: Nested spies cause memory leaks
Seems like nested spies can cause memory leaks since such objects are kept in memory without purging. Not sure if it can be resolved at all. Should it be avoided? Is there a mention about this in docs? Anyway, the code speaks better and fortunately I’ve been able to create a self-contained sample.

BTW I can provide a `.hprof` file if you are interested.

#### Versions

```
org.mockito:mockito-core:2.22.0
org.mockito:mockito-inline:2.22.0
```
```
java version "1.8.0_181"
Java(TM) SE Runtime Environment (build 1.8.0_181-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.181-b13, mixed mode)
```

#### Gradle

Heap is set to 64 MB.

```groovy
tasks.withType<Test> {
    maxHeapSize = "64m"
    jvmArgs("-XX:+HeapDumpOnOutOfMemoryError")

    failFast = true
}
```
```
$ ./gradlew :module:cleanTestDebugUnitTest :module:testDebugUnitTest --tests "com.github.sample.NestedSpiesMemoryLeakSpec"
```

#### Code

```kotlin
package com.github.sample

import com.jakewharton.rxrelay2.BehaviorRelay
import io.reactivex.functions.Consumer
import org.jetbrains.spek.api.Spek
import org.jetbrains.spek.api.dsl.it
import org.junit.platform.runner.JUnitPlatform
import org.junit.runner.RunWith
import org.mockito.Mockito

@RunWith(JUnitPlatform::class)
class NestedSpiesMemoryLeakSpec : Spek({

    repeat(10_000) { iteration ->

        it("iteration [$iteration]") {
            Mockito.spy(Service())
        }

    }

}) {

    class Service {
        // Remove Mockito.spy and OOM disappears.
        val value = Mockito.spy(Consumer<Int> {
            // This closure keeps a reference to Service.
            streams.size
        })

        // See at as a mass to fill the RAM.
        val streams = (0..1_000).map { BehaviorRelay.create<Int>() }
    }

}
```
```
> Task :module:testDebugUnitTest
java.lang.OutOfMemoryError: GC overhead limit exceeded
Dumping heap to java_pid23350.hprof ...
Heap dump file created [99857779 bytes in 0.356 secs]

com.github.sample.NestedSpiesMemoryLeakSpec > it iteration [187] STANDARD_ERROR
    java.lang.OutOfMemoryError: GC overhead limit exceeded
    	at com.jakewharton.rxrelay2.BehaviorRelay.<init>(BehaviorRelay.java:99)
    	at com.jakewharton.rxrelay2.BehaviorRelay.create(BehaviorRelay.java:77)
    	at com.github.sample.NestedSpiesMemoryLeakSpec$Service.<init>(NestedSpiesMemoryLeakSpec.kt:32)
    	at com.github.sample.NestedSpiesMemoryLeakSpec$1$1$1.invoke(NestedSpiesMemoryLeakSpec.kt:17)
    	at com.github.sample.NestedSpiesMemoryLeakSpec$1$1$1.invoke(NestedSpiesMemoryLeakSpec.kt:12)
    	at org.jetbrains.spek.engine.Scope$Test.execute(Scope.kt:102)
    	at org.jetbrains.spek.engine.Scope$Test.execute(Scope.kt:80)
    	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$5(NodeTestTask.java:105)
    	at org.junit.platform.engine.support.hierarchical.NodeTestTask$$Lambda$82/547193480.execute(Unknown Source)
    	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:72)
    	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:95)
    	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:71)

com.github.sample.NestedSpiesMemoryLeakSpec > it iteration [187] FAILED
    java.lang.OutOfMemoryError

> Task :module:testDebugUnitTest FAILED
```

#### Eclipse Memory Analyzer

screen shot 2018-11-14 at 19 09 53
screen shot 2018-11-14 at 19 08 59
screen shot 2018-11-14 at 19 09 09

---

Seems like this happens:

* `Service` is a spy.
* `Service` contains a `Consumer`, it is a `spy` as well.
* `Consumer` is a closure and has an implicit reference to `Service`.
* Mockito keeps both spies and cannot remove them from memory since there is a cross-reference (I guess).

### LINKED ISSUE #1533: Cross-references and a single spy cause memory leak
Not entirely sure, but I think that Mockito does not handle cross-references well. I have a semi-complex sample that proves this, but at the same time I’m not sure it is a Mockito failure and especially a fixable one. Suggestions how to avoid this behavior in general will be very helpful!

BTW I can provide a `.hprof` file if you are interested.

#### Versions

```
org.mockito:mockito-core:2.22.0
org.mockito:mockito-inline:2.22.0
```
```
java version "1.8.0_181"
Java(TM) SE Runtime Environment (build 1.8.0_181-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.181-b13, mixed mode)
```

#### Gradle

Heap is set to 64 MB.

```groovy
tasks.withType<Test> {
    maxHeapSize = "64m"
    jvmArgs("-XX:+HeapDumpOnOutOfMemoryError")

    failFast = true
}
```
```
$ ./gradlew :module:cleanTestDebugUnitTest :module:testDebugUnitTest --tests "com.github.sample.SubscriptionMemoryLeakSpec"
```

#### Code

```kotlin
package com.github.sample

import com.jakewharton.rxrelay2.BehaviorRelay
import com.jakewharton.rxrelay2.PublishRelay
import io.reactivex.disposables.CompositeDisposable
import org.jetbrains.spek.api.Spek
import org.jetbrains.spek.api.dsl.it
import org.junit.platform.runner.JUnitPlatform
import org.junit.runner.RunWith
import org.mockito.Mockito

@RunWith(JUnitPlatform::class)
class SubscriptionMemoryLeakSpec : Spek({

    repeat(1_000) { iteration ->

        it("iteration $iteration") {
            // Remove Mockito.spy and OOM disappears (even without component.unbind).
            val service = Mockito.spy(Service())
            val memoryConsumingService = MemoryConsumingService()

            val component = Component(service, memoryConsumingService)

            component.bind()

            // Uncomment the following line and OOM disappears.
            // component.unbind()
        }

    }

}) {

    class Service {
        val stream = PublishRelay.create<Unit>().toSerialized()
    }

    class MemoryConsumingService {
        // See at as a mass to fill the RAM.
        val streams = (0..1_000).map { BehaviorRelay.create<Int>() }
    }

    class Component(
            private val service: Service,
            private val memoryConsumingService: MemoryConsumingService
    ) {

        private val disposable = CompositeDisposable()

        fun bind() {
            disposable += service.stream.subscribe {
                // This closure keeps a reference to Component.
                memoryConsumingService.streams.size
            }
        }

        fun unbind() = disposable.clear()
    }

}
```
```
> Task :module:testDebugUnitTest
java.lang.OutOfMemoryError: GC overhead limit exceeded
Dumping heap to java_pid31753.hprof ...
Heap dump file created [96660586 bytes in 0.332 secs]

com.github.sample.SubscriptionMemoryLeakSpec > it iteration 260 STANDARD_ERROR
    java.lang.OutOfMemoryError: GC overhead limit exceeded
    	at java.util.concurrent.locks.ReentrantReadWriteLock$Sync.<init>(ReentrantReadWriteLock.java:338)
    	at java.util.concurrent.locks.ReentrantReadWriteLock$NonfairSync.<init>(ReentrantReadWriteLock.java:669)
    	at java.util.concurrent.locks.ReentrantReadWriteLock.<init>(ReentrantReadWriteLock.java:240)
    	at java.util.concurrent.locks.ReentrantReadWriteLock.<init>(ReentrantReadWriteLock.java:230)
    	at com.jakewharton.rxrelay2.BehaviorRelay.<init>(BehaviorRelay.java:99)
    	at com.jakewharton.rxrelay2.BehaviorRelay.create(BehaviorRelay.java:77)
    	at com.github.sample.SubscriptionMemoryLeakSpec$MemoryConsumingService.<init>(SubscriptionMemoryLeakSpec.kt:41)
*** java.lang.instrument ASSERTION FAILED ***: "!errorOutstanding" with message can't create byte arrau at JPLISAgent.c line: 813
    	at com.github.sample.SubscriptionMemoryLeakSpec$1$1$1.invoke(SubscriptionMemoryLeakSpec.kt:21)
    	at com.github.sample.SubscriptionMemoryLeakSpec$1$1$1.invoke(SubscriptionMemoryLeakSpec.kt:14)
    	at org.jetbrains.spek.engine.Scope$Test.execute(Scope.kt:102)
    	at org.jetbrains.spek.engine.Scope$Test.execute(Scope.kt:80)
    	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$5(NodeTestTask.java:105)
    	at org.junit.platform.engine.support.hierarchical.NodeTestTask$$Lambda$82/1204954813.execute(Unknown Source)
    	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:72)
    	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:95)
    	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:71)
    	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService$$Lambda$89/1263089654.accept(Unknown Source)
    	at java.util.ArrayList.forEach(ArrayList.java:1257)
    	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.invokeAll(SameThreadHierarchicalTestExecutorService.java:38)
    	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$5(NodeTestTask.java:110)
    	at org.junit.platform.engine.support.hierarchical.NodeTestTask$$Lambda$82/1204954813.execute(Unknown Source)
    	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:72)
    	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:95)
    	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:71)
    	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService$$Lambda$89/1263089654.accept(Unknown Source)
    	at java.util.ArrayList.forEach(ArrayList.java:1257)
    	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.invokeAll(SameThreadHierarchicalTestExecutorService.java:38)
    	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$5(NodeTestTask.java:110)
    	at org.junit.platform.engine.support.hierarchical.NodeTestTask$$Lambda$82/1204954813.execute(Unknown Source)
    	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:72)
    	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:95)
    	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:71)

com.github.sample.SubscriptionMemoryLeakSpec > it iteration 260 FAILED
    java.lang.OutOfMemoryError


261 tests completed, 1 failed
> Task :module:testDebugUnitTest FAILED
82 actionable tasks: 5 executed, 77 up-to-date
```

#### Eclipse Memory Analyzer

screen shot 2018-11-15 at 11 57 52
screen shot 2018-11-15 at 11 58 42
screen shot 2018-11-15 at 11 59 17
screen shot 2018-11-15 at 11 59 58
screen shot 2018-11-15 at 12 01 25

---

Seems like this happens:

* `Service` is a spy and is being passed to `Component` along with `MemoryConsumingService`.
* Since `Service` is a spy it is being held by Mockito.
* `Component` subscribes to `Service.stream`. The `subscribe` closure captures `Component`, `Service` and `MemoryConsumingService`. Due to RxJava specifics `Service.stream` holds all of these thanks to the closure.
* Since `Service` is being held by Mockito (since it is a spy) and by itself (due to the `subscribe` closure) Mockito does not release it.

However:

* Releasing the subscription via `component.unbind()` removes a reference, so Mockito releases it and there is no OOM.
* Avoiding making `Service` a spy eliminates OOM as well, i. e. cross-references are not an issue for the JVM itself.

### LINKED ISSUE #1614: Memory leak in mockito-inline calling method on mock with at least a mock as parameter
# Summary
We found a memory leak with mockito-inline. A short example is shown as below. A more detailed example can be found at the end.

```java
class BigClass {
  void accept(SmallClass small) {}
}

class SmallClass {
  void accept(BigClass big) {}
}

void leak() {
  BigClass a = Mockito.mock(BigClass.class);
  SmallClass b = Mockito.mock(SmallClass.class);

  // Remove any one of the 2 lines below will stop leaking
  a.accept(b);
  b.accept(a);
}

```

# Analysis
Looking at the [memory dump][MemDump]. Mock of `BigClass` and `SmallClass` are held as a weak ref in a map of type `WeakConcurrentMap$WithInlinedExpunction` used to map the mock and its invocation handler. When the mock can be reclaimed by GC the map will remove the record from it.

In the case with memory leak, the mock instance of `BigClass` was held as a strong reference by `rawArguments` and `arguments` in [`InvocationMatcher`][InvocationMatcher],  as `invocation` in [`InterceptedInvocation`][InterceptedInvocation], as `invocationForStubbing` in [`InvocationContainerImpl`][InvocationContainerImpl], as `invocationContainer` in [`MockHandlerImpl`][MockHandlerImpl]. That eventually leads to a value in the `WeakConcurrentMap` for `SmallClass`. Similar thing happens to the mock instance of `SmallClass`. That creates a ring of reference and no mock can be reclaimed by GC because they are all referenced transitively by a value in the map.

Similar things can also happen for stubbed methods, saved in `stubbed` in `InvocationContainerImpl`.

Subclass mock makers don't suffer from it because there is no map from mock to handler -- it's just a strong reference. GC can handle non-accessible rings well, but GC doesn't know the mock map purging semantic in inline mock makers.

# Potential Solution
Unlike #1533 where converting `spiedInstance` to a weak reference may be an acceptable solution, we can't convert arguments in stubbing method calls into weak references because there are stubbing calls with an object (or a mock) that doesn't have strong ref anywhere else than arguments in Mockito.

Therefore I failed to see a solution that's transparent to callers. The possible solution below is the one that I think has the least change.

The possible solution is we can somehow reset the mock when their lives end, which clears the stubbing records. Maybe we can tie their lives to a `MockitoSession`. We can track all mocks created after a session is created, and reset them when the session is finished. Now we only have events for mock creation in the same thread, we may need to expand that to include other threads.

We may be able to set `spiedInstance` to `null` when their lives end, which should also be able to fix #1532 and #1533. That would be a solution with least risk.

Of course any solution that's transparent to callers is still more desirable.

# A Detailed Example
See GitHub project [MockitoMethodCallMemLeak][MockitoMethodCallMemLeak]. One can open it in IntelliJ and the run configuration is already configured (with mem dump at OOM and 4M java heap size).

[MemDump]: https://github.com/ttanxu/MockitoMethodCallMemLeak/blob/master/java_pid12535.hprof
[InvocationMatcher]: https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/internal/invocation/InvocationMatcher.java
[InterceptedInvocation]: https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/internal/invocation/InterceptedInvocation.java
[InvocationContainerImpl]: https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/internal/stubbing/InvocationContainerImpl.java
[MockHandlerImpl]: https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/internal/handler/MockHandlerImpl.java
[MockitoMethodCallMemLeak]: https://github.com/ttanxu/MockitoMethodCallMemLeak

ttanxu:
In certain specific, rare scenarios (issue #1619) inline mocking causes memory leaks. There is no clean way to mitigate this problem completely. Hence, we introduced a new API to explicitly clear mock state (only make sense in inline mocking!):

```java
public class ExampleTest {

     @After
     public void clearMocks() {
         Mockito.framework().clearInlineMocks();
     }

     @Test
     public void someTest() {
         //...
     }
 }
```

Due to the introduction of map from weak reference from mock instance to
its invocation handler, Mockito became vulnerable to memory leaks as
there are multiple situations where Mockito could unintentionally hold
strong references to mock instances in the map record. The strong
references could be through spiedInstance for spies, and arguments used
to facilitate method stubbing.

Mockito could never know if the arguments passed in for method stubbing
are also strongly referenced somewhere else or not, so Mockito needs to
save a strong reference to these arguments to avoid premature GC.
Therefore to solve cyclic strong references through arguments Mockito
needs to explicitly know when mocks are out of their life, and clean
up all internal strong references associated with them.

Fixes #1532 and #1533.


TimvdLippe:
I will review this tomorrow. At a first glance, we are going to need to make some changes. Most notably the way we handle interfaces. (A new method on `MockMaker` is a breaking change). Will give you a detailed review tomorrow.

TimvdLippe:
(That said, the PR is much appreciated. Sorry if my initial reaction seemed not positive, we really do appreciate community PRs for these kind of issues!)

ttanxu:
Oh... Thanks for quick response. I didn't expect any feedback today.

I am not familiar with how to contribute to Mockito (i.e. don't know what can be done, and what can't). It's expected to have some back and forth for a change at this scale. Just please be as detailed as possible, so that I know how to make proper changes.

I just hope the overall approach won't be vetoed, as I basically don't have any other ideas on how to solve it.

src/main/java/org/mockito/plugins/MockMaker.java
```diff
@@ -101,6 +101,16 @@ void resetMock(
...
 
+    /**
+     * Cleans up internal state for specified {@code mock}. You may assume there won't be any interaction to this {@code mock}
+     * after this is called.
+     *
+     * @param mock the mock instance whose internal state is to be cleaned.
+     * @since 2.24.4
+     */
+    @Incubating
+    void cleanUpMock(Object mock);
```
TimvdLippe: Sadly this is a breaking change. Moreover, not all mockmakers actually need to clean up. I think it is better to have a separate interface for this and do an `instanceof` check instead.

ttanxu:
Thanks to @mockitoguy .

Just some context here. I tried to avoid new API at first, but `DefaultMockitoSession` seems (at least from tests) to be used within a thread, and changing its default behavior to listen on events from all threads will break tests for multi-threading. Therefore I came up with the new API solution.

I didn't know how much impact there would be to add a new method in `MockMaker` interface. That requires everyone providing an implementation when they uprev their Mockito to a version that has the new method, but I have no clue how much work it could be. If it's too noisy maybe we can have a default impl (if only Java 8+ is supported) or a brand new interface called `InlineMockMaker` to add that method. I personally want to expose a chance to mock makers to clean up as we don't know how they handle the mapping from mock to its handler, which may have other places that could cause memory leaks other than those 2 places we found. We need better documentation on that, but I have to admit I'm not good at writing API docs so I'll defer to any of your suggestions in terms of documentation.

I hope the review won't later become a whole refactor of current implementation.

src/main/java/org/mockito/plugins/MockMaker.java
ttanxu: Done.

Not a significant change at all.

mockitoguy:
I reviewed the implementation and most of it is really good. It is also a very high quality PR, thank you! Curious, are you facing this problem at G.?

I have a couple of concerns / design choices to discuss:
 
a) Can we make the fix more seamless? I'm not sure if this is possible but let's discuss. Goals: users don't have to remember to use trackAndCleanUpMocks() _when_ inline mocker is in use. For example (brainstorming):
 	- we introduce a new interface that InlineByteBuddyMockMaker implements, say: RequiresCleaningMocks
 	- we expose a new method like framework.getPlugins().getPlugin(MockMacker.class), this way we can find out in runtime what MockMaker is in use.
 	- in Mockito Session, we check if the mock maker is RequiresCleaningMocks
 	- in session, we call RequiresCleaningMocks.cleanUpMock()
 
b) MockMaker.cleanUpMock() is an incompatible change so we cannot ship this PR in its current form. MockMaker is typically provided at runtime from a different Jar (example: dexmaker, powermockito). I have a couple ideas but first let's explore a)
 
Great tests, deep change in the internals of Mockito, thank you very much for the contribution!

ttanxu:
Thanks for your review @mockitoguy .

Dexmaker is incubating a feature in inline-dexmaker to create spies without creating new instances, so that we can stub final members, and we (Android framework) found these mem leaks when using that. I then found #1533 and reproduced #1614 that I confirmed it also happens with ByteBuddy mock makers so I started the fix with help from @TimvdLippe .

First of all I really didn't see a solution w/o tying mock lives with something, and I think `MockitoSession` is the most natural choice for the "something". So technically it's already not transparent to properly use inline mock makers because sessions are necessary to make it work while it's optional to use subclass mock makers.

I also first tried to avoid `trackAndCleanUpMocks()`. The difficulty actually comes from the fact MockitoListeners are ThreadLocal before this change, and it's a requirement to track mocks generated from other threads because [`FailOnTimeout`][FailOnTimeOut] basically runs tests in different threads than the thread used to run Before and After. Before this change user can have multiple `DefaultMockitoSession`, as long as they are in different threads (`ThreadsRunAllTestsHalfManualTest` also verifies that behavior) and not interfering with each other. However if we convert `DefaultMockitoSession` to have "thread global" influence `ThreadRunAllTestsHalfManualTest` will fail because cross-threads mock cleanup will invalidate mocks created in tests running in a different thread.

I think I have a way (a bit hacky though) to make it work, but I don't think its complication outweighs the new API (given that it's already not transparent to users). It doesn't encourage clean tests in multi-threading scenario either imho. However I can still put my thought here and let you decide if we should take it:

- Don't reject registration of global listeners if it's from a different thread for the same type
- Only clean up mocks when the last `MockTracker` is unregistered

Then it basically treats the life of any `DefaultMockitoSession` as mock lives. It doesn't allow user to precisely control mock lives but with some care users should still be able to avoid OOM.

[FailOnTimeOut]: https://github.com/junit-team/junit4/blob/master/src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java#L120

ttanxu:
Oh... I already added `InlineMockMaker` interface so not all mock makers need to implement cleanUpMock(). I'm not sure if the plugin way is still preferred over current change because for most dexmakers they're still just implementing the basic `MockMaker` interface and nothing should break. While they can choose to implement `InlineMockMaker` if clean up is necessary.

ttanxu:
Just to be clear that I'm waiting for @mockitoguy 's response on if we would like to go for that approach in exchange of removing trackAndCleanUpMocks() API, just in case we're waiting for each other and it's going nowhere.

TimvdLippe:
As a sidenote: the refactoring with the new InlineMockMaker interface is :+1: for me :smile:

mockitoguy:
Hey @ttanxu, so you you're saying that the approach I suggested in a) wouldn't work?

mockitoguy:
> Oh... I already added `InlineMockMaker` interface so not all mock makers need to implement cleanUpMock()

Great! I'll review again this weekend.

mockitoguy:
> As a sidenote: the refactoring with the new InlineMockMaker interface is 👍 for me 😄

@TimvdLippe, I'm with you. It's a good idea to have a separate interface and stay compatible.

ttanxu:
Hey @mockitoguy , what you suggested could solve the same issue that is solved by `InlineMockMaker` in current PR (the MockMaker contract breakage).

`trackAndCleanUpMocks()` was introduced for other reason, but could use different tech to remove its necessity. I explained the detail in previous comments (mostly because of the multi-thread implementation of [`FailOnTimeout`][1] in JUnit).

[1]: https://github.com/junit-team/junit4/blob/master/src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java#L120

mockitoguy:
@ttanxu, thank you for explanation. I'm looking into it.

mockitoguy:
My comment is short but I put a lot of consideration into this. Given all arguments, I agree that a new API is the best way. Before we ship it, I want to bounce one more idea. What if we tie MockMaker with MockitoSession? When the session completes it tears down its MockMaker/cleans up all mocks. We could use ThreadLocal. Thoughts?

ttanxu:
Correct me if I'm wrong.

It seems you're suggesting one `MockMaker` per thread. The critical part of this approach is the mock map in the inline mock maker needs to be a `ThreadLocal`, which means mocks won't look have handlers tied to them in threads other than the thread where they were created, and thus won't appear to be a mock and method invocations on it might trigger exceptions. That's not likely a compatible change to existing use cases because if [`FailOnTimeout`][1] is used Befores and Afters are executed in a different thread than the test method.

If we still only have one `MockMaker` across all threads then the structure is pretty much the same as current change except that instead of calling `InlineMockMaker#clearUp()` we just tear down the entire `MockMaker`. I'm not sure if this is better than calling `clearUp()`. It does sound like a viable solution w/o introduction to `InlineMockMaker` as we can just swap the instance of `MockMaker` in `PluginRegistry`. The con is that's more of a systematic refactor than `InlineMockMaker#clearUp()` approach and it reduces the cleanness of current code (we can't use final for `mockMaker` in `PluginRegistry` anymore). BTW this doesn't help avoid a new API `trackAndClearUpMocks()` in `MockitoSession`. I'll leave this decision to you.

[1]: https://github.com/junit-team/junit4/blob/master/src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java#L120

mockitoguy:
Great points, thank you for taking the time to explain. Why do we need to keep track of specific mocks? Can we just clear all?

ttanxu:
Oh... You say we don't have to save created mocks in a list in `MockTracker`? I guess we can just clear all mocks. Especially considering there can't be multiple mock-tracking sessions open at the same time, mocks created before the session is opened may be leaked.

Removing `MockTracker` can simply this change a lot, especially we don't need global listeners. I'll update the change.

ttanxu:
Actually, I have a different thought that I might want to keep `MockTracker` (sorry for back and forth)... The thought of tracking specific mocks stemmed from the thought that usually mocks have different lives. E.g. mocks held by static variables, created in BeforeClass, in Before and in individual tests may need to have different lives. It would be nice if we can keep mocks created in BeforeClass for later reuse.

Current impl is half baked because there can only be one tracking session, but I want to keep `cleanUp(Object)` signature (or we can have `cleanUp(Object...)` instead so that we can clean up everything if Object array is `null`) for `InlineMockMaker` just in case we want to support the case above.

With that said I'll change `cleanUp(Object)` to `cleanUp(Object...)` and use `null` array for clean up everything.

ttanxu:
Oh... Besides, if we just want to clear everything, it doesn't seem natural to me to add the API to MockitoSession, maybe we can add an API to Mockito?

mockitoguy:
Thank you for considering! I want us to strive for simplicity. I'll think about the new API for Mockito (or MockitoFramework).

If we have a static method like "Mockito.framework().clearAllMocks()" would it work for you? 

> E.g. mocks held by static variables, created in BeforeClass, in Before and in individual tests may need to have different lives

Do we know legit use cases for "static" mocks? It seems an anti-pattern.

ttanxu:
Finally cleaned up all remains from previous change. It's a lot simpler now. Please take a look.

I can think of a case for static mocks where constructing a mock may be expensive and someone wants to reuse it across tests. That's not an argument why static mocks have to be used though, they can definitely reconstruct those mocks.

src/main/java/org/mockito/plugins/InlineMockMaker.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Extension to {@link MockMaker} for mock makers that changes inline method implementations.
+ * @since 2.24.8
+ */
+@Incubating
+public interface InlineMockMaker extends MockMaker {
+    /**
+     * Cleans up internal state for specified {@code mock}. You may assume there won't be any interaction to this {@code mock}
+     * after this is called.
```
TimvdLippe: Lets add an additional sentence that this method can be called with null, indicating all mocks should be cleared

src/main/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMaker.java
```diff
@@ -271,6 +272,15 @@ public void resetMock(Object mock, MockHandler newHandler, MockCreationSettings
         }
     }
 
+    @Override
+    public void cleanUpMock(Object mock) {
```
TimvdLippe: We are never calling this method with an actual mock. Is there a situation this should be the case? Could you maybe provide an example for it? (If it is not necessary, we might as well make this method `clearAllMocks` as well)

ttanxu: I said that in comments before I updated the PR.

If we will go with clearAllMocks() we need an additional method if there is strong feature request for keeping some mocks alive later.

That might be over-engineering, but I'm not sure how open you are to introduce new method to incubating class or change existing incubating method signature in this case. It would require mock maker changes later.

Maybe we can change the public API to take in mocks as well? I might just do that quickly.

src/main/java/org/mockito/plugins/InlineMockMaker.java
ttanxu: Done.

src/main/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMaker.java
ttanxu: Updated the PR, but I'm open to discuss this forward. If you two believe cutting off the mock specific clear up feature is better then I'm willing to make another change.

I'm just afraid that there are cases where callers might want to keep some mocks, but still need to release some other of them.

TimvdLippe: I think it is fair to keep both. Since we are introducing an interface, we are most flexible if we keep them both in. That said, I think it is better to also introduce `clearAllMocks` on `InlineByteBuddyMockMaker` as well as rename `cleanUpMock` to `clearMock` to keep it consistent.

ttanxu: Updated them for consistency. Thanks.

mockitoguy:
Reviewing...

mockitoguy:
We are ready. Thank you @ttanxu and @TimvdLippe for patience! This is the best pragmatic step to resolve the problem. In the future, perhaps we can avoid having the weak map of references in a first place.

@ttanxu, I made some tweaks to tests, documentation, I've also renamed the method to "clearMocks() -> clearInlineMocks()", for consistency. Take a look and see if you're OK with this. Let us know and I'll merge this. Nice work with this change!

@TimvdLippe, I did some cosmetic changes, please review if you want. Thanks!

TimvdLippe:
I will review this on Monday in the office. Thanks for the polish!

src/main/java/org/mockito/MockitoFramework.java
```diff
@@ -92,4 +92,55 @@
...
     InvocationFactory getInvocationFactory();
+
+    /**
+     * Clears up internal state of all inline mocks.
+     * This method is only meaningful if inline mock maker is in use.
+     * Otherwise this method is a no-op and need not be used.
+     * <p>
+     * This method is useful to tackle subtle memory leaks that are possible due to the nature of inline mocking
+     * (issue <a href="https://github.com/mockito/mockito/pull/1619">#1619</a>).
+     * If you are facing those problems, call this method at the end of the test (or in "@After" method).
```
TimvdLippe: Shouldn't we update our JUnit runner and extension to handle this automatically? I think this is an improvement in any situation, so we can just go ahead and do it eagerly.

src/main/java/org/mockito/Mockito.java
```diff
@@ -1533,6 +1534,15 @@
...
  * For more information refer to {@link Mockito#lenient()}.
  * Let us know how do you find the new feature by opening a GitHub issue to discuss!
+ *
+ * <h3 id="47">47. <a class="meaningful_link" href="#clear_inline_mocks" name="clear_inline_mocks">New API for clearing mock state in inline mocking (Since 2.25.0)</a></h3>
+ *
+ * In certain specific, rare scenarios (issue <a href="https://github.com/mockito/mockito/pull/1619">#1619</a>)
+ * inline mocking causes memory leaks.
+ * There is no clean way to mitigate this problem completely.
+ * Hence, we introduced a new API to explicitly clear mock state (only make sense in inline mocking!).
+ * See example in {@link MockitoFramework#clearInlineMocks()}.
```
TimvdLippe:
```suggestion
 * See example usage in {@link MockitoFramework#clearInlineMocks()}.
```

src/main/java/org/mockito/Mockito.java
```diff
@@ -1533,6 +1534,15 @@
  *
  * For more information refer to {@link Mockito#lenient()}.
  * Let us know how do you find the new feature by opening a GitHub issue to discuss!
+ *
+ * <h3 id="47">47. <a class="meaningful_link" href="#clear_inline_mocks" name="clear_inline_mocks">New API for clearing mock state in inline mocking (Since 2.25.0)</a></h3>
+ *
+ * In certain specific, rare scenarios (issue <a href="https://github.com/mockito/mockito/pull/1619">#1619</a>)
+ * inline mocking causes memory leaks.
+ * There is no clean way to mitigate this problem completely.
```
TimvdLippe:
```suggestion
 * It was not possible to resolve this issue without introducing an additional API.
 * Any inline MockMaker should therefore implement the additional interface {@link org.mockito.plugins.InlineMockMaker} to allow Mockito to properly clean up mocks.
```

src/main/java/org/mockito/MockitoFramework.java
mockitoguy: This would be a breaking change in thread-based parallel tests that are supported by Mockito (with limitations, as we stated in the FAQ).

If you turn this on by default, some Mockito tests will start failing. @ttanxu, discussed this somewhere in this PR (but the thread became long :).

We can always make it later if we understand the use patterns better. I suggest we expose the this API first and observe / learn how it goes.

TimvdLippe: Ah okay, I missed that discussion. Thanks for the clarification.

src/test/java/org/mockito/internal/framework/DefaultMockitoFrameworkTest.java
```diff
@@ -112,5 +123,43 @@ public void run() {
                     "For more information, see the javadoc for RedundantListenerException class.");
     }
 
+    @Test
+    public void clears_all_mocks() {
+        //clearing mocks only works with inline mocking
+        assumeTrue(Plugins.getMockMaker() instanceof InlineMockMaker);
```
ttanxu: FWIW, one of the reason why I didn't use assume at first is that it can test user can safely call clearInlineMocks() and clearMock() even when mock maker is not a inline mock maker.

That's not very important I suppose, but still would be good to cover that in tests imo.

mockitoguy:
> test user can safely call clearInlineMocks() and clearMock() even when mock maker is not a inline mock maker

@ttanxu, great point!

mockitoguy:
I'm making the final tweaks based on the code review and I'm merging.

src/test/java/org/mockito/internal/framework/DefaultMockitoFrameworkTest.java
mockitoguy: +1 I'm fixing it now.

src/main/java/org/mockito/Mockito.java
mockitoguy: Sure!

src/main/java/org/mockito/Mockito.java
mockitoguy: I'm skipping this suggestion if that's OK. I prefer skipping those details in the high level javadoc. Also, I prefer the original text grammatically.

Feel free to redo the javadoc if you want, in a separate PR - I'm happy to review/merge it.

ming13:
This is awesome, thank you guys! I’ll try this API on our codebase and samples I’ve provided in original issues. Also — kudos for including tests reproducing the behavior I’ve described 😉

s2131:
Hi, Just to let you know I am using the new API already with v2.28.2 and for me it was breaking the Strict test with following exception:
`
org.mockito.exceptions.misusing.NotAMockException: Argument passed to Mockito.mockingDetails() should be a mock, but is an instance of class Cache$MockitoMock$149288076!

    at org.mockito.internal.runners.DefaultInternalRunner$1$2.testFinished(DefaultInternalRunner.java:63)
 ...
`
I had to move it to `@AfterClass` to mitigate the problem

ttanxu:
@s2131 You are not allowed to interact with mocks anymore after calling that API because all internal states of mocks are cleaned up after the API call. That includes all delayed actions or interactions from other threads.

Unfortunately it's never as easy as just calling that API.

PR #1624, opened 2019-02-19
Handle terminal type var with bounds

### LINKED ISSUE #1621: RETURNS_DEEP_STUBS causes "Raw extraction not supported for : 'null'" in some cases
When trying to mock the [Hadoop Mapper.Context](https://github.com/apache/hadoop/blob/release-2.6.0/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/Mapper.java#L105) class with Mockito 2.24.0, exception `org.mockito.exceptions.base.MockitoException: Raw extraction not supported for : 'null'` is thrown when calling one of the methods that returns a generic type.  

Actual mocking is taking place inside the [MRUnit library](https://mrunit.apache.org/), as seen in [MockMapContextWrapper](https://github.com/apache/attic-mrunit/blob/release-1.1.0/src/main/java/org/apache/hadoop/mrunit/internal/mapreduce/MockMapContextWrapper.java#L75): 

    final Mapper<KEYIN, VALUEIN, KEYOUT, VALUEOUT>.Context context = mock(org.apache.hadoop.mapreduce.Mapper.Context.class, Mockito.RETURNS_DEEP_STUBS);

Context is an abstract inner class of Mapper, and implements the [MapContext](https://github.com/apache/hadoop/blob/release-2.6.0/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/MapContext.java) interface, which extends the [TaskInputOutputContext](https://github.com/apache/hadoop/blob/release-2.6.0/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/TaskInputOutputContext.java) interface, which extends several other non-generic interfaces. 

Exception is thrown later when calling the [`context.getCurrentKey()`](https://github.com/apache/hadoop/blob/release-2.6.0/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/TaskInputOutputContext.java#L51) method on the mock, which returns generic type `KEYIN`.

This works fine with Mockito 1.

MRUnit project is abandoned, so it is not possible to change the mock.  It by default uses old Mockito 1.9.5, and works with 1.10.19.  I am attempting to exclude its dependencies on old Mockito and Powermock, and instead provide latest Mockito 2.24.0 and Powermock 2.0.0 so that my project using MRUnit and Mockito will not be permanently stuck on Mockito 1. 

The only time these problem methods appear to be [called within MRUnit is to set up stubbing](https://github.com/apache/attic-mrunit/blob/release-1.1.0/src/main/java/org/apache/hadoop/mrunit/internal/mapreduce/MockMapContextWrapper.java#L96), so it doesn't really matter what the return value is in this case.  Should this even work with Mockito, or was this just a bug in Mockito 1 that MRUnit exploited? 

----

Here is a minimal example that reproduces the issue:

```Java
package example;

import org.junit.Before;
import org.junit.Test;

import static org.mockito.Mockito.RETURNS_DEEP_STUBS;

public class Example {

    public interface FirstInterface<T> {
        boolean normal();
        T generic();
    }

    public interface SecondInterface<T> extends FirstInterface<T> { }

    public class TestClass<T> {
        public abstract class AbstractInner implements SecondInterface<T> { }
    }

    private TestClass.AbstractInner testMock;

    @Before
    public void setup() {
        testMock = mock(TestClass.AbstractInner.class, RETURNS_DEEP_STUBS);
    }

    @Test
    public void works() {
        testMock.normal();
    }

    @Test
    public void fails() {
        testMock.generic();
    }
}
```

https://github.com/zman0900/mockito-example has this code with a gradle build.

Both tests pass with Mockito 1.10.19, but one fails with 2.24.0.  Changing TestClass.AbstractInner to implement FirstInterface instead of SecondInterface causes the same failure with both Mockito 1 and 2.

bric3:
This PR aim to fix the issue described in #1621, the issue being that terminal type variable arguments were not looked up for their bounds.

Fixes #1621

### DIFF
--- a/src/test/java/org/mockitousage/serialization/DeepStubsSerializableTest.java
+++ b/src/test/java/org/mockitousage/serialization/DeepStubsSerializableTest.java
@@ -11,7 +11,11 @@ import java.util.Iterator;
 import java.util.List;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.Mockito.*;
+import static org.assertj.core.api.Assertions.fail;
+import static org.mockito.Mockito.RETURNS_DEEP_STUBS;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.withSettings;
 import static org.mockitoutil.SimpleSerializationUtil.serializeAndBack;
 
 public class DeepStubsSerializableTest {
@@ -45,7 +49,7 @@ public class DeepStubsSerializableTest {
         assertThat(deserialized_deep_stub.iterator().next().add("yes")).isEqualTo(true);
     }
 
-    @Test(expected = NullPointerException.class)
+    @Test
     public void should_discard_generics_metadata_when_serialized_then_disabling_deep_stubs_with_generics() throws Exception {
         // given
         ListContainer deep_stubbed = mock(ListContainer.class, withSettings().defaultAnswer(RETURNS_DEEP_STUBS).serializable());
@@ -53,10 +57,14 @@ public class DeepStubsSerializableTest {
 
         ListContainer deserialized_deep_stub = serializeAndBack(deep_stubbed);
 
-        // when stubbing on a deserialized mock
-        when(deserialized_deep_stub.iterator().next().get(42)).thenReturn("no");
-
-        // then revert to the default RETURNS_DEEP_STUBS and the code will raise a ClassCastException
+        try {
+            // when stubbing on a deserialized mock
+            // then revert to the default RETURNS_DEEP_STUBS and the code will raise a ClassCastException
+            when(deserialized_deep_stub.iterator().next().get(42)).thenReturn("no");
+            fail("Expected an exception to be thrown as deep stubs and serialization does not play well together");
+        } catch (ClassCastException e) {
+            assertThat(e).hasMessageContaining("java.util.List");
+        }
     }
 
     static class SampleClass implements Serializable {
--- a/src/test/java/org/mockitousage/serialization/DeepStubsSerializableTest.java
+++ b/src/test/java/org/mockitousage/serialization/DeepStubsSerializableTest.java
@@ -62,8 +62,8 @@ public class DeepStubsSerializableTest {
             // then revert to the default RETURNS_DEEP_STUBS and the code will raise a ClassCastException
             when(deserialized_deep_stub.iterator().next().get(42)).thenReturn("no");
             fail("Expected an exception to be thrown as deep stubs and serialization does not play well together");
-        } catch (ClassCastException e) {
-            assertThat(e).hasMessageContaining("java.util.List");
+        } catch (NullPointerException e) {
+            assertThat(e).hasMessage(null);
         }
     }
 
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -1685,6 +1685,7 @@ public class Mockito extends ArgumentMatchers {
 
     /**
      * Optional <code>Answer</code> to be used with {@link Mockito#mock(Class, Answer)}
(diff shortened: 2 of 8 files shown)


src/test/java/org/mockitousage/serialization/DeepStubsSerializableTest.java
```diff
@@ -45,18 +49,22 @@ public void should_serialize_and_deserialize_parameterized_class_mocked_with_dee
...
 
-        // when stubbing on a deserialized mock
-        when(deserialized_deep_stub.iterator().next().get(42)).thenReturn("no");
-
-        // then revert to the default RETURNS_DEEP_STUBS and the code will raise a ClassCastException
+        try {
+            // when stubbing on a deserialized mock
+            // then revert to the default RETURNS_DEEP_STUBS and the code will raise a ClassCastException
+            when(deserialized_deep_stub.iterator().next().get(42)).thenReturn("no");
+            fail("Expected an exception to be thrown as deep stubs and serialization does not play well together");
```
TimvdLippe: This is a regression. Can we not throw a CCE here? If not, we are re-introducing #357.

bric3: Nope we're not this change actually properly fix the issue

bric3: I'm working on the next commit for a complete fix. Anyway, with the build matrix, the unit tests do fail on JDK > 8.

bric3: @TimvdLippe Plus this test is about documenting the issue about deserialized mock.
Do you have a lot at google ?

TimvdLippe: I don't recall the actual numbers, but if I would have to guess it was probably between 50 and 200.

Note that this test is slightly different than the pattern we saw the regression on. The tests were doing something along the lines of `when(subject.performBla().performOtherThing().getValue()).thenReturn(something)`. The `getValue()` in this case returned `T`. Without the `return null`, it would throw a CCE on the `.getValue()` invocation. With the `return null`, it would safely pass that and properly set up the stub.

This test case goes one step further: `when(subject.performBla().performOtherThing().getValue().otherInvocation()).thenReturn(something)`. The `otherInvocation` has never worked. Previously, it would already throw a CCE on the `getValue()`. Now, it throws a NPE on the `otherInvocation()`. E.g. we didn't regress, we are just throwing an exception in a different way.

So the former I don't want to regress on. I am fine in the latter situation to make changes. (E.g. this test in particular). If we are speaking in terms of this test, we need `when(deserialized_deep_stub.iterator().next()).thenReturn(object)` to succeed.

TimvdLippe: Oh and if your fix does not regress in the first situation, that is more appropriate than my `return null`. So I am all for that!

bric3: I don't know the codebase your dealing with, but the resolver is tad more intelligent now, now there's still some ground that is not covered. For that reason I decided to add back the guard, with an additional check.

So that way it should keep the old fix places in cases the generic metadata support fails to resolve generic types.

TimvdLippe: Awesome, taking a look :smile:

PR #1629, opened 2019-02-25
Issues 1540 take2

### LINKED ISSUE #1540: MocktioExtension being too strict with stubs in `@BeforeEach` method
I've written a test case for a class which persists errors to a file and thus I have a handful of tests to prove that non-errors cases are not persisted to a file however the majority of the tests in the test case are related to triggering various error scenarios. 

As the majority of test cases (~70%) require mock to supply the persistence file I've included a stub for that in the setup method (annotated using `junit5` `@BeforeEach`). I'm running the tests using `MockitoExtension` which fails tests with `org.mockito.exceptions.misusing.UnnecessaryStubbingException` for the 30% of test which demonstrate that the file isn't use when messaging works. 

While it is true that the stub is unused in those test I think it is discouraging readability by duplicating the stubbing to the majority of the tests in the class. From the [UnnecessaryStubbingException Javadoc](https://static.javadoc.io/org.mockito/mockito-core/2.23.4/org/mockito/exceptions/misusing/UnnecessaryStubbingException.html) 

> Mockito JUnit Runner triggers UnnecessaryStubbingException only when none of the test methods use the stubbings. This means that it is ok to put default stubbing in a 'setup' method or in test class constructor. That default stubbing needs to be used at least once by one of the test methods.

 I expected the Extension to offer the same behaviour.

I've contrived a simple example below which fails. 

I think this is related to: https://github.com/mockito/mockito/issues/769 but I think it is more of an issues with the `MockitoExtension` than strictness it self.

Dependencies on java 8 (8.0.171-oracle): 
```
<dependency>
            <artifactId>junit-jupiter-api</artifactId>
            <version>5.3.1</version>
            <scope>test</scope>
        </dependency>
        <dependency>
            <artifactId>junit-jupiter-engine</artifactId>
            <version>5.3.1</version>
            <scope>test</scope>
        </dependency><dependency>
        <groupId>org.mockito</groupId>
        <artifactId>mockito-core</artifactId>
        <version>2.23.4</version>
        <scope>test</scope>
    </dependency>
        <dependency>
            <groupId>org.mockito</groupId>
            <artifactId>mockito-junit-jupiter</artifactId>
            <version>2.23.4</version>
            <scope>test</scope>
        </dependency>
```

```
package nz.samandems.mockito;

import org.junit.jupiter.api.BeforeEach;
import org.mockito.Mock;
import org.mockito.Mockito;

class MockitoLenienceyTestCase {

    @Mock
    private Dependency dependency;

    private SystemUnderTest systemUnderTest;


    @BeforeEach
    void setUp() {
        Mockito.when(dependency.doWork()).thenReturn(true);
        systemUnderTest = new SystemUnderTest(dependency);
    }


    @Test
    void shouldDoWork() {
        //Given

        //When
        systemUnderTest.doThing(true);

        //Then
        Mockito.verify(dependency).doWork();
    }

    @Test
    void shouldDoNoWork() {
        //Given

        //When
        systemUnderTest.doThing(false);

        //Then
        Mockito.verifyZeroInteractions(dependency);
    }


    private class SystemUnderTest {

        private final Dependency dependency;


        private SystemUnderTest(final Dependency dependency) {
            this.dependency = dependency;
        }


        public boolean doThing(boolean works) {
            if (works) {
                return dependency.doWork();
            }
            else {
                return false;
            }
        }
    }

    private interface Dependency {

        boolean doWork();
    }
}
```

SamBarker:
Not for _merge_.

**Pull request is exploratory for review of changes as a possible fix for #1540 .**

These changes allow the sharing of stubs created in methods annotated `@Before` when using the `MockitoExtension` but not used in each test method. I believe this encourages readability (see example outlined in #1540).

However this PR has a significant drawback, it can no longer fail individual `@Test` methods it now has to fail their container. I have not been able to work out a way to do so, perhaps there is a alternative JUnit API which could address this.

As this PR is not for merge I have not completed all items on the check list. They would be addressed by the final PR (if any) to fix #1540.
check list

* [x]  Read the [contributing guide](https://github.com/mockito/mockito/blob/release/2.x/.github/CONTRIBUTING.md)
* [x]  PR should be motivated, i.e. what does it fix, why, and if relevant how
* [ ]  If possible / relevant include an example in the description, that could help all readers
* [x]  Avoid other runtime dependencies
* [ ]  Meaningful commit history ; intention is important please rebase your commit history so that each
* [x]  The pull request follows coding style
* [ ]  Mention `Fixes #<issue number>` in the description _if relevant_
* [ ]  At least one commit should mention `Fixes #<issue number>` _if relevant_


mockitoguy:
> However this PR has a significant drawback, it can no longer fail individual @Test methods it now has to fail their container. I have not been able to work out a way to do so, perhaps there is a alternative JUnit API which could address this.

@SamBarker, why do you consider it a major drawback? Thank you for contribution!

SamBarker:
A concrete example if possibly helpful: See [session_checks_for_strict_stubs](https://github.com/SamBarker/mockito/blob/50bd6fca60f850ad030b4008bdae09674f76d8fa/subprojects/junit-jupiter/src/test/java/org/mockitousage/StrictnessTest.java#L46) 

Without this PR mockito/junit5 used to be able to mark `session_checks_for_strict_stubs` as failed. With this PR applied the test method is marked as `SUCCESSFUL` however the test class/suite is now marked as failed, reporting the `UnnecessaryStubbingException` as the cause. 

I feel like that is a behaviour change and is possibly miss-leading to users.

mockitoguy:
@SamBarker, this is acceptable. The most important is that the test actually fails. It is secondary "how" it fails. We can make the exception message beautiful :-)

mockitoguy:
This is a harder problem:

```Java
/**
 * Expected: "global" stubbing declared in "BeforeEach" must be used in at least one test method.
 * Otherwise Mockito throws a "unnecessary stubbing" exception.
 * 
 * Problem: when the user selects a single test to run ("test3") it unexpectedly fails with "unnecessary stubbing" exception. 
 */
class DemoTest {

    @Mock List mock;

    @BeforeEach void setup() {
        // "global" stubbing
        when(mock.get(0)).thenReturn(true);
    }

    @Test
    void test1() {
        // "global" stubbing is used here:
        System.out.println(mock.get(0));
    }

    @Test
    void test2() {
        // "global" stubbing is used here as well:
        System.out.println(mock.get(0));
    }

    @Test
    void test3() {
        // this test does not use "global" stubbing
        // Problem: fails if one runs this very test method in IDEA 
    }
}
```

For JUnit4 MockitoRunners we use "filtering" trick to work around this problem: https://github.com/mockito/mockito/blob/release/3.x/src/main/java/org/mockito/internal/runners/StrictRunner.java#L58 If filter is requested we don't validate strictness.

@SamBarker, what do you think about this problem?

SamBarker:
I think this speaks to why I don't like not failing at the test level. 

As things stand with this PR we have inconsistent behaviour between running the suite and running an individual test. If we apply the filtering trick we get inconsistent behaviour the other way round. Which potentially makes debugging issues with unused stubbing harder. 

Thinking through the trade offs, I suspect its better to use some marker to disabled Strictness checking when running a single test or subset of a suite. 

I'll try to make some time to do that.

mockitoguy:
> Thinking through the trade offs, I suspect its better to use some marker to disabled Strictness checking when running a single test or subset of a suite.

We can try this. Ideally, we only disable strict checking for the stubs declared in "setup()".

> we have inconsistent behavior between running the suite and running an individual test

This is acceptable given circumstances. To find out if "setup()" stub is used one needs to run all tests in given class. When one runs just one test, we cannot detect it.

SamBarker:
I've not found time to work on this for a year, and I suspect its too narrow a to the problems further elaborated in #1924

PR #1635, opened 2019-02-27
Fix handling of generics in ReturnsMocks

TimvdLippe:
ReturnsMocks was exhibiting the same problems as we previously had with
ReturnsSmartNulls. Extract that common behavior into a separate class
and thus fix the issues with ReturnsMocks.


TimvdLippe:
We found a regression in our test suites. Should be easy to fix, but I will do that tomorrow. Then I will merge. Thanks @raphw for the review 😄

PR #1646, opened 2019-03-04
Add back getArgument(int, Class) to circumvent compilation issues

### LINKED ISSUE #1609: Add back InvocationOnMock.getArgument<T>(int, Class<T>)
In https://github.com/mockito/mockito/issues/365 we replaced the existing `getArgumentAt<T>(int, Class<T>)` API with the current `getArgument(int)` API. However, when integrating this change within Google, we ran into numerous issues.

An example:

```java
invocation.getArgumentAt(0, Runnable.class).run()
```

gets rewritten to

```java
invocation.getArgument(0).run()
```

This does not compile, as the generic parameter is erasured to `Object` and thus does not have the `run` method.

Therefore, I would propose adding back an overload to `getArgument` that allows you to specify the `Class`. In general, users would not need this inside a variable assignment, but do need it in all other cases.

TimvdLippe:
Also add `@NotExtensible` to several of our interfaces to document they
are not intended to be subclassed.

Fixes #1609


TimvdLippe:
@mockitoguy could you take a look? We need this for a couple of our projects.

TimvdLippe:
We need this for internal projects. Since @mockitoguy gave approval for this in https://github.com/mockito/mockito/issues/1609#issuecomment-469279795 and other engineers reviewed this internally, I am merging this PR.

mockitoguy:
Good work, thank you! Sorry I was late to review.

PR #1648, opened 2019-03-04
Prevent NPE in findTypeFromGenericInArguments

TimvdLippe:
There was only a single test failing on this. I think the issue was RETURNS_SMART_NULLS in combination with an ArgumentCaptor, but
couldn't figure that part out. At least this fixes prevented the NPE.


mockitoguy:
Nice change, thank you!

PR #1653, opened 2019-03-05
Use JUnit 4 ComparisonFailure if present

### LINKED ISSUE #1649: Update ArgumentsAreDifferent for JUnit4+
From #1642:
>There is code in `org.mockito.internal.junit.ExceptionFactory` that checks to see if JUnit is on the classpath - if so, then it will throw `org.mockito.exceptions.verification.junit.ArgumentsAreDifferent` rather than `org.mockito.exceptions.verification.ArgumentsAreDifferent`. The JUnit version inherits from `junit.framework.ComparisonFailure` whereas the basic version inherits from `org.mockito.exceptions.base.MockitoAssertionError`.
>
>...it would make more sense to derive from a more-up-to-date version of `ComparisonFailure` if it is available, like `org.junit.ComparisonFailure` - perhaps this could be the subject of a different issue/PR)

This is that issue. Create another ArgumentsAreDifferent exception that derives from `org.junit.ComparisonFailure` rather than `junit.framework.ComparisonFailure`, and then use the most up-to-date-one that is available on the classpath. The up-to-date one has better reporting and diagnostics.

One minor irritant with this might be that if you are using Mockito with JUnit 4+ on the classpath but with the JUnit 3 runner, your Mockito assertions will be reported as errors instead of failures. However, I think that would be a pretty odd combination of tests/runners to use and probably not worth worrying about. The alternative would be to dynamically check which runner is being used, which is doable but a bit more tricky.

kriegfrj:
Fixes #1649.

I do have a minor concern about the way this has been implemented, which I discovered when I ran the full test suite - if there is any code out there running on JUnit 4+ that is relying on catching `org.mockito.exceptions.verification.junit.ArgumentsAreDifferent`, that code will now fail until it has been updated.

Possible alternative implementations:

1. Make `org.mockito.exceptions.verification.junit.ArgumentsAreDifferent` the JUnit 4 version and `org.mockito.exceptions.verification.junit3.ArgumentsAreDifferent` the JUnit 3 version. Disadvantage: By changing the class hierarchy of the existing class name, it may break implementations that rely on this class deriving from `junit.framework.ComparisonFailure` will be affected.
2. Abandon the old name entirely and use two new ones in packages `junit3` & `junit4`. This will force the compiler to fail on these usages - at least code writers will be aware and can update their code (the backward-compatible options above may leave some hidden bugs waiting to manifest at an inopportune time).
3. Drop the whole idea of upgrading to the latest `ComparisonFailure`.

Comments/thoughts welcome. I don't have a feel for how common it is to people to try and catch these exceptions directly and whether or not this will be an issue in practice vs theory only.

### DIFF
--- a/src/main/java/org/mockito/exceptions/verification/ArgumentsAreDifferent.java
+++ b/src/main/java/org/mockito/exceptions/verification/ArgumentsAreDifferent.java
@@ -21,7 +21,7 @@ public class ArgumentsAreDifferent extends MockitoAssertionError {
      * Three-arg constructor for compatibility with ExceptionFactory's three-arg
      * create method. This implementation simply ignores the second and third
      * arguments.
-     * 
+     *
      * @param message
      * @param wanted ignored
      * @param actual ignored
--- a/src/main/java/org/mockito/exceptions/verification/ArgumentsAreDifferent.java
+++ b/src/main/java/org/mockito/exceptions/verification/ArgumentsAreDifferent.java
@@ -17,6 +17,19 @@ public class ArgumentsAreDifferent extends MockitoAssertionError {
         super(message);
     }
 
+    /**
+     * Three-arg constructor for compatibility with ExceptionFactory's three-arg
+     * create method. This implementation simply ignores the second and third
+     * arguments.
+     * 
+     * @param message
+     * @param wanted ignored
+     * @param actual ignored
+     */
+    public ArgumentsAreDifferent(String message, String wanted, String actual) {
+        this(message);
+    }
+
     @Override
     public String toString() {
         return removeFirstLine(super.toString());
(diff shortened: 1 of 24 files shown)


kriegfrj:
Also, I tried to streamline the code a bit in `ExceptionFactory`. It could be streamlined a lot further if I could use Java 8 method handles, but it seems that the current source compatibility for this project is 1.6? Or have I got something wrong there - should I be able to use method handles?

TimvdLippe:
We can not change the package name, that is a breaking change. At most, we can introduce a new package for junit3. I am not following the original issue though. Is code not working in Junit 4 or 3? Because afaik the JUnit4 integration has been working ever since we wrote the runner. I would be surprised if that has been broken since forever.

mockitoguy:
I suggest we take the discussion to the ticket: https://github.com/mockito/mockito/issues/1649#issuecomment-469742700

Thank you for attempting to tackle this!

mockitoguy:
See the #1649 for details.

PR #1665, opened 2019-03-16
New public API: invocation.getArgumentsAsMatchers()

### LINKED ISSUE #1664: Make invocation to drive its printing
Right now the `toString()` of an `Invocation` is ignored when it's being printed for a verification error, this prevents mockito-scala from providing a custom representation of the arguments via the implicit context.

With a simple change to drive the printing from within the `Invocation` the problem can be solved

bbonanno:
- New public API: invocation.getArgumentsAsMatchers(), useful for advanced integrations, such as mockito-scala
- Some internal refactorings that provide temporary help with mockito-scala integration

Fixes #1664

### DIFF
--- a/src/main/java/org/mockito/internal/invocation/InterceptedInvocation.java
+++ b/src/main/java/org/mockito/internal/invocation/InterceptedInvocation.java
@@ -4,6 +4,7 @@
  */
 package org.mockito.internal.invocation;
 
+import org.mockito.ArgumentMatcher;
 import org.mockito.internal.invocation.mockref.MockReference;
 import org.mockito.internal.exceptions.VerificationAwareInvocation;
 import org.mockito.internal.reporting.PrintSettings;
@@ -13,8 +14,10 @@ import org.mockito.invocation.StubInfo;
 
 import java.lang.reflect.Method;
 import java.util.Arrays;
+import java.util.List;
 
 import static org.mockito.internal.exceptions.Reporter.cannotCallAbstractRealMethod;
+import static org.mockito.internal.invocation.ArgumentsProcessor.argumentsToMatchers;
 
 public class InterceptedInvocation implements Invocation, VerificationAwareInvocation {
 
@@ -119,6 +122,23 @@ public class InterceptedInvocation implements Invocation, VerificationAwareInvoc
         return (T) arguments[index];
     }
 
+    public MockReference<Object> getMockRef() {
+        return mockRef;
+    }
+
+    public MockitoMethod getMockitoMethod() {
+        return mockitoMethod;
+    }
+
+    public RealMethod getRealMethod() {
+        return realMethod;
+    }
+
+    @Override
+    public List<ArgumentMatcher> getArgumentsAsMatchers() {
+        return argumentsToMatchers(getArguments());
+    }
+
     @Override
     public <T> T getArgument(int index, Class<T> clazz) {
         return clazz.cast(arguments[index]);
@@ -154,7 +174,7 @@ public class InterceptedInvocation implements Invocation, VerificationAwareInvoc
     }
 
     public String toString() {
-        return new PrintSettings().print(ArgumentsProcessor.argumentsToMatchers(getArguments()), this);
+        return new PrintSettings().print(getArgumentsAsMatchers(), this);
     }
 
     public final static RealMethod NO_OP = new RealMethod() {
--- a/src/main/java/org/mockito/invocation/Invocation.java
+++ b/src/main/java/org/mockito/invocation/Invocation.java
@@ -51,9 +51,11 @@ public interface Invocation extends InvocationOnMock, DescribedInvocation {
     Object[] getRawArguments();
 
     /**
-     * Returns arguments wrapped into ArgumentMatchers
+     * Wraps each argument using ArgumentMatchers.eq() or AdditionalMatchers.arryEq()
+     * Used internally for the purposes of human-readable invocation printing.
      *
      * @return a list of ArgumentMatcher wrapping each of this invocation arguments
+     * @since 2.25.6
      */
     List<ArgumentMatcher> getArgumentsAsMatchers();
 
--- a/src/main/java/org/mockito/invocation/Invocation.java
+++ b/src/main/java/org/mockito/invocation/Invocation.java
@@ -51,10 +51,11 @@ public interface Invocation extends InvocationOnMock, DescribedInvocation {
     Object[] getRawArguments();
 
     /**
-     * Wraps each argument using ArgumentMatchers.eq() or AdditionalMatchers.arryEq()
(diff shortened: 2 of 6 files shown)


src/main/java/org/mockito/internal/invocation/MatchersBinder.java
```diff
@@ -28,10 +28,14 @@ public InvocationMatcher bindMatchers(ArgumentMatcherStorage argumentMatcherStor
...
         }
+        return createMatcher(invocation, matchers);
+    }
+
+    protected InvocationMatcher createMatcher(Invocation invocation, List<ArgumentMatcher> matchers) {
         return new InvocationMatcher(invocation, matchers);
     }
 
-    private void validateMatchers(Invocation invocation, List<LocalizedMatcher> lastMatchers) {
+    protected void validateMatchers(Invocation invocation, List<LocalizedMatcher> lastMatchers) {
```
TimvdLippe: What is the reasoning for making this protected? It is an internal class, so it should not be subclassed.

src/main/java/org/mockito/internal/invocation/InterceptedInvocation.java
```diff
@@ -119,6 +119,18 @@ public Method getMethod() {
         return (T) arguments[index];
     }
 
+    public MockReference<Object> getMockRef() {
```
TimvdLippe: I don't see these being used anywhere else. Moreover, they are not on the interface `Invocation`. If we were to add them, they have to be on the interface.

src/main/java/org/mockito/internal/invocation/MatchersBinder.java
bbonanno: I currently need to extend it to replace the InvocationMatcher, so I made it protected just in case, but I'm happy to leave it private if you prefer so

src/main/java/org/mockito/internal/invocation/InterceptedInvocation.java
bbonanno: I'm calling it from mockito-scala, should I add it to the interface then?

bbonanno:
I agree that the PR has not a single purpose, but given the changes were minimal I thought it wouldn't be a problem

I can split it if you prefer so

src/main/java/org/mockito/invocation/DescribedInvocation.java
```diff
@@ -23,4 +25,12 @@
...
      */
     Location getLocation();
+
+    /**
+     * Describes the invocation in the human friendly way based on the provided settings.
+     *
+     * @param printSetting the settings to be considered.
+     * @return the description of this invocation.
+     */
+    String print(PrintSettings printSetting);
```
mockitoguy: We cannot do it. It's leaking internal API via a public API.

src/main/java/org/mockito/internal/invocation/InterceptedInvocation.java
mockitoguy: >should I add it to the interface then?

Please no. MockReference is an internal type and we should not leak it from a public interface.

src/main/java/org/mockito/internal/invocation/MatchersBinder.java
mockitoguy: @TimvdLippe, I'm ok with changing private method to protected on an internal type so that mockito-scala integration is easier in the short term. @bbonanno is aware that we don't guarantee compatibility for internal API in the long term.

bbonanno:
@mockitoguy @TimvdLippe Hey guys, thanks for your input

I took a different approach now, and the amount of changes required seems lower, the tl;dr story is that I kept the non-public-interface getters on `InterceptedInvocation` and I added a new one (which is public) that exposes the arguments as matchers, in this way we can delegate this transformation to the invocation, which makes it more consistent across the code, and also allows me to override said method to use my own matchers, which have their own toString, so I can customise how they are printed without having to mess that much with the internals of mockito-core

Please let me know your thoughts

src/main/java/org/mockito/invocation/Invocation.java
```diff
@@ -47,6 +50,13 @@
      */
     Object[] getRawArguments();
 
+    /**
+     * Returns arguments wrapped into ArgumentMatchers
+     *
+     * @return a list of ArgumentMatcher wrapping each of this invocation arguments
+     */
+    List<ArgumentMatcher> getArgumentsAsMatchers();
```
mockitoguy: This is good! We can add this method to the public API because it returns a public type.

mockitoguy: Can you add "since" tag because it's a new public API?

src/main/java/org/mockito/invocation/Invocation.java
```diff
@@ -47,6 +50,13 @@
      */
     Object[] getRawArguments();
 
+    /**
+     * Returns arguments wrapped into ArgumentMatchers
+     *
```
mockitoguy: Can you add a note that the method converts each argument using ArgumentMatchers.eq() or AdditionalMatchers.arryEq() and that it is used internally for the purposes of human-readable invocation printing.

I don't want users to confuse this new method with the MatchableInvocation.getMatchers() method.

src/main/java/org/mockito/invocation/Invocation.java
bbonanno: :)

bbonanno:
@mockitoguy I've improved the JavaDoc

src/main/java/org/mockito/invocation/Invocation.java
```diff
@@ -51,9 +51,11 @@
     Object[] getRawArguments();
 
     /**
-     * Returns arguments wrapped into ArgumentMatchers
+     * Wraps each argument using ArgumentMatchers.eq() or AdditionalMatchers.arryEq()
```
mockitoguy: Can you use Javadoc link? Thank you :)

bbonanno: Ups, sorry about that

mockitoguy:
Restarted the build. If it is happy, I'll merge!

bbonanno:
@TimvdLippe @mockitoguy Sorry to bother guys, but could you release this to maven central? or otherwise, when is the next notable version coming out?
Cheers

TimvdLippe:
Travis should be on its way: https://github.com/mockito/mockito/commit/8824e3e57daedce988c2cfc8d4c2b85bca488dbf

bbonanno:
@TimvdLippe much appreciated

TimvdLippe:
The build failed though (#1682) and I am not sure if it has actually pushed the artifact to Maven Central.

mockitoguy:
Thank you!

PR #1669, opened 2019-03-18
Add ErrorProne subproject

### LINKED ISSUE #1668: Add subproject for errorprone Mockito plugins
We heavily rely on ErrorProne plugins to prohibit certain usage of Mockito. Examples include disallowing the usages of Mockito internals or extending interfaces annotated by `@NotExtensible`.

These plugins are beneficial to the Java community and we would like to open source them to the project. The extra subproject would contain several checkers (and tests) that implement several safeguards. Users can add the checkers to their project to their benefit.

For example for Gradle, the configuration would be as follows (using https://github.com/tbroyer/gradle-errorprone-plugin#custom-error-prone-checks):

```gradle
dependencies {
  errorprone("org.mockito:mockito-errorprone:$mockitoVersion")
}
```

This would allow us to make issues like #1616 and #1613 compiler errors rather than runtime failures.

TimvdLippe:
These checkers allow us to enforce API contracts that the Java compiler
otherwise would not enforce. As a first example, the added checker
enforces that users can not subclass interfaces that are annotated by
`@NotExtensible`.

Fixes #1668

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -8,7 +8,7 @@ versions.bytebuddy = '1.9.7'
 versions.junitJupiter = '5.1.1'
 versions.errorprone = '2.3.2'
 
-libraries.junit4 = 'junit:junit:4.13-beta-1'
+libraries.junit4 = 'junit:junit:4.12'
 libraries.junitJupiterApi = "org.junit.jupiter:junit-jupiter-api:${versions.junitJupiter}"
 libraries.junitPlatformLauncher = 'org.junit.platform:junit-platform-launcher:1.1.0'
 libraries.junitJupiterEngine = "org.junit.jupiter:junit-jupiter-engine:${versions.junitJupiter}"
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -6,8 +6,9 @@ def versions = [:]
 
 versions.bytebuddy = '1.9.7'
 versions.junitJupiter = '5.1.1'
+versions.errorprone = '2.3.2'
 
-libraries.junit4 = 'junit:junit:4.12'
+libraries.junit4 = 'junit:junit:4.13-beta-1'
 libraries.junitJupiterApi = "org.junit.jupiter:junit-jupiter-api:${versions.junitJupiter}"
 libraries.junitPlatformLauncher = 'org.junit.platform:junit-platform-launcher:1.1.0'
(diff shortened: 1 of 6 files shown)


TimvdLippe:
Build is failing because of an issue after upgrading to JUnit 4.13 (https://github.com/junit-team/junit4/issues/1496#issuecomment-473972675). That was necessary, as errorprone itself relies on it.

gradle/dependencies.gradle
```diff
@@ -6,8 +6,9 @@ def versions = [:]
 
 versions.bytebuddy = '1.9.7'
 versions.junitJupiter = '5.1.1'
+versions.errorprone = '2.3.2'
 
-libraries.junit4 = 'junit:junit:4.12'
+libraries.junit4 = 'junit:junit:4.13-beta-1'
```
bric3: I think the JUnit dependency should be kept at 4.12, we should avoid depending on a beta version.

subprojects/errorprone/errorprone.gradle
```diff
@@ -0,0 +1,20 @@
...
+dependencies {
+    compile project.rootProject
+    compile libraries.errorprone
+
+    testCompile libraries.junit4
+    testCompile libraries.errorproneTestApi
+}
+
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
```
bric3: Shall we keep a lower target, to allow benefit for older platforms ?

TimvdLippe: ErrorProne must run on Java 8: https://errorprone.info/docs/installation

gradle/dependencies.gradle
TimvdLippe: The checker API requires 4.13 sadly. However, I will revert it to just run 4.13 for the tests on this project. 4.13 will be released next week.

TimvdLippe:
CI is finally green again :tada:

subprojects/errorprone/errorprone.gradle
bric3: OK that makes sense ;)

gradle/dependencies.gradle
bric3: OK waiting for junit 4.13 then

mockitoguy:
We'll see if shipkit correctly publishes it. It should. If not, I'll fix it :)

gradle/dependencies.gradle
mockitoguy: I'm OK shipping with 4.13 beta because this new jar is completely decoupled from other jars. Hope that helps!

PR #1670, opened 2019-03-20
Add support for Answer6 and VoidAnswer6

perlun:
Regretfully, I ran into a case where I needed to mock a six-argument method. One can of course claim that such methods _ought not to exist_, but sadly, this is not always the reality and I think that some pragmatism is useful in tools like Mockito, so here is my attempt at adding support for it.

Of course, it's possible to use the plain `Answer` interface in this case, but adding a dedicated `Answer6` interface makes it more convenient on the user; `Answer` is a quite raw interface.

(I'm not sure if this should target the 2.x release or `master` also, but feel free to cherry-pick as needed after review/merge.)

### DIFF
--- a/src/main/java/org/mockito/AdditionalAnswers.java
+++ b/src/main/java/org/mockito/AdditionalAnswers.java
@@ -16,11 +16,13 @@ import org.mockito.stubbing.Answer2;
 import org.mockito.stubbing.Answer3;
 import org.mockito.stubbing.Answer4;
 import org.mockito.stubbing.Answer5;
+import org.mockito.stubbing.Answer6;
 import org.mockito.stubbing.VoidAnswer1;
 import org.mockito.stubbing.VoidAnswer2;
 import org.mockito.stubbing.VoidAnswer3;
 import org.mockito.stubbing.VoidAnswer4;
 import org.mockito.stubbing.VoidAnswer5;
+import org.mockito.stubbing.VoidAnswer6;
 
 import static org.mockito.internal.stubbing.answers.AnswerFunctionalInterfaces.toAnswer;
 
@@ -486,6 +488,45 @@ public class AdditionalAnswers {
      */
     @Incubating
     public static <A, B, C, D, E> Answer<Void> answerVoid(VoidAnswer5<A, B, C, D, E> answer) {
+        return toAnswer(answer);
+    }
+
+    /**
+     * Creates an answer from a functional interface - allows for a strongly typed answer to be created
+     * idiomatically in Java 8
+     *
+     * @param answer interface to the answer - which is expected to return something
+     * @param <T> return type
+     * @param <A> input parameter type 1
+     * @param <B> input parameter type 2
+     * @param <C> input parameter type 3
+     * @param <D> input parameter type 4
+     * @param <E> input parameter type 5
+     * @param <F> input parameter type 6
+     * @return the answer object to use
+     * @since 2.26.0
+     */
+    @Incubating
+    public static <T, A, B, C, D, E, F> Answer<T> answer(Answer6<T, A, B, C, D, E, F> answer) {
+        return toAnswer(answer);
+    }
+
+    /**
+     * Creates an answer from a functional interface - allows for a strongly typed answer to be created
+     * idiomatically in Java 8
+     *
+     * @param answer interface to the answer - a void method
+     * @param <A> input parameter type 1
+     * @param <B> input parameter type 2
+     * @param <C> input parameter type 3
+     * @param <D> input parameter type 4
+     * @param <E> input parameter type 5
+     * @param <F> input parameter type 6
+     * @return the answer object to use
+     * @since 2.26.0
+     */
+    @Incubating
+    public static <A, B, C, D, E, F> Answer<Void> answerVoid(VoidAnswer6<A, B, C, D, E, F> answer) {
         return toAnswer(answer);
     }
 }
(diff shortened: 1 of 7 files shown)


TimvdLippe:
While we could simply ship this, my concerns are mostly "where do we stop?". If we ship with 6 arguments, we should also ship with 7. There does not seem to be a good reason not to.

However, I would prefer to limit to 5. These should cover all common cases. If we ship more, we are generating more code than is actually useful (and maintainable). It should be quite straightforward to write your own 6-answer, or use the lack thereof as an incentive to refactor you code to just use 5 arguments.

perlun:
> While we could simply ship this, my concerns are mostly "where do we stop?". If we ship with 6 arguments, we should also ship with 7. There does not seem to be a good reason not to.

Sure, that's a valid concern. We have to draw the line somewhere, I'm perfectly fine with that.

>  It should be quite straightforward to write your own 6-answer

That's actually how it started but I thought I'd go the extra mile and suggest this for inclusion in Mockito proper.

> or use the lack thereof as an incentive to refactor you code to just use 5 arguments.

Unfortunately, not that easy in this case. The method in question is an existing interface used for fetching TV recordings for a given customer, filtered by a given set of channels, a server ID (which can be null), a start time and end time and an optional flag determining if the result should be sorted or not.

In languages like Ruby, I could easily just add a hash parameter and provide `n` number of (optional) parameters, but in Java I'd have to wrap up a "parameter class" just for this sake and it feels kind of an overkill design if you ask me.

Simply put, real-world scenarios are not always so easy. Motivating change in systems simply because "the tools we use don't support more than 5 arguments" is perhaps also not so easy either.

----

Like you say though, we _can_ keep this as a local class in our project but I do think it makes some sense here since other users could find it useful as well. I think we could (in Mockito) consider expanding the "5-argument limit" to 10 in the long run. Yes, methods with 10 arguments are quite absurd but I'm not sure we will "win the war" against bad code simply by being too strict in this area.

TimvdLippe:
> That's actually how it started but I thought I'd go the extra mile and suggest this for inclusion in Mockito proper.

That is much appreciated, thanks for opening the discussion!

> but in Java I'd have to wrap up a "parameter class" just for this sake and it feels kind of an overkill design if you ask me.

That is inherently subjective. I would disagree and regard that as an appropriate solution. As a Mockito owner we can be selective on the APIs we ship, while keeping maintainability in mind. In this case, I don't think adding more convenience Answer classes is worth the extra maintenance cost.

mockitoguy:
> While we could simply ship this, my concerns are mostly "where do we stop?". If we ship with 6 arguments, we should also ship with 7. There does not seem to be a good reason not to.

@TimvdLippe, you're exactly right that is really hard to set a limit here! Thing is that this code has minimum complexity and is extremely easy to maintain (compiler!). I suggest we keep shipping new "increments" so long there are motivated customers that can send us a beautiful, well described, well tested PRs such as this one :)

perlun:
@mockitoguy Thanks, much appreciated. :) I fixed the trivial checkstyle problem and rebased on top of latest `release-2.x` from the main repo.

----

(Btw, running checkstyle locally for me fails w/ Java 11. It works on Java 8. The strange thing is that the Travis CI jobs run correctly even on the newer JDK versions so I wonder if there's some special conf going on there to make it work.)

```
$ ./gradlew check
Starting a Gradle Daemon (subsequent builds will be faster)
Parallel execution is an incubating feature.

> Configure project :
  Building version '2.25.2' (value loaded from 'version.properties' file).

> Task :checkstyleMain FAILED

* What went wrong:
Execution failed for task ':checkstyleMain'.
> Unable to create a Checker: configLocation {/home/per/git/3rd-party/mockito/config/checkstyle/checkstyle.xml}, classpath {/home/per/git/3rd-party/mockito/build/classes/java/main:/home/per/git/3rd-party/mockito/build/resources/main:/home/per/.gradle/caches/modules-2/files-2.1/net.bytebuddy/byte-buddy/1.9.7/8fea78fea6449e1738b675cb155ce8422661e237/byte-buddy-1.9.7.jar:/home/per/.gradle/caches/modules-2/files-2.1/net.bytebuddy/byte-buddy-agent/1.9.7/8e7d1b599f4943851ffea125fd9780e572727fc0/byte-buddy-agent-1.9.7.jar:/home/per/.gradle/caches/modules-2/files-2.1/org.objenesis/objenesis/2.6/639033469776fd37c08358c6b92a4761feb2af4b/objenesis-2.6.jar:/home/per/.gradle/caches/modules-2/files-2.1/junit/junit/4.12/2973d150c0dc1fefe998f834810d68f278ea58ec/junit-4.12.jar:/home/per/.gradle/caches/modules-2/files-2.1/org.hamcrest/hamcrest-core/1.3/42a25dc3219429f0e5d060061f71acb49bf010a0/hamcrest-core-1.3.jar}.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 5s
21 actionable tasks: 1 executed, 20 up-to-date
```

mockitoguy:
@TimvdLippe, https://github.com/mockito/mockito/pull/1670#issuecomment-474924864 - thoughts?

TimvdLippe:
> @TimvdLippe, [#1670 (comment)](https://github.com/mockito/mockito/pull/1670#issuecomment-474924864) - thoughts?

I stand by my opinion that this is not worth the maintenance increase, but if you want to take it on that I am all for it :smile: Don't regard my review as blocking.

mockitoguy:
@TimvdLippe, thank you!

PR #1676, opened 2019-03-26
Pinned ByteBuddy to 1.9.10

mockitoguy:
It seems that new version of BB breaks the build for Java 9+. I'm still investigating. Opening the PR for visibility - we can see if the build is happy now.


mockitoguy:
I'm going to merge this. It appears that there are consistent build failures on Java 9+ when ByteBuddy 1.9.11 is used. 1.9.10 works.

PR #1678, opened 2019-03-30
Create interface and extension point for MockHandlerFactory

emmostrom:
This creates a public interface for MockHandlerFactory and allows for custom/experimental  MockHandlers to be used only for specific mocked objects.

Example:

// default factory
... = Mockito.mock(IMethods.class);

// alternate factory
MockHandlerFactory altFactory = new MyFactory();
... = Mockito.mock(IMethods.class, Mockito.withSettings().mockHandlerFactory(altFactory));

This will make it easier to create a fine-grained extension that does not affect every mocked object.


TimvdLippe:
I like this proposal, but I do have a fundamental question: what is your usecase that warrants multiple MockHandlers? Could you provide a small project which validates that?

emmostrom:
If I understand your question you want to know why you would in a single project want to use two or more MockHandlers as opposed to just being able to change at a global level what the default handler is.  I can give you an example MockHandler similar to one I've used before; but I'm not sure it would answer the question because I think it is more of a philosophical API question.  [Here](https://www.linkedin.com/pulse/concurrent-testing-mockito-edward-mostrom) is an article I just put up that talks about a MockHandler I find useful but not necessarily one I would want to use for everything.

TimvdLippe:
I read the article, but it is unclear to me what specifically the `ConcurrentMock` can provide over a regular mock? If Mockito can't work concurrently, I would say we should fix that instead.

TimvdLippe:
Closing this PR as we require a concrete use case. Please file an issue first to have that discussion.

PR #1680, opened 2019-04-02
Fix handling of generic Void type for doNothing()

TimvdLippe:
When the return type bound to a generic type resolved to Void,
`doNothing()` would still throw an exception. Update the `isVoid`
implementation to also handle generic return types.


raphw:
Good catch.

PR #1683, opened 2019-04-09
Eagerly compute stackTraceLine in LocationImpl

TimvdLippe:
On typical tests with large stacktraces and many invocations on mocks,
the storing of Throwable in `stackTraceHolder` would result in
OutOfMemoryErrors. Eager computation of the stackTraceLine resulted in a
~2/3rd memory reduction of a typical test (1.5GB -> 400MB).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
 
     dependencies {
         classpath 'gradle.plugin.nl.javadude.gradle.plugins:license-gradle-plugin:0.14.0'
-        classpath 'ru.vyarus:gradle-animalsniffer-plugin:1.4.4' //for 'java-compatibility-check.gradle'
+        classpath 'ru.vyarus:gradle-animalsniffer-plugin:1.5.0' //for 'java-compatibility-check.gradle'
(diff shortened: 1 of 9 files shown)


TimvdLippe:
It will be a tiny bit slower, but I think this minimal compared to the amount of time saved by not invoking the GC. I observed GC times of multiple seconds trying to clean up all of the Throwables when a mock was garbage-collected.

TimvdLippe:
Btw, I am planning on investing some more time in this to investigate all hot paths and figure out other performance improvements. That should hopefully negate (if not all) computation regressions this might generate.

raphw:
Good idea, did you encounter this error in a test setting or is this a hypothetical? I am mainly wondering about the real-life impact. I do not think that the throughput would be an issue here over the saved memory consumption.

TimvdLippe:
Seems like I broke quite a few tests. Investigating...

mockitoguy:
Thank you!

felixdesouza:
I just ran into this on 2.23.4 yesterday 🙄, spent a few hours debugging this! OOMs/seg faults were pretty cryptic! I was about to write a fix, but then I stumbled onto this so thanks! 😄

PR #1685, opened 2019-04-15
[DO NOT MERGE] Update to Java 8

TimvdLippe:
This also removes all unnecessary configuration for Java 6 as well as
references to the release/2.x branch.

**DO NOT MERGE. This PR is for review only. Once approved, we will start releasing with Shipkit from this branch**


TimvdLippe:
I have manually cancelled the release jobs. Not sure how to move forward with that. Probably checkout a new branch from 3.x and then reset 3.x to 2.x for now, so that any future pushes won't retrigger the release logic.

.travis.yml
```diff
@@ -48,7 +48,7 @@ script:
   # We are using && below on purpose
   # ciPerformRelease must run only when the entire build has completed
   # This guarantees that no release steps are executed when the build or tests fail
-  - ./gradlew build idea -s -PcheckJavaCompatibility && ./gradlew ciPerformRelease
+  - ./gradlew build idea -s && ./gradlew ciPerformRelease
```
szpak: Probably in the long run we want to check the compatibility with Java 8

TimvdLippe: We can revisit that later, if we can no longer build with Java 8 on Travis. For now, we are building on JDK 8, so we should be compatible by default.

TimvdLippe:
I will release Mockito 3 next week if no reviews are posted, given the discussion with approvals we had in Slack.

TimvdLippe:
The first beta has been published. Closing this PR.

PR #1686, opened 2019-04-15
[DO NOT MERGE] Add test about wrong annotations for Guice injectors

TimvdLippe:
This is intended to show the problem we observed with our Guice integration. We should be able to construct a test case without Guice once we are able to determine what the issue is.

This is a problem with the ByteBuddy mockmaker, as it is passing with the CGLibMockmaker. We are not sure what the issue is, but it is related to annotations being persisted and generic types being removed from method parameters. @raphw do you maybe know what might be going on here?

CC @erichrison

### DIFF
--- a/gradle/java-library.gradle
+++ b/gradle/java-library.gradle
@@ -15,8 +15,8 @@ generatePomFileForJavaLibraryPublication.doLast {
 
 bintrayUpload.enabled = true
 
-sourceCompatibility = 1.6
-targetCompatibility = 1.6
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
 
 test {
     include "**/*Test.class"
--- a/settings.gradle
+++ b/settings.gradle
@@ -8,6 +8,7 @@ include 'junit-jupiter'
 include 'junitJupiterExtensionTest'
 include 'module-test'
 include 'memory-test'
+include 'guice-test'
 include 'errorprone'
 
 rootProject.name = 'mockito'
--- a/subprojects/guice-test/guice-test.gradle
+++ b/subprojects/guice-test/guice-test.gradle
@@ -1,0 +1,16 @@
+plugins {
+    id 'java'
+}
+description = "Test suite for Guice integration"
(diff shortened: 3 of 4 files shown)


raphw:
I think this would almost be expected behavior. If Mockito creates a mock of `Injectable<Void>`, Byte Buddy is actually asked to create a mock for `Injectable` (raw) as Mockito is not passing the generic `Type` but the `Class`. By the Java specification, a raw type cannot have generic members. To fullfil this Byte Buddy overrides `setStrings(List strings)` with a raw signature. This probably puts Guice off but is fully correct according to the spec. (Try implementing a raw subclass with your IDE and override the method, I bet the same thing happens.)

This is actually important to correctly resolve the type hierarchy for generic types with raw members in the hierarchy and cglib can implement incorrect overrides if such types are blended together which is why I chose to remain spec compliant.

The best solution would be to provide Byte Buddy the `Type` to mock rather then the `Class`. In this case, everything would work as expected.

TimvdLippe:
Shipping an API based on `Type` would make sense to me. We might even want to consider making that the default, if it can resolve all those generics issues we have been facing.

For now, I think we should investigate if we can somehow configure a mock to force the generics to be specified (like CGLib would I suppose).

raphw:
We could attempt a generification of the `Class` before supplying it to Byte Buddy. For this, one would just resolve any generic variable to its lower bound, e.g. `Object` in your example.

From this, Byte Buddy will resolve generic types as the generated type is no longer 'raw'. Do however note that this can cause rather confusing virtual dispatch behavior if a user is expecting a more specialized generic type.

Ideally, we should however retrofit the byte code provider API to supply `Type` rather then `Class` where the latter is a subclass of the former. Doing so, we can create prober mocks where also generic return variables are resolved and therefore mocked correctly.

TimvdLippe:
Closing this as WAI for now. We should revisit this overall issue to investigate a non-Class based API that can properly handle generics.

PR #1690, opened 2019-04-17
Add ErrorProne MockitoInternalUsage checker

TimvdLippe:
This checker will warn on any usage of a Mockito internal symbol. Since
we do not guarantee API stability on symbols in our internal packages,
users should not rely on them. This checker can be used to enforce that
they are unused.


TimvdLippe:
I will merge this after #1693

PR #1693, opened 2019-04-22
Make `MockitoNotExtensible` service-loadable

### LINKED ISSUE #1692: The new Error Prone plugin isn't service-loadable
The `MockitoNotExtensible` plugin added in #1669 is pretty cool. However, when `mockito-errorprone` is added to the annotation processor classpath, Error Prone does not recognize/find the plugin. This is because `MockitoNotExtensible` is not declared in a `META-INF/services/com.google.errorprone.bugpatterns.BugChecker` file on the classpath.

Stephan202:
This way Error Prone can find the plugin when `mockito-errorprone` is added to the annotation processor classpath.

Fixes #1692.

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -23,6 +23,8 @@ libraries.bytebuddyandroid = "net.bytebuddy:byte-buddy-android:${versions.bytebu
 libraries.errorprone = "com.google.errorprone:error_prone_core:${versions.errorprone}"
 libraries.errorproneTestApi = "com.google.errorprone:error_prone_test_helpers:${versions.errorprone}"
 
+libraries.autoservice = "com.google.auto.service:auto-service:1.0-rc5"
+
 // objenesis 3.x with full Java 11 support requires Java 8, bump version when Java 7 support is dropped
 libraries.objenesis = 'org.objenesis:objenesis:2.6'
 
--- a/subprojects/errorprone/errorprone.gradle
+++ b/subprojects/errorprone/errorprone.gradle
@@ -4,6 +4,9 @@ apply from: "$rootDir/gradle/java-library.gradle"
 apply from: "$rootDir/gradle/dependencies.gradle"
 
 dependencies {
+    compileOnly libraries.autoservice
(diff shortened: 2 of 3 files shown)


Stephan202:
NB: in this PR I chose to use [Google AutoService](https://github.com/google/auto/tree/master/service) as is customary for Error Prone plugins. If preferred I can instead introduce an explicit ` META-INF/services/com.google.errorprone.bugpatterns.BugChecker` file.

Stephan202:
@TimvdLippe sure! Will have a look at what's the simplest way to achieve this.

TimvdLippe:
@Stephan202 Were you able to take a look at this? I wanted to fix the same in #1690.

Stephan202:
@TimvdLippe I considered several approaches and concluded that the best way forward would be to see whether/how support for this kind of "real"/"full" test can be added to Error Prone's `CompilationTestHelper` and `BugCheckerRefactoringTestHelper` classes. The idea would be that they get another static factory method which relies on service loading rather than an explicitly provided checker class. This would also benefit other non-Google EP users, such as @uber's [NullAway](https://github.com/uber/NullAway).

But then "life happened", so I didn't yet get to the point of actually trying that implementation. Note that even if it would work, this PR would be blocked on those changes getting merged and released as part of Error Prone. (But I assume you could help getting the changes reviewed by the EP team.)

So that leaves the question of timelines. I can try (but won't commit) to spend some time on this later this week, likely in the weekend. Alternatively, if you agree with the suggested solution, you could have a crack at an Error Prone PR and I'd be happy to be the first reviewer.

W.r.t. whether this PR should be blocked on all that: I have manually tested the changes in this PR, so functionally we could just merge this for now. The approach taken here is [exactly](https://github.com/uber/NullAway/blob/c3979b4241f80411dbba6aac3ff653acbb88d00b/nullaway/src/main/java/com/uber/nullaway/NullAway.java#L142) what NullAway does. (But "just merging" is not in the spirit of the Mockito project, so I totally undestand not wanting to go that route.)

TimvdLippe:
Let's merge it for now and hope we don't regress in the future.

PR #1695, opened 2019-04-27
Fix javadoc typo and use HTML list

Marcono1234:
The javadoc for `InvocationOnMock.getArgument(int index, Class<T> clazz);` had two typos:
- "to directly **pas** the result"
- "**clazz** to cast the argument to"

Additionally I changed the list to be a proper HTML list, the current version just collapsed it into one text block: https://static.javadoc.io/org.mockito/mockito-core/2.27.0/org/mockito/invocation/InvocationOnMock.html#getArgument-int-java.lang.Class-

It might also be worth it adding the actual HTML paragraphs into the javadoc (for the other methods as well).


TimvdLippe:
Travis is failing on unrelated issues. Will take a look next week.

PR #1698, opened 2019-04-30
Fix ClonesArguments for null and arrays

tombrown52:
The ClonesArguments class works for many cases, but fails when trying to clone an argument that is an array, or is null. This PR fixes those edge cases. This does not directly resolve any open issue, but it is related to #1153, #1469

check list

### DIFF
--- a/src/test/java/org/mockitousage/stubbing/CloningParameterTest.java
+++ b/src/test/java/org/mockitousage/stubbing/CloningParameterTest.java
@@ -12,6 +12,7 @@ import org.mockitoutil.TestBase;
 import java.util.List;
 
 import static org.junit.Assert.assertNotNull;
+import static org.mockito.ArgumentMatchers.*;
 import static org.mockito.Mockito.*;
 
 public class CloningParameterTest extends TestBase {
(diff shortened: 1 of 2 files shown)


tombrown52:
@TimvdLippe I believe I have the right branch this time. Please review

src/test/java/org/mockitousage/stubbing/CloningParameterTest.java
```diff
@@ -26,12 +27,12 @@ public void shouldVerifyEvenIfArgumentsWereMutated() throws Exception {
         businessLogic(emailSender);
 
         // then
-        verify(emailSender).sendEmail(1, new Person("Wes"));
+        verify(emailSender).sendEmail(eq(1), eq(new Person("Wes")), (Person)any(), (Person[])any());
```
TimvdLippe: Could you revert these changes and instead add a new method with multiple arguments. (and thus add a new test to use that method separately)

tombrown52:
@TimvdLippe I made new tests instead of extending the existing one. Please review again.

PR #1699, opened 2019-04-30
Fix JaCoCo code coverage report

### LINKED ISSUE #1689: Coverage report is broken since Gradle 5
We upgraded to Gradle 5 in #1683 to fix our Java 11 build. However, this has now broken our coverage computation because of a breaking change in Gradle 4.10: https://github.com/gradle/gradle/pull/6419

There is no guidance from the Gradle team on how to migrate to a proper configuration. I have spend a couple of hours today to try to fix it, with no result. The TLDR is that Gradle now throws away the `test.exec` after executing the tests in a subproject. Thus, it will only have the coverage from the last subproject it ran.

I have looked online and there are other projects that do something similar (https://github.com/grpc/grpc-java/blob/master/all/build.gradle, https://github.com/groovy/groovy-core/blob/master/gradle/jacoco/jacoco.gradle) but none of the approaches worked for us.

marcphilipp:
Instead of collecting execution data of all test tasks of all
subprojects in a single file, each subproject's test task writes its own
execution data file. The `mockitoCoverage` task is configured to include
all execution data files into account that exists when it starts. This
is done to avoid exceptions for missing files for skipped test tasks for
the `android` and `errorprone` subprojects.

Fixes #1689.

---

### DIFF
--- a/gradle/root/coverage.gradle
+++ b/gradle/root/coverage.gradle
@@ -22,15 +22,12 @@ task mockitoCoverage(type: JacocoReport) {
                 if (currentProject != rootProject) {   //already automatically enhanced in mockito main project as "java" plugin was applied before
                     applyTo test
                 }
+                test.jacoco.destinationFile = file("${currentProject.buildDir}/jacoco/test.exec")
             }
 
-            currentProject.tasks.withType(Test) { testTask ->
-                mockitoCoverage.dependsOn testTask
-            }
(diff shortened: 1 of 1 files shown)


marcphilipp:
I think so, I just didn’t want to mix up this PR.

TimvdLippe:
Yes agreed. Let's tackle 1 problem at the time. I will clean that up tomorrow.

TimvdLippe:
Hurray! I can't express how happy I am seeing this fix and so quickly.

Issue #1359, opened 2018-04-03
Update Gradle files to Kotlin


TimvdLippe:
Whenever I have to update our Gradle files, I have trouble expressing it in Groovy. Since Gradle now also supports Kotlin files, it might be better to upgrade our files to be built with Kotlin instead. This would be type-safe and have improved editor support (e.g. autocomplete).

PR #1704, opened 2019-05-05
Migrate junitJupiterExtensionTest subproject to kotlin dsl

epeee:
Another step towards #1359.
This pr migrates a very first subproject to kotlin dsl.


subprojects/junitJupiterExtensionTest/junitJupiterExtensionTest.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+tasks.named<Test>("test") {
+    useJUnitPlatform()
+}
+
+val Project.libraries
+    get() = rootProject.extra["libraries"] as Map<String, Any>
+
+fun Project.library(name: String) =
```
TimvdLippe: Should these functions be declared in `dependencies.gradle`?

epeee: yes, indeed, we have to move them to a better place. I will do so while migrating the next subproject if that's ok for you.

TimvdLippe: Sounds like a plan 😄

Issue #1631, opened 2019-02-26
Regression: @Spy with @InjectMock are not injected anymore


arnor2000:
Hi,

 - [x] The mockito message in the stacktrace have useful information, but it didn't help
    java.lang.NullPointerException
	at org.test.TestedService.servivce(TestedService.java:16)
	at org.test.TestedServiceTest.mockSpyAndInjectAll(TestedServiceTest.java:33)
 - [x] The problematic code (if that's possible) is copied here;
    Fix of #1587 (Remove unnecessary loop from InjectingAnnotationEngine) has introduce a regression.
 - [x] Provide versions (mockito / jdk / os / any other relevant information)
    Issue can be reproduce starting from Mockito version 2.23.15.
 - [x] Provide a [Short, Self Contained, Correct (Compilable), Example](http://sscce.org) of the issue
    Please find attached [inject-spy.zip](https://github.com/mockito/mockito/files/2905749/inject-spy.zip).
    execute `./gradlew check`
- [x] Read the [contributing guide](https://github.com/mockito/mockito/blob/master/.github/CONTRIBUTING.md)

We are experiencing a regression behavior caused by fix of #1587 (Remove unnecessary loop from InjectingAnnotationEngine) which prevent us to update from Mockito 2.23.4 to newer versions.

```java
public class TestedServiceTest extends AbstractTest {

    @InjectMocks
    private TestedService testedService;

    @InjectMocks
    @Spy
    private SpiedCompositionService spiedCompositionService;

    @Mock
    private MockedDao mockedDao;
    ...
}
```

As you can see in this snippet and example project attached, with Mockito 2.23.14 when we try to inject a mock in a spy and this spy in the tested object, it works as expected and everything is injected. But when we switch to Mockito 2.23.15, this behavior doesn't work anymore, the spy is not injected to tested object and obviously causes a NullPointerException.

We also figured out injection of injected spies was working only because test class extends has an abstract test class.

I understand this test case is not a state of the art way to do unit test, our use case is more a component test because several classes work together to do a coherent unit of work and it makes more sense to test them together.

We are relying on the behavior provided up to Mockito 2.23.14 and the only workaround I see is to expose attributes to package visibility and do injection by hand.

The loop removed by #1587 was injecting injected spies newly instantiated by first loop. The depth of injection was relying on the depth of test hierarchy.

jpasquers:
[...]

Issue #1518, opened 2018-10-13
Deprecate InjectMocks


TimvdLippe:
I think no one of the core team is a fan of this feature. The implementation is not as full-fletched as other dependency injection mechanisms. At the time of writing this issue, there are 17 open issues describing various problems with `@InjectMocks`. Moreover, as it solves a problem unrelated to mocking and there are different tools available to solve this use case, I propose we deprecate `@InjectMocks` and schedule for removal in Mockito 3/4. (I think 4 is best, to keep 3 a minimal breaking change version to ease upgrades).

ChristianSchwarz:
Good idea! 
A lib relate to this topic: [Jukito ](https://github.com/ArcBees/Jukito) handles the dependency injection for Mockito using Guice.

bric3:
I agree however, we should have a proper replacement already in place.
By the way dropping this may bring havoc in the world.

TimvdLippe:
For simple cases, manual constructing the object should work. For more complicated cases, I think Jukito is appropriate.

An option is to, in the deprecation message, ask for feedback if either of these solutions is not feasible for the user. That way we know if there are actually use cases we can't cover (and thus would have to keep supporting `@InjectMocks`)

mockitoguy:
Our users love this feature! (just talk to them, you'll see :)

We already had this discussion before the feature was introduced. @bric3, do you remember? It was contentious back then, we knew that we cannot compete with fully fledged dependency injection framework. We documented this and moved on. Over time, it paid off because I heard multiple times that our users really like this feature.

Does deprecating this feature helps our community? There are many users (like myself ;) who are happy with the current state of InjectMocks. I have never had problems with it. I have not seen a request from the community to get rid of this feature.

@TimvdLippe, what is the problem you're are trying to solve? It's ok to have 17 tickets - users can help us fix them. We can also close those tickets if we don't want the fixes.

TimvdLippe:
The problem with all these issues is that the expectation of the user is broken. By having an inheritently broken feature in our API, we are not living up to our promises imo. Rather than having something that sometimes work, I think it is better to let the task be solved by those frameworks that are built to solve them.

msangel:
[Jukito](https://github.com/ArcBees/Jukito) seems to be a good choice for google-guice based apps. How about spring boot autoconfiguration? Will it work?

bric3:
@mockitoguy Yes I do remember, you were quite lucid back then and I was quite young.
[...]

PR #1710, opened 2019-05-10
Fixes #1516 and #1631 : Allows @Spy with @InjectMocks to be injected into other @InjectMocks

### LINKED ISSUE #1516: @Spy @InjectMocks dependency A->spy(B)->C not injected correctly when using constructor dependency injection
I am trying to do a test on class A, with a dependency A->B to be spied, and a transitive dependency B->C

This code is typical in a spring application, when A is a high-level service, C is a low-level one (like a Repository), and B is simply a helper service class that delegate most calls to C

To use @Spy on instance of B, requires to mock C, and inject C into B... 
a typical test code should be 
`
   @RunWith(MockitoJUnitRunner.class)
   class Test {
        @InjectMocks A a;
	@Spy @InjectMocks  B b; 
        @Spy C c;

       @Test  public void testA() { .. }
}
`

This generally works fine with standard spring injections for code like 
`
   @Component // or @Service, @Controller, ..
   public class B {
      @Autowired // or @Inject
      private C c;
   }
`

... BUT fails  when using Constructor dependency injection for fields !!!
This is absurd since constructor injection is precisely a best practice for enforcing code stability regarding  test

`
   @Component
   public class B {
      private final C c;
      public B(C c) { this.c = c;  }
   }
`

There are many strange situations, while testing different combinations of annotations on real class (@Inject, @Autowired..), and @Spy @InjectMocks in test class

Either 
case 1/ spy NOT injected in  a.b   ... so causing NPE in spy execution
case 2/ spy NOT injected in b.c ... but injected in a.b  ... so causing NPE in spy execution
case 3/ MockitoException Unable to initialize @Spy annotated field 'b', Please ensure that the type 'CtorB' has a no-arg constructor

Notice also that the final desired scenario would be to use lombok generated full constructor.. but it also fails !!
`
	@InjectMocks 
	A a; // ERROR.. a.b == null
	
	@Spy @InjectMocks
	B b; // b instanciated, but not injected into a !!!

	@Spy
	C c;

    // with..
    @AllArgsConstructor(onConstructor_={@Inject}) // equivalent of :  @Inject public A(B b) { this.b = b; }
    public class A {
	private final B b;
    }
   // or also 
   @AllArgsConstructor
   public class A {
	private final B b;
   }
`

I have joined a full maven test project that reproduce many of these scenarios
[test-mockito.zip](https://github.com/mockito/mockito/files/2466624/test-mockito.zip)

You can also browse it (up-to-date) here : 
[https://github.com/Arnaud-Nauwynck/test-snippets/tree/master/test-mockito]

arnor2000:
check list

This PR aims to fix #1516 and #1631. I know and understand dependency injection is not priority of Mockito team but discussion on #1518 convinces me to propose a fix. Like @mockitoguy I think DI in Mockito is useful and most of reported issues denote users high expectations for a helpful feature rather than the fact it's broken.

Basically this PR allows fields annotated with @Spy and @InjectMocks to be injected into other fields annotated with @InjectMocks and so on.
It may also be a response to #174, the need to a @Real annotation, actually I think most usage of @Spy and @InjectMocks combination is a workaround to such annotation.

Because PR ended to change a bunch of classes, I will highlight main changes.

"Entry point" of changes is InjectingAnnotationEngine, I add a loop of mock injection. The loop uses variants of existing field injection strategies documented in DefaultInjectionEngine.

To facilitate field injection strategies variants (StrictConstructorInjection and LenientPropertyAndSetterInjection), I merge ConstructorArgumentResolver (contract to find constructor arguments) and ConstructorInstantiator (contract to find and instantiate constructor) interfaces into an unique ConstructorResolver interface representing a strategy to find a constructor and its arguments. Constructor instantiation part comes back into FieldInitializer.

The fact to merge these two interfaces into ConstructorResolver leads to a better separation of concerns between finding constructor and its arguments in one side and instantiation, initialization and field injection on the other side. isResolvable() allows some tuning between the inclination of strategy implementations to fail with exception or handle failure softly. Moreover it's opening future smarter behaviors, for example BiggestConstructorResolver could try to resolve biggest constructor matching actual available mocks.

I hope you will appreciate this work.


TimvdLippe:
It's awesome that you worked on this issue, but this PR is quite large. It will take us quite some time to review and we are busy with some other things as well. Hopefully we can give your a proper code review soon. (I had a quick glance and the code seems good, so that's a good sign 😄 )

bric3:
Hi @arnor2000 thanks for contributing on this, I will take a look as soon as I have some free time.

arnor2000:
@bric3, OK I cherry picked changes to release/3.x into PR #1711 without any conflict. I let you decide to close one or other PR at your convenience.

TimvdLippe:
Superseded by #1711

PR #1715, opened 2019-05-23
Fixes #298: replaces Objenesis references from stubbing classes with plugin equivalents.

### LINKED ISSUE #298: ObjenesisHelper references in MockMaker-independent code
A couple of references to org.objenesis classes in code that is otherwise independent of bytecode-based mocks:

src/main/java/org/mockito/internal/stubbing/answers/ClonesArguments.java
src/main/java/org/mockito/internal/stubbing/answers/ThrowsExceptionClass.java

Both of these classes use ObjenesisHelper.newInstance(Class), which differs from Class.newInstance() in that no constructors are called. Is that necessary for these classes? If not, then updating to Class.newInstance() makes Mockito fully platform-independent (at least for platforms that define a MockMaker). If it is necessary, can this call be moved into MockMaker?

I'm happy to make the changes once the team decides on the best approach.

tomball:
This enables the iOS Mockito plugin
(https://github.com/google/j2objc/tree/master/testing/mockito) to upgrade to v2.

### DIFF
--- a/src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
+++ b/src/main/java/org/mockito/internal/stubbing/BaseStubbing.java
@@ -4,6 +4,8 @@
  */
 package org.mockito.internal.stubbing;
 
+import org.mockito.creation.instance.Instantiator;
+import org.mockito.internal.configuration.plugins.Plugins;
 import org.mockito.internal.stubbing.answers.CallsRealMethods;
 import org.mockito.internal.stubbing.answers.Returns;
 import org.mockito.internal.stubbing.answers.ThrowsException;
@@ -12,7 +14,6 @@ import org.mockito.stubbing.OngoingStubbing;
 
 import static org.mockito.internal.exceptions.Reporter.notAnException;
 import static org.mockito.internal.progress.ThreadSafeMockingProgress.mockingProgress;
-import static org.objenesis.ObjenesisHelper.newInstance;
 
 public abstract class BaseStubbing<T> implements OngoingStubbing<T> {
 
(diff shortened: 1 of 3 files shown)


tomball:
Please bump the Maven version to 2.27.1 so j2objc's build can reference this change.

tomball:
Should I also Objenesis references should be moved out of core Mockito? I did the minimum to keep the patch small, but a more complete change would:

 1. Move [ObjenesisInstantiator](https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/internal/creation/instance/ObjenesisInstantiator.java#L11) to a bytebuddy-like subpackage, such as org.mockito.internal.creation.instance.objenesis, and
 1. Initialize [DefaultInstantiatorProvider](https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/internal/creation/instance/DefaultInstantiatorProvider.java#L13) so it uses the plugin API to define its default.

That small re-org, plus the new unit test, should make it much more obvious what is core and what belongs in bytecode-based Mockito implementations.

TimvdLippe:
> Should I also Objenesis references should be moved out of core Mockito?

Let's put that in a follow up issue/PR and keep this PR small. I want to get this patch in, but want to have the regression test in here now. That way, if we do decide to do the follow up PR, we can refactor safely.

tomball:
Sounds good. I should have the test done later today. It's a little more involved than NoJUnitDependenciesTest because the default plugins are getting loaded. I'll see if a set of stub plugins can be substituted (may be useful for future tests), and if not, update the test's classloader to not test the list of default plugin classes.

tomball:
Test added. It's not perfect since DefaultMockitoPlugins has bytebuddy plugin names hardwired, so classes that trigger that class loading needed to be removed. To fully test plugin loading code, we'd have to make DefaultMockitoPlugins.DEFAULT_PLUGINS settable so a set of plugin stub classes can be substituted. This seems like a useful, but bigger change that you likely want with this PR.

TimvdLippe:
I think this test is fine for now. Thanks for adding it!

PR #1723, opened 2019-06-06
Change implementation of ExceptionFactory to explicitly test for dependent classes

### LINKED ISSUE #1716: mockito-android 2.26+ tries to use missing opentest4j exceptions
It appears that Mockito `2.26+` on Android tries to use exceptions from OpenTest4j, even when OpenTest4j is not used, resulting in tests crashing due to exceptions classes not being found.

Steps to reproduce:

1. Open [MockitoAndroidDemo](https://github.com/mockito/mockito/files/3215946/MockitoAndroidDemo.zip) project in Android Studio
2. Run `ExampleInstrumentedTest.java` test on android device/emulator
3. Test will crash with `java.lang.NoClassDefFoundError: Failed resolution of: Lorg/opentest4j/AssertionFailedError`

Lowering version to `2.25.0` fixes the issue

kriegfrj:
The original implementation of conditional OpenTest4J support (#1667) relied on some implicit behaviour of the Java classloader to generate and catch the exception at the right time. It seems that this behaviour is not always exactly replicated in all environments - for example, in #1716, where Mockito was being used in an instrumentation test that was being run on an Android emulator.

The new implementation is a bit more direct in how it tests for the dependent classes. The existing test cases still pass, and @matejdro confirmed that this fixed his problem.

Fixes #1716.

### DIFF
--- a/src/main/java/org/mockito/internal/junit/ExceptionFactory.java
+++ b/src/main/java/org/mockito/internal/junit/ExceptionFactory.java
@@ -21,29 +21,16 @@ public class ExceptionFactory {
         ExceptionFactoryImpl theFactory = null;
 
         try {
-            theFactory = new ExceptionFactoryImpl() {
-                @Override
-                public AssertionError create(String message, String wanted, String actual) {
-                    return new org.mockito.exceptions.verification.opentest4j.ArgumentsAreDifferent(message, wanted, actual);
-                }
-            };
-        } catch (Throwable onlyIfOpenTestIsNotAvailable) {
+            Class.forName("org.opentest4j.AssertionFailedError");
+            theFactory = org.mockito.exceptions.verification.opentest4j.ArgumentsAreDifferent::new;
+        } catch (ClassNotFoundException onlyIfOpenTestIsNotAvailable) {
             try {
-                theFactory = new ExceptionFactoryImpl() {
-                    @Override
-                    public AssertionError create(String message, String wanted, String actual) {
-                        return new org.mockito.exceptions.verification.junit.ArgumentsAreDifferent(message, wanted, actual);
-                    }
-                };
-            } catch (Throwable onlyIfJUnitIsNotAvailable) {
+                Class.forName("junit.framework.ComparisonFailure");
+                theFactory = org.mockito.exceptions.verification.junit.ArgumentsAreDifferent::new;
+            } catch (ClassNotFoundException onlyIfJUnitIsNotAvailable) {
             }
         }
-        factory = (theFactory == null) ? new ExceptionFactoryImpl() {
-            @Override
-            public AssertionError create(String message, String wanted, String actual) {
-                return new ArgumentsAreDifferent(message, wanted, actual);
-            }
-        } : theFactory;
+        factory = (theFactory == null) ? ArgumentsAreDifferent::new : theFactory;
     }
 
     /**


matejdro:
I just want to note that all `mockito-android` builds after `2.26` are unusable without `OpenTest4J` installed until this patch. So maybe you could create patch release for just Android? Not sure what the timeline for 3.x is.

TimvdLippe:
@matejdro I am aiming for releasing Mockito 3 today.

matejdro:
Okay never mind then :-)

src/main/java/org/mockito/internal/junit/ExceptionFactory.java
```diff
@@ -21,21 +21,23 @@ private ExceptionFactory() {
...
+                Class.forName("junit.framework.ComparisonFailure");
                 theFactory = new ExceptionFactoryImpl() {
                     @Override
                     public AssertionError create(String message, String wanted, String actual) {
                         return new org.mockito.exceptions.verification.junit.ArgumentsAreDifferent(message, wanted, actual);
                     }
                 };
-            } catch (Throwable onlyIfJUnitIsNotAvailable) {
+            } catch (ClassNotFoundException onlyIfJUnitIsNotAvailable) {
             }
```
tlubz: something like this may be a little more robust, and also DRYer:

```java
        ExceptionFactoryImpl theFactory = null;

        String[] paths = {
            "org.mockito.exceptions.verification.opentest4j.ArgumentsAreDifferent",
            "org.mockito.exceptions.verification.junit.ArgumentsAreDifferent"
        };
        Constructor<?> constructor = null;
        for(String path : paths) {
            try {
                Class<?> c = Class.forName(path);
                constructor = c.getConstructor(String.class, String.class, String.class);
            } catch (ClassNotFoundException | NoSuchMethodException e) {
                // skip and try next
            }
        }

        if(constructor != null) {
            final Constructor constructor1 = constructor;
            theFactory = (message, wanted, actual) -> {
                try {
                    return (AssertionError) constructor1.newInstance(message, wanted, actual);
                } catch (IllegalAccessException | InstantiationException | InvocationTargetException e) {
                    throw new RuntimeException("failed to instantiate AssertionError", e);
                }
            };
        } else {
            theFactory = (message, wanted, actual) -> {
                return new ArgumentsAreDifferent(message, wanted, actual);
            };
        }
```

kriegfrj: Hi @tlubz, thanks a bunch for your comments!

Lambdas would certainly be an improvement. I originally wrote it using lambdas and method references, but the main code base was still on 1.6 compliance and so I had to ditch the lambdas (it seems we've since moved on). With lambdas and method references, the code is a lot more compact:

```java
    static {
        ExceptionFactoryImpl theFactory = null;
        try {
            theFactory = org.mockito.exceptions.verification.opentest4j.ArgumentsAreDifferent::new;
        } catch (Throwable t) {
            try {
                theFactory = org.mockito.exceptions.verification.junit.ArgumentsAreDifferent::new;
            } catch (Throwable t2) {
            }
        }
        factory = (theFactory == null) ? ArgumentsAreDifferent::new
                : theFactory;
```

Re: your desire to apply the DRY principle: this is a worthy aim, but because it requires the use of reflection it comes at the expense of compile-time type checking. Both will work, but my personal preference (given the low number of repetitions) was to sacrifice DRY to retain the compile-time type checking. Arguably my compile-time safety "pro" doesn't matter as much because you can catch errors at test time with good unit tests (as we have).

Re: being more robust: this is debatable. Unless you've successfully tried it with `AndroidJUnitRunner`, I suspect that your proposed implementation will still hit issue #1716. The problem is that the actual `org.mockito.exceptions.verification.opentest4j.ArgumentsAreDifferent` and `org.mockito.exceptions.verification.junit.ArgumentsAreDifferent` classes are always found, so they never throw `ClassNotFoundException` when you try and load them. However, when the classloader attempts to *resolve* the class (ie, find all its dependent classes), *this* fails if the appropriate supertype assertion class isn't on the classpath. Failing resolution causes a `NoClassDefFoundError` to be thrown rather than a `ClassNotFoundException`. In a regular JVM environment, this `NoClassDefFoundError` propagates up and is caught by the surrounding `catch (Throwable)`, but it seems that `AndroidJUnitRunner` on the emulator behaves differently - it notices the `NoClassDefFoundError`, logs it, and aborts the test. Hence the patch in this PR tested specifically for the existence of the supertype assertion class (using reflection) before attempted to instantiate the dependent `ArgumentsAreDifferent` class.

So I think for your solution to work, you'd also have to add an explicit test for `org.opentest4j.AssertionError` et al, as this PR did.

kriegfrj:
What's happened with this PR - did it ever get merged into 3.x or into 2.x? It doesn't seem to have been merged as far as I can see. Are you waiting for me to rebase it onto 3.x and resubmit?

TimvdLippe:
This should be rebased onto 3.x: https://github.com/mockito/mockito/pull/1723#pullrequestreview-246417736

kriegfrj:
I'm so sorry, I didn't realise you were waiting for me to do the rebase. I'll get it fine in the next 14 hours or so.

TimvdLippe:
No problem at all, don't rush! We have plenty of time to get this in, there is no deadline. Work on this whenever you have the time to do so, don't force it.

src/main/java/org/mockito/internal/junit/ExceptionFactory.java
tlubz: Great feedback. Thanks for taking the time. mostly what prompted by suggestion is that I was concerned that `Class.forName(...);` may be optimized away by an aggressive optimizer if its result is unused. Though my concerns may be unfounded, since it throws checked exceptions, so it has a potential side-effect. 🤷‍♂

kriegfrj:
Ok, rebased it onto `release/3.x` and changed the base branch of my pull request.

I also changed it to use lambdas per @tlubz's suggestion.

Issue #977, opened 2017-03-06
Misleading documentation of verifyZeroInteractions


ffissore:
## Problem

The current behavior of verifyZeroInteractions can be misleading, see the #989.

In current version of Mockito (2.x), we can improve the javadoc for verifyZeroInteractions to clearly describe how it works (and mention the behavior of strict stubs, too).

Please contribute!

## Original report

Original issue reported by @ffissore:

According to the [docs](http://static.javadoc.io/org.mockito/mockito-core/2.7.13/org/mockito/Mockito.html#verifyZeroInteractions(java.lang.Object...)) `verifyZeroInteractions` is like `verify(mock, never()).method()` when the mock has only one method.

The tests below should both raise an exception. Instead, `testVerifyZeroInteractions` raises none and fails

```java
package test;

import org.junit.Before;
import org.junit.Rule;
import org.junit.Test;
import org.mockito.Mock;
import org.mockito.MockitoAnnotations;
import org.mockito.exceptions.verification.NeverWantedButInvoked;
import org.mockito.exceptions.verification.NoInteractionsWanted;
import org.mockito.junit.MockitoJUnit;
import org.mockito.junit.MockitoRule;
import org.mockito.quality.Strictness;

import static org.junit.Assert.assertEquals;
import static org.mockito.Mockito.*;

public class MockitoTest {

   private static class Collaborator {

      public String doStuff(int number) {
         throw new RuntimeException();
      }

   }

   private static class Container {

      private final Collaborator collaborator;

      public Container(Collaborator collaborator) {
         this.collaborator = collaborator;
      }

      public String doStuff(int number) {
         return collaborator.doStuff(number);
      }

   }


   @Rule
   public MockitoRule rule = MockitoJUnit.rule().strictness(Strictness.STRICT_STUBS);

   @Mock
   private Collaborator collaborator;

   @Before
   public void setUp() throws Exception {
      MockitoAnnotations.initMocks(this);
   }

   // passes
   @Test(expected = NeverWantedButInvoked.class)
   public void testVerifyNever() throws Exception {
      Container container = new Container(collaborator);
      when(collaborator.doStuff(42)).thenReturn("yeah");

      String output = container.doStuff(42);
      assertEquals("yeah", output);

      verify(collaborator, never()).doStuff(42);
   }

   // fails
   @Test(expected = NoInteractionsWanted.class)
   public void testVerifyZeroInteractions() throws Exception {
      Container container = new Container(collaborator);
[...]

PR #1732, opened 2019-06-25
verifyZeroInteractions guarantees zero interactions instead of just delegating to verifyNoMoreInteractions

ffissore:
With #995 I provided a fix to #977 and #989
However that PR targeted `master`, which eventually became `old-3.x`, and its code didn't make it into `release/3.x`.
#977 and #989 should be reopened.

I have back-ported that PR to target `release/2.x`.
`verifyZeroInteractions` is marked as `@Deprecated` and a new `verifyNoInteractions` is added. There are no breaking changes AFAIK, so this code is backwards compatible.
Since I don't know if/when this code will ship, javadocs contain a placeholder `Since 2.x.x`

### DIFF
--- a/src/main/java/org/mockito/BDDMockito.java
+++ b/src/main/java/org/mockito/BDDMockito.java
@@ -251,7 +251,9 @@ public class BDDMockito extends Mockito {
         /**
          * @see #verifyZeroInteractions(Object...)
          * @since 2.1.0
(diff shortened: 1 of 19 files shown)


ffissore:
While looking around, I've found https://github.com/mockito/mockito/pull/1723#pullrequestreview-246417736, so I'll close this PR and open a new one targeting release/3.x

PR #1733, opened 2019-06-25
 verifyNoInteractions guarantees no interactions with mock

ffissore:
With #995 I provided a fix to #977 and #989
However that PR targeted `master`, which eventually became `old-3.x`, and its code didn't make it into `release/3.x`.
#977 and #989 should be reopened.

I have back-ported that PR to target `release/3.x`.
`verifyZeroInteractions` is marked as `@Deprecated` and a new `verifyNoInteractions` is added. There are no breaking changes AFAIK, so this code is backwards compatible.
Since I don't know if/when this code will ship, javadocs contain a placeholder `Since 3.x.x`

### DIFF
--- a/src/main/java/org/mockito/BDDMockito.java
+++ b/src/main/java/org/mockito/BDDMockito.java
@@ -251,7 +251,9 @@ public class BDDMockito extends Mockito {
         /**
          * @see #verifyZeroInteractions(Object...)
          * @since 2.1.0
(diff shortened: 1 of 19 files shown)


TimvdLippe:
Thanks for this and good catch! I still have to release Mockito 3, but wasn't able to because of issues with our release mechanism. I will try that next week after which we can merge this PR 😄

TimvdLippe:
Mockito 3 was released so I am merging :tada:

Issue #1736, opened 2019-07-07
(Possible) memory leak in MockitoJUnitRunner


awillinger:
Hi,

while debugging a memory leak issue affecting the test execution at work, I came across a (potential) memory leak present within the MockitoJUnitRunner.

Basically, when using the runner, it creates a DefaultInternalRunner that itself subclasses the default BlockJUnit4ClassRunner, injecting itself into its lifecycle as well as registering an anonymous RunListener in order to de-register Mockito-internal run listeners (and provide some other features).

Now, this wouldn't usually be a problem, however, the DefaultInternalRunner never unregisters the RunListener, while the internal subclass (referenced through the Listener) still holds a strong reference to the test class instance.

This in turn has caused issues with our tests, as we initialize a JPA EntityManager for certain tests that are (in certain cases) also kept track of in the tested class. Now, the RunNotifier itself is a singleton, and eventually holds a reference to the test class instance during the whole execution, preventing a cleanup and (in our case) causing 800+ MB of unnecessary memory usage.

Example (from a simple, isolated test case):
![javaw_2019-07-07_20-57-37](https://user-images.githubusercontent.com/2893404/60772659-ec1af000-a0f9-11e9-82ec-aa61cfab5d00.jpg)

While browsing through some other issues I have noticed that the JUnitRunner itself shouldn't really be used anymore. Therefore, I am wondering if it's even worth fixing at this point? If yes, I will prepare a repository with some test classes demonstrating the problem.

Versions used:
Mockito Core: 2.28.2
JUnit: 4.12

TimvdLippe:
The listener is intended to be removed: https://github.com/mockito/mockito/blob/220494473031298300a4c47c247c8b778cc39164/src/main/java/org/mockito/internal/runners/DefaultInternalRunner.java#L62 Not sure why it is not removed. If you can, please open a PR on Mockito with reproduction test and (if possible) a fix.

awillinger:
Actually, the problematic line is this one:

https://github.com/mockito/mockito/blob/220494473031298300a4c47c247c8b778cc39164/src/main/java/org/mockito/internal/runners/DefaultInternalRunner.java#L73

As in, the registration within the core JUnit RunNotifier (not in Mockito's internal).
Would it also be okay if I simply created a new repository with some test cases I created and linking it here instead of making a PR? As I already created a setup which is easily reproducible.

TimvdLippe:
[...]

PR #1755, opened 2019-08-11
Test case and fix for memory leak issue described in #1736

awillinger:
Added a testcase as wanted here: https://github.com/mockito/mockito/issues/1736#issuecomment-511144519

Please keep in mind that this test currently fails, as I did not work on a fix yet.
For more information, please see issue #1736.

### DIFF
--- a/src/main/java/org/mockito/internal/junit/DefaultTestFinishedEvent.java
+++ b/src/main/java/org/mockito/internal/junit/DefaultTestFinishedEvent.java
@@ -5,12 +5,12 @@
 package org.mockito.internal.junit;
 
 public class DefaultTestFinishedEvent implements TestFinishedEvent {
-    private final Object testClassInstance;
+    private final Class<?> testClass;
     private final String testMethodName;
     private final Throwable testFailure;
 
-    public DefaultTestFinishedEvent(Object testClassInstance, String testMethodName, Throwable testFailure) {
-        this.testClassInstance = testClassInstance;
+    public DefaultTestFinishedEvent(Class<?> testClass, String testMethodName, Throwable testFailure) {
+        this.testClass = testClass;
         this.testMethodName = testMethodName;
         this.testFailure = testFailure;
     }
@@ -22,6 +22,6 @@ public class DefaultTestFinishedEvent implements TestFinishedEvent {
 
     @Override
     public String getTestName() {
-        return testClassInstance.getClass().getSimpleName() + "." + testMethodName;
+        return testClass.getSimpleName() + "." + testMethodName;
     }
 }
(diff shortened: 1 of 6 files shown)


awillinger:
Build seems to be "correctly" failing now, as in, with an OOM because of the bug.

awillinger:
Sorry for the delay, I only just now managed to leave from work early so I can take a look at this again.

The fix for the initial issue (the leak in the RunNotifier) is an easy one, just need to set aside some time to implement/test it.

Not sure how to proceed with the one occurring within the Suite Runner - see last comment in the linked issue.. would have to remove the final modifier from the runner and clear it as well when the run finishes, unless you got a better idea?

TimvdLippe:
> unless you got a better idea

Sadly no :cry: Looking forward to your fix!

awillinger:
Finally managed to get around fixing this problem. 

In the diff itself, you will see the following 3 main changes:
* An extension of the memory-test.gradle, explicitly excluding all classes whose name contain the string "TestDefinition". I had to add that, as otherwise, JUnit would execute the Suite AS WELL AS the test class in itself, so 3 times in total instead of twice. With this fix, the test class will only ever run if it is included explicitly within a test suite.
Interestingly, there appears to be a memory leak within JUnit Suite classes as well, as in this case (suite + test class itself) the runner would still be referenced through a "BeforeMethods"-class - out of focus for this issue.
More info: https://discuss.gradle.org/t/how-to-run-a-testsuite-but-not-the-test-classes-directly/14576
* The test case itself, as requested in the issue, whereas the test class itself is named "TestDefinition", to avoid running it outside the suite. This simply allocates a Long-array which uses up ~127.9MB of RAM.
* The fix itself. Instead of resetting the runner / making the field non-final, I decided to simply drop the "target" field entirely, and pass through the testClass-Class direcltry to the event, as the only usage of the "target" field was a call to Object#getClass - which is not needed, as JUnit passes in the to-be-run class already.
I still added the RunNotifier#removeListener call to make the call site "symmetric".

mockitoguy:
I'll review this. Thank you!

TimvdLippe:
I was about to merge this (great issue description and fix!) but sadly Travis is failing with ```org.mockito.memorytest.ShouldFreeMemoryAfterExecutionFinishedSuite > org.mockito.memorytest.ShouldFreeMemoryAfterExecutionFinishedTestDefinition.classMethod FAILED
    java.lang.OutOfMemoryError```

Could you take another look and find out if we need to make additional changes? Thanks and sorry for the slow response.

awillinger:
Well, I did run a full Gradle build on my dev machine which completed successfully. Seems like only two of the environments (both JDK8s) are failing - one of which for an unrelated reason.

Is there a way to re-trigger Travis without actually pushing another commit? Maybe it got confused with the rebase-force-push shenanigans I was doing.

TimvdLippe:
I have restarted the jobs. Lets see if they are still problematic

TimvdLippe:
I am closing this PR for inactivity and Travis failures. Feel free to reopen once they are resolved 😄

PR #1761, opened 2019-08-16
Run Travis on Xenial

TimvdLippe:
Per https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment this is now the default. We previously set Trusty explicitly to fix our build failures. However, since then we switched to `openjdk` which should not have any issues on Xenial.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,6 @@ sudo: true
 
 language: java
 
-dist: trusty
-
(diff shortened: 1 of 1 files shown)


TimvdLippe:
CheckStyle is unhappy on Xenial with openjdk11. Not sure why this would suddenly break.

PR #1762, opened 2019-08-19
Fixes #1758: Method GenericTypeExtractor#findGenericInterface always …

### LINKED ISSUE #1758: Method GenericTypeExtractor#findGenericInterface always returns first  interface
The Method `findGenericInterface` of `GenericTypeExtractor` ( <https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/internal/util/reflection/GenericTypeExtractor.java>) always returns the first generic interface.

The relevant line seems to be:
<https://github.com/mockito/mockito/blob/220494473031298300a4c47c247c8b778cc39164/src/main/java/org/mockito/internal/util/reflection/GenericTypeExtractor.java#L64>

Maybe instead of 
`return sourceClass.getGenericInterfaces()[0];` 
it should be 
`return sourceClass.getGenericInterfaces()[i];`

Example:
```java
interface MyInterface extends Runnable, Consumer<Double> {
}

Type result = GenericTypeExtractor.findGenericInterface(MyInterface.class, Consumer.class);
System.out.println(result);
// output: interface java.lang.Runnable
// expected output: java.util.function.Consumer<java.lang.Double>
```
Mockito version: 2.x 3.x
Java version: jdk8u202-b08

serssp:
…returns first interface

### DIFF
--- a/src/test/java/org/mockito/internal/util/reflection/GenericTypeExtractorTest.java
+++ b/src/test/java/org/mockito/internal/util/reflection/GenericTypeExtractorTest.java
@@ -38,6 +38,7 @@ public class GenericTypeExtractorTest extends TestBase {
 
     interface Crazy extends Serializable, IDeeper, Cloneable {}
     class Crazier extends EvenDeeper implements Crazy {}
+    class SecondGeneric implements Serializable, IBase<Integer> {}
 
     @Test public void finds_generic_type() {
         assertEquals(Integer.class, genericTypeOf(IntImpl.class, Base.class, IBase.class));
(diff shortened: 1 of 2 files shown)


src/test/java/org/mockito/internal/util/reflection/GenericTypeExtractorTest.java
```diff
@@ -64,5 +65,7 @@
         assertEquals(Integer.class, genericTypeOf(IDeeper.class, Base.class, IBase.class));
         assertEquals(Integer.class, genericTypeOf(Crazy.class, Base.class, IBase.class));
         assertEquals(Integer.class, genericTypeOf(Crazier.class, Base.class, IBase.class));
+
+        assertEquals(Integer.class, genericTypeOf(SecondGeneric.class, Base.class, IBase.class));
```
TimvdLippe: These should have been separate test methods, but this is consistent. Feel free to send a follow-up PR to fix that.

PR #1779, opened 2019-09-14
Convenience factory method for creating stub objects

### LINKED ISSUE #1778: Convenience factory for stubs
A `stub(Class)` factory method on the `Mockito` class would be a convenient way to create stub-only objects:

```
List stub = stub(List.class);
```

Such a factory method improves the clarity of test code by succinctly communicating the test author's intent for the test-double.

gilday:
Fixes #1778

This improves test clarity when creating stubs by helping test authors succinctly communicate their intentions.

Adds a `stub(Class)` convenience method to `Mockito` which is an alias for `mock(stubClass, withSettings().stubOnly())`.

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -1881,6 +1881,24 @@ public class Mockito extends ArgumentMatchers {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mockitoguy:
See discussion in #1778

PR #936, opened 2017-02-07
[WIP] Unification of timeout(..) and after(..) API aka within()   

ChristianSchwarz:
----
**Please use this issue to review only the implementation. For API related discussions like why within() and not xy() please use ticket [#472 Unify after() and timeout() ](../issues/472) !**

----

__**Fixes**: [#472 Unify after() and timeout() ](../issues/472)__
Fixes: [#417 Add new WithinRange VerificationMode](../issues/417)
Fixes: [#345 Verification using After and ArgumentCaptor returns unexpected size of captured values list](../issues/345)
Fixes: [#1037 Duplicate arguments captured with multithreaded method call](../issues/1037)
Relates to: [#91 Document Mockito.after() method ](../issues/91)

### What
* This PR unifies the API of `timeout()` and `after()` into the new API `within()` that is accessible throu `Mockito.within(..)`. 
* This PR also includes an enhanced `atLeast()` API that allow to combine it with `atMost()` (see #417). 
Example:  `verify(mock, within(100,MILLISECONDS).atLeast(2).andAtMost(4)).simpleMethod();`

### How the behaviour differs from after/timeout
A within-verification always succeed or fail fast where applicable. Not all verification can fail or succeed fast. Eg. times(n) can only fail fast if more calls are detected but not succeed fast cause it needs to wait the full time if less calls than required where detected.

* Verification modes that check for a lower  bound will succeed fast if enough calls were detected.
Example:
```java
mock.methodCall();
mock.methodCall();
mock.methodCall();
verify(mock, within(100,MILLISECONDS).atLeast(2)).methodCall()
//^ succeed immediatly, doesn't wait the full 100ms
```
* Verification modes that check for a upper bound will fail fast if more calls were detected.
Example:
```java
mock.methodCall();
mock.methodCall();
mock.methodCall();
verify(mock, within(100,MILLISECONDS).times(2))).methodCall()
//^ fails immediatly, doesn't wait the full 100ms
```

### Compability
The within implementation is fully compatible cause it doesn't change the timeout or after API nor internal classes. 

### Implementation
Unlike _after_ and _timeout_ the _within_ implementation doesn't poll for invocations in a loop but uses an queue based approach. When the verification starts all invocations that where made until now where placed in that queue. New invocations are added to the queue using an invocation listener. The queue will be polled and its invocations are passed to the verification strategy (e.g. never, atMost ,..) it decides if the verification condition is satified or the verification failed. This allows to fail or succeed fast.

### Current state
The following verifications are working now:
[...]

PR #1787, opened 2019-09-27
Fixes #1786 : Clarify differences between 'timeout' and 'after'

### LINKED ISSUE #1768: StrictStubs doesn't work with @org.junit.experimental.categories.Category
mockito version 3.0.0

Can't use `org.junit.experimental.categories.Category` annotation  with `org.mockito.junit.MockitoJUnitRunner.StrictStubs#StrictStubs` runner to have all of this runner benefits. Because of `filterRequested` is always `true` in such case so this runner don't report about unused stubs.

```public class StrictRunner implements InternalRunner {

    private final Class<?> testClass;
    private final InternalRunner runner;
    private boolean filterRequested;

    /**
     * @param runner - the runner to wrap around
     * @param testClass - for reporting purposes
     */
    public StrictRunner(InternalRunner runner, Class<?> testClass) {
        this.runner = runner;
        this.testClass = testClass;
    }

    public void run(RunNotifier notifier) {
        //TODO need to be able to opt in for full stack trace instead of just relying on the stack trace filter
        UnnecessaryStubbingsReporter reporter = new UnnecessaryStubbingsReporter();
        FailureDetector listener = new FailureDetector();

        Mockito.framework().addListener(reporter);
        try {
            // add listener that detects test failures
            notifier.addListener(listener);
            runner.run(notifier);
        } finally {
            Mockito.framework().removeListener(reporter);
        }

        if (!filterRequested && listener.isSuccessful()) {
            //only report when:
            //1. if all tests from given test have ran (filter requested is false)
            //   Otherwise we would report unnecessary stubs even if the user runs just single test from the class
            //2. tests are successful (we don't want to add an extra failure on top of any existing failure, to avoid confusion)
            reporter.validateUnusedStubs(testClass, notifier);
        }
    }

    public Description getDescription() {
        return runner.getDescription();
    }

    public void filter(Filter filter) throws NoTestsRemainException {
        filterRequested = true;
        runner.filter(filter);
    }
}```

### LINKED ISSUE #1786: Difference between 'timeout' and 'after' not immediately clear
The first few lines of the JavaDoc for both methods do not exactly specify their difference. From `timeout` you have to jump to `after` where at the end of the JavaDoc the difference is explained.

I think this can be improved. The first sentence in `timeout` could be more specific to highlight the difference directly.

SimplyDanny:
Tries to improve the JavaDoc for `timeout` to differentiate it from `after` directly in the first sentence.

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -2820,14 +2820,13 @@ public class Mockito extends ArgumentMatchers {
     }
 
     /**
-     * Verification will be triggered after given amount of millis, allowing testing of async code.
+     * Verification will be triggered over and over until the given amount of millis, allowing testing of async code.
      * Useful when interactions with the mock object did not happened yet.
-     * Extensive use of after() method can be a code smell - there are better ways of testing concurrent code.
+     * Extensive use of {@code timeout()} method can be a code smell - there are better ways of testing concurrent code.
      * <p>
      * See also {@link #after(long)} method for testing async code.
      * Differences between {@code timeout()} and {@code after} are explained in Javadoc for {@link #after(long)}.
-     * <p>
-     * Extensive use of {@code timeout()} method can be a code smell - there are better ways of testing concurrent code.
+     *
      * <pre class="code"><code class="java">
      *   //passes when someMethod() is called no later than within 100 ms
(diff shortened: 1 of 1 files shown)


ChristianSchwarz:
The root cause can be fixed by PR #936 "Unification of timeout(..) and after(..) API"

PR #1788, opened 2019-09-28
Upgraded JUnit Jupiter 5.1.1 -> 5.4.2

### LINKED ISSUE #1630: Mockito JUnit Jupiter extension does not correctly support parallel test execution
Sorry for not being able to provide the required information, but we just noticed that the MockitoExtension (mockito-junit-jupiter 2.24.5) leads to spurious test failures (e.g. due to calls to mocks not being recorded) when using "junit.jupiter.execution.parallel.enabled=true" with surefire. Prior we used our much simpler MockitoExtension which works fine in the same situation:

```java
public class MockitoExtension implements Extension, TestInstancePostProcessor, AfterEachCallback {
   @Override
   public void postProcessTestInstance(Object testInstance, ExtensionContext context) {
      MockitoAnnotations.initMocks(testInstance);
   }

   @Override
   public void afterEach(ExtensionContext extensionContext) throws Exception {
   }
}
```

mockitoguy:
We need higher version of JUnit Jupiter to reproduce and fix #1630

Added a test to enhance coverage

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -5,7 +5,7 @@ ext {
 def versions = [:]
 
 versions.bytebuddy = '1.9.10'
-versions.junitJupiter = '5.1.1'
+versions.junitJupiter = '5.4.2'
 versions.errorprone = '2.3.2'
 
 libraries.junit4 = 'junit:junit:4.12'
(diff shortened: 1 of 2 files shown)


mockitoguy:
Coverage drop is a red herring. I verified the new test runs as part of CI. Merging!

PR #1789, opened 2019-09-28
Fixed JUnit Jupiter parallel issue

### LINKED ISSUE #1630: Mockito JUnit Jupiter extension does not correctly support parallel test execution

mockitoguy:
Fixes #1630

This fix improves Mockito JUnit Jupiter extension. However, it does not completely resolve all kinds of parallel issues when nested test classes are used. I'll open a separate ticket for it.

### DIFF
--- a/subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/MockitoExtension.java
+++ b/subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/MockitoExtension.java
@@ -5,14 +5,20 @@
 package org.mockito.junit.jupiter;
 
 
-import org.junit.jupiter.api.extension.*;
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
+import org.junit.jupiter.api.extension.ParameterContext;
+import org.junit.jupiter.api.extension.ParameterResolutionException;
+import org.junit.jupiter.api.extension.ParameterResolver;
+import org.junit.jupiter.api.extension.TestInstancePostProcessor;
 import org.mockito.Mock;
 import org.mockito.Mockito;
 import org.mockito.MockitoSession;
+import org.mockito.internal.configuration.MockAnnotationProcessor;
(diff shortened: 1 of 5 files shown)


subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/MockitoExtension.java
```diff
@@ -190,16 +196,24 @@ public void beforeEach(final ExtensionContext context) {
     private void collectParentTestInstances(ExtensionContext context, Set<Object> testInstances) {
         Optional<ExtensionContext> parent = context.getParent();
 
+        boolean firstParent = true;
```
TimvdLippe: Instead of having a boolean here, can we pass in `context.getParent()` on line 164. E.g. we pass the first parent in already, but then we keep all the other logic the same. That seems a little bit nicer to me and easier to follow. The comment can then be placed before line 164 instead.

mockitoguy: Please push a commit to this branch - it's simpler!

TimvdLippe: Sadly I am not at my working computer. I can do a push on Monday if you prefer.

mockitoguy: This works :-) Push on Monday. Have a nice weekend!

TimvdLippe: Done PTAL

mockitoguy:
This is much nicer, thank you!!!

PR #1790, opened 2019-09-30
Enable spotless to automatically format our source code

TimvdLippe:
Spotless [1] has been used by various popular open source projects,
including JUnit [2]. While working on Mockito inside of Google, I
discovered that our formatting is inconsistent. We can use Spotless to
automatically take care of that.

To be able to format the source code, run

./gradlew spotlessApply

This check runs on Travis to make sure code remains formatted. This
should also reduce the amount of trouble with the CheckStyle linter, as
Spotless will automatically resolve these issues.

On purpose, the spotless configuration is quite minimal. The current
configuration is mostly a reflection on the current state of affairs. We
can later discuss different configuration options and how we can
integrate them.

[1]: https://github.com/diffplug/spotless/tree/master/plugin-gradle
[2]: https://github.com/junit-team/junit5/blob/3f491f3148d2c745808fc75b0a802b60243e104e/build.gradle.kts#L124-L144

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ script:
   # We are using && below on purpose
   # ciPerformRelease must run only when the entire build has completed
   # This guarantees that no release steps are executed when the build or tests fail
-  - ./gradlew build idea -s && ./gradlew ciPerformRelease
+  - ./gradlew spotlessCheck && ./gradlew build idea -s && ./gradlew ciPerformRelease
 
 after_success:
   #Generates coverage report:
--- a/src/main/java/org/mockito/internal/matchers/apachecommons/EqualsBuilder.java
+++ b/src/main/java/org/mockito/internal/matchers/apachecommons/EqualsBuilder.java
@@ -2,9 +2,6 @@
  * Copyright (c) 2007 Mockito contributors
  * This program is made available under the terms of the MIT License.
  */
-
-//NON-STANDARD LICENCE HEADER HERE - THAT'S OK
-//Class comes from Apache Commons Lang, added some tiny changes
 package org.mockito.internal.matchers.apachecommons;
 
 import java.lang.reflect.AccessibleObject;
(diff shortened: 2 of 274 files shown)


src/main/java/org/mockito/internal/matchers/apachecommons/EqualsBuilder.java
```diff
@@ -14,6 +11,7 @@
 import java.util.Collections;
 import java.util.List;
 
+// Class comes from Apache Commons Lang, added some tiny changes
```
szpak: I wonder if it is ok to re-license the class on the Apache license to MIT.

.travis.yml
```diff
@@ -50,7 +50,7 @@ script:
   # We are using && below on purpose
   # ciPerformRelease must run only when the entire build has completed
   # This guarantees that no release steps are executed when the build or tests fail
-  - ./gradlew build idea -s && ./gradlew ciPerformRelease
+  - ./gradlew spotlessCheck && ./gradlew build idea -s && ./gradlew ciPerformRelease
```
szpak: It could be probably implemented to fail fast in the build `./gradlew spotlessCheck build idea`, but that is simpler to achieve the same effect.

src/main/java/org/mockito/internal/matchers/apachecommons/EqualsBuilder.java
TimvdLippe: I am not sure. We should probably get rid of these classes and https://static.javadoc.io/org.mockito/mockito-core/3.1.0/org/mockito/ArgumentMatchers.html#refEq-T-java.lang.String...- as a whole. Filed #1800 to track that effort.

.travis.yml
TimvdLippe: I wanted to keep it clear that we fail fast on formatting only. We can probably later clean this up if we can enforce the correct dependency of the formatter and the build.

PR #1805, opened 2019-10-22
Added Tidelift info to README.md

mockitoguy:
Added Tidelift blurp info to README.md. It's one of the Tidelift tasks and I think it's ok.


mockitoguy:
> Is this exact wording necessary?

No. However, this wording has the most marketing "power" according to data from Tidelift (they have evidence).

> Can we maybe move it down a bit

We can. My take on this:
 - If someone is looking for specific information (i.e. because has questions about versions), he will find it regardless if versioning is 1st or 3rd paragraph.
 - the marketing info needs to be really high so that folks who don't look for this information see it :-)

Perhaps we move this marketing blurp above everything so that folks not interested in this can quickly move down to find the info they want.

Thank you for feedback!

mockitoguy:
@TimvdLippe, I'll push this PR if that's OK. If you want to change the message, please submit a PR with how you would prefer it (I'm OK merging it).

PR #1806, opened 2019-10-22
Fixed JUnit5 concurrency bug

mockitoguy:
Fixed JUnit Jupiter MockitoExtension so that it behaves correctly in a parallel execution with nested tests. I found this issue recently when working on the extension. ThreadLocal here is OK but I'm happy to use a different solution if someone has a better idea.

Added a new test that was failing prior to this change.

### DIFF
--- a/subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/MockitoExtension.java
+++ b/subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/MockitoExtension.java
@@ -5,13 +5,6 @@
 package org.mockito.junit.jupiter;
 
 
-import static org.junit.jupiter.api.extension.ExtensionContext.Namespace.create;
-import static org.junit.platform.commons.support.AnnotationSupport.findAnnotation;
-
-import java.lang.reflect.Parameter;
-import java.util.LinkedHashSet;
-import java.util.Optional;
-import java.util.Set;
 import org.junit.jupiter.api.extension.AfterEachCallback;
 import org.junit.jupiter.api.extension.BeforeEachCallback;
 import org.junit.jupiter.api.extension.ExtensionContext;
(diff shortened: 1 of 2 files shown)


mockitoguy:
> Travis is not happy about it :(

I made him happy now :-)

mockitoguy:
Thanks for review! I wanted to double check the ThreadLocal idea is something we are OK pushing forward. Will merge later today.

TimvdLippe:
FYI @marcphilipp for this change.

subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/MockitoExtension.java
```diff
@@ -160,7 +165,7 @@ public void beforeEach(final ExtensionContext context) {
         Set<Object> testInstances = new LinkedHashSet<>();
         testInstances.add(context.getRequiredTestInstance());
 
-        this.collectParentTestInstances(context, testInstances);
+        this.collectParentTestInstances(testInstances);
```
marcphilipp: This is not necessary anymore. Since you're on 5.4.2, you can remove the `ThreadLocal` and simply call

```java
List<Object> testInstances = context.getRequiredTestInstances().getAllInstances();
```

subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/MockitoExtension.java
```diff
@@ -133,6 +136,8 @@ private MockitoExtension(Strictness strictness) {
         this.strictness = strictness;
     }
 
+    private static final ThreadLocal<Queue<Object>> TEST_INSTANCE_HIERARCHIES = ThreadLocal.withInitial(ConcurrentLinkedQueue::new);
+
```
marcphilipp: `TEST_INSTANCE` is now unused.

subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/MockitoExtension.java
mockitoguy: Fantastic code review feedback!!! Many thanks!!!

mockitoguy: Also, thank you for this sweet new API :-)

mockitoguy:
Updated per great code review feedback (much simpler code). When Travis CI is happy I'll merge.

subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/MockitoExtension.java
marcphilipp: You're welcome! 😊

PR #68, opened 2014-08-02
Added custom failure message to Mockito.verify. Issue 482

geoffschoeman:
Issue 482: passing custom failure message to Mockito.verify

PR #1807, opened 2019-10-22
Add default implementation for VerificationMode#description

TimvdLippe:
This method was added in Mockito 2 ([1]) and was implemented in all
relevant subclasses. However, most subclasses do not implement a custom
description.

Since we are now building on Java 8, we can ship this as a default
method and remove all duplicate subclass implementations. As a
side-effect, it allows users of Mockito 1 to migrate to Mockito 3
without additional breakages. (They could implement it, but
VerificationModeFactory is internal and would require additional
boilerplate)

[1]: https://github.com/mockito/mockito/pull/68


mockitoguy:
I'll review by the end of today. Looks good at first sight. I'd like to remind myself about our VerificationMode hierarchy :)

mockitoguy:
Thank you!

PR #1588, opened 2019-01-10
Fixes #1587 : Remove unnecessary loop from InjectingAnnotationEngine

LihMeh:
Removes InjectingAnnotationEngine::processInjectMocks method that does an unnecessary loop.
Fixes https://github.com/mockito/mockito/issues/1587

PR #1810, opened 2019-10-24
Revert "Fixes #1587 : Remove unnecessary loop from InjectingAnnotationEngine"

### LINKED ISSUE #1587: InjectingAnnotationEngine does an unnecessary loop.
processInjectMocks method of InjectingAnnotationEngine does unnecessary work. It might call injectMocks several times while it is sufficient to call injectMocks once.

    private void processInjectMocks(final Class<?> clazz, final Object testInstance) {
        Class<?> classContext = clazz;
        while (classContext != Object.class) {
            injectMocks(testInstance);
            classContext = classContext.getSuperclass();
        }
    }

Inside, injectMocks loops classes by himself.

    public void injectMocks(final Object testClassInstance) {
        Class<?> clazz = testClassInstance.getClass();
        Set<Field> mockDependentFields = new HashSet<Field>();
        Set<Object> mocks = newMockSafeHashSet();

        while (clazz != Object.class) {
            new InjectMocksScanner(clazz).addTo(mockDependentFields);
            new MockScanner(testClassInstance, clazz).addPreparedMocks(mocks);
            onInjection(testClassInstance, clazz, mockDependentFields, mocks);
            clazz = clazz.getSuperclass();
        }

        new DefaultInjectionEngine().injectMocksOnFields(mockDependentFields, mocks, testClassInstance);
    }

Why not delete the unnecessary loop?

tokuhirom:
Reverts mockito/mockito#1588
Related to #1631 

Here's a pseudo code to describe the issues around this issue.

# Before 2.23.15:	

```
while (clazz != Object.clazz) {
	val mocks = aggregateSpyOrMockFields(clazz.getMockFields());
  val fields = aggregateInjectMocksFields(clazz);
	val ret = tryConstructorInjection(clazz, fields, mock);
  if (! ret.fieldWasInitializedUsingConstructorInjection()) {
    trySetterOrPropertyInjection(clazz, fields, mock);
  }
	clazz = clazz.getSuperclass();
}
```
# After 2.23.15:
```
	val mocks = aggregateSpyOrMockFields(clazz.getMockFields());
  val fields = aggregateInjectMocksFields(clazz);
	val ret = tryConstructorInjection(clazz, fields, mock);
  if (! ret.fieldWasInitializedUsingConstructorInjection()) {
    trySetterOrPropertyInjection(clazz, fields, mock);
  }
```

# Details

Before 2.23.15, if the `@InjectMocks` field has a field that not listed in the constructor argument will fill by setter/property injection at the 2nd loop 😃
It may not the expected behaviors by Mockito authors, but some users' code depends on this behavior.

My suggestion is to revert the [Fixes #1587 : Remove unnecessary loop from InjectingAnnotationEngine by LihMeh · Pull Request #1588 · mockito/mockito · GitHub](https://github.com/mockito/mockito/pull/1588) at this time.

### DIFF
--- a/src/main/java/org/mockito/internal/configuration/InjectingAnnotationEngine.java
+++ b/src/main/java/org/mockito/internal/configuration/InjectingAnnotationEngine.java
@@ -39,7 +39,15 @@ public class InjectingAnnotationEngine implements AnnotationEngine, org.mockito.
      */
     public void process(Class<?> clazz, Object testInstance) {
         processIndependentAnnotations(testInstance.getClass(), testInstance);
-        injectMocks(testInstance);
+        processInjectMocks(testInstance.getClass(), testInstance);
+    }
+
+    private void processInjectMocks(final Class<?> clazz, final Object testInstance) {
+        Class<?> classContext = clazz;
+        while (classContext != Object.class) {
+            injectMocks(testInstance);
+            classContext = classContext.getSuperclass();
+        }
     }
 
     private void processIndependentAnnotations(final Class<?> clazz, final Object testInstance) {
--- a/src/test/java/org/mockito/internal/configuration/InjectingAnnotationEngineTest.java
+++ b/src/test/java/org/mockito/internal/configuration/InjectingAnnotationEngineTest.java
@@ -1,0 +1,51 @@
+package org.mockito.internal.configuration;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.Spy;
+import org.mockito.junit.MockitoJUnitRunner;
+
+class I { }
+
+@RunWith(MockitoJUnitRunner.class)
+public class InjectingAnnotationEngineTest extends I {
+    @InjectMocks
+    Target target;
+    @Mock
+    Foo foo;
+    @Spy
+    Bar bar = new Bar();
+
+    /*
+     If the test case has super classes, the @InjectMocks field has a field that not listed in the constructor argument
+     will fill by setter/property injection .
+
+     https://github.com/mockito/mockito/issues/1631
+    */
+    @Test
+    public void injectMocks() {
+        Assert.assertNotNull(target.bar);
+    }
+
+    public static class Target {
+        private final Foo foo;
+        private Bar bar;
+
+        public Target(Foo foo) {
+            this.foo = foo;
+        }
+
+        public Bar getBar() {
+            return bar;
+        }
+    }
+
+    public static class Foo {
+    }
+
+    public static class Bar {
+    }
+}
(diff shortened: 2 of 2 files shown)


tokuhirom:
Why travis fails?

TimvdLippe:
This needs a rebase. That should fix Travis (presumably)

tokuhirom:
Rebased.

tokuhirom:
@TimvdLippe 
Added test case.

src/test/java/org/mockito/internal/configuration/InjectingAnnotationEngineTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /*
+     If the test case has super classes, the @InjectMocks field has a field that not listed in the constructor argument
+     will fill by setter/property injection .
+
+     https://github.com/mockito/mockito/issues/1631
+    */
+    @Test
+    public void injectMocks() {
+        Assert.assertNotNull(target.bar);
```
TimvdLippe: Can we assert here on `target.getBar()`. Also please add another assertion with `assertEquals(target.foo, foo)`

tokuhirom: Thanks. applied your feedback.

src/main/java/org/mockito/internal/configuration/InjectingAnnotationEngine.java
```diff
@@ -39,7 +39,15 @@
      */
     public void process(Class<?> clazz, Object testInstance) {
         processIndependentAnnotations(testInstance.getClass(), testInstance);
-        injectMocks(testInstance);
+        processInjectMocks(testInstance.getClass(), testInstance);
+    }
+
+    private void processInjectMocks(final Class<?> clazz, final Object testInstance) {
+        Class<?> classContext = clazz;
+        while (classContext != Object.class) {
```
1-alex98, line 47: Why would you iterate over the super classes of the Test class ???

src/test/java/org/mockito/internal/configuration/InjectingAnnotationEngineTest.java
```diff
@@ -0,0 +1,60 @@
...
+    Bar bar = new Bar();
+
+    /*
+     If the test case has super classes, the @InjectMocks field has a field that not listed in the constructor argument
+     will fill by setter/property injection .
+
+     https://github.com/mockito/mockito/issues/1631
+    */
+    @Test
+    public void injectMocks() {
```
1-alex98, line 33: This behaviour does not reflect the docs of the annotation

src/test/java/org/mockito/internal/configuration/InjectingAnnotationEngineTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.Spy;
+import org.mockito.junit.MockitoJUnitRunner;
+
+class I { }
```
1-alex98, line 15: This test only works if I is extended... seems to make little sense

PR #1811, opened 2019-10-25
Apply consistent import ordering with spotless

TimvdLippe:
Spotless allows us to enforce a consistent import order. Configure it
such that static imports are first and non-static imports afterwards.
Within imports, java imports first, then javax, then all others.

This is consistent with the existing CONTRIBUTING.md guidelines, apart
from the fact that static imports are first. (File has been updated). I
would propose that makes more sense, but happy to discuss.

In any case, spotless already found inconsistent ordering within imports
themselves, so that is already providing value for consistency.

I have also updated the CONTRIBUTING guidelines to explain how
`./gradlew spotlessApply` will fix a lot of issues without the
developer having to worry about it. That should hopefully help newcomers
make changes more confidently.

(The only 2 files that are interesting to you are the `build.gradle` and
`.github/CONTRIBUTING.md`. All other file changes were automatically
performed after running `./gradlew spotlessApply` locally.)


.github/CONTRIBUTING.md
```diff
@@ -92,13 +94,13 @@ If you want to check if your code complies to the style guide, you can run:
...
+1. blank line
 1. `import java.*`
-2. `import javax.*`
-3. `import all other imports`
-4. blank line
-5. `import static java.*`
-6. `import static javax.*`
-7. `import static all other imports`
+1. `import javax.*`
+1. `import all other imports`
```
bric3: I think we should also leverage [IntelliJ IDEA's `.editorconfig` extensions](https://www.jetbrains.com/help/idea/configuring-code-style.html#editorconfig) for that.

Issue #1808, opened 2019-10-23
ArgumentMatcher any(Class<T>) doesn't work during stubbing


ideskov:
Hey everybody,

I just stumbled upon a very weird issue with ArgumentMatchers. A bit of background - I'm testing a REST endpoint in a Spring Boot application written in Kotlin and using WebFlux. I have managed to reproduce the issue in a small project and pushed it here -> https://github.com/ideskov/mockito-arg-matcher

Essentially, when I try to do
```Kotlin
Mockito.`when`(mock.foo(any(Bar::class.java))).thenReturn("Mocked done".toMono())
```
I get
```
org.mockito.exceptions.misusing.InvalidUseOfMatchersException: 
Misplaced or misused argument matcher detected here:

-> at ideskov.mockito.MockitoHandlerTest.test mockito argument matchers(MockitoHandlerTest.kt:26)

You cannot use argument matchers outside of verification or stubbing.
Examples of correct usage of argument matchers:
    when(mock.get(anyInt())).thenReturn(null);
    doThrow(new RuntimeException()).when(mock).someVoidMethod(anyObject());
    verify(mock).someMethod(contains("foo"))

This message may appear after an NullPointerException if the last matcher is returning an object 
like any() but the stubbed method signature expect a primitive argument, in this case,
use primitive alternatives.
    when(mock.get(any())); // bad use, will raise NPE
    when(mock.get(anyInt())); // correct usage use

Also, this error might show up because you use argument matchers with methods that cannot be mocked.
Following methods *cannot* be stubbed/verified: final/private/equals()/hashCode().
Mocking methods declared on non-public parent classes is not supported.
```
If I replace the argument with a String and do ```anyString()``` instead of ```any(Class<T>)```, it works just fine.

Not sure if it's a problem with Kotlin/Reactor and have no idea how to even start investigating. Any help would be greatly appreciated.

TimvdLippe:
Could you send us a PR with a failing test case that reproduces your problem in isolation? Then we can debug and see what is going on :smile:

TimvdLippe:
Could you send us a PR with a failing test case that reproduces your problem in isolation? Then we can debug and see what is going on :smile:

ideskov:
Solution discussed in linked PR. Thanks

santunioni:
where is the linked pr?

PR #1812, opened 2019-10-27
Demonstrate problem with ArgumentMatchers and Kotlin (#1808)

ideskov:
An attempt to demonstrate a problem with ArgumentMatchers in Kotlin described in #1808 

As mentioned in the issue, I get a
```
java.lang.IllegalStateException: ArgumentMatchers.any(InputArgument::class.java) must not be null
```
And then an explanation that I'm using argument matching outside of stubbing/verification.

If I replace the argument type with a string and use ```anyString()``` instead of ```any(Class<T>)```, it works fine.

I might as well be missing something, any help would be greatly appreciated.

### DIFF
--- a/subprojects/kotlinTest/src/test/kotlin/org/mockito/kotlin/ArgumentMatchersTest.kt
+++ b/subprojects/kotlinTest/src/test/kotlin/org/mockito/kotlin/ArgumentMatchersTest.kt
@@ -1,0 +1,30 @@
+package org.mockito.kotlin
+
+import org.hamcrest.core.IsEqual
+import org.junit.Assert.assertThat
+import org.junit.Test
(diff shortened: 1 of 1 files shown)


TimvdLippe:
I think `mockito-kotlin` (https://github.com/nhaarman/mockito-kotlin) solves that problem. Does that work for you?

ideskov:
Thanks for the help @TimvdLippe That library worked for this example but I needed a workaround for Spring's ```@MockBean```. Turns out there's a bunch of libraries specific to Kotlin which solve that problem.

I'll now close this PR and the issue linked to it. Thanks

SaliniSasidharan:
Is this solved?

Issue #1097, opened 2017-05-28
Strict mocks support


szczepiq:
## Introduction

**Supporting "strict" mocks will help with test debuggability in certain scenarios.**

This design note explains the rationale and suggested implementation plan to support "strict" mocks. **Please contribute!** If you want to help our community and implement this feature, please read the design spec carefully and let us know if something is unclear. This feature has been asked for many years now and we think that now we are ready to nail it. We already started to make Mockito stricter (#769).

We believe that this design note is pretty awesome and you could get started on implementing this feature in no time. Seriously. Read it, code it and give back a pull request :) If something is unclear, you have feedback about the design or better ideas - leave us a comment.

## Rationale

Mockito mocks are loose by default - methods can be freely invoked on a mock, non-void methods will return default empty value. "Strict" mock works differently, you are not allowed to invoke a method on a strict mock unless this method was previously "configured". This is how traditional mocking frameworks like JMock or EasyMock work by default. Strict mocks offer better debuggability in certain scenarios, however overusing strict mocks leads to overspecified, brittle tests. Strict VS loose mock explanation is further described in [an article on Mockito blog](http://blog.mockito.org/2017/01/clean-tests-produce-clean-code-strict.html).

**Mockito mocks will remain loose by default but we want to offer better support for strict mocks.**

Useful links:

- previous design and discussion in #649
- initial discussion in #303

### New behavior

Strict mocking can be enabled per test method execution and introduces following behavior:
 - unstubbed, non-void method fails immediately with "I am unstubbed!" exception
 - such method is implicitly marked as verified for the verifyNoMoreInteractions() calls
 - unused stubs are detected at the end of test and trigger failure
 - requires to use doReturn API for stubbing mocks (limitation of the syntax, already known gotcha when using [test "spy"](https://static.javadoc.io/org.mockito/mockito-core/2.8.9/org/mockito/Mockito.html#13) or [strict stubs](https://static.javadoc.io/org.mockito/mockito-core/2.8.9/org/mockito/exceptions/misusing/PotentialStubbingProblem.html))

### Options and questions

 - Q: Do we separate behavior for voids / non-voids? A: Mockito API makes clear distinction between stubbing (typically non-void methods) and verification (typically void methods). It is a design choice to drive clarity and understanding of ASK vs TELL style interactions (this design spec does not explain this topic further). Unless we hear more feedback, we want the strict mock to only fail early on non-void methods.
[...]

PR #1814, opened 2019-11-01
Strict mocks support

mockitoguy:
Implemented support for strict mocks (#1097)

The PR is mergeable but I'd like to collect feedback first. What do you think?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,5 +91,6 @@ jobs:
     - name: Build and publish
       run: ./gradlew bintrayUpload githubRelease --scan
       env:
+        MAVEN_CENTRAL_RELEASE: ${{contains(toJSON(github.event.commits.*.message), '[ci maven-central-release]')}}
         GH_WRITE_TOKEN: ${{secrets.GH_WRITE_TOKEN}}
         BINTRAY_API_KEY: ${{secrets.BINTRAY_API_KEY}}
--- a/gradle/shipkit.gradle
+++ b/gradle/shipkit.gradle
@@ -28,7 +28,7 @@ ext.mavenCentralRelease = shouldReleaseToCentral(project)
  * To test this logic, run build with '-i' (info logging) and inspect the build output.
  */
 static boolean shouldReleaseToCentral(project) {
-    boolean centralReleaseByCommit = System.getenv("TRAVIS_COMMIT_MESSAGE")?.contains("[ci maven-central-release]")
+    boolean centralReleaseByCommit = 'true' == System.getenv("MAVEN_CENTRAL_RELEASE")
     boolean centralReleaseByProjectProperty = project.hasProperty("maven-central-release")
     boolean centralRelease = centralReleaseByCommit || centralReleaseByProjectProperty
     def message = """Bintray publish was using following settings:


src/main/java/org/mockito/quality/Strictness.java
```diff
@@ -89,5 +94,49 @@
...
+     * If you are convinced you need {@code STRICT_MOCKS} mocking then read on.
+     * <p>
+     * This mode requires the developer to declare all stubbings for <strong>non-void</strong> methods
+     * invoked in the production code.
+     * Please remember that <strong>void methods</strong> are ignored and you are not required to stub them.
+     * <p>
+     * {@code STRICT_MOCKS} mocking is similar to traditional mocking with EasyMock,
+     * where undeclared invocations would fail early.
+     * {@code STRICT_MOCKS} mocking is useful for users migrating from EasyMock
+     * who are emotionally attached to stricter mocking.
```
TimvdLippe: Let's not use words like `emotionally` here.
```suggestion
     * who prefer the behavior of stricter mocking.
```

src/main/java/org/mockito/quality/Strictness.java
```diff
@@ -89,5 +94,49 @@
...
+     * clean tests and improved debuggability.
+     * {@code STRICT_MOCKS} mocking may increase brittleness of your tests
+     * and force you to declare additional stubbings that are irrelevant in given test method.
+     * This makes tests bigger and hence harder to read and comprehend.
+     * At times, {@code STRICT_MOCKS} mocking can be useful for troubleshooting tests or production code
+     * that is legacy or overcomplicated.
+     * <p>
+     * For discussion about {@code STRICT_MOCKS} mocking see ticket <a href="https://github.com/mockito/mockito/issues/1097">#1097</a>.
+     * <p>
+     * You can enable {@code STRICT_MOCKS} mocking via our JUnit4 support ({@link MockitoJUnit}) or {@link MockitoSession}
```
TimvdLippe: The JUnit4 support is annotated with `@Deprecated` in this PR. If we want to deprecate it, we should not mention it in this javadoc.

mockitoguy:
> I would prefer to tackle the stricts stubs issue before introducing more strict mocks support.

That's fair. What issue you have in mind?

> However, we have seen users running into issues with strict stubs and I fear strict mocks suffer the same fate.

By "issues" you mean bugs or known limitations of the API?

Thank you for reviewing!

src/main/java/org/mockito/quality/Strictness.java
mockitoguy: +1

src/main/java/org/mockito/quality/Strictness.java
mockitoguy: I deprecated only one specific JUnit4 runner subclass, not the entire JUnit4 support.

mockitoguy:
Just run "./gradlew spotlessJavaApply" and it worked :-) Thanks!

TimvdLippe:
> > I would prefer to tackle the stricts stubs issue before introducing more strict mocks support.
> 
> That's fair. What issue you have in mind?
> 
> > However, we have seen users running into issues with strict stubs and I fear strict mocks suffer the same fate.
> 
> By "issues" you mean bugs or known limitations of the API?
> 
> Thank you for reviewing!

I was mostly concerned with the "it reported this invocation was unused" in a test with a lot of setup code. I now realize that strict mocks are the inverse. You have to specify *everything*. So I think my initial concern was ungrounded.

I am okay with shipping this as `@Incubating` to collect user feedback.

mockitoguy:
I'll close for now. Will reopen if we feel it is needed.

andreisilviudragnea:
Why was this abandoned? The tests in our codebase highly benefited from the strict stubbing support and this was the next evolution we were waiting for, so the tests are even easier to understand and reason about.

robertotru:
@mockitoguy hello, 
I hope you are doing well.

I'm wondering... why was this abandoned?

Issue #1817, opened 2019-11-06
MockitoHamcrest#argThat does not allow "super" matchers


spjoe:
IMHO to really follow the hamcrest matchers design philosophy the argThat interface should look like this: `public static <T> T argThat(Matcher<? super T> matcher)`, but unfortunately it is
https://github.com/mockito/mockito/blob/68bc593ebb79efa515a211f8cdff77c7c3746a6a/src/main/java/org/mockito/hamcrest/MockitoHamcrest.java#L60

this change would allow a statement like the following:
`verify(mock).aMethod(argThat(nullValue()))`

I see that the default value is derived from the concrete matcher type, which of course would not work anymore. To circumvent this problem, it would be possible to provide a overload for primitives types.

Or provide a new argThat method, for example argThatSuper which always returns null, but of course would not work for primitive types. Than the existing argThat would keep its behavior.

TimvdLippe:
Rather than making changes to `MockitoHamcrest`, I think our stance is that we are deprecating it. I will send a PR to do that instead. Users should use `ArgumentMatchers.argThat` instead.

Note that https://github.com/mockito/mockito/blob/68bc593ebb79efa515a211f8cdff77c7c3746a6a/src/main/java/org/mockito/ArgumentMatchers.java#L1204-L1207 probably has the same issue? If so, feel free to send a PR to update the definition and add regression tests.

spjoe:
Thanks for the fast response @TimvdLippe 

I am very sorry to hear that hamcrest matcher support may get deprecated :-(

With the hamcrest matcher support in mockito it was very easy to share matcher implantations over multiple usecases.

For example I use the same matcher in:
* mockito verify
* junit assertThat
* awaitility until

For me dropping the support would mean to rewrite (and maintain) a tone of matchers as ArgumentMatcher.

Please do not drop Hamcrest Matcher support.

Best Regards,
spjoe

TimvdLippe:
We are keeping this deprecated, but won't delete it for now.

PR #1819, opened 2019-11-06
Deprecate MockitoHamcrest

TimvdLippe:
This class was used during the migration period from Mockito 1 and
Mockito 2, but is no longer necessary. To be able to remove our
dependency on Hamcrest, we should remove MockitoHamcrest.

In response to #1817

### DIFF
--- a/src/main/java/org/mockito/hamcrest/MockitoHamcrest.java
+++ b/src/main/java/org/mockito/hamcrest/MockitoHamcrest.java
@@ -44,7 +44,10 @@ import org.mockito.internal.hamcrest.HamcrestArgumentMatcher;
  * If you have an idea how to fix the problem, let us know via the mailing list or the issue tracker.
  *
  * @since 2.1.0
+ *
+ * @deprecated Use the regular {@link org.mockito.ArgumentMatchers#argThat(ArgumentMatcher)} instead
  */
+@Deprecated
 public class MockitoHamcrest {
 
     /**
@@ -55,7 +58,9 @@ public class MockitoHamcrest {
      * @param matcher decides whether argument matches
      * @return <code>null</code> or default value for primitive (0, false, etc.)
      * @since 2.1.0
+     * @deprecated Use {@link org.mockito.ArgumentMatchers#argThat(ArgumentMatcher)} instead
      */
+    @Deprecated
     @SuppressWarnings("unchecked")
     public static <T> T argThat(Matcher<T> matcher) {
         reportMatcher(matcher);
@@ -70,7 +75,9 @@ public class MockitoHamcrest {
      *
      * @param matcher decides whether argument matches
      * @return <code>0</code>.
+     * @deprecated Use {@link org.mockito.ArgumentMatchers#charThat(ArgumentMatcher)} instead
      */
+    @Deprecated
     public static char charThat(Matcher<Character> matcher) {
         reportMatcher(matcher);
         return 0;
@@ -84,7 +91,9 @@ public class MockitoHamcrest {
      *
      * @param matcher decides whether argument matches
      * @return <code>false</code>.
+     * @deprecated Use {@link org.mockito.ArgumentMatchers#booleanThat(ArgumentMatcher)} instead
      */
+    @Deprecated
     public static boolean booleanThat(Matcher<Boolean> matcher) {
         reportMatcher(matcher);
         return false;
@@ -98,7 +107,9 @@ public class MockitoHamcrest {
      *
      * @param matcher decides whether argument matches
      * @return <code>0</code>.
+     * @deprecated Use {@link org.mockito.ArgumentMatchers#byteThat(ArgumentMatcher)} instead
      */
+    @Deprecated
     public static byte byteThat(Matcher<Byte> matcher) {
         reportMatcher(matcher);
         return 0;
@@ -112,7 +123,9 @@ public class MockitoHamcrest {
      *
      * @param matcher decides whether argument matches
      * @return <code>0</code>.
+     * @deprecated Use {@link org.mockito.ArgumentMatchers#shortThat(ArgumentMatcher)} instead
      */
+    @Deprecated
     public static short shortThat(Matcher<Short> matcher) {
         reportMatcher(matcher);
         return 0;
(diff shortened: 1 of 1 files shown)


TimvdLippe:
Friendly ping on this PR.

TimvdLippe:
I will submit this PR next week if there are no objections.

TimvdLippe:
There were no objections, so I am merging this.

mockitoguy:
Sorry I'm late here! I suggest we don't deprecate it. The use case (as documented in the Javadoc) is the following: "Hamcrest integration is provided so that users can take advantage of existing Hamcrest matchers". That use case is still valid (perhaps the docs need to be clearer to address #1817).

For code that does not use Hamcrest, users should be following standard, non-Hamcrest argThat() API.

Hope that helps!

TimvdLippe:
> For code that does not use Hamcrest, users should be following standard, non-Hamcrest argThat() API.

The problem is that this project requires Hamcrest as compile-time dependency to be built. This is a problem for users who do not want to use Hamcrest, but still build the source code (as they can't use the pre-built jars)

If we want to keep supporting Hamcrest users with this API, I would suggest a `mockito-hamcrest` artifact that takes care of the integration. That way, those users who do *not* want to use Hamcrest don't rely on it as well.

WDYT?

mansours:
I'm novice to testing because I have to wear many hats (and I avoided getting into it like the plague in the past). Returning to Spring development in my career, and discovering Mockito this year changed my [integration testing] life. I rely heavily on googling stack overflow examples until it become memorized (if ever I do).  

It seems like there is a much larger documented wealth of knowledge on hamcrest on stack overflow (maybe cause its been around a long time), and so its easy to find solutions to basic challenges while someone new to Mockito gets used to it.

For example hasEntry from this SO helped me today https://stackoverflow.com/questions/2580369/using-mockito-how-do-i-match-against-the-key-value-pair-of-a-map

I like TimvdLippe's suggestion because it leverages giving your users the ability to tap into that knowledge/solutions while slimming down the dependencies for the core Mockito code base.

Thanks for all your contributions in this area, and letting us stand on your shoulders.

TimvdLippe:
Thanks for that response @mansours ! I will look into introducing a `mockito-hamcrest` artifact and figure out a way forward. Tomorrow I can revert this PR given the points raised by @mockitoguy and make that work.

mockitoguy:
> This is a problem for users who do not want to use Hamcrest, but still build the source code (as they can't use the pre-built jars)

Can you elaborate this problem? We use "compileOnly" dependency for Hamcrest, just like we do for JUnit4 or OpenTest4j. This should not cause problems for consumers.

TimvdLippe:
Yes we are facing the same issues with JUnit4 and OpenTest4J. JUnit4 is not an issue atm, as we are also using it, but OpenTest4J is currently an issue.

We are not able to include jars, as we require source code to be included and built (for security analyses). Thus, the inclusion of the dependencies during compile time does lead to problems when we want to build Mockito in isolation.

I am currently investigating solutions for JUnit4 and OpenTest4J and how that could be decoupled. I went ahead with this PR, as it was my understanding with Mockito 2 is that we wanted to remove our dependency of Hamcrest. Therefore, I assumed that deprecation was appropriate. Especially as there were no objections posted on this PR for almost 2 weeks, I understood that this was okay.

I will revert the PR for us to revisit, but I would like to come to a decision on our Hamcrest support and whether we want to support it or not at all. When we come to a decision, I would like to update our JavaDoc and wiki on that to clarify that for our users.

mockitoguy:
> Especially as there were no objections posted on this PR for almost 2 weeks, I understood that this was okay

Of course! Thank you for waiting 2 weeks. Don't block on us.

> Yes we are facing the same issues with JUnit4 and OpenTest4J. JUnit4 is not an issue atm, as we are also using it, but OpenTest4J is currently an issue.

Can you elaborate the problem? Is it a Google mono-repo use case? (I don't object reworking the artifacts - I want understanding for me and others :))

TimvdLippe:
> Can you elaborate the problem? Is it a Google mono-repo use case? (I don't object reworking the artifacts - I want understanding for me and others :))

This is for every repository that includes third_party code that we need to have the original source code for security analyses. That includes the mono repository, as well as some other repositories that we have.

Since the Hamcrest API is not type-safe (as it requires casting with `Object` vs `T` as argument type), we would like to move away from unsafe type casts. The `ArgumentMatcher` interface does not inherit that problem, but the older Hamcrest API does.

By having the Hamcrest-compatible API in the Mockito artifact, users can still use Mockito in combination with the Hamcrest API.

I would personally say that, while the Hamcrest API was originally useful for defining reusable matchers, the lack of type-safety does more harm than good. I have received good responses when I finished the migration of the `ArgumentMatcher` to be typed and it already caught bugs in our tests. I think we should promote the type-safe `ArgumentMatcher` and thus deprecate the old API, but happy to disagree on that part and leave Hamcrest compatibility in for example a separate artifact.

mockitoguy:
> that we need to have the original source code for security analyses

So the problem is that in addition to Mockito source, you have to pull in Hamcrest source for the analysis? (I still don't feel I fully understand this use case).

> By having the Hamcrest-compatible API in the Mockito artifact, users can still use Mockito in combination with the Hamcrest API.

That's a fair argument.

> leave Hamcrest compatibility in for example a separate artifact.

To keep backwards compatibility we would need "mockito-core" -> "mockito-hamcrest". Are you thinking of reversing the dependency in future major version?

TimvdLippe:
Revert of this PR is in https://github.com/mockito/mockito/commit/ad2f3526a347f37dd00c53ee664c5cd9f3738814

TimvdLippe:
> So the problem is that in addition to Mockito source, you have to pull in Hamcrest source for the analysis? (I still don't feel I fully understand this use case).

Correct. Since jars can be built with any arbitrary code, we prefer to not check in a jar, as we cant be certain that it was built from the original source code. That's why we check in the original source code and build that instead.

> To keep backwards compatibility we would need "mockito-core" -> "mockito-hamcrest". Are you thinking of reversing the dependency in future major version?

Yes, but since Maven does not allow a cyclic dependency we have to figure out what the possibilities are.

PR #1820, opened 2019-11-06
Indent with spaces using spotless

TimvdLippe:
Discovered in #1818 that (if new contributors are unaware of our indentation with 4 spaces), spotless did not fix the formatting. `indentWithSpaces(4)` does just that and it actually found other formatting issues.


TimvdLippe:
Yes that is correct. I think at that point you need a formatter like `google-java-formatter` or `eclipse-formatter` to fix these kind of "aesthetics".

PR #1823, opened 2019-11-13
fix a word mistake

scadgek:


bric3:
Thank you for the fix ;)

PR #1832, opened 2019-11-27
Add matchers for incompatible type matchers

TimvdLippe:
We discovered that users run into issues with using the wrong Mockito
matcher for arguments. Examples include `any(Integer.class)` instead of
`anyInt()` and `anyInt()` instead of `anyFloat()`. Users then run into
cryptic run-time errors that are difficult to understand.

These ErrorProne checkers make these a compile warning, to warn the user
before hand. They also provide the appropriate fixes that can be
directly applied.

### DIFF
--- a/subprojects/errorprone/src/main/java/org/mockito/errorprone/bugpatterns/AbstractMockitoAnyForPrimitiveType.java
+++ b/subprojects/errorprone/src/main/java/org/mockito/errorprone/bugpatterns/AbstractMockitoAnyForPrimitiveType.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright (c) 2017 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.errorprone.bugpatterns;
+
+import static com.google.errorprone.matchers.Description.NO_MATCH;
+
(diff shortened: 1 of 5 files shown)


ChristianSchwarz:
@TimvdLippe 
Why not fix the implementation of `any(Class<T> type)` ? `any(..)` can check if the passed type is a primitive and then delegate to the required matcher.

Sample:

```
public static <T> T any(Class<T> type) {
        if (type==int.class || type==Integer.class){ 
            return anyInt()
        }

        reportMatcher(new InstanceOf.VarArgAware(type, "<any " + type.getCanonicalName() + ">"));
        return defaultValue(type);
}
```

TimvdLippe:
@ChristianSchwarz That would lead to a lot of confusing code. E.g. half of the matchers would be `anyInt` and other half would be `any(Integer.class)`. I would also be hesitant to introducing an API that takes the primitives as classes.

For now, we can keep this checker and we can revisit your suggestion in the future. Do you mind filing a bug?

TimvdLippe:
I am merging this for now, as it upstreams a checker that prevents real issues on runtime. We can change the Mockito API, but I would rather prevent users from running into runtime exceptions when we can.

Issue #1827, opened 2019-11-24
Cannot mock this class: class java.io.InputStream with Java 13


adamretter:
## Versions
* Mockito: 3.1.0
* JDK: 13.0.1+9
* OS: macOS 10.13
* XCode: 10.1-19

## Code causing the error
```
InputStream in = mock(InputStream.class);
```

## Error:
```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.015 s <<< FAILURE! - in uk.gov.nationalarchives.droid.core.interfaces.archive.ZipEntryRequestFactoryTest
[ERROR] testNewZipRequest(uk.gov.nationalarchives.droid.core.interfaces.archive.ZipEntryRequestFactoryTest)  Time elapsed: 0.005 s  <<< ERROR!
org.mockito.exceptions.base.MockitoException: 
Mockito cannot mock this class: class java.io.InputStream.
If you're not sure why you're getting this error, please report to the mailing list.
Java               : 13
JVM vendor name    : Oracle Corporation
JVM vendor version : 13.0.1+9
JVM name           : OpenJDK 64-Bit Server VM
JVM version        : 13.0.1+9
JVM info           : mixed mode, sharing
OS name            : Mac OS X
OS version         : 10.13.6
You are seeing this disclaimer because Mockito is configured to create inlined mocks.
You can learn about inline mocks and their limitations under item #39 of the Mockito class javadoc.
Underlying exception : org.mockito.exceptions.base.MockitoException: Could not modify all classes [class java.lang.Object, class java.io.InputStream, interface java.lang.AutoCloseable, interface java.io.Closeable]
	at uk.gov.nationalarchives.droid.core.interfaces.archive.ZipEntryRequestFactoryTest.testNewZipRequest(ZipEntryRequestFactoryTest.java:70)
Caused by: org.mockito.exceptions.base.MockitoException: Could not modify all classes [class java.lang.Object, class java.io.InputStream, interface java.lang.AutoCloseable, interface java.io.Closeable]
	at uk.gov.nationalarchives.droid.core.interfaces.archive.ZipEntryRequestFactoryTest.testNewZipRequest(ZipEntryRequestFactoryTest.java:70)
Caused by: java.lang.IllegalStateException: 
Byte Buddy could not instrument all classes within the mock's type hierarchy
This problem should never occur for javac-compiled classes. This problem has been observed for classes that are:
 - Compiled by older versions of scalac
[...]

Issue #1734, opened 2019-06-26
Cannot mock wrapper types, String.class or Class.class


moleksyuk:
Hi, 
Mockito can't not mock `String.class` even with `org.mockito:mockito-inline:2.28.2`, but mocks custom final classes successfully.

Version:
JUnit: `org.junit.jupiter:junit-jupiter:5.4.2`
Mockito-core: `org.mockito:mockito-core:2.28.2`
Mockito-inline: `org.mockito:mockito-inline:2.28.2`

**Steps to reproduce:**
1. Download example project and unzip it: [mockito.zip](https://github.com/mockito/mockito/files/3329584/mockito.zip)
2. Go to project directory and execute such command:
```bash
./gradlew clean build
```

**Actual result:**
```


Cannot mock/spy class java.lang.String
Mockito cannot mock/spy because :
 - Cannot mock wrapper types, String.class or Class.class
org.mockito.exceptions.base.MockitoException: 
Cannot mock/spy class java.lang.String
Mockito cannot mock/spy because :
 - Cannot mock wrapper types, String.class or Class.class
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:628)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.invoke(ExecutableInvoker.java:117)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeTestMethod$7(TestMethodTestDescriptor.java:184)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.invokeTestMethod(TestMethodTestDescriptor.java:180)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:127)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:68)
[...]

PR #1833, opened 2019-11-28
Add annotation to mark a type as DoNotMock

TimvdLippe:
Mocking types that users not own [1] or are severely complicating test
logic [2] leads to brittle or wrong tests. In particular, the
StackOverflow answer is wrong, as the contract of java.util.Map is
violated. When a new key is added to the Map, the stubbed return would be wrong.
In Google we have used the DoNotMock annotation via ErrorProne [3]
to annotate these types, as well as an internal list of types that can't
be mocked (this includes several java.util types). We are using a custom
Mockmaker to enforce this on run-time.

Based on our successful experience with DoNotMock (we have seen a large
reduction in bad/broken tests for types involved), we are proposing to
open source this into Mockito itself.

The DoNotMock annotation can be added to any type, e.g. classes and
interfaces. If, in the type hierarchy of the class-to-be-mocked, there
is a type that is annotated with DoNotMock, Mockito will throw a
DoNotMockException.

This would help preventing issues such as #1827 and #1734 which is
in-line with the guidance on our wiki [1]. A follow-up change would
allow us to define external types (like the java.util types) that can't
be mocked. (We can't add the annotation to the types, as they live in the
JDK instead.)

[1]: https://github.com/mockito/mockito/wiki/How-to-write-good-tests#dont-mock-a-type-you-dont-own
[2]: https://stackoverflow.com/a/15820143
[3]: https://errorprone.info/api/latest/com/google/errorprone/annotations/DoNotMock.html


mockitoguy:
Great submission, thank you!

> we are proposing to open source this into Mockito itself.

Fantastic idea. THANK you!

> We are using a custom Mockmaker to enforce this on run-time.

Wow! Interesting use of this API. Can you enumerate other interesting use cases Google has extended Mockito behavior via our public API? I'd love us to offer first class support for such features.

> A follow-up change would
allow us to define external types (like the java.util types) that can't
be mocked.

+1

mockitoguy:
Request: can we add a new method to override this just in case? Here's a use case: let's say someone incorrectly sets DoNotMock annotation on a type and ships the code in a library. The consumers of the library don't have a good way to remove DoNotMock annotation if they don't own the library code. Thoughts?

Example API:

```java
mock(Foo.class, withSettings().overrideDoNotMock())
```

src/main/java/org/mockito/internal/MockitoCore.java
```diff
@@ -60,11 +63,32 @@ public boolean isTypeMockable(Class<?> typeToMock) {
...
+        while (mockTypeToCheck != null && mockTypeToCheck != Object.class) {
+            unmockableTypes.add(mockTypeToCheck);
+            Class<?>[] interfaces = mockTypeToCheck.getInterfaces();
+            if (interfaces != null) {
+                unmockableTypes.addAll(Arrays.asList(interfaces));
+            }
+            mockTypeToCheck = mockTypeToCheck.getSuperclass();
+        }
+
+        if (unmockableTypes.stream().anyMatch(clazz -> clazz.isAnnotationPresent(DoNotMock.class))) {
```
mockitoguy: Do we want to honor this specific annotation (Mockito API) or _any_ annotation with name "DoNotMock".

The latter seems more useful. For example: a library owner can mark types with any "DoNotMock" annotation without adding compile dependency on mockito-core.

src/test/java/org/mockitousage/annotation/DoNotMockTest.java
```diff
@@ -0,0 +1,73 @@
...
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Mockito.mock;
+
+import org.junit.Test;
+import org.mockito.DoNotMock;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+import org.mockito.exceptions.misusing.DoNotMockException;
+
+public class DoNotMockTest {
```
mockitoguy: Great test coverage.

TimvdLippe:
> Wow! Interesting use of this API. Can you enumerate other interesting use cases Google has extended Mockito behavior via our public API? I'd love us to offer first class support for such features.

Yes I am auditing/discovering usages of Mockito API and reviewing them whether they are worth keeping (and open sourcing) or should be changed. This is one of many :smile: 

> Request: can we add a new method to override this just in case? Here's a use case: let's say someone incorrectly sets DoNotMock annotation on a type and ships the code in a library. The consumers of the library don't have a good way to remove DoNotMock annotation if they don't own the library code. Thoughts?

Yes we already have the concept of "grandfathered" mock violations. E.g. all existing mocks of a type that you want to add `@DoNotMock` to will not immediately break. This allows us to put a hold on new violations while users can (over time) remove violations. I am still trying to come up with a scalable solution for implementing that logic, which will solve your use case as well.

src/main/java/org/mockito/internal/MockitoCore.java
TimvdLippe: I would like to keep this for now and open it up later to match on name. If we would implement that now, I would break all users of ErrorProne that rely on `@DoNotMock` but have grandfathered mock violations.

If we can gather feedback first, we can see adoption patterns and change it to match on the name later. WDYT?

mockitoguy: How about a happy medium so that we don't need to revisit this in the future (and cause breaking changes :) - we can match for "org.mockito.DoNotMock" by String. This should be backwards compatible for G. and other companies / projects. Thoughts?

src/main/java/org/mockito/internal/MockitoCore.java
```diff
@@ -60,11 +63,32 @@ public boolean isTypeMockable(Class<?> typeToMock) {
...
         return mock;
     }
 
+    private void checkDoNotMockAnnotation(Class<?> typeToMock,
+        MockCreationSettings<?> creationSettings) {
+        ArrayList<Class<?>> unmockableTypes = new ArrayList<>(
+            creationSettings.getExtraInterfaces());
+
+        Class<?> mockTypeToCheck = typeToMock;
+        while (mockTypeToCheck != null && mockTypeToCheck != Object.class) {
```
mockitoguy: I played with the code and noticed that "mockTypeToCheck != null" part of this statement is either: a) not needed or b) not covered with tests.

Do we need this part?

raphw:
Really interesting feature, also with regards to returning mocks from methods, I think. But very much +1 on the override functionality. There is always a "but" that might need consideration.

src/main/java/org/mockito/internal/MockitoCore.java
TimvdLippe: It is covered by mocking an interface that directly has the `DoNotMock` annotation. I added a test to cover this case.

src/main/java/org/mockito/internal/MockitoCore.java
TimvdLippe: Fixed

TimvdLippe:
It is now possible to provide a custom `DoNotMockEnforcer` where you have full control over what types can be mocked. See the `extTest` test for an example.

Given the positive response, I will now spend some time integrating the changes in Google and see if it all builds. Given that is all positive and you have no more review comments, I will merge this PR.

mockitoguy:
> It is now possible to provide a custom DoNotMockEnforcer

Curious, are you going to use this new plugin at G? If we don't have a user, I suggest we avoid complexity (it can be added later if needed).

Thank you for implementing name matching for the annotation!!!

TimvdLippe:
> > It is now possible to provide a custom DoNotMockEnforcer
> 
> Curious, are you going to use this new plugin at G? If we don't have a user, I suggest we avoid complexity (it can be added later if needed).

Yes we will immediately use it. We have some additional classes that we are annotating as well as our "grandfathering violations" system.

Note that there is no other nice way of overriding the method as requested in https://github.com/mockito/mockito/pull/1833#issuecomment-559627262 (a method call per mock creation would mean that we would need to change every existing mock in our codebase to include it, which is very cumbersome)
> Thank you for implementing name matching for the annotation!!!

Glad you like it :smile:

src/main/java/org/mockito/internal/MockitoCore.java
```diff
@@ -60,11 +69,36 @@ public boolean isTypeMockable(Class<?> typeToMock) {
...
         return mock;
     }
 
+    private void checkDoNotMockAnnotation(Class<?> typeToMock, MockCreationSettings<?> creationSettings) {
+        checkDoNotMockAnnotationForType(typeToMock);
+        creationSettings.getExtraInterfaces().forEach(MockitoCore::checkDoNotMockAnnotationForType);
+    }
+
+    private static void checkDoNotMockAnnotationForType(Class<?> type) {
+        if (type == null) {
```
mockitoguy: I suggest to drop this conditional, type cannot be null (unless I'm wrong :-)

src/main/java/org/mockito/internal/MockitoCore.java
```diff
@@ -50,6 +56,9 @@
 @SuppressWarnings("unchecked")
 public class MockitoCore {
 
+    private static final DoNotMockEnforcer DO_NOT_MOCK_ENFORCER = Plugins.getDoNotMockEnforcer();
+    private static final Set<Class<?>> SAFE_DONOTMOCK_ENFORCED_CLASSES = Collections.synchronizedSet(new HashSet<>());
```
mockitoguy: I suggest "MOCKABLE_CLASSES"

src/main/java/org/mockito/internal/MockitoCore.java
```diff
@@ -60,11 +69,36 @@ public boolean isTypeMockable(Class<?> typeToMock) {
...
+        }
+
+        DO_NOT_MOCK_ENFORCER.checkTypeForDoNotMockViolation(type).ifPresent(message -> {
+            throw new DoNotMockException(message);
+        });
+
+        checkDoNotMockAnnotationForType(type.getSuperclass());
+        Arrays.stream(type.getInterfaces()).forEach(MockitoCore::checkDoNotMockAnnotationForType);
+
+        SAFE_DONOTMOCK_ENFORCED_CLASSES.add(type);
```
mockitoguy: I feel uneasy about this perf tweak. Do we need it? Does it really make things faster? Static state that keeps growing with every mocked type may bite us later.

Did you find this perf tweak useful at G?

TimvdLippe: I am pretty sure this is optimization is necessary as we can have large type hierarchies. I have asked internally for more background, but I expect this optimization to be necessary to not have a large performance impact for existing applications.

src/main/java/org/mockito/internal/MockitoCore.java
TimvdLippe: This will be `null` for `.getSuperClass()` of interfaces and `Object.class`. The tests will fail if we remove it.

src/main/java/org/mockito/plugins/DoNotMockEnforcer.java
```diff
@@ -0,0 +1,25 @@
...
+     * If this type is allowed to be mocked. Return an empty optional if the enforcer allows
+     * this type to be mocked. Return a message if there is a reason this type can not be mocked.
+     *
+     * Note that Mockito performs traversal of the type hierarchy. Implementations of this class
+     * should therefore not perform type traversal themselves but rely on Mockito.
+     *
+     * @param type The type to check
+     * @return Optional message if this type can not be mocked, or an empty optional if type can be mocked
+     */
+    Optional<String> checkTypeForDoNotMockViolation(Class<?> type);
```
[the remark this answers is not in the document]
TimvdLippe: I felt like this was the nicer way of handling state, since we have to stop type traversal. I will mark it as Incubating and might change later, depending on how well it integrates with our internal system.

src/main/java/org/mockito/internal/configuration/DefaultDoNotMockEnforcer.java
```diff
@@ -0,0 +1,28 @@
...
+
+public class DefaultDoNotMockEnforcer implements DoNotMockEnforcer {
+
+    @Override
+    public Optional<String> checkTypeForDoNotMockViolation(Class<?> type) {
+        return Arrays.stream(type.getAnnotations()).filter(
+            annotation -> annotation.annotationType().getName().endsWith("org.mockito.DoNotMock"))
+            .findFirst()
+            .map(annotation -> {
+                String exceptionMessage = type + " is annotated with @DoNoMock and can't be mocked.";
```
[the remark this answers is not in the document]
TimvdLippe: Yes that's better :)

src/main/java/org/mockito/internal/MockitoCore.java
TimvdLippe: :+1:

src/main/java/org/mockito/internal/MockitoCore.java
mockitoguy: Adding perf tweaks "on hunch", without evidence is not good for the codebase (growing complexity :-). I don't mind this change but I'd love to get your assurance that this is useful.

You have the approval so ship this change at will!

TimvdLippe: I will double check before merging. I haven't had time during the Christmas vacation to do this work. I probably get to this late next week.

TimvdLippe:
I have done an initial pass at integrating this internally with good results! We are doing additional verification to confirm everything is working as expected, but thus far it is working out.

grimreaper:
Hey, this is a fantastic feature and I'd love for this to make it in. Can I gently nudge the and reviewers to finish this up?

TimvdLippe:
I still have to integrate this in the Google monorepository and figure out whether it is working correctly. Sadly I haven't had the time to do so.

TimvdLippe:
#2026 is another instance where we could benefit from `DoNotMock` to prevent users from running into problematic situations.

TimvdLippe:
There are a fair few issues in our backlog about mocking classes like `File`, `Method` and `Map`, all of which cause undefined behavior in Mockito internals. Therefore, I think it is worth reviving this PR and see if we can land this to avoid people running into this problem. It would be a breaking change though, so we probably need an incubation timeline to get people to test it out.

src/main/java/org/mockito/internal/MockitoCore.java
bric3: Maybe leave this as a comment

src/main/java/org/mockito/DoNotMock.java
```diff
@@ -0,0 +1,36 @@
...
+ * <p>When marking a type {@code @DoNotMock}, you should always point to alternative testing
+ * solutions such as standard fakes or other testing utilities.
+ *
+ * Mockito enforces {@code @DoNotMock} with the {@link org.mockito.plugins.DoNotMockEnforcer}.
+ *
+ * If you want to use a custom {@code @DoNotMock} annotation, the {@link org.mockito.plugins.DoNotMockEnforcer}
+ * will match on annotations with a type ending in "org.mockito.DoNotMock". You can thus place
+ * your custom annotation in {@code com.my.package.org.mockito.DoNotMock} and Mockito will enforce
+ * that types annotated by {@code @com.my.package.org.mockito.DoNotMock} can not be mocked.
+ */
```
bric3, line 25: It would be nice to add a simple code snippet with the use of the reason. 

Since annotation are likely to be how user discover this feature, it may makes sense to me to also describe how to make non-mockable the non-annotatable types like `File`.

src/main/java/org/mockito/internal/configuration/DefaultDoNotMockEnforcer.java
```diff
@@ -0,0 +1,30 @@
...
+import org.mockito.plugins.DoNotMockEnforcer;
+
+public class DefaultDoNotMockEnforcer implements DoNotMockEnforcer {
+
+    @Override
+    public String checkTypeForDoNotMockViolation(Class<?> type) {
+        for (Annotation annotation : type.getAnnotations()) {
+            if (annotation.annotationType().getName().endsWith("org.mockito.DoNotMock")) {
+                String exceptionMessage =
+                    type + " is annotated with org.mockito.@DoNoMock and can't be mocked.";
```
bric3, line 19: Shouldn't it be `@org.mockito.DoNotMock` ? Maybe adding the actual annotation might be useful as well.

src/main/java/org/mockito/DoNotMock.java
```diff
@@ -0,0 +1,36 @@
...
+@Target({TYPE})
+@Retention(RUNTIME)
+@Documented
+public @interface DoNotMock {
+    /**
+     * The reason why the annotated type should not be mocked.
+     *
+     * <p>This should suggest alternative APIs to use for testing objects of this type.
+     */
+    String value() default "Create a real instance instead.";
```
bric3, line 35: Maybe rename this as "reason", does an "alternative" field could make sense ?

bric3 (approved):
This is a nice feature in my opinion. And it can really help improve the quality of large code base.

About third party types, could it make sense to have a programmatic way to fill the non-mockable types in the default plugin implementation.

src/main/java/org/mockito/DoNotMock.java
TimvdLippe: If we used `value`, then we can have `@DoNotMock("Special reason")` rather than `@DoNotMock(reason = "Special reason")`. Still, I think `reason` is clearer and also is better to read when reading the whole annotation usage. 👍

src/main/java/org/mockito/internal/configuration/DefaultDoNotMockEnforcer.java
TimvdLippe: I think this is the official notation, but I have changed it to yours as that is probably more intuitive.

src/main/java/org/mockito/DoNotMock.java
TimvdLippe: I have added an example, good suggestion 👍 

I think we can ship this without adding classes like `File` in Mockito. Then once we know the enforcer is working as expected, we can make the switch in the next major version of Mockito to by default disallow mocking these classes.

src/main/java/org/mockito/internal/MockitoCore.java
TimvdLippe: Done

TimvdLippe:
Thanks everyone for your input and feedback! I will be publishing Mockito 4.1.0 that includes this feature 🎉

PR #1849, opened 2019-12-17
Performance optimization by using Method.getParameterCount() where possible

### LINKED ISSUE #1848: Use Method.getParameterCount() where possible
Hi,

as of Mockito 3.x and its Java 8 requirement we could make use of `Method.getParameterCount()` to avoid unnecessary cloning caused by `Method.getParameterTypes()`.

I will shortly open a PR with the proposed changes.

Cheers,
Christoph

dreis2211:
Hi,

as described in #1848 we could make use of `Method.getParameterCount()` in order to save some overhead from unnecessary cloning caused by `Method.getParameterTypes()`.

As this is my first contribution to Mockito, let me know if I can do something to help.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
+++ b/src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
@@ -40,8 +40,7 @@ public class ConstructorInstantiator implements Instantiator {
         List<Constructor<?>> matchingConstructors = new LinkedList<Constructor<?>>();
         try {
             for (Constructor<?> constructor : cls.getDeclaredConstructors()) {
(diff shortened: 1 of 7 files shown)


TimvdLippe:
After importing this change into Google, I realized that this change is breaking all downstream Android applications. Should we revert this change similarly to #1845 (see full postmortem at https://github.com/mockito/mockito/wiki/Android-Java-8-%60java.time.Duration%60-postmortem ), given that it does not change our public API?

dreis2211:
That's very unfortunate.

mockitoguy:
> Should we revert this change similarly to #1845

Absolutely.

PR #1850, opened 2019-12-22
Fixes  #522 : Basic test suite for OSGi framework

### LINKED ISSUE #522: Provide OSGi test project to validate OSGi headers automatically.
As stated in #82, we should provide a project for OSGi to validate correct exports.

theogimonde:
This is a proposal for a basic test project to ensure that the Mockito artifacts can be consumed from an OSGi environment. For example, reverting the fix for #678 is caught with the following exception:

```
java.lang.IllegalStateException: Failed to install bundle: mockito-core-3.2.3.jar
	at org.mockito.osgitest.OsgiTest.setUp(OsgiTest.java:55)
	... (snipped)
Caused by: org.osgi.framework.BundleException: Error occurred installing a bundle.
	at org.eclipse.osgi.storage.Storage.install(Storage.java:739)
	at org.eclipse.osgi.internal.framework.BundleContextImpl.installBundle(BundleContextImpl.java:187)
	at org.eclipse.osgi.internal.framework.BundleContextImpl.installBundle(BundleContextImpl.java:179)
	at org.mockito.osgitest.OsgiTest.setUp(OsgiTest.java:53)
	... 40 more
Caused by: java.lang.IllegalArgumentException: invalid range "[1.6.0": invalid format
	at org.osgi.framework.VersionRange.<init>(VersionRange.java:173)
	at org.eclipse.osgi.container.builders.OSGiManifestBuilderFactory.addPackageImports(OSGiManifestBuilderFactory.java:373)
	at org.eclipse.osgi.container.builders.OSGiManifestBuilderFactory.getPackageImports(OSGiManifestBuilderFactory.java:352)
	at org.eclipse.osgi.container.builders.OSGiManifestBuilderFactory.createBuilder(OSGiManifestBuilderFactory.java:111)
	at org.eclipse.osgi.container.builders.OSGiManifestBuilderFactory.createBuilder(OSGiManifestBuilderFactory.java:78)
	at org.eclipse.osgi.storage.Storage.getBuilder(Storage.java:784)
	at org.eclipse.osgi.storage.Storage.getBuilder(Storage.java:766)
	at org.eclipse.osgi.storage.Storage.install(Storage.java:706)
	... 43 more
Caused by: java.util.NoSuchElementException
	at java.util.StringTokenizer.nextToken(StringTokenizer.java:349)
	at org.osgi.framework.VersionRange.<init>(VersionRange.java:157)
	... 50 more
```

The OSGi test project consists of three source sets:

1. A regular JUnit4 `test` source set with a single class OsgiTest. Starts an OSGi framework, installs mockito-core, byte-buddy and objenesis bundles as well as two test bundles, and executes some simple Mockito test cases inside the OSGi framework.
2. A source set `testBundle` that's built to an OSGi bundle using the `Bundle` task from the Bnd Gradle Plugin. Contains the test classes that run the Mockito test cases and thus depends on the Mockito bundle.
3. A source set `otherBundle` that's also built to an OSGi bundle but with no other dependencies. Used for testing mocking of classes from other bundles.

Some further technical details:

- The `test` source set only has compile dependencies on the OSGi APIs and JUnit4, i.e. not on Mockito. The OSGi framework is created using the standard Framework API. At runtime Equinox is added as a dependency to supply an actual framework implementation, but the intention is that any framework implementation could be used.
- The test cases are run by loading classes from the test bundle that all implement `Runnable`. This is to provide a simple bridge using a common JDK class. The reason for setting up such a bridge is that I want to drive the tests using a test runner that Gradle supports, for example to get good test reports, but the test bundles don't know anything about JUnit4 inside the OSGi framework. I didn't want to explore adding an OSGified version of JUnit into the mix at this point since that would also entail solving test discovery inside an OSGi framework.
- The OSGi bundles are built using the [Bnd Gradle Plugin](https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd.gradle/README.md#gradle-plugin-for-workspace-builds) instead of the deprecated `osgi` plugin in Gradle. The reason is both future proofing, to not impede a migration to Gradle 6.0 in this new sub project, and that I believe it better mirrors how other projects would consume the Mockito artifacts. Further it paves the way for replacing the usage of the deprecated plugin for the creating the OSGi manifests in the root project.

Since this is my first contribution I'll be happy to fix anything that goes against the grains.

### DIFF
--- a/subprojects/osgi-test/src/test/java/org/mockito/osgitest/OsgiTest.java
+++ b/subprojects/osgi-test/src/test/java/org/mockito/osgitest/OsgiTest.java
@@ -4,7 +4,9 @@
  */
 package org.mockito.osgitest;
 
-import org.junit.*;
+import org.junit.runner.RunWith;
+import org.junit.runners.Suite;
+import org.junit.runners.model.RunnerBuilder;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.BundleException;
@@ -20,14 +22,16 @@ import java.util.*;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import org.junit.AfterClass;
 
 import static org.junit.Assert.fail;
 
-public class OsgiTest {
+@RunWith(OsgiTest.class)
+public class OsgiTest extends Suite {
 
     private static final FrameworkFactory frameworkFactory = ServiceLoader.load(FrameworkFactory.class).iterator().next();
     private static final String STORAGE_TEMPDIR_NAME = "osgi-test-storage";
-    private static final List<String> EXTRA_SYSTEMPACKAGES = Arrays.asList("sun.misc", "sun.reflect");
+    private static final List<String> EXTRA_SYSTEMPACKAGES = Arrays.asList("org.junit", "sun.misc", "sun.reflect");
 
     private static final List<Path> TEST_RUNTIME_BUNDLES = splitPaths(System.getProperty("testRuntimeBundles"));
     private static final String TEST_BUNDLE_SYMBOLIC_NAME = "testBundle";
@@ -38,8 +42,11 @@ public class OsgiTest {
     private static Framework framework;
     private static Bundle testBundle;
 
-    @BeforeClass
-    public static void setUp() throws IOException, BundleException {
+    public OsgiTest(Class<?> osgiTestClass, RunnerBuilder builder) throws Exception {
+        super(builder, osgiTestClass, setUpClasses());
+    }
+
+    private static Class<?>[] setUpClasses() throws Exception {
         frameworkStorage = Files.createTempDirectory(STORAGE_TEMPDIR_NAME);
         Map<String, String> configuration = new HashMap<>();
         configuration.put(Constants.FRAMEWORK_STORAGE, frameworkStorage.toString());
@@ -62,10 +69,25 @@ public class OsgiTest {
             fail("Test bundle not found.");
         }
         framework.start();
+        try {
+            // Manual start to get a better exception if the bundle cannot be resolved
+            testBundle.start();
+        } catch (BundleException e) {
+            throw new IllegalStateException("Failed to start test bundle.", e);
+        }
+        return getTestClasses();
+    }
+
+    private static Class<?>[] getTestClasses() throws Exception {
+        return new Class<?>[] {
+            loadTestClass("SimpleMockTest"),
+            loadTestClass("MockNonPublicClassFailsTest"),
+            loadTestClass("MockClassInOtherBundleTest")
+        };
     }
 
     @AfterClass
-    public static void tearDown() throws IOException, BundleException, InterruptedException {
+    public static void tearDown() throws Exception {
         try {
             if (framework != null) {
                 framework.stop();
@@ -78,23 +100,8 @@ public class OsgiTest {
         }
     }
 
-    @Test
-    public void testSimpleMock() throws BundleException, InterruptedException, ClassNotFoundException, IllegalAccessException, InstantiationException, IOException {
-        createTestRunnable("SimpleMockTest").run();
-    }
-
-    @Test
-    public void testMockNonPublicClassFails() throws BundleException, InterruptedException, ClassNotFoundException, IllegalAccessException, InstantiationException, IOException {
-        createTestRunnable("MockNonPublicClassFailsTest").run();
-    }
-
-    @Test
-    public void testMockClassInOtherBundle() throws BundleException, InterruptedException, ClassNotFoundException, IllegalAccessException, InstantiationException, IOException {
-        createTestRunnable("MockClassInOtherBundleTest").run();
-    }
-
-    private Runnable createTestRunnable(String className) throws InstantiationException, IllegalAccessException, ClassNotFoundException {
-        return (Runnable) testBundle.loadClass("org.mockito.osgitest.testbundle." + className).newInstance();
+    private static Class<?> loadTestClass(String className) throws Exception {
+        return testBundle.loadClass("org.mockito.osgitest.testbundle." + className);
     }
 
     private static List<Path> splitPaths(String paths) {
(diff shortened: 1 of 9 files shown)


subprojects/osgi-test/src/test/java/org/mockito/osgitest/OsgiTest.java
```diff
@@ -0,0 +1,122 @@
...
+            }
+        } finally {
+            if (frameworkStorage != null) {
+                deleteRecursively(frameworkStorage);
+            }
+        }
+    }
+
+    @Test
+    public void testSimpleMock() throws BundleException, InterruptedException, ClassNotFoundException, IllegalAccessException, InstantiationException, IOException {
```
TimvdLippe: Nit: let's make this `throws Exception` to be more concise.

subprojects/osgi-test/src/testBundle/java/org/mockito/osgitest/testbundle/MockClassInOtherBundleTest.java
```diff
@@ -0,0 +1,22 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class MockClassInOtherBundleTest implements Runnable {
+
+    @Override
+    public void run() {
+        Methods methods = mock(Methods.class);
+        when(methods.intReturningMethod()).thenReturn(42);
+        if (methods.intReturningMethod() != 42) {
```
TimvdLippe: Nit: use `assertThat` from AssertJ or `assertThrows` from JUnit 5 to verify the exception. That makes the test clearer and provides better error messages when the test fails.

subprojects/osgi-test/src/testBundle/java/org/mockito/osgitest/testbundle/MockNonPublicClassFailsTest.java
```diff
@@ -0,0 +1,24 @@
...
+public class MockNonPublicClassFailsTest implements Runnable {
+
+    static class NonPublicClass {}
+
+    @Override
+    public void run() {
+        try {
+            NonPublicClass nonPublicClass = mock(NonPublicClass.class);
+            throw new AssertionError();
+        } catch(MockitoException e) {
```
TimvdLippe: Nit: don't use `try-catch` clauses in tests. Instead, use `assertThat` or `assertThrows`

subprojects/osgi-test/src/testBundle/java/org/mockito/osgitest/testbundle/MockClassInOtherBundleTest.java
TimvdLippe: (In this case, an `assertEquals(methods.intReturningMethod(), 42)` would be better.

subprojects/osgi-test/src/test/java/org/mockito/osgitest/OsgiTest.java
```diff
@@ -0,0 +1,122 @@
...
+        createTestRunnable("MockNonPublicClassFailsTest").run();
+    }
+
+    @Test
+    public void testMockClassInOtherBundle() throws BundleException, InterruptedException, ClassNotFoundException, IllegalAccessException, InstantiationException, IOException {
+        createTestRunnable("MockClassInOtherBundleTest").run();
+    }
+
+    private Runnable createTestRunnable(String className) throws InstantiationException, IllegalAccessException, ClassNotFoundException {
+        return (Runnable) testBundle.loadClass("org.mockito.osgitest.testbundle." + className).newInstance();
```
TimvdLippe: Can we use the JUnit suite API to run these tests? For an example, see https://github.com/mockito/mockito/blob/release/3.x/src/test/java/org/mockitousage/junitrunner/StrictRunnerTest.java and the `JUnitCore` API.

theogimonde:
> This is great work, thanks! Only some nits and a question whether we can use the JUnit API to handle the tests. That should improve the assertions and simplify the loading of the tests.

Thanks for the speedy review! It was intentional not to use the JUnit API since it's not readily available to consume as an OSGi bundle in this project. However, I have committed a new version where the `org.junit` package is provided to the OSGi bundles as a "system package" and thus is loaded from the regular class loader.

It's still tricky to get the test cases to run inside the bundles, since they aren't visible to Gradle in the `test` source set. In fact the test classes are only available during runtime _after_ starting the OSGi framework and installing their bundles. To overcome that issue I've inherited from `Suite` and fed it test classes loaded through the class loader of the test bundle.

Of course using proper asserts is preferable now that they are available so I fixed that. To get a clean diff from the first commit I did not yet consolidate the test classes, i.e. there's still one test class with a single `test()` method for each previous runnable, which might look a bit wasteful. Input on that is welcome, and also if you think that the added complexity of using a suite and the way JUnit is provided to the inside of OSGi is acceptable.

Technical details:
- The framework needs to be started in the suite constructor to be able to load the test classes. Thus `@BeforeClass` can no longer be used. However `@AfterClass` still works for the tear down.
- Using JUnit4 inside the test bundles makes Bnd add a corresponding `Import-Package: org.junit` to its manifest, the one that's resolved by the extra system package entry. This additional package import should not affect the effectiveness of the test for detecting manifest issues. In fact it might even make it more like other usages of Mockito.

subprojects/osgi-test/src/test/java/org/mockito/osgitest/OsgiTest.java
theogimonde: Makes sense. I collapsed multiple exception specifiers into `Exception` where applicable.

subprojects/osgi-test/src/testBundle/java/org/mockito/osgitest/testbundle/MockClassInOtherBundleTest.java
theogimonde: Using `assertEquals(expected, actual)` now that it's available.

subprojects/osgi-test/src/test/java/org/mockito/osgitest/OsgiTest.java
theogimonde: I did use the suite API to make the tests run directly in the Gradle runner, which I think is better than using an explicit `JUnitCore` runner in this situation.

subprojects/osgi-test/src/testBundle/java/org/mockito/osgitest/testbundle/MockNonPublicClassFailsTest.java
theogimonde: Using `@Test(expected = MockitoException.class)` now that it's available.

PR #1854, opened 2020-01-02
Fixes #1853: Allow @MockitoSettings to be inherited

### LINKED ISSUE #1853: Allow @MockitoSettings to be inherited
- [X] The mockito message in the stacktrace have useful information, but it didn't help
 - [X] The problematic code (if that's possible) is copied here;
 - [X] Provide versions (mockito / jdk / os / any other relevant information)
 - [X] Provide a [Short, Self Contained, Correct (Compilable), Example](http://sscce.org) of the issue
 - [X] Read the [contributing guide](https://github.com/mockito/mockito/blob/release/3.x/.github/CONTRIBUTING.md)

I have a base test class which among other things mocks a "serviceLocator" and a very common class used by many other services:

```java
public abstract class DomainServiceTests<T> {
	@Mock
	protected ServiceLocator sl;

	@Spy
	protected UserEnvironmentService userEnvironment;

	@BeforeEach
	public void setup() throws Exception {
		Mockito.when(sl.get(UserEnvironmentService.class)).thenReturn(userEnvironment);
	}
}
```

But whenever any test try to mock the get method of this serviceLocator with a different class I get an UnnecessaryStubbingException:

```
org.mockito.exceptions.misusing.UnnecessaryStubbingException:
    Unnecessary stubbings detected.
    Clean & maintainable test code requires zero unnecessary code.
    Following stubbings are unnecessary (click to navigate to relevant line of code):
      1. -> at *.DomainServiceTests.setup(DomainServiceTests.java:24)
    Please remove unnecessary stubbings or use 'lenient' strictness. More info: javadoc for UnnecessaryStubbingException class.
```

Then I specify the strictness in the base test class 

```java
@MockitoSettings(strictness = Strictness.WARN)
public abstract class DomainServiceTests<T> {
```
But even though the strictness is not seen by mockito.

I have debugged the code and found that the strictness is retrieved by searching the annotation on the method and the class of the test

https://github.com/mockito/mockito/blob/b2b66f430a5bf890135c0c25872bf9869eb09aec/subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/MockitoExtension.java#L143-L145

This eventually calls 

https://github.com/junit-team/junit5/blob/5e41ebe612fd67e905f96f1dd3184a071b65be17/junit-platform-commons/src/main/java/org/junit/platform/commons/util/AnnotationUtils.java#L108-L112
```java
public static <A extends Annotation> Optional<A> findAnnotation(AnnotatedElement element, Class<A> annotationType) {
	Preconditions.notNull(annotationType, "annotationType must not be null");
	boolean inherited = annotationType.isAnnotationPresent(Inherited.class);
	return findAnnotation(element, annotationType, inherited, new HashSet<>());
}
```

So if you annotate `MockitoSettings` with `java.lang.annotation.Inherited` it will work (I turned the inherited variable to true using the debugger to confirm).

For now I will have to annotate every test class with
```java
@MockitoSettings(strictness = Strictness.WARN)
```

I am using  Gradle 7.0 with JUnit 5.5.2

Thanks

fabianMendez:
- [X] Read the [contributing guide](https://github.com/mockito/mockito/blob/release/3.x/.github/CONTRIBUTING.md)

### DIFF
--- a/subprojects/junit-jupiter/src/test/java/org/mockitousage/StrictnessTest.java
+++ b/subprojects/junit-jupiter/src/test/java/org/mockitousage/StrictnessTest.java
@@ -118,6 +118,26 @@ class StrictnessTest {
         assertThat(result.getThrowable().get()).isInstanceOf(UnnecessaryStubbingException.class);
     }
 
+    @MockitoSettings(strictness = Strictness.WARN)
+    static class BaseWarnStubs {}
+
+    static class InheritedWarnStubs extends BaseWarnStubs {
+        @Mock
+        private Function<Integer, String> rootMock;
+
+        @Test
+        void should_execute_successfully_on_warn_stubs_inherited_from_base_class() {
+            Mockito.when(rootMock.apply(10)).thenReturn("Foo");
+        }
+    }
+
+    @Test
+    void inherits_strictness_from_base_class() {
+        TestExecutionResult result = invokeTestClassAndRetrieveMethodResult(InheritedWarnStubs.class);
+
+        assertThat(result.getStatus()).isEqualTo(TestExecutionResult.Status.SUCCESSFUL);
+    }
+
     private TestExecutionResult invokeTestClassAndRetrieveMethodResult(Class<?> clazz) {
         LauncherDiscoveryRequest request = LauncherDiscoveryRequestBuilder.request()
             .selectors(
--- a/subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/MockitoSettings.java
+++ b/subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/MockitoSettings.java
@@ -7,6 +7,7 @@ package org.mockito.junit.jupiter;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.quality.Strictness;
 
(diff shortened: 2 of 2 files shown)


fabianMendez:
> 
> 
> Could you add a regression test in the junit-jupiter-test testsuite?

I've just amended the commit with a test

subprojects/junit-jupiter/src/test/java/org/mockitousage/StrictnessTest.java
```diff
@@ -118,6 +118,26 @@ void by_default_configures_strict_stubs_in_runner() {
...
 
+    @MockitoSettings(strictness = Strictness.WARN)
+    static class BaseWarnStubs {}
+
+    static class InheritedWarnStubs extends BaseWarnStubs {
+        @Mock
+        private Function<Integer, String> rootMock;
+
+        @Test
+        void should_throw_an_exception_on_strict_stubs_configured_by_default() {
```
TimvdLippe: Nit: test name should be updated to state that it properly inherits from the MockitoSettings on the superclass.

fabianMendez: Done

PR #1858, opened 2020-01-14
Add TestRule to MockitoJUnit

TimvdLippe:
JUnit has two types of test rules: MethodRule and TestRule. The existing
MockitoJUnit rule returns a MockitoRule which extends MethodRule.
However, since this is not a TestRule, some features of JUnit do not
nicely integrate with Mockito. For example, the RuleChain feature of
JUnit only works on TestRules.

Therefore, add a separate entrypoint to MockitoJUnit which returns a
TestRule that can be integrated with RuleChain. It introduces a separate
interface to make this distinction clear. If we would change the
existing MockitoRule to also extend TestRule, the JUnit logic would no
longer run the MethodRule [1] and would thus break existing users.

[1]: https://github.com/junit-team/junit4/blob/2df7e0882128d551565f87f688bbe745d85aacba/src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java#L402

### DIFF
--- a/src/main/java/org/mockito/internal/junit/JUnitRule.java
+++ b/src/main/java/org/mockito/internal/junit/JUnitRule.java
@@ -11,7 +11,7 @@ import org.mockito.plugins.MockitoLogger;
 import org.mockito.quality.Strictness;
 
 /** Internal implementation. */
-public class JUnitRule extends AbstractJUnitRule implements MockitoRule {
+public final class JUnitRule extends AbstractJUnitRule implements MockitoRule {
 
     /** @param strictness how strict mocking / stubbing is concerned */
     public JUnitRule(MockitoLogger logger, Strictness strictness) {
--- a/src/main/java/org/mockito/internal/junit/JUnitRule.java
+++ b/src/main/java/org/mockito/internal/junit/JUnitRule.java
@@ -14,18 +14,14 @@ import org.mockito.junit.MockitoRule;
 import org.mockito.plugins.MockitoLogger;
 import org.mockito.quality.Strictness;
 
-/**
- * Internal implementation.
- */
+/** Internal implementation. */
 public class JUnitRule implements MockitoRule {
 
     private final MockitoLogger logger;
     private Strictness strictness;
     private MockitoSession session;
 
-    /**
-     * @param strictness how strict mocking / stubbing is concerned
-     */
+    /** @param strictness how strict mocking / stubbing is concerned */
     public JUnitRule(MockitoLogger logger, Strictness strictness) {
         this.logger = logger;
         this.strictness = strictness;
@@ -33,15 +29,20 @@ public class JUnitRule implements MockitoRule {
 
     @Override
     public Statement apply(final Statement base, final FrameworkMethod method, final Object target) {
+        return createStatement(base, target.getClass().getSimpleName() + "." + method.getName(), target);
+    }
+
+    protected Statement createStatement(final Statement base, final String methodName, final Object target) {
         return new Statement() {
             public void evaluate() throws Throwable {
                 if (session == null) {
-                    session = Mockito.mockitoSession()
-                                     .name(target.getClass().getSimpleName() + "." + method.getName())
-                                     .strictness(strictness)
-                                     .logger(new MockitoSessionLoggerAdapter(logger))
-                                     .initMocks(target)
-                                     .startMocking();
+                    session =
+                        Mockito.mockitoSession()
+                            .name(methodName)
+                            .strictness(strictness)
+                            .logger(new MockitoSessionLoggerAdapter(logger))
+                            .initMocks(target)
+                            .startMocking();
                 } else {
                     MockitoAnnotations.initMocks(target);
                 }
--- a/src/main/java/org/mockito/internal/junit/JUnitRule.java
+++ b/src/main/java/org/mockito/internal/junit/JUnitRule.java
@@ -6,25 +6,16 @@ package org.mockito.internal.junit;
 
 import org.junit.runners.model.FrameworkMethod;
 import org.junit.runners.model.Statement;
-import org.mockito.Mockito;
-import org.mockito.MockitoAnnotations;
-import org.mockito.MockitoSession;
-import org.mockito.internal.session.MockitoSessionLoggerAdapter;
 import org.mockito.junit.MockitoRule;
 import org.mockito.plugins.MockitoLogger;
 import org.mockito.quality.Strictness;
 
 /** Internal implementation. */
-public class JUnitRule implements MockitoRule {
-
-    private final MockitoLogger logger;
-    private Strictness strictness;
-    private MockitoSession session;
+public class JUnitRule extends AbstractJUnitRule implements MockitoRule {
 
     /** @param strictness how strict mocking / stubbing is concerned */
     public JUnitRule(MockitoLogger logger, Strictness strictness) {
-        this.logger = logger;
-        this.strictness = strictness;
+        super(logger, strictness);
     }
 
     @Override
@@ -32,49 +23,12 @@ public class JUnitRule implements MockitoRule {
         return createStatement(base, target.getClass().getSimpleName() + "." + method.getName(), target);
     }
 
-    protected Statement createStatement(final Statement base, final String methodName, final Object target) {
-        return new Statement() {
-            public void evaluate() throws Throwable {
-                if (session == null) {
-                    session =
-                        Mockito.mockitoSession()
-                            .name(methodName)
-                            .strictness(strictness)
-                            .logger(new MockitoSessionLoggerAdapter(logger))
-                            .initMocks(target)
-                            .startMocking();
-                } else {
-                    MockitoAnnotations.initMocks(target);
-                }
-                Throwable testFailure = evaluateSafely(base);
-                session.finishMocking(testFailure);
-                if (testFailure != null) {
-                    throw testFailure;
-                }
-            }
-
-            private Throwable evaluateSafely(Statement base) {
-                try {
-                    base.evaluate();
-                    return null;
-                } catch (Throwable throwable) {
-                    return throwable;
-                }
-            }
-        };
-    }
-
     public MockitoRule silent() {
         return strictness(Strictness.LENIENT);
     }
 
     public MockitoRule strictness(Strictness strictness) {
-        this.strictness = strictness;
-        // session is null when this method is called during initialization of
-        // the @Rule field of the test class
-        if (session != null) {
-            session.setStrictness(strictness);
-        }
+        super.setStrictness(strictness);
         return this;
     }
 }
--- a/src/main/java/org/mockito/internal/junit/JUnitRule.java
+++ b/src/main/java/org/mockito/internal/junit/JUnitRule.java
@@ -11,16 +11,18 @@ import org.mockito.plugins.MockitoLogger;
 import org.mockito.quality.Strictness;
 
 /** Internal implementation. */
-public final class JUnitRule extends AbstractJUnitRule implements MockitoRule {
+public final class JUnitRule implements MockitoRule {
+
+    private final JUnitSessionStore sessionStore;
 
     /** @param strictness how strict mocking / stubbing is concerned */
     public JUnitRule(MockitoLogger logger, Strictness strictness) {
-        super(logger, strictness);
+        this.sessionStore = new JUnitSessionStore(logger, strictness);
     }
 
     @Override
     public Statement apply(final Statement base, final FrameworkMethod method, final Object target) {
-        return createStatement(base, target.getClass().getSimpleName() + "." + method.getName(), target);
+        return sessionStore.createStatement(base, target.getClass().getSimpleName() + "." + method.getName(), target);
     }
 
     public MockitoRule silent() {
@@ -28,7 +30,7 @@ public final class JUnitRule extends AbstractJUnitRule implements MockitoRule {
     }
 
     public MockitoRule strictness(Strictness strictness) {
-        super.setStrictness(strictness);
+        sessionStore.setStrictness(strictness);
         return this;
     }
 }
--- a/src/main/java/org/mockito/internal/junit/JUnitTestRule.java
+++ b/src/main/java/org/mockito/internal/junit/JUnitTestRule.java
@@ -6,26 +6,23 @@ package org.mockito.internal.junit;
 
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
-import org.mockito.exceptions.base.MockitoException;
 import org.mockito.junit.MockitoTestRule;
 import org.mockito.plugins.MockitoLogger;
 import org.mockito.quality.Strictness;
 
-public final class JUnitTestRule extends AbstractJUnitRule implements MockitoTestRule {
+public final class JUnitTestRule implements MockitoTestRule {
 
     private final Object testInstance;
+    private final JUnitSessionStore sessionStore;
 
     public JUnitTestRule(MockitoLogger logger, Strictness strictness, Object testInstance) {
-        super(logger, strictness);
+        this.sessionStore = new JUnitSessionStore(logger, strictness);
         this.testInstance = testInstance;
     }
 
     @Override
     public Statement apply(Statement base, Description description) {
-        if (description.isSuite()) {
-            throw new MockitoException("JUnitTestRule can not be used as a @ClassRule.");
-        }
-        return createStatement(base, description.getDisplayName(), this.testInstance);
+        return sessionStore.createStatement(base, description.getDisplayName(), this.testInstance);
     }
 
     public MockitoTestRule silent() {
@@ -33,7 +30,7 @@ public final class JUnitTestRule extends AbstractJUnitRule implements MockitoTes
     }
 
     public MockitoTestRule strictness(Strictness strictness) {
-        super.setStrictness(strictness);
+        sessionStore.setStrictness(strictness);
         return this;
     }
 }
--- a/src/main/java/org/mockito/internal/junit/JUnitTestRule.java
+++ b/src/main/java/org/mockito/internal/junit/JUnitTestRule.java
@@ -11,7 +11,7 @@ import org.mockito.junit.MockitoTestRule;
 import org.mockito.plugins.MockitoLogger;
 import org.mockito.quality.Strictness;
 
-public class JUnitTestRule extends JUnitRule implements MockitoTestRule {
+public class JUnitTestRule extends AbstractJUnitRule implements MockitoTestRule {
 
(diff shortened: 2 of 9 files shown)


mockitoguy:
Thank you for the PR! Can you elaborate on the motivation?

> since this is not a TestRule, some features of JUnit do not
nicely integrate with Mockito. For example, the RuleChain feature of
JUnit only works on TestRules

Can you show an example test that benefits from implementing TestRule? Also, I'm curious what is driving this feature - do you have tests at work that need it or was there a user request?

I don't object the feature. I'd like to understand better why we need this. Thanks!

TimvdLippe:
Please see the `JUnitTestRuleIntegratesWithRuleChainTest` that shows how it can be used with `RuleChain`. The test code will not compile if you omit the `this`, as `MockitoRule` inherits from `MethodRule`, whereas `RuleChain` needs a `TestRule`. In other words, if Mockito does not expose a subclass from `TestRule`, those users can not use Mockito in combination with `RuleChain`, as it will not compile.

TimvdLippe:
The background for this change is that we are moving Google to use the open source rule, rather than our internal one. During that migration, we discovered that we could not migrate to the rule if it was used in combination with a subset of the JUnit features, one of which is the `@RuleChain`. This PR would fix the integration and allow Google (as well as other users) to use the Mockito Rule in combination with all JUnit features.

mockitoguy:
> The background (...)

Thank you! This is a very valuable information that helps understanding why we need this feature!

mockitoguy:
I haven't reviewed the code yet but I'll do it today or tomorrow. Is that OK?

TimvdLippe:
Yes that's okay. It will also be reviewed internally.

src/main/java/org/mockito/internal/junit/JUnitRule.java
```diff
@@ -6,74 +6,29 @@
...
-/**
- * Internal implementation.
- */
-public class JUnitRule implements MockitoRule {
-
-    private final MockitoLogger logger;
-    private Strictness strictness;
-    private MockitoSession session;
+/** Internal implementation. */
+public final class JUnitRule extends AbstractJUnitRule implements MockitoRule {
```
mockitoguy: Abstract class hierarchies are complex. Is there a way to share the code without the abstract class? Delegation or composition? Thoughts?

src/main/java/org/mockito/internal/junit/JUnitTestRule.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public JUnitTestRule(MockitoLogger logger, Strictness strictness, Object testInstance) {
+        super(logger, strictness);
+        this.testInstance = testInstance;
+    }
+
+    @Override
+    public Statement apply(Statement base, Description description) {
+        if (description.isSuite()) {
+            throw new MockitoException("JUnitTestRule can not be used as a @ClassRule.");
```
mockitoguy: double checking - do we have coverage for this case?

src/main/java/org/mockito/junit/MockitoJUnit.java
```diff
@@ -33,6 +35,19 @@ public static MockitoRule rule() {
...
+    /**
+     * Creates a rule instance that initiates &#064;Mocks and is a {@link TestRule}. Use this method
+     * if you need to explicitly need a {@link TestRule}, for example if you need to compose
+     * multiple rules using a {@link org.junit.rules.RuleChain}.
+     * See {@link MockitoRule}.
+     *
+     * @param testInstance The instance to initiate mocks for
+     * @return the rule instance
+     */
+    public static MockitoTestRule rule(Object testInstance) {
```
mockitoguy: Can we rename to "testRule"? It is a bit not intuitive when overloaded method returns a different type.

Also, the identical test method name may lead to inconsistencies - some tests using ```rule(this)``` and some ```rule()``` with no good reason.

src/main/java/org/mockito/junit/MockitoJUnit.java
```diff
@@ -33,6 +35,19 @@ public static MockitoRule rule() {
         return new JUnitRule(Plugins.getMockitoLogger(), Strictness.WARN);
     }
 
+    /**
+     * Creates a rule instance that initiates &#064;Mocks and is a {@link TestRule}. Use this method
```
mockitoguy: I suggest to make the recommended use stronger - just to avoid confusion which method to use:

"Use this method only when you explicitly need (...) Otherwise always prefer 'rule()' method".

src/main/java/org/mockito/junit/MockitoTestRule.java
```diff
@@ -0,0 +1,94 @@
...
+import org.junit.rules.TestRule;
+import org.mockito.Incubating;
+import org.mockito.MockSettings;
+import org.mockito.Mockito;
+import org.mockito.exceptions.misusing.PotentialStubbingProblem;
+import org.mockito.quality.MockitoHint;
+import org.mockito.quality.Strictness;
+
+/**
+ * Equivalent to {@link MockitoRule}, but also inherits from {@link TestRule}. For more information,
```
mockitoguy: Using "also" here makes me a bit confused. How about this: "but inherits different JUnit4 base interface - {@link TestRule}"

src/main/java/org/mockito/junit/MockitoTestRule.java
```diff
@@ -0,0 +1,94 @@
...
+     * <pre class="code"><code class="java">
+     * public class ExampleTest {
+     *
+     *     &#064;Rule
+     *     public MockitoRule rule = MockitoJUnit.rule().silent();
+     *
+     * }
+     * </code></pre>
+     *
+     * @since 2.1.0
```
mockitoguy: let's update since tag

src/main/java/org/mockito/junit/MockitoTestRule.java
```diff
@@ -0,0 +1,94 @@
...
+import org.mockito.quality.Strictness;
+
+/**
+ * Equivalent to {@link MockitoRule}, but also inherits from {@link TestRule}. For more information,
+ * please see the documentation on {@link MockitoRule}.
+ */
+public interface MockitoTestRule extends TestRule {
+
+    /**
+     * Rule will not report stubbing warnings during test execution. By default, stubbing warnings are
```
mockitoguy: I'd write a single sentence here and link to the MockitoRule.silent(). This way it's more DRY but still useful.

src/main/java/org/mockito/junit/MockitoTestRule.java
```diff
@@ -0,0 +1,94 @@
...
+     *
+     * "Strict stubs" are planned to be the default for Mockito v3 We are very eager to hear feedback
+     * about "strict stubbing" feature, let us know by commenting on GitHub <a
+     * href="https://github.com/mockito/mockito/issues/769">issue 769</a>. Strict stubbing is an
+     * attempt to improve testability and productivity with Mockito. Tell us what you think!
+     *
+     * @since 2.3.0
+     */
+    @Incubating
+    MockitoTestRule strictness(Strictness strictness);
```
mockitoguy: as my previous comment, let's update the since tag, and make the doc more DRY.

src/test/java/org/mockitousage/junitrule/JUnitTestRuleIntegratesWithRuleChainTest.java
```diff
@@ -0,0 +1,103 @@
...
+import org.junit.runners.model.Statement;
+import org.mockito.Mock;
+import org.mockito.internal.util.MockUtil;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.quality.Strictness;
+import org.mockitousage.IMethods;
+import org.mockitousage.junitrule.MutableStrictJUnitTestRuleTest.StrictByDefault;
+import org.mockitoutil.JUnitResultAssert;
+
+public class JUnitTestRuleIntegratesWithRuleChainTest {
```
mockitoguy: Can you double-check this test? It seems it has few bugs.

src/test/java/org/mockitousage/junitrule/JUnitTestRuleIntegratesWithRuleChainTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    JUnitCore runner = new JUnitCore();
+
+    @Test public void rule_can_be_changed_to_strict() {
+        //when
+        Result result = runner.run(MutableStrictJUnitTestRuleTest.LenientByDefault.class);
```
mockitoguy: MutableStrictJUnitTestRuleTest?

src/test/java/org/mockitousage/junitrule/JUnitTestRuleIntegratesWithRuleChainTest.java
```diff
@@ -0,0 +1,103 @@
...
+import org.mockito.quality.Strictness;
+import org.mockitousage.IMethods;
+import org.mockitousage.junitrule.MutableStrictJUnitTestRuleTest.StrictByDefault;
+import org.mockitoutil.JUnitResultAssert;
+
+public class JUnitTestRuleIntegratesWithRuleChainTest {
+
+    JUnitCore runner = new JUnitCore();
+
+    @Test public void rule_can_be_changed_to_strict() {
```
mockitoguy: test name does not correspond to the test class name - unsure what are testing here?

src/test/java/org/mockitousage/junitrule/JUnitTestRuleIntegratesWithRuleChainTest.java
```diff
@@ -0,0 +1,103 @@
...
+        //when
+        Result result = runner.run(MutableStrictJUnitTestRuleTest.LenientByDefault.class);
+
+        //then
+        JUnitResultAssert.assertThat(result)
+            .succeeds(1)
```
mockitoguy: take a look at other methods on JUnitResultAssert - we can make specific assertions on the names of failed tests, etc.

src/test/java/org/mockitousage/junitrule/JUnitTestRuleIntegratesWithRuleChainTest.java
```diff
@@ -0,0 +1,103 @@
...
+        private AtomicBoolean called = new AtomicBoolean(false);
+
+        @Test
+        public void creates_mocks_in_correct_rulechain_ordering() {
+            assertThat(MockUtil.isMock(mock)).isTrue();
+            assertThat(called.get()).isTrue();
+        }
+    }
+
+    public static class StrictByDefault {
```
mockitoguy: If we are testing RuleChain use case here, I don't think we need 2 test methods.

src/test/java/org/mockitousage/junitrule/JUnitTestRuleIntegratesWithRuleChainTest.java
```diff
@@ -0,0 +1,103 @@
...
+        //when
+        Result result = runner.run(StrictByDefault.class);
+
+        //then
+        JUnitResultAssert.assertThat(result)
+            .succeeds(1)
+            .fails(1, RuntimeException.class);
+    }
+
+    public static class LenientByDefault {
```
mockitoguy: LenientByDefault?

src/test/java/org/mockitousage/junitrule/MockitoJUnitTestRuleTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoTestRule;
+
+public class MockitoJUnitTestRuleTest {
+
+    @Rule
+    public MockitoTestRule mockitoRule = MockitoJUnit.rule(this);
+
+    // Fixes #1578: Protect against multiple execution.
```
mockitoguy: Is this Mockito ticket?

src/test/java/org/mockitousage/junitrule/MutableStrictJUnitTestRuleTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    JUnitCore runner = new JUnitCore();
+
+    @Test public void rule_can_be_changed_to_strict() throws Throwable {
+        //when
+        Result result = runner.run(LenientByDefault.class);
+
+        //then
+        JUnitResultAssert.assertThat(result)
+                .succeeds(1)
```
mockitoguy: Can we make the assertion stronger and check if the right test method failed, and specific exception vs. RuntimeException. Check out the methods on JUnitResultAssert. Thanks!

src/test/java/org/mockitousage/junitrule/MutableStrictJUnitTestRuleTest.java
```diff
@@ -0,0 +1,75 @@
...
+import org.junit.runner.JUnitCore;
+import org.junit.runner.Result;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoTestRule;
+import org.mockito.quality.Strictness;
+import org.mockitousage.IMethods;
+import org.mockitoutil.JUnitResultAssert;
+
+public class MutableStrictJUnitTestRuleTest {
```
mockitoguy: LOVE this test case.

src/main/java/org/mockito/internal/junit/JUnitRule.java
TimvdLippe: Done

src/main/java/org/mockito/junit/MockitoJUnit.java
TimvdLippe: Done

src/main/java/org/mockito/junit/MockitoJUnit.java
TimvdLippe: Done

src/main/java/org/mockito/junit/MockitoTestRule.java
TimvdLippe: Done

src/main/java/org/mockito/junit/MockitoTestRule.java
TimvdLippe: Done

src/main/java/org/mockito/junit/MockitoTestRule.java
TimvdLippe: Done

src/main/java/org/mockito/junit/MockitoTestRule.java
TimvdLippe: Done

src/main/java/org/mockito/internal/junit/JUnitTestRule.java
TimvdLippe: Done

TimvdLippe: (Turns out you can only use `@ClassRule` on static fields, which doesn't work if you need to specify `this`.)

src/test/java/org/mockitousage/junitrule/JUnitTestRuleIntegratesWithRuleChainTest.java
TimvdLippe: Good catch. Copy-paste bug. Fixed!

src/test/java/org/mockitousage/junitrule/JUnitTestRuleIntegratesWithRuleChainTest.java
TimvdLippe: Done

src/test/java/org/mockitousage/junitrule/JUnitTestRuleIntegratesWithRuleChainTest.java
TimvdLippe: done

src/test/java/org/mockitousage/junitrule/JUnitTestRuleIntegratesWithRuleChainTest.java
TimvdLippe: This is taken from `MutableStrictJUnitTestRuleTest`.

src/test/java/org/mockitousage/junitrule/JUnitTestRuleIntegratesWithRuleChainTest.java
TimvdLippe: I wanted to make sure that with the rulechain, the order is still correct.

src/test/java/org/mockitousage/junitrule/JUnitTestRuleIntegratesWithRuleChainTest.java
TimvdLippe: Yes

src/test/java/org/mockitousage/junitrule/MockitoJUnitTestRuleTest.java
TimvdLippe: This was taken from `MockitoJUnitRuleTest` and it felt like a useful comment for future reference.

src/test/java/org/mockitousage/junitrule/MutableStrictJUnitTestRuleTest.java
TimvdLippe: This was taken from `MutableStrictJUnitRuleTest`. Can we update these assertions in a separate PR to make them consistent?

src/test/java/org/mockitousage/junitrule/MutableStrictJUnitTestRuleTest.java
TimvdLippe: :smile:

TimvdLippe:
@mockitoguy do you have time to take another look? I would like to integrate this internally to start a migration.

TimvdLippe:
@mockitoguy Based on https://github.com/mockito/mockito/pull/1858#pullrequestreview-344403753 I deduce you have highlevel approval for the new API. Engineers internally have been asking for this quite frequently now, so I would like to merge this.

I would like to merge this tomorrow, if you don't have any comments until then. If you do, please let me know, thanks!

mockitoguy:
I'll take a look later today. If you don't hear from me, merge at will :-)

src/test/java/org/mockitousage/junitrule/MutableStrictJUnitTestRuleTest.java
mockitoguy: >This was taken from MutableStrictJUnitRuleTest. Can we update these assertions in a separate PR to make them consistent?

Absolutely!

PR #1859, opened 2020-01-14
Add NotExtensible to OngoingStubbing

TimvdLippe:
This interface is not intended to be subclassed and should therefore be
marked as such.


mockitoguy:
Absolutely!

PR #1862, opened 2020-01-22
Revert "Performance optimization by using Method.getParameterCount() where possible"

TimvdLippe:
Reverts mockito/mockito#1849

It breaks all downstream Android applications, see https://github.com/mockito/mockito/pull/1849#issuecomment-577139089


TimvdLippe:
Approved per https://github.com/mockito/mockito/pull/1849#issuecomment-578564595

PR #1863, opened 2020-01-22
Adds the Official Gradle Wrapper Validation GitHub Action

bric3:
The idea is to alert when a PR introduces non official gradle wrapper
to prevent _social engineering supply chain attack_.

See: https://github.com/gradle/wrapper-validation-action

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,3 +6,9 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 indent_style = space
 indent_size = 4
+
+[{*.yml,*.yaml}]
+indent_size = 2
+ij_continuation_indent_size = 2
+ij_yaml_keep_indents_on_empty_lines = false
+ij_yaml_keep_line_breaks = true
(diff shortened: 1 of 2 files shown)


bric3:
> Do we need to enable the GitHub actions somehow?

I'm not sure about that. I believe the workflow detection is automatic, but it's the first time I do this before enableing github actions.

TimvdLippe:
I think GitHub actions are still in beta, but I am not sure tbh. We could check this and try to social engineer ourselves by checking in a corrupt gradle-wrapper. If it doesn't fail, we know we will need to enable it somehow.

bric3:
As far as I know GA work very well. Some well known projects uses it, like bytebuddy ;)
And I think this is a quick win in security. Making a PR that bump gradle with a fancy jar is really easy.

TimvdLippe:
Oh yeah, to be clear: I am heavily in favor of this change. Just trying to figure out what are the steps necessary to make this happen :smile:

bric3:
I think just merging it would be enough. And if it doesn't work we can always remove the change.

Seems like adding YAML reduces code coverage elsewhere :D

bric3:
Github actions are automatically enabled : https://github.com/mockito/mockito/actions?query=workflow%3A%22Validate+Gradle+Wrapper%22

PR #1865, opened 2020-01-23
Expose MockitoExtension to ServiceLoader to support autodetection fea…

### LINKED ISSUE #1864: Allow MockitoExtension to be registered automatically
JUnit 5 supports [automatic extension registration](https://junit.org/junit5/docs/current/user-guide/#extensions-registration-automatic). This requiers a tiny change in junit-juniper module.

georgy:
Fixes #1864

Add org.junit.jupiter.api.extension.Extension service definition for MockitoExtension

### DIFF
--- a/subprojects/junit-jupiter/src/main/resources/META-INF/services/org.junit.jupiter.api.extension.Extension
+++ b/subprojects/junit-jupiter/src/main/resources/META-INF/services/org.junit.jupiter.api.extension.Extension
@@ -1,0 +1,1 @@
+org.mockito.junit.jupiter.MockitoExtension


georgy:
@TimvdLippe I see your point, personally I never used services this way before. Both options are possible: user can configure service in their jar or it can be configured in junit-juniper. My understanding of ServiceLoader is that second option is what I would call "by design": if it is on classpath it can be loaded. It is always possible to stick to the good old @ExtendWith (by not enabling autodetection).

georgy:
An alternative would be to build an artifact with a classifier, like 'autodetection'. This way user can opt in for autodetection without the need for additional config.

TimvdLippe:
Having such an artifact feels like more indirection. It would be easier to add the file yourself.

I am closing this per https://github.com/mockito/mockito/pull/1865#pullrequestreview-347321421 Thank you for opening the PR and discussing this feature, sorry to not make it in!

PR #1867, opened 2020-01-30
Upgrade assertj-core to version 3.15.0

scordio:


mockitoguy:
Nice upgrade, thanks!

PR #1870, opened 2020-02-15
 adding once and twice verification

phablotassio:
Hey guys, I adding twice and once verification for mocks, the idea is simplify because the majority of tests usage times(1) or times(2). how exist never() , I thought it would be a good idea to add this, as it is much more readable.

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -2715,6 +2715,49 @@ public class Mockito extends ArgumentMatchers {
     }
 
     /**
+     * Alias to <code>times(1)</code>, see {@link Mockito#times(int)}
+     * <p>
+     * Verifies that interaction did not happen. E.g:
+     * <pre class="code"><code class="java">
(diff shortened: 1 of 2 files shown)


grimreaper:
Coming at this diff as a random stranger (not a committer): `once` might be nice as a an alias for `times(1)` and makes sense as an alias as it is fairly common. Having twice leads to the question of why not thrice, quad, quince, or frice. Adding aliases for the sake of aliases doesn't seem like a great idea unless the use cases are incredibly common.

danielceinos:
A scalable solution for this use cases could be this utility extension:

```kotlin
  fun Int.times() : VerificationMode = times(this)
```

TimvdLippe:
Sorry for not getting earlier on this PR. I think most of the maintainers are hesitant to accept this PR based on https://github.com/mockito/mockito/pull/1870#issuecomment-612702183 Therefore, I am closing this PR. Sorry for that and thanks for proposing this!

PR #1889, opened 2020-03-17
Fixes #1887 : correct @deprecated javadoc

### LINKED ISSUE #1887: Deprecation doc is wrong
https://github.com/mockito/mockito/blob/560adf377677a4a2c658a0b24e2f9c2e0b373c20/src/main/java/org/mockito/Mockito.java#L2252 should suggest using `verifyNoMoreInteractions` instead of `verifyNoInteractions` as discussed [here](https://github.com/mockito/mockito/issues/989)

biergit:

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -2249,7 +2249,7 @@ public class Mockito extends ArgumentMatchers {
      * This method has the same behavior as {@link #verifyNoMoreInteractions(Object...)}.
      *
      * @param mocks to be verified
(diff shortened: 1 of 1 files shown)


TimvdLippe:
My apologies. Looking at the PR in question again (#995) I now realize that the current javadoc is correct. The old behavior was indeed `verifyNoMoreInteractions`, but we want users to use `verifyNoInteractions` instead if their intent was to have zero interactions on a mock.

Sorry for this oversight and thank you for opening the PR nonetheless!

PR #1890, opened 2020-03-20
Fix mocks throwing same instance with throwable class

### LINKED ISSUE #1875: thenThrow(Class) no longer creates new instances
As of #1162, `thenThrow(Class)` no longer creates new instances for each invocation.  We had tests that were reliant on that behavior; for example:
```
	@Test
	public void thenThrow() {
		Supplier<Object> r = mock(Supplier.class);
		when(r.get()).thenThrow(RuntimeException.class);
		RuntimeException first = null;
		for (int i = 0; i < 2; i++) {
			try {
				r.get();
			} catch (RuntimeException e) {
				if (first == null) {
					first = e;
				} else {
					first.addSuppressed(e);
				}
			}
		}
		if (first != null) {
			throw first;
		}
	}
```
These tests now fail:
```
java.lang.IllegalArgumentException: Self-suppression not permitted
	at java.lang.Throwable.addSuppressed(Throwable.java:1043)
	at example.Test.thenThrow(Test.java:10)
```

dean-burdaky:
Fixes mocks throwing the same Throwable instance consecutively after specifying the mock behaviour with `doThrow(Class)` or `thenThrow(Class)`.

This is achieved by moving the instantiation of the Throwable into a new Answer called ThrowsExceptionClass - much like ThrowsException. Then `thenThrow(Class)` calls `thenAnswer(Answer)` with a new instance of ThrowsExceptionClass rather than `thenThrow(Throwable)`. Similar for `doThrow(Class)`, `doThrow(Throwable)` and `doAnswer(Answer)`.

Fixes #1875 

@TimvdLippe Just have a quick inquiry: The test in UninstantiableThrowableTest is failing because it expects an InstantiationError to be thrown in the middle of stubbing (check PR #1549). The new code doesn't instantiate in the middle of stubbing anymore. How do I ignore the test? Or do I just delete the test file? I'm used to using an annotation to ignore tests. The PR should be ready after that's sorted and any other suggested changes.


src/test/java/org/mockito/internal/stubbing/answers/AbstractThrowsExceptionTest.java
```diff
@@ -0,0 +1,106 @@
...
+    public void should_fill_in_exception_stacktrace() {
+        AbstractThrowsException ate = instantiateFixture(new Exception());
+
+        Throwable throwable = Assertions.catchThrowable(() -> ate.answer(createMethodInvocation()));
+        assertNotNull("Should have raised an exception.", throwable);
+        assertThat(throwable.getStackTrace()[0].getClassName()).isEqualTo(AbstractThrowsException.class.getName());
+        assertThat(throwable.getStackTrace()[0].getMethodName()).isEqualTo("answer");
+    }
+
+    @Test(expected = MockitoException.class)
```
TimvdLippe: Instead of `expected = `, use `assertThrows` and assert on the exception that it is of the correct exception class and it has the correct message.

src/main/java/org/mockito/internal/stubbing/answers/ThrowsExceptionClass.java
```diff
@@ -0,0 +1,25 @@
...
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.internal.stubbing.answers;
+
+import java.io.Serializable;
+
+import org.mockito.creation.instance.Instantiator;
+import org.mockito.internal.configuration.plugins.Plugins;
+
+public class ThrowsExceptionClass extends AbstractThrowsException implements Serializable {
```
TimvdLippe: Let's name this `ThrowsExceptionForClassType` to make it clear we are passing in the `.class` itself.

dean-burdaky: The change has been done. The class, references and associated variables have been updated.

src/test/java/org/mockito/internal/stubbing/answers/AbstractThrowsExceptionTest.java
dean-burdaky: I've replaced `expected = ` but I can't find the import for `assertThrows`. I think the tests are configured for JUnit 4 only and `assertThrows` was introduced in 5. If you have Jupiter configured in another test source then I can migrate it to there and do the necessary alterations.

TimvdLippe: Ah I misremembered the name. In AssertJ it is called `assertThatThrownBy`: https://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#exception-assertion

PR #1892, opened 2020-03-22
Make JARs build reproducibly

### LINKED ISSUE #1891: Build is not reproducible
Repeated subsequent builds of the project create outputs which are not byte-for-byte identical, and so cannot be reproduced locally (see: [Reproducible Builds](https://reproducible-builds.org/))

This is due to:
 * File modification timestamps in JARs.
 * Non-deterministic file ordering in JARs.
 * OSGi bundle embedding timestamps in JAR manifests.

Given the build environment under which a version was created, I would like to be able to compare the checksums published on Bintray and/or Maven Central with a locally built version against the same environment, to assure myself that the artifact being pulled has actually been generated from this source code.

(I have a PR to address this which I am happy to raise)

mrwilson:
Fixes #1891 

## Overview

This PR makes the java bytecode generated by this project [reproducible](https://reproducible-builds.org/). This means that under identical builds condition (for example Java version), repeated builds should provide the same output byte-for-byte.

It achieves this by:

* using Gradle's [reproducible archives](https://docs.gradle.org/current/userguide/working_with_files.html#sec:reproducible_archives) functionality to fix timestamps and file ordering within JARs
* re-archiving JARs that are non-deterministic because they contain timestamps in their Manifests

Motivation for this change is best explained on the previously linked Reproducible Builds page.

## Confirming the change

```shell
#!/bin/bash -e

rm -f checksums*

./gradlew clean build -x test --parallel

find . -name '*.jar' \
    | grep '/build/libs/' \
    | sort \
    | xargs sha256sum > checksums-1.txt

./gradlew clean build -x test --parallel

find . -name '*.jar' \
    | grep '/build/libs/' \
    | sort \
    | xargs sha256sum > checksums-2.txt

diff checksums-1.txt checksums-2.txt
```

The diff should be empty, that is to say, both independent runs of the build should have generated byte-for-byte identical outputs, with the exception of the JavaDocs jars

## Omitting Javadocs

Despite Gradle allegedly supporting `noTimestamp` through [StandardJavadocDocletOptions](https://docs.gradle.org/5.3/javadoc/org/gradle/external/javadoc/StandardJavadocDocletOptions.html), I was unable to actually get this to work and remove the timestamped build comment in the javadoc. 

I'll happily fold this into the PR if I (or someone else) can get it to work.

---

### DIFF
--- a/gradle/mockito-core/osgi.gradle
+++ b/gradle/mockito-core/osgi.gradle
@@ -1,3 +1,12 @@
+import java.nio.file.attribute.FileTime
+import java.util.jar.Attributes
+import java.util.jar.JarEntry
+import java.util.jar.JarFile
+import java.util.jar.JarOutputStream
+import java.util.jar.Manifest
+
+import static java.util.Calendar.FEBRUARY
+
 apply plugin: 'osgi'
 
 afterEvaluate {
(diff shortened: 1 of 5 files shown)


gradle/mockito-core/osgi.gradle
```diff
@@ -31,5 +40,90 @@ afterEvaluate {
...
+                .renameTo jar.archivePath
+        }
+
+    }
+}
+
+
+/*
+
+The OSGi plugin in Gradle 5.3 has a known issue (https://github.com/gradle/gradle/issues/6187) where it embeds
```
TimvdLippe: Nit: Use full Java block comments (e.g. lines starting with `*`)

gradle/mockito-core/osgi.gradle
```diff
@@ -31,5 +40,90 @@ afterEvaluate {
...
+a Manifest entry "Bnd-LastModified" (set to the last modified time of the build directory) even if you ask it not to
+using `-noextraheaders` or `-removeheaders`.
+
+It cannot be pinned or removed, and is a source of non-determinism in the build.
+
+This class addresses the problem by rewriting the JAR and removing that entry from the Manifest. A side-effect of this
+is having to set `lastModifiedTime` for each entry to a fixed value, else this would also introduce non-determinism.
+
+The fixed value is the same as the default value for Gradle's fixed timestamps, and the rationale behind it is detailed
+in https://github.com/gradle/gradle/blob/master/subprojects/core/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java
```
TimvdLippe: Nit: change this link to https://github.com/gradle/gradle/blob/58538513a3bff3b7015718976fe1304e23f40694/subprojects/core/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java#L42-L57

mrwilson:
Sure! Where's the best place to put it?

I could include something similar to this script and have it executed as part of the `script:` block in Travis, post build but pre-release, so it would prevent release if it failed.

TimvdLippe:
Probably best to have a new Gradle task that can verify the determinism. I suppose we can put it on this line (https://github.com/mockito/mockito/blob/fd6dfea618126e0555a12038e7071eae3ba3234d/.travis.yml#L53) after `./gradlew build idea -s`. There we can build a new jar in a separate directory and then diff it against the one we just build. (Make sure to not do any unnecessary work such as tests, to keep the build speedy enough)

mrwilson:
I've added a new Gradle task, `checkReproducibility`, which kicks off two sub-builds and compares the checksums of the output archives.

Let me know what you think! 😄

gradle/mockito-core/reproducible-build.gradle
```diff
@@ -0,0 +1,67 @@
...
+    doLast {
+        def first = new File("checksums-1.txt").text.split('\n')
+        def second = new File("checksums-2.txt").text.split('\n')
+
+        assert first.sort() == second.sort()
+    }
+}
+
+task getChecksums1(type: GradleBuild) {
+    tasks = ['clean', 'build']
```
TimvdLippe: This runs the tests as well, which is not what we want. We only want to package the jar, to not incur a high cost on our CI. Can we only package the jar and verify the checksums on that?

mrwilson: Oh, ofc, we can run `assemble` instead of `build` which creates the artifacts but doesn't run the tests. My mistake!

TimvdLippe:
Confirmed that if I remove the fixes you applied to the jars, the checksums do not match and the build fails.

I did update the checksums locations to a separate directory to keep our root directory a bit cleaner. Once CI is happy I will merge this.

PR #1896, opened 2020-03-24
Feature/returns empty java8 time

mikolajczykmarcin:
Im not sure why I see the commit [https://github.com/mockito/mockito/commit/428713153a9fc38e0e2547b255b40e62e9dc8479](url) that was already merged in changes, but:
when creating this PR I had in mind that I would merge with branch 2.x (I didnt know back then that v2.x is maintenance only) hence the reflection-based methods (also to keep the coding style of the ReturnsEmptyValues class)


TimvdLippe:
@kluever Could you review this please? You are the expert on the these APIs 😄

src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValues.java
```diff
@@ -125,6 +134,24 @@ Object returnValueFor(Class<?> type) {
...
+            return JavaEightUtil.emptyLocalDateTime();
+        } else if ("java.time.LocalTime".equals(type.getName())) {
+            return JavaEightUtil.emptyLocalTime();
+        } else if ("java.time.OffsetDateTime".equals(type.getName())) {
+            return JavaEightUtil.emptyOffsetDateTime();
+        } else if ("java.time.OffsetTime".equals(type.getName())) {
+            return JavaEightUtil.emptyOffsetTime();
+        } else if ("java.time.ZonedDateTime".equals(type.getName())) {
+            return JavaEightUtil.emptyZonedDateTime();
         }
```
kluever: Do we want a default for `java.time.ZoneId` (`ZoneOffset.UTC`) and `java.time.ZoneOffset` (same)?

kluever:
Wow, it's crazy how much effort has to be done to maintain pre-Java8 compatibility!

Left a comment about potential other defaults for other `java.time` types, but otherwise LGTM.

TimvdLippe:
Yes it is very unfortunate that Android does not ship these Java 8 API's 😭

src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValues.java
mikolajczykmarcin: fyi ZonedDateTime is already coded ;) I can do the rest if you think its needed

src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesTest.java
```diff
@@ -158,6 +158,191 @@ private void verify_empty_Stream_is_returned(String streamFqcn) throws Exception
...
+        final int nano = (Integer) instantClass.getMethod("getNano").invoke(instant);
+        final long epochSecond = (Long) instantClass.getMethod("getEpochSecond").invoke(instant);
+
+        // then
+        assertEquals("nano of empty " + fqcn, 0, nano);
+        assertEquals("epoch second of empty " + fqcn, 0L, epochSecond);
+    }
+
+    @Test
+    public void should_return_empty_local_date() throws Exception {
```
TimvdLippe: Seems like this one is failing on JDK 8: https://travis-ci.org/github/mockito/mockito/jobs/666534233#L914

src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValues.java
kluever: It's tricky to figure out where to draw the line. I can see an argument for just doing `Duration` and `Period`, given that there is really no "empty" Instant, LocalDate, LocalTime, etc (the unix epoch is rather arbitrary).

That being said, I'd suggest either doing *only* `Duration` and `Period`, or doing all of the `java.time` types. Up to you.

mikolajczykmarcin: ok, it was tough decision to delete all the code I already made, but I agree that it's hard to define what an empty Instant, LocalDate etc. is. I left only period and duration. the rest is in git history if some day you will want to restore it ;)

kluever: Thanks - I understand it's a tough decision, but it's always easier to add than to remove :-)

src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesTest.java
mikolajczykmarcin: yes, EPOCH static was from jdk11 apparently. anyway, after giving a second thought, I decided to leave only period and duration, because it's not obvious that empty date is epoch.

src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValues.java
```diff
@@ -42,6 +42,15 @@
  * Returns an {@code java.util.stream.Stream#empty() empty Stream} for Stream. Similarly for primitive stream variants.
  * </li>
  * <li>
+ * Returns an {@code java.time.Duration.ZERO zero Duration} for empty Duration and {@code java.time.Period.ZERO zero Period} for empty Period.
+ * </li>
+ * <li>
+ * Returns an {@code java.time.Instant.EPOCH epoch Instant} for empty Instant. Similarly EPOCH (UTC where possible) for empty {@code java.time.LocalDate LocalDate}, {@code java.time.LocalDateTime LocalDateTime}, {@code java.time.OffsetDateTime OffsetDateTime}, {@code java.time.ZonedDateTime ZonedDateTime}
+ * </li>
+ * <li>
+ * Returns an {@code java.time.LocalTime.MIN min LocalTime} for empty LocalTime and similarly for {@code java.time.OffsetTime OffsetTime} with UTC zone offset.
```
kluever: Nuke this?

src/main/java/org/mockito/internal/util/JavaEightUtil.java
```diff
@@ -19,6 +20,15 @@
     private static Object emptyOptionalDouble;
     private static Object emptyOptionalInt;
     private static Object emptyOptionalLong;
+    private static Object emptyDuration;
+    private static Object emptyPeriod;
+    private static Object emptyInstant;
```
kluever: Nuke most of these?

kluever:
A few more cleanups, but otherwise LGTM

src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValues.java
mikolajczykmarcin: oops, removed :-)

src/main/java/org/mockito/internal/util/JavaEightUtil.java
mikolajczykmarcin: oops, removed :-)

PR #1897, opened 2020-03-30
Add gradleupdate badge for auto detection of Gradle wrapper updates

scordio:
After this badge is added, @gradleupdate will send a pull request for the latest version of Gradle wrapper if it is outdated.

See https://github.com/int128/gradleupdate.


TimvdLippe:
Hm, I am not sure what to think of this. I would prefer to only sporadically update Gradle, as it has happened in the past that it broke our coverage tracking for example. Therefore a human has to be in the loop for the upgrade, as it isn't a simple "accept this PR and everything will be awesome".

scordio:
I see the point. I assumed the pull request CI would highlight potential issues due to the upgrade, so merging an automated PR would be safe. If this is not the case, let's avoid it and close this PR 🙂

TimvdLippe:
Closing this for now per https://github.com/mockito/mockito/pull/1897#issuecomment-605946080 We can reconsider once we are confident that automatic Gradle upgrades don't inadvertently break our Gradle integrations such as code coverage. Thanks for creating this PR!

scordio:
There is also a chance that dependabot will handle this, see dependabot/feedback#237.

PR #1901, opened 2020-04-07
#1900 Configure mocks for annotation interfaces with default values from annotation

### LINKED ISSUE #1900: Configure mocks for annotation interfaces with default values from annotation
Mockito.mock() can quite conveniently be used to not only create instances of interfaces, but also to create instances of [annotation interfaces](https://docs.oracle.com/javase/tutorial/java/annotations/declaring.html). Although normally those instances are created by using the annotations in source code (and subsequently using `class.getAnnotations()`, often in framework code) there are some use cases where the annotation instances are created by a framework - in particular OSGi Metatype Service (see https://osgi.org/specification/osgi.cmpn/7.0.0/service.metatype.html#d0e17824 for details) is one of them. For this use case, the a mock for an annotation interface should be automatically set up to return the annotation-defined default values.

The following JUnit Tests illustrates what is needed:

```
public @interface AnnotationInterface {
    String strAttWithDefault() default "default";
    String strAttWithoutDefault();
    String[] strArrAtt() default { "default1", "default2" };
    int intAtt() default 1;
    long longAtt() default 2L;
}

@Mock
AnnotationInterface mockedAnnotationInterface;

@Test
public void shouldMockDefaultValuesFromAnnoationInterface() throws Exception {
    // no further action should be needed to have the default values returned
    assertEquals("default", mockedAnnotationInterface.strAttWithDefault());
    assertNull(mockedAnnotationInterface.strAttWithoutDefault());
    assertArrayEquals(new String[] { "default1", "default2" }, mockedAnnotationInterface.strArrAtt());
    assertArrayEquals(new String[] { "default1", "default2" }, mockedAnnotationInterface.strArrAtt());
    assertEquals(1, mockedAnnotationInterface.intAtt());
    assertEquals(2L, mockedAnnotationInterface.longAtt());
}
```

ghenzler:
Closes #1900

### DIFF
--- a/src/main/java/org/mockito/internal/MockitoCore.java
+++ b/src/main/java/org/mockito/internal/MockitoCore.java
@@ -15,6 +15,7 @@ import static org.mockito.internal.util.MockUtil.typeMockabilityOf;
 import static org.mockito.internal.verification.VerificationModeFactory.noInteractions;
 import static org.mockito.internal.verification.VerificationModeFactory.noMoreInteractions;
 
+import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.List;
 
@@ -62,6 +63,11 @@ public class MockitoCore {
         MockCreationSettings<T> creationSettings = impl.build(typeToMock);
         T mock = createMock(creationSettings);
         mockingProgress().mockingStarted(mock, creationSettings);
+        
+        if(typeToMock.isAnnotation()) {
+            configureMockWithAnnotationDefaults(typeToMock, mock);
+        }
+        
         return mock;
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/mockito/internal/MockitoCore.java
```diff
@@ -62,6 +63,11 @@ public boolean isTypeMockable(Class<?> typeToMock) {
         MockCreationSettings<T> creationSettings = impl.build(typeToMock);
         T mock = createMock(creationSettings);
         mockingProgress().mockingStarted(mock, creationSettings);
+        
+        if(typeToMock.isAnnotation()) {
+            configureMockWithAnnotationDefaults(typeToMock, mock);
+        }
```
ghenzler, line 69: This works well but I'm not sure if this is the best position to hook in this functionality - I'm happy to adjust if there is a better position to perform this.

src/main/java/org/mockito/internal/MockitoCore.java
```diff
@@ -62,6 +63,11 @@ public boolean isTypeMockable(Class<?> typeToMock) {
         MockCreationSettings<T> creationSettings = impl.build(typeToMock);
         T mock = createMock(creationSettings);
         mockingProgress().mockingStarted(mock, creationSettings);
+
+        if(typeToMock.isAnnotation()) {
```
ST-DDT, line 67:
```suggestion
        if (typeToMock.isAnnotation()) {
```

PR #1922, opened 2020-05-03
[build] add ben-manes dependency upgrade finder

grimreaper:
Problem

The mockito build system has several pieces which are out of date. This
includes dependencies

Solution

This plugin adds a new task dependencyUpdates which identifies
dependencies which may be out of date


grimreaper:
Sure. I'll do that shortly.

grimreaper:
actually can you point me at the doc to update? I'm looking around and don't see a good place to put it.

Its just adds a new task:
∴./gradlew dependencyUpdates

TimvdLippe:
We should add it to the CONTRIBUTING.md.

grimreaper:
done

PR #1926, opened 2020-05-07
Remove ListUtil and Fields classes

andreisilviudragnea:
- Removed ListUtil class in favor of Java 8 Streams
- Removed unused Fields class


src/main/java/org/mockito/internal/invocation/InvocationsFinder.java
```diff
@@ -116,17 +120,20 @@ public static Location getLastLocation(List<Invocation> invocations) {
...
+        List<Invocation> verifiedOnly = invocations
+            .stream()
+            .filter(context::isVerified)
+            .collect(Collectors.toList());
 
         if (verifiedOnly.isEmpty()) {
             return null;
         } else {
-            return verifiedOnly.getLast();
+            return verifiedOnly.get(verifiedOnly.size() - 1);
```
grimreaper, line 131: This can be `.reduce((first, second) -> second).orElse(null)`

TimvdLippe:
I have to check whether this breaks Android projects 😢

andreisilviudragnea:
Yes, I would expect someone to use it outside of Mockito. Maybe just not using it internally anymore.

src/main/java/org/mockito/internal/invocation/InvocationsFinder.java
salujayatharth: Isnt the existing:
- more readable
- performant, _getLast_ is O(1) for array and linked list
Sorry for the silly question.

TimvdLippe:
Sadly we can't adopt this PR for now, given that it breaks our Android users. However much I would **love** to merge, it sadly breaks too many users for now. Hopefully we can revisit this in the future once the Android ecosystem has caught up with Java.

Apologies for the late response.

andreisilviudragnea:
@TimvdLippe What about still deprecating them and leaving them around? What did you mean by `our Android users`. I want to check those usages to see if the solution in #2593 is good.

PR #1928, opened 2020-05-08
JUnit 5 strict stubs check should not suppress the regular test failure

andreisilviudragnea:
If the test fails, MockitoExtension should not check for strict stubs at the end of the test, because the possible UnnecessaryStubbingException will end up as a suppressed exception on the test's initial failure.

### DIFF
--- a/subprojects/junit-jupiter/src/test/java/org/mockitousage/StrictnessTest.java
+++ b/subprojects/junit-jupiter/src/test/java/org/mockitousage/StrictnessTest.java
@@ -4,6 +4,12 @@
  */
 package org.mockitousage;
 
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+import static org.junit.platform.engine.discovery.DiscoverySelectors.selectClass;
+
+import java.util.function.Function;
+
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


mockitoguy:
Good idea. Will review shortly. Thank you for the contribution!

subprojects/junit-jupiter/src/test/java/org/mockitousage/StrictnessTest.java
```diff
@@ -53,6 +53,28 @@ void session_checks_for_strict_stubs() {
...
+        }
+    }
+
+    @Test
+    void session_does_not_check_for_strict_stubs_on_test_failure() {
+        TestExecutionResult result = invokeTestClassAndRetrieveMethodResult(StrictStubsNotReportedOnTestFailure.class);
+
+        assertThat(result.getStatus()).isEqualTo(TestExecutionResult.Status.FAILED);
+        Throwable throwable = result.getThrowable().get();
+        assertThat(throwable).isInstanceOf(AssertionError.class);
```
mockitoguy, line 74: Can you make the assertion stronger by using a more specific exception? You can create a custom exception type in the test.

andreisilviudragnea: Ok, I see you already did it.

mockitoguy: Yes ;)

Issue #1912, opened 2020-04-18
Different mocks are used for @Mock and @InjectMock in the same test class with JUnit 5 extension when using `SpringExtension.class` along with `MockitoExtension.class`


karthikaiselvan:
> I'm seeing this same bug when using `SpringExtension.class` along with `MockitoExtension.class`:
> 
> ```
> @ExtendWith({MockitoExtension.class, SpringExtension.class})
> public class InjectMocksTest {
>   static class TacticalStation {}
> 
>   static class SpaceShip {
>     private final TacticalStation tacticalStation;
> 
>     public SpaceShip(final TacticalStation tacticalStation) {
>       this.tacticalStation = tacticalStation;
>     }
> 
>     public TacticalStation getTacticalStation() {
>       return tacticalStation;
>     }
>   }
> 
>   @InjectMocks
>   private SpaceShip spaceShip;
> 
>   @Mock
>   private TacticalStation tacticalStation;
> 
>   @Test
>   public void shouldInjectMocks() {
>     assertThat(tacticalStation).isNotNull();
>     assertThat(spaceShip.getTacticalStation()).isEqualTo(tacticalStation);
>     //Bug ↑ - different mocks are use in @Mock and @InjectMock
>   }
> }
> ```
> 
> This is using Mockito 3.1.0, JUnit 5.6.0, and Spring Boot 5.3.0.RELEASE.

Yes I am also facing this issue. Mine is also exactly the same scenario. Below are my dependencies,
org.junit.jupiter:junit-jupiter-engine:5.6.0
org.springframework.boot:spring-boot-starter:2.2.6.RELEASE
org.springframework.boot:spring-boot-starter-test:2.2.6.RELEASE
org.mockito:mockito-junit-jupiter:3.1.0

_Originally posted by @karthikaiselvan in https://github.com/mockito/mockito/issues/1346#issuecomment-615879855_

andreisilviudragnea:
The reason why the test fails is that both `MockitoTestExecutionListener` from `spring-boot-test` and the `MockitoExtension` call `MockitoAnnotations.init()` on the test class instance.

Only the first `init()` call sets the value of the field `spaceShip`, with an object that refers the `@Mock`-annotated field `tacticalStation` injected by the first `init()` call.

[...]

Issue #1882, opened 2020-03-03
Erroneously failed tests using @InjectMocks with @MockitoExtension


iamdaniu:
Using Mockito 2.23.4 on JUnit 5, Jupiter version 5.6.0.

We encountered the following construct which erroneously fails unit tests.

Two separate interfaces:

```
interface InterfaceOne {
    String getValueOne();
}

interface InterfaceTwo {
    String getValueTwo();
}
```

and one class implementing both of them.

```
class ImplementorBoth implements InterfaceOne, InterfaceTwo {
    private String valueOne;
    private String valueTwo;

    public String getValueOne() {
        return valueOne;
    }
    public String getValueTwo() {
        return valueTwo;
    }
}
```

Now we have a service that needs both interfaces. It initially only needed one of them, but when we added the other, we injected the actual implementation class rather than the interface.

```
public class ExampleService {
    private final InterfaceOne oneProvider;
    private final InterfaceTwo twoProvider;

    public ExampleService(InterfaceOne oneProvider, ImplementorBoth twoProvider) {
        this.oneProvider = oneProvider;
        this.twoProvider = twoProvider;
    }

    String serviceMethod() {
        if (oneProvider.getValueOne() != null && twoProvider.getValueTwo() != null) {
            return oneProvider.getValueOne() + twoProvider.getValueTwo();
        }
        return "";
    }
}
```

The above is the MCVE version of our actual code. The code behaves correctly, but it incorrectly breaks the following test.

```
@ExtendWith(MockitoExtension.class)
class ExampleServiceTest {
    @InjectMocks private ExampleService sut;

    @Mock private InterfaceOne interfaceOne;
    @Mock private ImplementorBoth interfaceTwo;

    @Test void serviceMethodOneTwo() {
        when(interfaceOne.getValueOne()).thenReturn("one");
        when(interfaceTwo.getValueTwo()).thenReturn("two");
        assertThat(sut.serviceMethod()).isEqualTo("onetwo");
    }

    @Test void serviceMethodYesNo() {
        when(interfaceOne.getValueOne()).thenReturn("yes");
        when(interfaceTwo.getValueTwo()).thenReturn("no");
        assertThat(sut.serviceMethod()).isEqualTo("yesno");
    }

    @Test void testNullOne() {
        when(interfaceTwo.getValueTwo()).thenReturn("no");
        assertThat(sut.serviceMethod()).isEmpty();
    }
}
```
Note that the individual test cases pass; they only fail when run together.

[...]

PR #1936, opened 2020-05-20
[Proposal] New InjectMocksNewInstance annotation

andreisilviudragnea:
I have discovered the `@InjectMocks` annotation yesterday and I was eager to use it. However, I have learnt very fast of its many problems.

One of them is the behavior of not overriding the value of a non-null field, no matter if it is already initialized by the test code or when calling `MockitoAnnotations.init()` more than once on a test instance.

Sometimes, calling `MockitoAnnotations.init()` more than once on the same test instance is accidental, as in #1912, or intended and unavoidable. Two examples are:
- using TestNG with `MockitoSession`. In TestNG, the same test instance is reused by all test methods. Using `MockitoSession` in a `@BeforeMethod` and `@AfterMethod` in order to enable Strict Stubs support and `MockitoAnnotations.init()` in order to initialize the mocks is common practice. However, if `@InjectMocks` is used, only the first test in the test class will use the mocks injected in the `@InjectMocks`-annotated field.
- using JUnit 5 in a test class annotated with `@TestInstance(TestInstance.Lifecycle.PER_CLASS)`, which is essentially the same scenario as TestNG example above.

I propose adding a new `@InjectMocksNewInstance` annotation, which always initializes the annotated field with an object having mocks injected by the same logic as `@InjectMocks` (either constructor injection or property and setter injection).

The new annotation does not suffer from the bug-which-became-a-feature from #1631.
It solves the two cases indicated above and also #1912, if used instead of `@InjectMocks`.

However, the new annotation suffers from the same problems as the old one when it comes to the brittle support for constructor injection, as seen in #1882.

This is just a draft of my idea. I created this draft before checking the many issues related to `@InjectMocks`, some of them unknown to me prior to this PR. I believe the intention of `@InjectMocks` is good, but its implementation is too unstable and unsafe to use. I believe that it should be replaced with a new annotation, with more clearly defined semantics, or just deprecated and removed completely, as suggested in #1518. The actual form of the `@InjectMocks` annotation works well only in very specific scenarios.

In my opinion, a big problem of `@InjectMocks` annotation is that it works with Strict Stubs over JUnit 4, it sometimes works with Strict Stubs over JUnit 5 and never works with Strict Stubs over TestNG.

### DIFF
--- a/src/main/java/org/mockito/InjectMocksNewInstance.java
+++ b/src/main/java/org/mockito/InjectMocksNewInstance.java
@@ -1,0 +1,18 @@
+/*
+ * Copyright (c) 2020 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@Documented
+@Target(FIELD)
+@Retention(RUNTIME)
+public @interface InjectMocksNewInstance {
+}
--- a/src/main/java/org/mockito/MockitoAnnotations.java
+++ b/src/main/java/org/mockito/MockitoAnnotations.java
@@ -54,17 +54,17 @@ import org.mockito.plugins.AnnotationEngine;
 public class MockitoAnnotations {
 
     /**
-     * Initializes objects annotated with Mockito annotations for given testClass:
+     * Initializes objects annotated with Mockito annotations for given testInstance:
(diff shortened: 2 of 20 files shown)


TimvdLippe:
This is quite a big feature for `@InjectMocks`, where we generally want to avoid building another dependency injection framework. Given that the maintainers are time-constrained and it implements new injection mechanisms, I think this PR will likely not be merged. Just wanted to give you a heads up on that.

Therefore, I would encourage you to explore alternatives that do not significantly increase the API surface, but could also address the issues you linked.

andreisilviudragnea:
I think I will just try to avoid it completely.

TimvdLippe:
Closing this per https://github.com/mockito/mockito/pull/1936#issuecomment-636465679 Thanks for the suggestion!

PR #1945, opened 2020-06-15
[Travis] Replace JDK 9/10 with 14

TimvdLippe:
9/10 are EOL since September 2018. 14 is the currently supported
Java version.


TimvdLippe:
Yeah I am working on resolving the build issue. We need to upgrade our Gradle version. Sadly that turns out to be a whole rabbithole, but I think I am close.

epeee:
@TimvdLippe it looks like we also have to use errorprone 2.4.0 since this is the first version which actually supports java 14.

> Task :compileJava
/home/travis/build/mockito/mockito/src/main/java/org/mockito/invocation/InvocationOnMock.java:57: error: An unhandled exception was thrown by the Error Prone static analysis plugin.
    <T> T getArgument(int index);
          ^
     Please report this at https://github.com/google/error-prone/issues/new and include the following:
     error-prone version: 2.3.2
     BugPattern: TypeParameterUnusedInFormals
     Stack Trace:
     java.lang.NoSuchFieldError: bound
  	at com.google.errorprone.bugpatterns.TypeParameterUnusedInFormals.matchMethod(TypeParameterUnusedInFormals.java:73)
  	at com.google.errorprone.scanner.ErrorProneScanner.visitMethod(ErrorProneScanner.java:890)

TimvdLippe:
@epeee Yes indeed. Were you able to get Gradle working on your machine?!

epeee:
yes. Furthermore, also the latest travis build (https://travis-ci.org/github/mockito/mockito/jobs/698954197) is failing with the message posted previously.

TimvdLippe:
Interesting, because I was looking at https://travis-ci.org/github/mockito/mockito/jobs/698954215 which is failing for a completely different reason :O

TimvdLippe:
I wonder maybe if the build cache is not working on one job, but is working on the other. Not sure how to diagnose this issue though.

TimvdLippe:
Oh, on the push Travis build, both tasks failed on the ErrorProne issue, but not the Gradle compilation issue. So out of 3 tasks, only the 1 on the PR build is failing.

TimvdLippe:
I purged the corrupt Travis cache and was able to get a passing build. See https://github.com/gradle/gradle/issues/13444#issuecomment-645642759 for more info.

PR #1946, opened 2020-06-15
Move spotless check to separate build task

TimvdLippe:
This means the tests can run in parallel of the formatting changes.
Therefore, formatting changes would not block obtaining your test
results, which should hopefully reduce the amount of Travis builds
necessary to work on a community PR.

This is also in preparation of #1934 which requires spotless to
run on JDK11 minimum.


.travis.yml
```diff
@@ -28,6 +28,9 @@ matrix:
     env: SKIP_RELEASE=true
   - jdk: openjdk14
     env: SKIP_RELEASE=true MOCK_MAKER=mock-maker-inline
+  - jdk: openjdk11
+    script: ./gradlew spotlessCheck
```
mockitoguy, line 32: Does ^^ mean that on openjdk11 we only run "spotlessCheck" task?

.travis.yml
```diff
@@ -28,6 +28,9 @@ matrix:
...
+    after_success: true
```
mockitoguy, line 33: What does ^^ do?

.travis.yml
TimvdLippe: It means that we have 1 task that only runs spotless, and that task runs on opendjk11. We still have the other openjdk11 builds to verify that the tests pass etc... I will add a comment for clarification.

.travis.yml
TimvdLippe: To make sure we don't run the coverage report task, since that's not what spotless generates. I will add a clarifying comment.

PR #1947, opened 2020-06-15
Replace anonymous class with lambda expression

bhawna94:
[Refactor] This PR replaces the anonymous class with a lambda expression


bhawna94:
@TimvdLippe @vin @davidkarlsen @tokuhirom 
Can you please review the change and let me know your thoughts :)

PR #1948, opened 2020-06-15
Update gradle 6

TimvdLippe:
This included a couple of breaking changes:
- The build plugin is now part of the Gradle Enterprise plugin
- The osgi plugin was removed, for which we have to use the bndtools
plugin instead
- Kotlin version has to be updated, including their coroutine
library versions

Gradle 6 is required for us to build on Java 14.

Supersedes and closes #1923


.travis.yml
```diff
@@ -34,6 +34,9 @@ matrix:
     env: SKIP_RELEASE=true
   - jdk: openjdk11
     env: SKIP_RELEASE=true MOCK_MAKER=mock-maker-inline
+  - jdk: openjdk8
+    script: ./check_reproducibility.sh
```
mrwilson, line 38:
```suggestion
  - name: "Check reproducibility" 
    jdk: openjdk8
    script: ./check_reproducibility.sh
```

Taking a look in Travis right now, it looks like "just" another Jdk 8 build and this would allow people to identify this particular job at a glance if it starts failing.

mrwilson:
Ah, I've hit this before (doing stuff with `GradleExec` in 6).

Running it as a separate task is a good idea too 👍 

I've added one small suggestion.

This script works for the JUnit 5 project, so if it's passing the builds then I'm content that it's behaving as expected.

TimvdLippe:
Thanks for the review, I will fix the Travis name. Meanwhile, CodeCov is not happy with this Gradle change, so I gotta figure why that is broken. (This happened with a previous Gradle upgrade as well)

.travis.yml
TimvdLippe: I am going to do this in a follow-up, as we can do some more cleanup of the Travis tasks. Thanks for the suggestion 😄

TimvdLippe:
Coverage changes are as expected. We no longer include the android subproject in our files and the Kotlin experimental test suite has been upgraded. We can clean that up in a follow-up PR.

Given that this is mostly a mechanical change, I am going to go ahead and merge this. There is plenty of things we can clean up in follow-up PRs, but this PR is already getting quite big.

PR #1949, opened 2020-06-16
Fixes #1712 : prepend description to AssertionError thrown in verification

### LINKED ISSUE #1712: "description" not printing when verify args don't match
Mockito version 2.27.0

I have a test that uses `description` inside a `verify` method. The `description` is correctly printing when the test fails because we don't interact with the mock

```
import static org.mockito.Mockito.description;
import static org.mockito.Mockito.spy;

import org.junit.Test;

public class ExampleTest {
    @Test
    public void shouldFailWithDescription() {
        Dependency dependency = spy(Dependency.class);
        SystemUnderTest systemUnderTest = new SystemUnderTest();
        systemUnderTest.doSomething(dependency);
        verify(dependency, description("Failed to call doSomethingElse")).doSomethingElse(false);
    }

    static class SystemUnderTest {
        void doSomething(Dependency dependency) {
            // dependency.doSomethingElse(true); // <---- TEST FAILS BECAUSE WE DON'T INTERACT WITH MOCK
        }
    }

    static class Dependency {
        void doSomethingElse(boolean value) {
            // ...
        }
    }
}
```

Results in this failure message:

```
org.mockito.exceptions.base.MockitoAssertionError: Failed to call doSomethingElse <---- SEE OUR DESCRIPTION

Wanted but not invoked:
dependency.doSomethingElse(false);
-> at com.example.ExampleTest$Dependency.doSomethingElse(ExampleTest.java:28)
Actually, there were zero interactions with this mock.
```

In this case the test failed because there were no interactions with this mock. However if we update the SUT so that it calls the method with the wrong argument:

```
public class ExampleTest {
    @Test
    public void shouldFailWithDescription() {
        Dependency dependency = spy(Dependency.class);
        SystemUnderTest systemUnderTest = new SystemUnderTest();
        systemUnderTest.doSomething(dependency);
        verify(dependency, description("Failed to call doSomethingElse")).doSomethingElse(false);
    }

    static class SystemUnderTest {
        void doSomething(Dependency dependency) {
             dependency.doSomethingElse(true);  // <---- TEST FAILS BECAUSE WE PASS WRONG ARGS
        }
    }

    static class Dependency {
        void doSomethingElse(boolean value) {
            // ...
        }
    }
}
```

The test still fails. But this time it **does not include the description**:

```
Argument(s) are different! Wanted: <----- WE SHOULD SEE DESCRIPTION
dependency.doSomethingElse(false);
-> at com.example.ExampleTest$Dependency.doSomethingElse(ExampleTest.java:28)
dependency.doSomethingElse(true);
-> at com.example.ExampleTest$SystemUnderTest.doSomething(ExampleTest.java:21)

Comparison Failure: 
Expected :dependency.doSomethingElse(false);
Actual   :dependency.doSomethingElse(true);
```

adrianriley:
Extended MockitoAssertionError to wrap instance of AssertionError in addition to MockitoAssertionError, so that the various possible exceptions thrown for non-matching arguments have description prepended to the message.

### DIFF
--- a/src/main/java/org/mockito/exceptions/base/MockitoAssertionError.java
+++ b/src/main/java/org/mockito/exceptions/base/MockitoAssertionError.java
@@ -46,6 +46,18 @@ public class MockitoAssertionError extends AssertionError {
         unfilteredStackTrace = error.getUnfilteredStackTrace();
     }
 
+    /**
+     * Creates a copy of the given assertion error with the custom failure message prepended.
+     * @param error The assertion error to copy
+     * @param message The custom message to prepend
+     * @since 3.3.13
+     */
+    public MockitoAssertionError(AssertionError error, String message) {
+        super(message + "\n" + error.getMessage());
+        unfilteredStackTrace = error.getStackTrace();
+        super.setStackTrace(unfilteredStackTrace);
+    }
+
     public StackTraceElement[] getUnfilteredStackTrace() {
         return unfilteredStackTrace;
     }
--- a/src/main/java/org/mockito/internal/verification/Description.java
+++ b/src/main/java/org/mockito/internal/verification/Description.java
@@ -41,6 +41,8 @@ public class Description implements VerificationMode {
 
         } catch (MockitoAssertionError e) {
             throw new MockitoAssertionError(e, description);
+        } catch (AssertionError e) {
+            throw new MockitoAssertionError(e, description);
         }
     }
 }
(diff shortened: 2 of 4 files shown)


TimvdLippe:
Oh, it seems like you need to run `./gradlew spotlessApply` locally to fix the formatting issues.

TimvdLippe:
I think you need to rebase on master to pick up the latest changes. Your Spotless configuration locally appears to be out-of-date.

TimvdLippe:
Are you running `./gradlew spotlessApply`? I can take a look on Monday why it is not formatting as expected.

adrianriley:
Thanks Tim, I've rebased now, hopefully should fix the formatting issue. Just squashed the commits.

PR #1954, opened 2020-06-18
Move performRelease step into separate Travis job

TimvdLippe:
Currently, all jobs run this step. However, only one of them actually
performs the logic. This means that all jobs are spending ~1 min
doing nothing, except the one that actually performs the task at hand.

Since we only merge PRs that are passing the build, we don't actually
need to gate our release procedure on passing tests.

All in all, this should reduce our Travis CI build time by about
1 minute per job, and make it clearer that we have a separate
release task, as well as cleanup our env declarations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,15 +19,13 @@ matrix:
   include:
   - jdk: openjdk8
   - jdk: openjdk8
-    env: SKIP_RELEASE=true MOCK_MAKER=mock-maker-inline
+    env: MOCK_MAKER=mock-maker-inline
   - jdk: openjdk11
-    env: SKIP_RELEASE=true
   - jdk: openjdk11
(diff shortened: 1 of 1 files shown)


TimvdLippe:
You can see the Travis job names and times in https://travis-ci.org/github/mockito/mockito/builds/699652692

TimvdLippe:
> This means that we can publish a version of mocking without running tests ;-)

Given that we use the PR-model where all tests must pass, I don't think that is the case. Since we are paying a cost for running the release tasks on every job, I think we can extract them out.

E.g. if there would be a possibility where the tests would fail on `release/3.x` and the release should not happen, then that would mean we haven't gone through the proper PR-model. I would then rather work on solutions to make that model work. WDYT?

Issue #1802, opened 2019-10-10
Exception "The mock object was garbage collected."


basdebakker:
We received the following exception while running tests using Mockito, either version 3.0.0 or 3.1.0, and I'm filing this bug as requested:
```
Exception in thread "main" java.lang.IllegalStateException: The mock object was garbage collected. This should not happen in normal circumstances when using public API. Typically, the test class keeps strong reference to the mock object and it prevents getting the mock collected. Mockito internally needs to keep weak references to mock objects to avoid memory leaks for certain types of MockMaker implementations. If you see this exception using Mockito public API, please file a bug. For more information see issue #1313.
	at org.mockito.internal.invocation.mockref.MockWeakReference.get(MockWeakReference.java:32)
	at org.mockito.internal.invocation.InterceptedInvocation.getMock(InterceptedInvocation.java:106)
	at org.mockito.internal.stubbing.InvocationContainerImpl.invokedMock(InvocationContainerImpl.java:157)
	at org.mockito.internal.stubbing.OngoingStubbingImpl.<init>(OngoingStubbingImpl.java:22)
	at org.mockito.internal.handler.MockHandlerImpl.handle(MockHandlerImpl.java:83)
	at org.mockito.internal.handler.NullResultGuardian.handle(NullResultGuardian.java:29)
	at org.mockito.internal.handler.InvocationNotifierHandler.handle(InvocationNotifierHandler.java:35)
	at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.doIntercept(MockMethodInterceptor.java:61)
	at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.doIntercept(MockMethodInterceptor.java:49)
	at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor$DispatcherDefaultingToRealMethod.interceptAbstract(MockMethodInterceptor.java:126)
	at org.mockito.codegen.Runnable$MockitoMock$1840149894.run(Unknown Source)
	at MockitoTest.runTest(MockitoTest.java:13)
	at MockitoTest.main(MockitoTest.java:6)
```

I've reduced the problem to the following test program:
```java
import org.mockito.*;

public class MockitoTest {

  public static void main(String[] args) {
    while (true) runTest();
  }

  private static void runTest() {
    Node list = createList();
    while (list != null) {
      Node next = list.next;
      list.object.run();
      list = next;
    }
  }

  private static Node createList() {
    Node node = null;
[...]

PR #1955, opened 2020-06-19
Mockito #1013: Defines and implements API for static mocking.

### LINKED ISSUE #1013: Enable mocking static methods in Mockito
# Static methods mocking with Mockito

This is a placeholder ticket for enabling mocking static methods in Mockito. Example action items that are totally negotiable and can ran in parallel. We're looking for someone who can lead this effort.

- [x] Research + discuss whether it is a good idea to enable static methods mocking in Mockito. The theory is that it is useful for legacy code, which is most code in the world.
- [x] Research on how other mocking frameworks do that and whether it is considered useful feature for their users.
- [x] Design and present for discussion an API for static mocking (slightly relevant ticket: #643)
- [x] Work with @raphw / ByteBuddy to come up with hacky prototype (the hackier, the better!)
- [x] Mold the prototype with the API, remove enough rough edges so that the feature is good enough for incubating rollout
- [ ] SHIP IT!

raphw:
Fixes #1013 - allows for static method mocking in Mockito.

### DIFF
--- a/src/main/java/org/mockito/Captor.java
+++ b/src/main/java/org/mockito/Captor.java
@@ -15,9 +15,16 @@ import java.lang.annotation.*;
  *
  *    &#64;Captor ArgumentCaptor&lt;AsyncCallback&lt;Foo&gt;&gt; captor;
  *
+ *    private AutoCloseable closeable;
+ *
  *    &#64;Before
- *    public void init(){
- *       MockitoAnnotations.initMocks(this);
+ *    public void open() {
+ *       MockitoAnnotations.openMocks(this);
+ *    }
+ *
+ *    &#64;After
+ *    public void release() throws Exception {
+ *       closeable.close();
  *    }
  *
  *    &#64;Test public void shouldDoSomethingUseful() {
--- a/src/main/java/org/mockito/Captor.java
+++ b/src/main/java/org/mockito/Captor.java
@@ -19,7 +19,7 @@ import java.lang.annotation.*;
  *
  *    &#64;Before
  *    public void open() {
- *       MockitoAnnotations.openMocks(this);
+ *       closeable = MockitoAnnotations.openMocks(this);
  *    }
  *
  *    &#64;After
--- a/src/main/java/org/mockito/MockedStatic.java
+++ b/src/main/java/org/mockito/MockedStatic.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2007 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito;
+
+import org.mockito.stubbing.OngoingStubbing;
+import org.mockito.verification.VerificationMode;
+
+import static org.mockito.Mockito.*;
+
+/**
+ * Represents an active mock of a type's static methods. The mocking only affects the thread
+ * on which this static mock was created and it is not safe to use this object from another
+ * thread. The static mock is released when this object's {@link MockedStatic#close()} method
+ * is invoked. If this object is never closed, the static mock will remain active on the
+ * initiating thread. It is therefore recommended to create this object within a try-with-resources
+ * statement unless when managed explicitly, for example by using a JUnit rule or extension.
+ * <p>
+ * If the {@link Mock} annotation is used on fields or method parameters of this type, a static mock
+ * is created instead of a regular mock. The static mock is activated and released upon completing any
+ * relevant test.
+ *
+ * @param <T> The type being mocked.
+ */
+@Incubating
+public interface MockedStatic<T> extends AutoCloseable {
+
+    /**
+     * See {@link Mockito#when(Object)}.
+     */
+    <S> OngoingStubbing<S> when(Verification verification);
+
+    /**
+     * See {@link Mockito#verify(Object)}.
+     */
+    default void verify(Verification verification) {
+        verify(times(1), verification);
+    }
+
+    /**
+     * See {@link Mockito#verify(Object, VerificationMode)}.
+     */
+    void verify(VerificationMode mode, Verification verification);
+
+    /**
+     * See {@link Mockito#reset(Object[])}.
+     */
+    void reset();
+
+    /**
+     * See {@link Mockito#clearInvocations(Object[])}.
+     */
+    void clearInvocations();
+
+    /**
+     * {@link Mockito#verifyNoMoreInteractions(Object...)}.
+     */
+    void verifyNoMoreInteractions();
+
+    /**
+     * See {@link Mockito#verifyNoInteractions(Object...)}.
+     */
+    void verifyNoInteractions();
+
+    @Override
+    void close();
+
+    /**
+     * Releases this static mock and is non-operational if already released.
+     */
+    void closeOnDemand();
+
+    interface Verification {
+
+        void apply() throws Throwable;
+    }
+}
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -105,6 +105,7 @@ import org.mockito.verification.VerificationWithTimeout;
  *      <a href="#45">45. New JUnit Jupiter (JUnit5+) extension</a><br/>
  *      <a href="#46">46. New <code>Mockito.lenient()</code> and <code>MockSettings.lenient()</code> methods (Since 2.20.0)</a><br/>
  *      <a href="#47">47. New API for clearing mock state in inline mocking (Since 2.25.0)</a><br/>
+ *      <a href="#48">48. New API for mocking static methods (Since 3.4.0)</a><br/>
  * </b>
  *
  * <h3 id="0">0. <a class="meaningful_link" href="#mockito2" name="mockito2">Migrating to Mockito 2</a></h3>
@@ -461,7 +462,7 @@ import org.mockito.verification.VerificationWithTimeout;
  * runner:
  *
  * <pre class="code"><code class="java">
- * MockitoAnnotations.initMocks(testClass);
+ * MockitoAnnotations.openMocks(testClass);
  * </code></pre>
  *
  * You can use built-in runner: {@link MockitoJUnitRunner} or a rule: {@link MockitoRule}.
@@ -863,7 +864,7 @@ import org.mockito.verification.VerificationWithTimeout;
  * should only use partial mocks as a last resort. See point 16 about partial mocks.
  *
  * <p>
- * All new annotations are <b>*only*</b> processed on {@link MockitoAnnotations#initMocks(Object)}.
+ * All new annotations are <b>*only*</b> processed on {@link MockitoAnnotations#openMocks(Object)}.
  * Just like for &#064;{@link Mock} annotation you can use the built-in runner: {@link MockitoJUnitRunner} or rule:
  * {@link MockitoRule}.
  * <p>
@@ -907,7 +908,7 @@ import org.mockito.verification.VerificationWithTimeout;
  * Mockito will now try to instantiate &#064;{@link Spy} and will instantiate &#064;{@link InjectMocks} fields
  * using <b>constructor</b> injection, <b>setter</b> injection, or <b>field</b> injection.
  * <p>
- * To take advantage of this feature you need to use {@link MockitoAnnotations#initMocks(Object)}, {@link MockitoJUnitRunner}
+ * To take advantage of this feature you need to use {@link MockitoAnnotations#openMocks(Object)}, {@link MockitoJUnitRunner}
  * or {@link MockitoRule}.
  * <p>
  * Read more about available tricks and the rules of injection in the javadoc for {@link InjectMocks}
@@ -1144,7 +1145,7 @@ import org.mockito.verification.VerificationWithTimeout;
  *
  * <ul>
  *     <li>Annotating the JUnit test class with a <code>&#064;{@link org.junit.runner.RunWith}({@link MockitoJUnitRunner}.class)</code></li>
- *     <li>Invoking <code>{@link MockitoAnnotations#initMocks(Object)}</code> in the <code>&#064;{@link org.junit.Before}</code> method</li>
+ *     <li>Invoking <code>{@link MockitoAnnotations#openMocks(Object)}</code> in the <code>&#064;{@link org.junit.Before}</code> method</li>
  * </ul>
  *
  * Now you can choose to use a rule :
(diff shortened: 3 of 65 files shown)


raphw:
I thought about it but I am pretty confident that this is what it's gonna be! I suggest we put the APIs as incubating as we have done with the inline mock maker. If we commit to the API, even in a separate artifact, I think the Android folks will still go for it and have a similar expectation on it's stability.

You think a separate artifact would be cleaner?

TimvdLippe:
I mostly prefer a conservative route if we have that option. But keeping as Incubating is okay with me.

raphw:
I think I'd prefer the single artifact. It's fully opt-in and does not touch any existing code. I think this way it will be used more and we'll find out the adoption quicker.

mockitoguy:
I just saw this and cannot believe my eyes :-) Will try to review shortly!!!

raphw:
Awesome, just discovered a remaining recursion on `thenCallRealMethod` but I think I already know how to fix it.

raphw:
And it works, too!

src/main/java/org/mockito/MockedStatic.java
```diff
@@ -0,0 +1,209 @@
...
+        }
+    }
+
+    private void assertNotClosed() {
+        if (closed) {
+            throw new MockitoException(
+                    join(
+                            "The static mock created at",
+                            location.toString(),
+                            "is already resolved and cannot longer be used"));
```
TimvdLippe, line 200:
```suggestion
                            "is already resolved and can no longer be used"));
```

src/main/java/org/mockito/Mockito.java
```diff
@@ -1541,6 +1542,29 @@
...
+ *
+ * To make sure a static mock remains temporary, it is recommended to define the scope within a try-with-resources construct.
+ * In the following example, the <code>Foo</code> type's static method would return <code>foo</code> unless mocked:
+ *
+ * <pre class="code"><code class="java">
+ * assertEquals("foo", Foo.method());
+ * try (MockedStatic<Foo> mocked = mockStatic(Foo.class)) {
+ * mocked.when(Foo::method).thenReturn("bar");
+ * assertEquals("bar", Foo.method());
+ * mocked.verify(Foo::method);
```
TimvdLippe, line 1560:
```suggestion
 *     mocked.when(Foo::method).thenReturn("bar");
 *     assertEquals("bar", Foo.method());
 *     mocked.verify(Foo::method);
```

src/main/java/org/mockito/MockitoAnnotations.java
```diff
@@ -37,8 +37,14 @@
  *
  *   public class SampleBaseTestCase {
  *
+ *       private AutoCloseable closeable;
+ *
  *       &#064;Before public void initMocks() {
- *           MockitoAnnotations.initMocks(this);
+ *           closeable = MockitoAnnotations.initMocks(this);
```
TimvdLippe, line 43: This is potentially a breaking change, if compilations downstream require the result of any `AutoCloseable` to be used. I would have to check for example internally if that is the case. Can we maybe separate this out into a separate method called `MockitoAnnotations.initWithStaticMocks`? That would make it clearer that a particular test is allowed to use static mocks. By using a separate method, users can write static analyses that verify that `.close` is called.

src/main/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMaker.java
```diff
@@ -300,19 +314,41 @@ public MockHandler getHandler(Object mock) {
     public void resetMock(Object mock, MockHandler newHandler, MockCreationSettings settings) {
         MockMethodInterceptor mockMethodInterceptor =
                 new MockMethodInterceptor(newHandler, settings);
-        mocks.put(mock, mockMethodInterceptor);
-        if (mock instanceof MockAccess) {
-            ((MockAccess) mock).setMockitoInterceptor(mockMethodInterceptor);
+        if (mock instanceof Class<?>) {
+            Map<Class<?>, MockMethodInterceptor> interceptors = mockedStatics.get();
```
TimvdLippe, line 318: I am not really following this refactoring. Could you elaborate on why this is necessary? Would it maybe be possible to separate these changes into its own PR to ease reviewing?

src/main/java/org/mockito/MockitoAnnotations.java
raphw: I thought about this, too. It is source compatible but not binary compatible. However, I do not think that this is a very commonly used API by derivative tools and direct users do not have to apply any changes until using static mocks in their custom tooling what likely requires them to revisit this API. In a way, explicit life-cycles are a new concept in Mockito that get introduced by this change, it will require some upstream adoption and in such a case, I think that breakage is the most explicit way to communicate this to users; I always prefer this over silent or runtime failure. I suggest therefore to make this visible in the release notes and by the versioning scheme via releasing a new feature version when merging this.

src/main/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMaker.java
raphw: It's easier when looking at the raw file, the changefile cuts parts of the interesting segment. Basically, to avoid breaking the existing implementation, static mocks are represented by the `Class` value for the static mock. This is also done in the listerners etc. what is possible since it is currently impossible to mock a `Class` instance, even with the inline-mock maker.

We cannot store static mocks in the same concurrent map as the regular mocks since they are thread local. Therefore, we always need to make a distinction if a mock is an instance of `Class` or not to see if a static or a regular mock is requested. The advantage of this solution is that the current API can mainly remain untouched, keeping the changeset to a minimum.

raphw:
I thought about splitting the `MockitoAnnotations` into offering two methods. However, since we introduce a life-cycle, the "old" method can no longer be used safely. It's a rather uncommon API to use manually, therefore I believe that a breaking change is a good way to notify users that they need to pay attention. Most people do however use the JUnit runner, rule and extension, I think and won't be affected.

I considered splitting up the PR but that way I could not implement static mocks as a running feature without putting all pieces together. I therefore suggest that we release it as one change, that way Mockito stays consistent in its offered features and we hopefully get frictionless adoption. I am afraid that this change has a certain inherent complexity that would not go away even when splitting up.

TimvdLippe:
> It's a rather uncommon API to use manually, therefore I believe that a breaking change is a good way to notify users that they need to pay attention.

I have seen numerous usages in Google. Before we make this change, I would like to verify that this is indeed safe to do. I am not sure when I have time to do so, but I think someday next week should be possible.

raphw:
I have not considered wide usage. In this case, I think it's better to deprecate `initMocks` in favor of `openMocks` where the former does not misbehave unless you are using the new static mocks. This way, there is no API breakage and Mockito continues to well-behave for unchanged user code.

I think such a life-cycle will be useful for other cases, too. For instance, I think it opens a solution for https://github.com/mockito/mockito/issues/1802 where I otherwise cannot see a solution.

TimvdLippe:
`initMocks` is used widely in the Android ecosystem, as it (allegedly) has better integration with ClassLoader like Robolectric. I think since then, we have addressed the issues in Mockito, but the usage of `initMocks` remains prevalent.

src/main/java/org/mockito/Captor.java
```diff
@@ -15,9 +15,16 @@
  *
  *    &#64;Captor ArgumentCaptor&lt;AsyncCallback&lt;Foo&gt;&gt; captor;
  *
+ *    private AutoCloseable closeable;
+ *
  *    &#64;Before
- *    public void init(){
- *       MockitoAnnotations.initMocks(this);
+ *    public void open() {
+ *       MockitoAnnotations.openMocks(this);
```
TimvdLippe, line 22:
```suggestion
 *       closeable = MockitoAnnotations.openMocks(this);
```

raphw: Awesome! We could also update Byte Buddy some time soon to get initial Java 16 support.

src/main/java/org/mockito/plugins/MockMaker.java
```diff
@@ -108,6 +109,26 @@
...
+     * @param settings Mock creation settings like type to mock, extra interfaces and so on.
+     * @param handler See {@link org.mockito.invocation.MockHandler}.
+     *                <b>Do not</b> provide your own implementation at this time. Make sure your implementation of
+     *                {@link #getHandler(Object)} will return this instance.
+     * @param <T> Type of the mock to return, actually the <code>settings.getTypeToMock</code>.
+     * @return A control for the static mock.
+     * @since 3.4.0
+     */
+    @Incubating
+    <T> StaticMockControl<T> createStaticMock(
```
TimvdLippe, line 129: This is a breaking change and breaks Dexmaker. I recommend adding a default implementation with ```throw new MockitoException("The subclass byte code generator cannot create static mocks");```

src/main/java/org/mockito/MockedStatic.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Override
+    void close();
+
+    /**
+     * Releases this static mock and is non-operational if already released.
+     */
+    void closeOnDemand();
+
+    @FunctionalInterface
```
TimvdLippe, line 74: This is not available on Android builds and therefore will fail. I think we can remove this annotation and the rest will still work?

raphw: You are right, I commented it out. It is mainly for documentation purposes and to fail compilation if the interfaces does not define a single abstract method.

src/main/java/org/mockito/plugins/MockMaker.java
raphw: You are right, I added a default method which throws an explaining exception. I also had to move the static mock control into a non-internal package. Otherwise nobody else could implement the API.

src/main/java/org/mockito/Captor.java
TimvdLippe: Could you apply this suggestion?

src/main/java/org/mockito/MockedStatic.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Override
+    void close();
+
+    /**
+     * Releases this static mock and is non-operational if already released.
+     */
+    void closeOnDemand();
+
+    // @FunctionalInterface - not supported on Android
```
TimvdLippe, line 74:
```suggestion
```

TimvdLippe:
@raphw You also need to run `./gradlew spotlessApply` locally to format the sourcecode.

src/main/java/org/mockito/internal/configuration/MockAnnotationProcessor.java
```diff
@@ -41,6 +50,29 @@ public static Object processAnnotationForMock(Mock annotation, Class<?> type, St
...
+            return Mockito.mockStatic(inferStaticMock(genericType.get(), name), mockSettings);
+        } else {
+            return Mockito.mock(type, mockSettings);
+        }
+    }
+
+    private static Class<?> inferStaticMock(Type type, String name) {
+        if (type instanceof ParameterizedType) {
+            ParameterizedType parameterizedType = (ParameterizedType) type;
+            return (Class<?>) parameterizedType.getRawType();
```
andrei-ivanov, line 64: Shouldn't this use `getActualTypeArguments()[0]` instead?
It's failing for me now since `getRawType()` returns `MockedStatic`

TimvdLippe: Do you mind opening a PR with regression test + fix? Thanks!

andrei-ivanov: Hmm, seems this was easy to fix locally with my suggestion, but I now get another error, which I don't know why it happens:
```
org.mockito.exceptions.misusing.NotAMockException: Argument passed to Mockito.mockingDetails() should be a mock, but is an instance of class java.lang.Class!

	at org.mockito.junit.jupiter.MockitoExtension.afterEach(MockitoExtension.java:186)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeAfterEachCallbacks$11(TestMethodTestDescriptor.java:255)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor$$Lambda$316/0000000000000000.invoke(Unknown Source)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeAllAfterMethodsOrCallbacks$12(TestMethodTestDescriptor.java:271)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor$$Lambda$1809/0000000000000000.execute(Unknown Source)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeAllAfterMethodsOrCallbacks$13(TestMethodTestDescriptor.java:271)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor$$Lambda$314/0000000000000000.accept(Unknown Source)
```
The class with the static method ends up there...

raphw: I isolated the error. The problem is that the mocks for static methods are scoped. This means that they are released at the end of the try-with-resources block what implies that the mocks are no longer valid at the end of the method resulting in this error. 

@TimvdLippe For now I think the easiest is to exclude static mocks from the post processing. The more correct approach would be to execute the "after each" logic upon releasing the mock rather then at the end of the test. I will look into it but need to understand the post-processing first.

andrei-ivanov: Btw, my setup was with and without a try-with-resources block, since I understood from the documentation of `@Mock` that it should work with that as well, which would simplify things (and since I already had other fields that were annotated with `@Mock`)

andrei-ivanov:
```java
public class TenantScopeTest {
    @Mock
    private UserPrincipal principal;

    @Mock
    private MockedStatic<UserPrincipalUtil> userPrincipalUtil;

    @Test
    public void opsShouldBeCachedInstancesByNameAndTenant() {
        String tenant = "x";
        when(principal.getMainTenant()).thenReturn(tenant);

        userPrincipalUtil.when(UserPrincipalUtil::requireUserPrincipal).thenReturn(principal);

        //....
}
```

raphw: @andrei-ivanov I have not had the time to test it yet (we should have unit tests for the runners using both mock makers) but maybe the 'runner-fix' branch already fixes the problem.

andrei-ivanov: I still get the `NotAMockException` with the `runner-fix` branch

andrei-ivanov: Let me know when you want to do run some more tests, I'm eager to get rid of PowerMock 😁

jacky1193610322: I also have this problem, when the mockMaker is InlineByteBuddyMockMaker, I use SpyBean to spy a spring bean, and then verify the bean, it throws NotAMockException. I debug and find the class type is not contains mockito, but it contains when I use the ByteBuddyMockMaker.

jacky1193610322:
I also have this problem, when the mockMaker is InlineByteBuddyMockMaker, I use SpyBean to spy a spring bean, and then verify the bean, it throws NotAMockException. I debug and find the class type is not contains mockito, but it contains when I use the ByteBuddyMockMaker. and it seems not a spring bug.

PR #1968, opened 2020-07-12
Fixes #1967: Correctly handle mocks with limited life-cycle in listeners.

### LINKED ISSUE #1967: Static method mocks incompatible with MockitoExtension (NotAMockException)
Hello,

I am experimenting with the newly released functionality of mocking static methods.
It appears that static mocks are not compatible with the `MockitoExtension` designed for JUnit 5.
Following error message is produced when using the extension:
```
org.mockito.exceptions.misusing.NotAMockException: Argument passed to Mockito.mockingDetails() should be a mock, but is an instance of class java.lang.Class!
```
Full Stacktrace: [static-mock-mockito-extension-stacktrace.txt](https://github.com/mockito/mockito/files/4908237/static-mock-mockito-extension-stacktrace.txt)

I have created following minimal example which is composed of a `Utils` class (containing the static method) and a `UtilsUser` class (calls the static Utils method). `UtilsUserTest` tests the functionality of the `UtilsUser` class.

```java
public class Utils {
    public static long millis() {
        return System.currentTimeMillis();
    }
}
```

```java
public class UtilsUser {
    public long use() {
        return Utils.millis();
    }
}
```

```java
import org.mockito.InjectMocks;
import org.mockito.MockedStatic;
import org.mockito.Mockito;

class UtilsUserTest {
    @InjectMocks
    private UtilsUser utilsUser;

    @Test
    void use() {
        try (MockedStatic<Utils> mockedUtils = Mockito.mockStatic(Utils.class)) {
            mockedUtils.when(Utils::millis).thenReturn(17L);
            assertEquals(17L, utilsUser.use());
        }
    }
}
```
When converting this test to a test which does not use the extension (removing `@ExtendWith`, removing `@InjectMocks`, adding `= new UtilsUser()` after the field declaration), the test passes.

Can you tell me if there is something special I have to accomodate for when using the static mocks in JUnit Jupiter?
Or is this indeed a bug (which you can hopefully reproduce)?

I am using mockito-inline and mockito-junit-jupiter in version 3.4.0.

Thanks in advance and thanks for the new static mocks functionality!
David

raphw:
Due to the limited lifetime of static mocks, they cannot be validated within a Mockito session since they might have expired when the validation is applied. This patch excludes static mocks from all validation. To support static mocks in the regular session validation, we would rather need to trigger the validation upon release of any static mock.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -10,6 +10,7 @@ include("deprecatedPluginsTest",
     "android",
     "junit-jupiter",
     "junitJupiterExtensionTest",
+    "junitJupiterInlineMockMakerExtensionTest",
     "module-test",
     "memory-test",
     "errorprone",
@@ -18,7 +19,7 @@ include("deprecatedPluginsTest",
 
 rootProject.name = "mockito"
 
-val koltinBuildScriptProject = hashSetOf("junitJupiterExtensionTest")
+val koltinBuildScriptProject = hashSetOf("junitJupiterExtensionTest", "junitJupiterInlineMockMakerExtensionTest")
 
 fun buildFileExtensionFor(projectName: String) =
     if (projectName in koltinBuildScriptProject) ".gradle.kts" else ".gradle"
--- a/src/main/java/org/mockito/MockedStatic.java
+++ b/src/main/java/org/mockito/MockedStatic.java
@@ -63,6 +63,16 @@ public interface MockedStatic<T> extends AutoCloseable {
      */
     void verifyNoInteractions();
 
+    /**
+     * Checks if this mock is closed.
+     *
+     * @return {@code true} if this mock is closed.
+     */
+    boolean isClosed();
+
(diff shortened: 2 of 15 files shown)


raphw:
It only tests part of the issue. For actually testing it, we would need to run the JUnit tests with the inline mock maker and create subprojects for it. I am going on a two week vacation tomorrow, so I did not want to open that barrel of wine tonight but I validated the working manually. If you have some time, could you add additional subprojects? I validated the runners manually, however.

TimvdLippe:
Ah okay. I got confused as the commit mentions that the issue is fixed, so that's why I thought we had test the whole scenario. Please go on vacation, we can figure it out later 😄

raphw:
I added an additional project for the JUnit Jupiter extension and added a JUnit rule for the inline mock maker tests. I hope this should avoid any future regressions.

PR #1971, opened 2020-07-13
Add support for creating constructors without using Objenesis (and unsafe API)

raphw:
This PR adds code to constructors to allow short-wiring a constructor without avoiding to call it, thus making the use of Objenesis obsolete. This is desired since Objenesis relies on Unsafe API which is deprecated and will be removed in a future release.

The idea is as follows: Any constructor for a class:

```java
class Foo extends Bar {
  Foo() {
    super(somethingWithSideeffect());
  }
}
```

is rewritten as follows:

```java
class Foo {
  Foo() {
    if (MockDispatcher.isMockedConstruction()) {
      super(null);
      return;
    }
    super(somethingWithSideeffect());
  }
}
```

The mock dispatcher then applies a thead-local check to see if the current construction is supposed to be short-wired and suppresses the original construction in such a case. The check is repeated along the super class hierarchy until reaching the `Object` constructor which is by definition side-effect free. As a result, a mock was created without triggering any user code and without relying on Unsafe API.

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMaker.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMaker.java
@@ -9,6 +9,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.instrument.Instrumentation;
+import java.lang.reflect.Constructor;
 import java.lang.reflect.Modifier;
 import java.util.Arrays;
 import java.util.Map;
@@ -20,6 +21,7 @@ import java.util.jar.JarOutputStream;
 
 import net.bytebuddy.agent.ByteBuddyAgent;
 import org.mockito.Incubating;
+import org.mockito.creation.instance.InstantiationException;
 import org.mockito.creation.instance.Instantiator;
 import org.mockito.exceptions.base.MockitoException;
 import org.mockito.exceptions.base.MockitoInitializationException;
@@ -93,7 +95,8 @@ import static org.mockito.internal.util.StringUtil.*;
  * support this feature.
  */
 @Incubating
-public class InlineByteBuddyMockMaker implements ClassCreatingMockMaker, InlineMockMaker {
+public class InlineByteBuddyMockMaker
+        implements ClassCreatingMockMaker, InlineMockMaker, Instantiator {
 
     private static final Instrumentation INSTRUMENTATION;
 
@@ -188,6 +191,8 @@ public class InlineByteBuddyMockMaker implements ClassCreatingMockMaker, InlineM
     private final DetachedThreadLocal<Map<Class<?>, MockMethodInterceptor>> mockedStatics =
             new DetachedThreadLocal<>(DetachedThreadLocal.Cleaner.INLINE);
 
+    private final ThreadLocal<Boolean> mockConstruction = ThreadLocal.withInitial(() -> false);
+
     public InlineByteBuddyMockMaker() {
         if (INITIALIZATION_ERROR != null) {
             String detail;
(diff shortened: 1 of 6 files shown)


src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodAdvice.java
```diff
@@ -344,6 +366,122 @@ boolean checkSelfCall(Object value) {
...
+                    }
+                }
+                if (current != null) {
+                    final MethodDescription.InDefinedShape selected = current;
+                    return new MethodVisitor(OpenedClassReader.ASM_API, methodVisitor) {
+                        @Override
+                        public void visitCode() {
+                            super.visitCode();
+                            Label label = new Label();
+                            super.visitLdcInsn(identifier);
```
TimvdLippe, line 411: For these opcodes, could you add the pseudo Java code in a comment? Maybe even with the JVM bytecode in a comment as well. That would help with maintaining the overview, as reading the opcode instruction methods themselves can be quite difficult.

src/main/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMaker.java
```diff
@@ -202,16 +207,25 @@ public InlineByteBuddyMockMaker() {
...
     @Override
     public <T> T createMock(MockCreationSettings<T> settings, MockHandler handler) {
         Class<? extends T> type = createMockType(settings);
 
-        Instantiator instantiator = Plugins.getInstantiatorProvider().getInstantiator(settings);
         try {
-            T instance = instantiator.newInstance(type);
+            T instance;
+            try {
+                // We attempt to use the "native" mock maker first that avoids Objenesis and Unsafe
```
TimvdLippe, line 222: In which case can this fail and do we have to fallback to `Objenesis`? E.g. ideally if we go for this pattern and we can avoid `Objenesis`, I would like to remove it as a dependency as a whole.

src/main/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMaker.java
```diff
@@ -395,6 +409,60 @@ public String nonMockableReason() {
...
+            if (Modifier.isPublic(constructor.getModifiers())) {
+                selected = constructor;
+                break;
+            }
+        }
+        Class<?>[] types = selected.getParameterTypes();
+        Object[] arguments = new Object[types.length];
+        int index = 0;
+        for (Class<?> type : types) {
+            if (type == boolean.class) {
```
TimvdLippe, line 430: Nit: extract these if-statements in a separate method for readability.

src/main/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMaker.java
bric3: We could do that in a second phase.

src/main/java/org/mockito/internal/creation/bytebuddy/InlineByteBuddyMockMaker.java
```diff
@@ -202,16 +207,25 @@ public InlineByteBuddyMockMaker() {
...
     public <T> T createMock(MockCreationSettings<T> settings, MockHandler handler) {
         Class<? extends T> type = createMockType(settings);
 
-        Instantiator instantiator = Plugins.getInstantiatorProvider().getInstantiator(settings);
         try {
-            T instance = instantiator.newInstance(type);
+            T instance;
+            try {
+                // We attempt to use the "native" mock maker first that avoids Objenesis and Unsafe
+                instance = newInstance(type);
```
bric3, line 223: Why not _hiding_ this behind the interface ? The implementing class could still be this class.
And tweak the _plug-in selector_to use this one first. And use the fallback instantiator otherwise.

src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodAdvice.java
bric3: Maybe at some point this could ship in bytebuddy.

PR #1975, opened 2020-07-16
Implements mocking for constructor invocations.

raphw:
Allows to declare all constructor invocations as mocking attempts.

### DIFF
--- a/src/main/java/org/mockito/MockedConstruction.java
+++ b/src/main/java/org/mockito/MockedConstruction.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2007 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito;
+
+import java.lang.reflect.Constructor;
+import java.util.List;
+
+/**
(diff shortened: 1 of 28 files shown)


msteinbeck:
Are there any plans on when this will be released? I'm really looking forward for this feature.

raphw:
Many of us are on vacation at the time but likely some time in August.

JojOatXGME:
As far as I can see, there is still no support to get the arguments of the constructor. Is it planed to be added at some point? Here is a small motivation. I usually use copy constructors if I want to copy objects. (I think `clone()` is more or less broken.) If I want to assert that some mocked object is a copy of another object, I would need to look at the constructor arguments when the object is created.

Beside that, I think it would improve usability when `MockedConstruction` had a method to change the preparation callback. Otherwise, it is impossible to set a preparation callback when `@Mock` is used.

raphw:
Good point, I refactored it to take a `Context` object in addition to the mock where one can get the invoked constructor including its arguments and an index for the count of the created mock for a given scope. This should offer any desired functionality.

raphw:
Already on master.

PR #1983, opened 2020-07-19
Replace usages of LinkedList with ArrayList/ArrayDeque

turbanoff:
It's known that ArrayList performance is almost always better than LinkedList:
1. Due to dense memory layout - elements of ArrayList are easier to cache by CPU. It gives really huge speedup compared to LinkedList
2. It has less overhead when it's empty: LinkedList has 4 fields (size, first, last, modCount) while ArrayList has 3 fields (size, elementData, modCount)
3. ArrayList is much more often used in application code and JDK, compared to LinkedList. So chances that its code will be already compiled by JIT when test/mock is executed is very high.

See examples of similar cleanup in other projects:
1. JDK https://bugs.openjdk.java.net/browse/JDK-8246048 
2. Spring https://github.com/spring-projects/spring-framework/issues/25650

This PR replaces usages of LinkedList with ArrayList and ArrayDeque where it's _convenient_. I intentionally didn't converted few places where LinkedList used as Dequeu to avoid making code more complex.

### DIFF
--- a/src/main/java/org/mockito/internal/InOrderImpl.java
+++ b/src/main/java/org/mockito/internal/InOrderImpl.java
@@ -6,7 +6,7 @@ package org.mockito.internal;
 
 import static org.mockito.internal.exceptions.Reporter.inOrderRequiresFamiliarMock;
 
-import java.util.LinkedList;
+import java.util.ArrayList;
 import java.util.List;
 
 import org.mockito.InOrder;
@@ -27,7 +27,7 @@ import org.mockito.verification.VerificationMode;
 public class InOrderImpl implements InOrder, InOrderContext {
 
     private final MockitoCore mockitoCore = new MockitoCore();
-    private final List<Object> mocksToBeVerifiedInOrder = new LinkedList<Object>();
+    private final List<Object> mocksToBeVerifiedInOrder = new ArrayList<>();
     private final InOrderContext inOrderContext = new InOrderContextImpl();
 
     public List<Object> getMocksToBeVerifiedInOrder() {
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
@@ -21,7 +21,6 @@ import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Iterator;
-import java.util.LinkedList;
 import java.util.Random;
 
 import net.bytebuddy.ByteBuddy;
@@ -215,7 +214,7 @@ class SubclassBytecodeGenerator implements BytecodeGenerator {
     }
 
     private <T> Collection<Class<? super T>> getAllTypes(Class<T> type) {
-        Collection<Class<? super T>> supertypes = new LinkedList<Class<? super T>>();
+        Collection<Class<? super T>> supertypes = new ArrayList<>();
         supertypes.add(type);
         Class<? super T> superType = type;
         while (superType != null) {
--- a/src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
+++ b/src/main/java/org/mockito/internal/creation/instance/ConstructorInstantiator.java
@@ -8,8 +8,8 @@ import static org.mockito.internal.util.StringUtil.join;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
+import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.LinkedList;
 import java.util.List;
 
 import org.mockito.creation.instance.InstantiationException;
@@ -39,7 +39,7 @@ public class ConstructorInstantiator implements Instantiator {
     }
 
     private <T> T withParams(Class<T> cls, Object... params) {
-        List<Constructor<?>> matchingConstructors = new LinkedList<Constructor<?>>();
+        List<Constructor<?>> matchingConstructors = new ArrayList<>();
         try {
             for (Constructor<?> constructor : cls.getDeclaredConstructors()) {
                 Class<?>[] types = constructor.getParameterTypes();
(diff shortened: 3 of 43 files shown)


turbanoff:
I've create this benchmark to check changes:
```
package org.mockito.pr;


import static org.mockito.ArgumentMatchers.eq;

import java.util.concurrent.TimeUnit;

import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.BenchmarkMode;
import org.openjdk.jmh.annotations.Fork;
import org.openjdk.jmh.annotations.Measurement;
import org.openjdk.jmh.annotations.Mode;
import org.openjdk.jmh.annotations.OutputTimeUnit;
import org.openjdk.jmh.annotations.Param;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.annotations.Warmup;
import org.openjdk.jmh.runner.Runner;
import org.openjdk.jmh.runner.RunnerException;
import org.openjdk.jmh.runner.options.Options;
import org.openjdk.jmh.runner.options.OptionsBuilder;

@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@Warmup(iterations = 8, time = 500, timeUnit = TimeUnit.MILLISECONDS)
@Measurement(iterations = 5, time = 500, timeUnit = TimeUnit.MILLISECONDS)
@Fork(1)
@State(Scope.Thread)
public class ArrayListBenchmark {

    @Param({"4", "400", "4000"})
    public int times;

    @Benchmark
    public IBench verifyInt() {
        IBench mock = mock(IBench.class);

        for (int i = 0; i < times; i++) {
            when(mock.intWork(eq(i))).thenReturn(i);
        }
        for (int i = 0; i < times; i++) {
            mock.intWork(i);
        }
        for (int i = 0; i < times; i++) {
            verify(mock, times(1)).intWork(eq(i));
        }
        return mock;
    }

    @Benchmark
    public IBench verifyLong() {
        IBench mock = mock(IBench.class);

        for (int i = 0; i < times; i++) {
            Long l = (long) i;
            when(mock.longWork(eq(l))).thenReturn(l);
        }
        for (int i = 0; i < times; i++) {
            Long l = (long) i;
            mock.longWork(l);
        }
        for (int i = 0; i < times; i++) {
            Long l = (long) i;
            verify(mock, times(1)).longWork(eq(l));
        }
        return mock;
    }

    public interface IBench {
        Long longWork(Long l);
        int intWork(int a);
    }

    public static void main(String[] args) throws RunnerException {
        Options opt = new OptionsBuilder()
                .include(ArrayListBenchmark.class.getSimpleName())
                .build();

        new Runner(opt).run();
    }
}
```

Results on my machine:
```
With LinkedList

Benchmark                      (times)  Mode  Cnt            Score            Error  Units
ArrayListBenchmark.verifyInt         4  avgt    5       560685,589 ±     107064,526  ns/op
ArrayListBenchmark.verifyInt       400  avgt    5    400371540,300 ±   21917662,331  ns/op
ArrayListBenchmark.verifyInt      4000  avgt    5  36659444000,400 ± 3578007977,198  ns/op
ArrayListBenchmark.verifyLong        4  avgt    5       567006,300 ±     170306,251  ns/op
ArrayListBenchmark.verifyLong      400  avgt    5    398900860,200 ±   26951228,399  ns/op
ArrayListBenchmark.verifyLong     4000  avgt    5  37868755020,200 ± 5303126213,561  ns/op

With ArrayList

Benchmark                      (times)  Mode  Cnt            Score            Error  Units
ArrayListBenchmark.verifyInt         4  avgt    5       543610,862 ±      92885,978  ns/op
ArrayListBenchmark.verifyInt       400  avgt    5    389968770,000 ±   20683156,985  ns/op
ArrayListBenchmark.verifyInt      4000  avgt    5  35712889840,000 ± 1644684325,891  ns/op
ArrayListBenchmark.verifyLong        4  avgt    5       547077,864 ±     114747,071  ns/op
ArrayListBenchmark.verifyLong      400  avgt    5    397587730,000 ±   19347863,897  ns/op
ArrayListBenchmark.verifyLong     4000  avgt    5  36479603000,000 ± 1615490251,500  ns/op
```

turbanoff:
@TimvdLippe what do you think?

In my opinion, while performance difference in benchmark is not dramatical, it's still worth it. LinkedList is used very rarely nowadays. So chances that it will be JIT compiled are low.

Opinion of performance engineer @kuksenko from Oracle: https://twitter.com/kuksenk0/status/1299541293438181376
>There is nothing wrong with general linked list. But in Java java.util.LinkedList is useless. There are very few scenarios when it may outperform ArrayList (practically on modern HW). I've never seen it.

TimvdLippe:
I am open to this refactoring, but we have a couple of other higher priority issues that take up our time on Mockito sorry. We will get to this PR when dust settles I think.

mockitoguy (approved):
Thanks!

mockitoguy:
@TimvdLippe, I'm OK with this change. My understanding of this quite niche domain is that ArrayLists are, on average, a better choice than LL. Using a consistent approach across the entire codebase makes sense to me. Are you OK to merge or you'd prefer to review yourself? Thanks!

mockitoguy:
@turbanoff, thanks for the PR, I'd like to get it merged.

TimvdLippe:
I am okay with merging, but I would like to first get confirmation that indeed improves performance. E.g. we need to replicate the findings by @turbanoff and maybe even turn it into a performance test in our regression test suite.

TimvdLippe:
I am closing this PR for now. The primary reason is that it it is large, has numerous merge conflicts and I am generally not a fan of refactoring large amount of code without clear benefit. The numbers provided appear to be marginally the same. Still, if there are obvious performance problems with Mockito, we should fix these. But then I would like to have a concrete user story, performance regression tests and a thorough solution to that. Blanket refactorings from one data structure to another I am not thrilled about, given our prior problems with internal refactorings that turned out to be really problematic for users/ourselves.

I am sorry that we kept it lingering for so long and that the response (from my side at least) hasn't been enthusiastic. I appreciate the effort put into tackling performance problems. We have run into performance problems before, but they were tackled mostly in ByteBuddy and were resolved as such.

For future code, I think avoid using `LinkedList` sounds fine to me. But let's not do that for all the existing code that is scattered throughout the codebase.

My apologies for the long wait and unsatisfying response. I should have probably earlier and more clearly expressed my objections to this particular approach, so for that I apologize.

turbanoff:
@TimvdLippe would it be Ok if I will create PR to update separate modules/folders?
For me (and most of java developers) it's clear that LinkedList is very bad at performance side. The sooner it will be eliminated from such widely used library the better.

PR #1996, opened 2020-08-07
Add invoker API to allow for alternative invocation modes to better support the module system.

raphw:
Adds a `MemberAccessor` abstraction for accessing fields, methods and constructors where the default implementation `ReflectionMemberAccessor` implements the current behavior of using reflection and `setAccessible`. 

Also, this PR adds a new implementation `ModuleMemberAccessor` where the instrumentation API is leveraged to open modules to Mockito before using method handles to access any such member. This way, module boundaries are no longer stopping Mockito from functioning on Java 9 and onwards. Since the instrumentation API is already used by the inline-mock-maker, it is enabled for this mock maker by default.


src/test/java/org/mockitousage/annotation/MockInjectionUsingSetterOrPropertyTest.java
```diff
@@ -50,10 +51,19 @@
...
     @Spy private TreeSet<String> searchTree = new TreeSet<String>();
 
+    private AutoCloseable session;
+
     @Before
     public void enforces_new_instances() {
-        // initMocks called in TestBase Before method, so instances are not the same
-        MockitoAnnotations.openMocks(this);
+        // openMocks called in TestBase Before method, so instances are not the same
+        session = MockitoAnnotations.openMocks(this);
```
TimvdLippe, line 59: We should update `openMocks` to use `@CheckReturnValue`. Let's fix that in a separate PR.

src/main/java/org/mockito/internal/util/reflection/InstrumentationMemberAccessor.java
```diff
@@ -0,0 +1,363 @@
...
+                    MethodHandles.publicLookup()
+                            .findStatic(
+                                    MethodHandles.class,
+                                    "privateLookupIn",
+                                    MethodType.methodType(
+                                            MethodHandles.Lookup.class,
+                                            Class.class,
+                                            MethodHandles.Lookup.class));
+        } catch (Throwable t) {
+            throw new MockitoInitializationException(
```
TimvdLippe, line 133: If this exception is thrown, should we automatically fall back to the reflection accessor? Or should we bail out at all times?

src/main/java/org/mockito/internal/util/reflection/InstrumentationMemberAccessor.java
```diff
@@ -0,0 +1,363 @@
...
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.lang.reflect.*;
+import java.util.*;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+import static org.mockito.internal.util.StringUtil.join;
+
+class InstrumentationMemberAccessor implements MemberAccessor {
```
TimvdLippe, line 24: There's quite a bit code there and I think the code coverage is not high enough, which is what CodeCov is catching. Is it feasible to increase the coverage here or is that not possible?

src/main/java/org/mockito/internal/util/reflection/InstrumentationMemberAccessor.java
```diff
@@ -0,0 +1,363 @@
...
+                                            method.getDeclaringClass(), DISPATCHER.getLookup()))
+                            .unreflect(method);
+            if (!Modifier.isStatic(method.getModifiers())) {
+                handle = handle.bindTo(target);
+            }
+            try {
+                return handle.invokeWithArguments(arguments);
+            } catch (Throwable t) {
+                throw new InvocationTargetException(t);
+            }
```
TimvdLippe, line 198: There is quite a bit of duplication across these methods. Does it maybe make sense to extract it out? The only difference I am spotting is the `unreflect` in here, where others use for example `unreflectGetter`.

src/test/java/org/mockitousage/annotation/MockInjectionUsingSetterOrPropertyTest.java
raphw: +1

src/main/java/org/mockito/internal/util/reflection/InstrumentationMemberAccessor.java
raphw: This would mean that Java 9 does not contain a method that is speced to be there. I think an exception is appropriate. The use is guarded by `ModuleMemberAccess`.

src/main/java/org/mockito/internal/util/reflection/InstrumentationMemberAccessor.java
raphw: I can add a few tests for the exception scenarios.

src/main/java/org/mockito/internal/util/reflection/InstrumentationMemberAccessor.java
raphw: I didn't want to factor it out too much since I don't think it makes the code easier to read and it would be harded to extend if you wanted to add a special case for one of the dispatchers at some point that does not apply to the other.

raphw:
Thanks. It should fix the warnings of illegal access together with https://github.com/mockito/mockito/pull/2002 - if using the inline mock maker.

PR #1998, opened 2020-08-09
Update objenesis to 3.1

### LINKED ISSUE #1992: Update to org.objenesis 3.x
Hi all,

we are slowly updating our product to Java 11; we've managed to move to Java 11 runtime, now we are trying to also compile with Java 11.

We ran into a problem with `org.easymock` (it doesn't work with Java 11 compiled classes, https://github.com/easymock/easymock/issues/218). The problem is resolved in 4.x versions.

Unfortunately, 4.x versions of `org.easymock` run only with `org.objenesis` 3.x, which `mockito` so far does not accept (upper bound for `org.objenesis` is 2.6).

Are there plans to update `mockito` to run with `org.objenesis` and if so are there known timelines?

Some more background:

We essentially wished to update to latest `mockito` and `easymock` earlier (during the Java 11 runtime move), but were unable to do so due to conflicting `org.objenesis` requirements. If we want to compile with Java 11 however, we are made to choose between  `easymock` or `mockito`, or waiting. We would prefer to wait, if a `mockito` update that runs with `org.objenesis` 3.x would be available soon.

Sorry if this should be an e-mail message and not an issue in the tracker, or if this is a duplicate (I didn't find much about `org.objenesis` in the issue list).

Best regards and thanks,
Simeon

epeee:
see http://objenesis.org/notes.html

fixes #1992

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -25,8 +25,7 @@ libraries.errorproneTestApi = "com.google.errorprone:error_prone_test_helpers:${
 
 libraries.autoservice = "com.google.auto.service:auto-service:1.0-rc5"
 
-// objenesis 3.x with full Java 11 support requires Java 8, bump version when Java 7 support is dropped
-libraries.objenesis = 'org.objenesis:objenesis:2.6'
+libraries.objenesis = 'org.objenesis:objenesis:3.1'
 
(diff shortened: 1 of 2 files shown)


trancexpress:
Thanks!

trancexpress:
Thanks!

PR #2006, opened 2020-08-17
Introduce animal sniffer

raphw:
Introduces animal sniffer with exclusion of inline-mock-maker classes which would never be present on Android. Avoids calling invoke/invokeExact methods of handles directly but rather puts the invocations into generated code to avoid breaking Android builds.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,6 +21,7 @@ plugins {
     id 'eclipse'
     id 'com.github.ben-manes.versions' version '0.28.0'
     id 'biz.aQute.bnd.builder' version '5.1.0'
+    id 'ru.vyarus.animalsniffer' version '1.5.1'
 }
 
 description = 'Mockito mock objects library core API and implementation'
@@ -98,6 +99,14 @@ dependencies {
     testRuntime configurations.compileOnly
 
     testUtil sourceSets.test.output
+
+    signature 'org.codehaus.mojo.signature:java18:1.0@signature'
+    signature 'net.sf.androidscents.signature:android-api-level-24:7.0_r2@signature'
+}
+
+animalsniffer {
+    sourceSets = [sourceSets.main]
+    annotation = 'org.mockito.internal.SuppressSignatureCheck'
 }
 
 wrapper {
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -98,6 +99,14 @@ dependencies {
     testRuntime configurations.compileOnly
 
     testUtil sourceSets.test.output
+
+    signature 'org.codehaus.mojo.signature:java18:1.0@signature'
+    signature 'net.sf.androidscents.signature:android-api-level-24:7.0_r2@signature'
```
TimvdLippe, line 104: So for my understanding, this says that we are supporting Android API levels 24 and over? If so, we should document that somewhere as well. In the Twitter thread, they proposed level 21 and over. If we set this to 21, do we need to make a lot of changes? Iirc, Mockito mostly builds in Google so I think we would be able to do so?

raphw: We should, indeed. Android level 21 does not support types such as `Optional` or `Function`. We can go that far but it's already part of the public API. I guess 3.5.0 is not super-wide spread yet but I'd like to avoid releasing 4.0 just for that.

What do you think?

TimvdLippe: Hm, iirc Mockito 3.4.0 builds just fine internally and also includes Optional. I can take a look tomorrow as to why that works.

raphw: It looks though, that Android can easily desugar these few types: https://developer.android.com/studio/write/java8-support - it only does not work with method handle types which are not exposed. I just asked for a validation of this branch of the user who reported this. By requiring API level 24, we do not leak API that cannot be desugared at least. But yes, we should document this!

raphw: Also, with code shrinking, it should already work with 3.5.0: https://developer.android.com/studio/build/shrink-code#shrink-code

TimvdLippe:
@raphw Can we publish a new release to Maven central for #2007?

raphw:
Doesn't Android consume JCenter?

darakeon:
@raphw : https://blog.gradle.org/jcenter-shutdown

PR #2021, opened 2020-08-24
Constructor dispatch

raphw:
Fixes recursive constructor invocation.
Avoids reflective access warning when using instrumentation-based member accessor.

### DIFF
--- a/subprojects/inline/inline.gradle
+++ b/subprojects/inline/inline.gradle
@@ -12,3 +12,7 @@ tasks.javadoc.enabled = false
 //required by the "StressTest.java" and "OneLinerStubStressTest.java"
 test.maxHeapSize = "256m"
 retryTest.maxHeapSize = "256m"
+
+test {
+    jvmArgs '--illegal-access=debug'
+}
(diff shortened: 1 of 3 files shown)


subprojects/inline/inline.gradle
```diff
@@ -12,3 +12,7 @@ tasks.javadoc.enabled = false
 //required by the "StressTest.java" and "OneLinerStubStressTest.java"
 test.maxHeapSize = "256m"
 retryTest.maxHeapSize = "256m"
+
+test {
+    jvmArgs '--illegal-access=debug'
```
TimvdLippe, line 17: This means that the build will now fail when an Illegal Access is performed? Nice!

raphw: It will log it in more detail but fail would be the much more sensible option. Weird I did not think of it, that's of course a much better value...

TimvdLippe:
I think something has gone wrong with this PR. Travis is quite red after this PR: https://travis-ci.org/github/mockito/mockito/builds Example broken build: https://travis-ci.org/github/mockito/mockito/builds/720600848

Seems like this PR was merged, but Travis hadn't finished yet?

PR #2034, opened 2020-09-03
Escape mock during method dispatch on mock to avoid premature garbage collection.

### LINKED ISSUE #1802: Exception "The mock object was garbage collected."

raphw:
Under heavy optimization mocks might get garbage collected during the dispatching of a mocked method if the mock instance is not used after this method dispatch. To avoid this, we escape the mock instance during the dispatch to make sure that the GC cannot collect the object.

Fixes #1802.

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodInterceptor.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodInterceptor.java
@@ -6,6 +6,8 @@ package org.mockito.internal.creation.bytebuddy;
 
 import static org.mockito.internal.invocation.DefaultInvocationFactory.createInvocation;
 
+import java.io.IOException;
+import java.io.ObjectInputStream;
 import java.io.ObjectStreamException;
 import java.io.Serializable;
 import java.lang.reflect.Method;
@@ -36,10 +38,17 @@ public class MockMethodInterceptor implements Serializable {
 
     private final ByteBuddyCrossClassLoaderSerializationSupport serializationSupport;
 
+    private transient ThreadLocal<Object> weakReferenceHatch = new ThreadLocal<>();
+
     public MockMethodInterceptor(MockHandler handler, MockCreationSettings mockCreationSettings) {
         this.handler = handler;
         this.mockCreationSettings = mockCreationSettings;
         serializationSupport = new ByteBuddyCrossClassLoaderSerializationSupport();
+    }
+
+    private void readObject(ObjectInputStream stream) throws IOException, ClassNotFoundException {
+        stream.defaultReadObject();
+        weakReferenceHatch = new ThreadLocal<>();
     }
 
     Object doIntercept(Object mock, Method invokedMethod, Object[] arguments, RealMethod realMethod)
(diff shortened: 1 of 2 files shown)


TimvdLippe (approved):
Seems like CheckStyle requires a license header. Other than that, nice fix! I like the extensive documentation as well 😄

raphw:
Indeed, force pushed a new version just now.

Let's hope the JIT stays dumb enough to avoid this error in the future - I was really certain that the stack allocation of the mock object would avoid the allocation but if the entire Mockito call stack gets inlined, that does no longer apply. Worst case, we need to make the Mockito call stack deeper then the inlining heuristic ;)

PR #2042, opened 2020-09-14
Mock resolver plugin

### LINKED ISSUE #1980: mockito3.4.0 InlineByteBuddyMockMaker can't work with spring cglib proxy, but ByteBuddyMockMaker works
I upgrade mockito to 3.4.0, and the mockMaker is InlineByteBuddyMockMaker, I use SpyBean to spy a spring bean, and then verify the bean, it throws NotAMockException.
image

then i debug, find that the mocks.get(mock) can't find it because the mock is CGLIB proxy, not the original type.

when I use the verify((AppGroupClusterAction) AopTestUtils.getTargetObject(appGroupClusterAction)) to verify. it works.

but the before mockMaker ByteBuddyMockMaker works. so can the InlineByteBuddyMockMaker be compatible with the before？

raphw:
Adds a plugin to allow for adding one or multiple mock resolvers. This way, instances that are provided to Mockito's DSL do no longer require to be the actual mocks but can also be proxies of mocks what is something already being applied by frameworks such as Spring.

This works today with the subclass mock maker by accident as we read the mock state from the mock instance via a method which gets proxied. If we ever find a more private appraoch this would however break Spring's Mockito use and it is already broken with the inline mock maker.

By this SPI, Spring could add an unproxy resolver to its Mockito build-up and make it's proxied mocks compatible with Mockito. Other frameworks could use the same approach if desired as this SPI is generic.

Closes #1980

### DIFF
--- a/src/main/java/org/mockito/internal/util/MockUtil.java
+++ b/src/main/java/org/mockito/internal/util/MockUtil.java
@@ -16,6 +16,7 @@ import org.mockito.mock.MockCreationSettings;
 import org.mockito.mock.MockName;
 import org.mockito.plugins.MockMaker;
 import org.mockito.plugins.MockMaker.TypeMockability;
+import org.mockito.plugins.MockResolver;
 
 import java.util.function.Function;
 
@@ -63,13 +64,16 @@ public class MockUtil {
         mockMaker.resetMock(mock, newHandler, settings);
     }
 
-    public static <T> MockHandler<T> getMockHandler(T mock) {
+    public static MockHandler<?> getMockHandler(Object mock) {
         if (mock == null) {
             throw new NotAMockException("Argument should be a mock, but is null!");
         }
 
-        if (isMock(mock)) {
-            return mockMaker.getHandler(mock);
+        mock = resolve(mock);
+
+        MockHandler handler = mockMaker.getHandler(mock);
+        if (handler != null) {
+            return handler;
         } else {
             throw new NotAMockException("Argument should be a mock, but is: " + mock.getClass());
         }
@@ -96,7 +100,20 @@ public class MockUtil {
         // Potentially we could also move other methods to MockitoMock, some other candidates:
         // getInvocationContainer, isSpy, etc.
         // This also allows us to reuse our public API MockingDetails
-        return mock != null && mockMaker.getHandler(mock) != null;
+        if (mock == null) {
+            return false;
+        }
+
+        mock = resolve(mock);
+
+        return mockMaker.getHandler(mock) != null;
+    }
+
+    private static Object resolve(Object mock) {
+        for (MockResolver mockResolver : Plugins.getMockResolvers()) {
+            mock = mockResolver.resolve(mock);
+        }
+        return mock;
     }
 
     public static MockName getMockName(Object mock) {
--- a/src/main/java/org/mockito/internal/util/MockUtil.java
+++ b/src/main/java/org/mockito/internal/util/MockUtil.java
@@ -56,11 +56,12 @@ public class MockUtil {
         return mock;
     }
 
-    public static <T> void resetMock(T mock) {
+    public static void resetMock(Object mock) {
         MockHandler oldHandler = getMockHandler(mock);
         MockCreationSettings settings = oldHandler.getMockSettings();
         MockHandler newHandler = createMockHandler(settings);
 
+        mock = resolve(mock);
         mockMaker.resetMock(mock, newHandler, settings);
     }
 
@@ -110,6 +111,9 @@ public class MockUtil {
     }
 
     private static Object resolve(Object mock) {
+        if (mock instanceof Class<?>) { // static mocks are resolved by definition
+            return mock;
+        }
         for (MockResolver mockResolver : Plugins.getMockResolvers()) {
             mock = mockResolver.resolve(mock);
         }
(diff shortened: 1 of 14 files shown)


src/main/java/org/mockito/internal/util/MockUtil.java
```diff
@@ -63,13 +64,18 @@ public static TypeMockability typeMockabilityOf(Class<?> type) {
...
             throw new NotAMockException("Argument should be a mock, but is null!");
         }
 
-        if (isMock(mock)) {
-            return mockMaker.getHandler(mock);
+        for (MockResolver mockResolver : Plugins.getMockResolvers()) {
+            mock = mockResolver.resolve(mock);
+        }
+
+        MockHandler handler = mockMaker.getHandler(mock);
```
TimvdLippe, line 76: Nit: extract this out in a private method and use that below as well.

subprojects/extTest/src/test/java/org/mockitousage/plugins/resolver/MockResolverTest.java
```diff
@@ -0,0 +1,50 @@
...
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.*;
+
+@MockitoSettings(strictness = Strictness.WARN)
+@ExtendWith(MockitoExtension.class)
+class MockResolverTest {
```
TimvdLippe, line 18: Can we add a test that uses multiple resolvers? I can think of the following scenarios:

1. Multiple resolvers, no match
2. Multiple resolvers, one match
3. Multiple resolvers, multiple match

For scenario 3, should we throw maybe? What if it is order-dependent? Should we continue iterating over the resolvers until none match?

TimvdLippe (commented):
Implementation and solution seems fine. I do want to wait for the Spring folks to try this out before committing to this API.

subprojects/extTest/src/test/java/org/mockitousage/plugins/resolver/MockResolverTest.java
raphw: I think multiple resolvers should be allowed. You can technically add several in your specified order and decompose multiple levels of nesting. It's of course an obscure scenario, but it makes sense to not forbid it since it would be extra work. I can add the tests you suggest.

src/main/java/org/mockito/internal/util/MockUtil.java
raphw: Fully agree, I hope for some feedback first, too.

subprojects/extTest/src/test/java/org/mockitousage/plugins/resolver/MockResolverTest.java
raphw: Hmm, registering multiple converters in a test is not quite that easy since it would require multiple modules. I don't think its worth the noise to add this functionality just for testing such a simple corner case.

TimvdLippe: Understood. How would that work when a user wants to integrate with multiple frameworks that all expose a resolver?

raphw: They would all expose their resolver and they would be invoked in their registration order. For instance, let's say Hibernate offered a Mockito-based mocking API but returns Hibernate proxies on mocks. Those proxies are by definition never Spring Beans. But if one passed a Spring bean or a Hibernate proxy, both would traverse the resolver chain and be unpacked accordingly.

I assume that normally, only one resolver would apply to any wrapped mock but since Mockito should remain agnostic to this, that would be the best approach in my eyes.

src/main/java/org/mockito/internal/util/MockUtil.java
```diff
@@ -55,21 +56,25 @@ public static TypeMockability typeMockabilityOf(Class<?> type) {
         return mock;
     }
 
-    public static <T> void resetMock(T mock) {
+    public static void resetMock(Object mock) {
```
TimvdLippe, line 59: Why were the generic parameters removed from this and the other methods? I understand that they are not used, but I wonder if there are unintended side-effects of this change.

raphw: Just as a cleanup on the side since I am touching the code already. The generic type is erased to `Object` and has no other function, that's why I removed it. No impact on source or binary compatibility.

src/main/java/org/mockito/plugins/MockResolver.java
```diff
@@ -0,0 +1,22 @@
...
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.plugins;
+
+/**
+ * A mock resolver offers an opportunity to resolve a mock from any instance that is
+ * provided to the {@link org.mockito.Mockito}-DSL. This mechanism can be used by
+ * frameworks that provide mocks that are implemented by Mockito but which are wrapped
+ * by other instances to enhance the proxy further.
+ */
```
TimvdLippe, line 12: This needs to be marked as `@Incubating`

TimvdLippe (approved):
LGTM with 1 nit

src/main/java/org/mockito/plugins/MockResolver.java
raphw: Arghs, I missed this and just hit merge. Sorry, we can still add this, though.

TimvdLippe: Yes let's still add it.

PR #2045, opened 2020-09-16
Prefer non-synthetic constructors in MockMethodAdvice.ConstructorShortcut

### LINKED ISSUE #2040: Mockito.spy(Activity).getBaseContext() returns null on Robolectric 4.4 and Java8
### Description
Since robolectric 4.4, Mockito.spy(Activity).getBaseContext() returns null

### Steps to Reproduce
```
$ git clone -b mockito_spy_robolectric_4_4 https://github.com/ganadist/VersionCodeDemo demo
$ cd demo

// switch to robolectric 4.3.1
$ git checkout HEAD~1
$ ./gradlew :app:testDevelopDebugUnitTest
BUILD SUCCESSFUL in 14s
28 actionable tasks: 28 executed

// switch to robolectric 4.4
$ git checkout mockito_spy_robolectric_4_4
$ ./gradlew :app:testDevelopDebugUnitTest
> Task :app:testDevelopDebugUnitTest FAILED
* What went wrong:
Execution failed for task ':app:testDevelopDebugUnitTest'.
```
Failed unittest result : https://scans.gradle.com/s/wbmmbnzatcyai/tests/:app:testDevelopDebugUnitTest/com.example.myapplication.UnitTest1/test1#1

### Robolectric & Android Version
- Compile Sdk : 29
- Target Sdk : 28
- Robolectric : 4.4 
- Mockito : 3.5.10
- Jvm : OpenJDK 8

### Link to a public git repo demonstrating the problem:
https://github.com/ganadist/VersionCodeDemo/blob/mockito_spy_robolectric_4_4/app/src/test/java/com/example/myapplication/UnitTest1.java#L27-L38

This issue is copied from https://github.com/robolectric/robolectric/issues/5916

hoisie:
After moving spy creation to instrumenting constructor chains in
ByteBuddyMockMaker, creating spies for Robolectric-instrumented Android
classes started failing in some situations due to fields not being
copied on base classes (specifically, ContextWrapper.mBase). The problem
is that AsmVisitorWrapper.ForDeclaredMethods does not visit the
synthetic constructrs that are added by Robolectric during runtime.
While the visibility issuer is still being explored, a workaround is to
prefer non-synthetic constructors when selecting which parent
constructor to call in MockMethodAdvice.ConstructorShortcut.

Fixes #2040

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodAdvice.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodAdvice.java
@@ -396,6 +396,13 @@ public class MockMethodAdvice extends MockMethodDispatcher {
                                 .asErasure()
                                 .getDeclaredMethods()
                                 .filter(isConstructor().and(not(isPrivate())));
+                MethodList<MethodDescription.InDefinedShape> nonSynthetics =
+                        constructors.filter(not(isSynthetic()));
+                // Prefer non-synthetic constructors as a workaround to synthetic Robolectric
+                // generated constructors not being visited by AsmVisitorWrapper.ForDeclaredMethods.
+                if (nonSynthetics.size() > 0) {
+                    constructors = nonSynthetics;
+                }
                 int arguments = Integer.MAX_VALUE;
                 boolean visible = false;
                 MethodDescription.InDefinedShape current = null;


raphw (approved):

TimvdLippe (commented):
Could we get a regression test as well? Not sure if it is worth it though.

raphw:
I hope it's temporary and we get to fix the root problem some time in the nesr future.

hoisie:
> Could we get a regression test as well? Not sure if it is worth it though.

I can try, though it would essentially recreating a Robolectric-like environment, so it may be challenging, considering there are so many moving parts to a Robolectric environment. I'll could try to create a custom child-first classloader that instuments a class to insert a constructor and see if I can replicate the failure.

FWIW I did add a regression test on the Robolectric side: https://github.com/robolectric/robolectric/blob/master/integration_tests/mockito-experimental/src/test/java/org/robolectric/integrationtests/mockito/experimental/MockitoSpyTest.java

TimvdLippe:
Let's keep the test on the Robolectric side, that makes sense to me. Thanks!

raphw:
I found the actual problem: https://github.com/mockito/mockito/pull/2046

We excluded synthetic constructors by Byte Buddy's default. *facepalm*

PR #2047, opened 2020-09-17
Fixes #1997 : make mockito-junit-jupiter workin in OSGiMockitoExtensi…

### LINKED ISSUE #1997: MockitoExtension cannot be used with JUnit 5 in OSGi integration tests
The Mockito junit-jupiter JAR is missing OSGi bundle manifest entries so it cannot be used in OSGi integration tests.

We run some OSGi integration tests (using bnd) that have `@Mock`s so it would be easier if these mocks can be initialized by annotating classes with `@MockitoExtension` instead of using `MockitoAnnotations.openMocks(this)`, see for example this [FirmwareUpdateServiceTest](https://github.com/openhab/openhab-core/blob/43cb68c9d753faf3a88cfd19c4fc015c8289f880/itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/internal/firmware/FirmwareUpdateServiceTest.java).

What may complicate adding the manifest entries is that `MockitoExtension` is [importing](https://github.com/mockito/mockito/blob/b87fad310bf59b94035028dc478089ded2a9198a/subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/MockitoExtension.java#L27-L29) internal classes which [aren't exported](https://github.com/mockito/mockito/blob/b87fad310bf59b94035028dc478089ded2a9198a/gradle/mockito-core/osgi.gradle#L19) by the mockito-core bundle. What may help is to make junit-jupiter a bundle fragment with mockito-core as host.

rotty3000:
…on

without
a) giving up the internal nature of some packages
b) resorting to substandard solutions like OSGi fragements

Signed-off-by: Raymond Augé <raymond.auge@liferay.com>

> Hey, 
> 
> Thanks for the contribution, this is awesome.
> As you may have read, project members have somehow an opinionated view on what and how should be
> Mockito, e.g. we don't want mockito to be a feature bloat.
> There may be a thorough review, with feedback -> code change loop.
> 
> Which branch : 
> - On mockito 3.x, make your pull request target `release/3.x`
> - On mockito 2.x, make your pull request target `release/2.x` (2.x is in maintenance mode)
>
> _This block can be removed_
> _Something wrong in the template fix it here `.github/PULL_REQUEST_TEMPLATE.md`

check list

### DIFF
--- a/gradle/mockito-core/osgi.gradle
+++ b/gradle/mockito-core/osgi.gradle
@@ -1,12 +1,3 @@
-import java.nio.file.attribute.FileTime
-import java.util.jar.Attributes
-import java.util.jar.JarEntry
-import java.util.jar.JarFile
-import java.util.jar.JarOutputStream
-import java.util.jar.Manifest
-
-import static java.util.Calendar.FEBRUARY
-
 apply plugin: 'biz.aQute.bnd.builder'
 
 jar {
@@ -14,9 +5,9 @@ jar {
     bnd(
         'Bundle-Name': 'Mockito Mock Library for Java. Core bundle requires Byte Buddy and Objenesis.',
         'Bundle-SymbolicName': 'org.mockito.mockito-core',
-        'Bundle-Version': project.version.replace('-', '.'),
+        'Bundle-Version': "\${version_cleanup;${project.version}}",
         '-versionpolicy': '[${version;==;${@}},${version;+;${@}})',
-        'Export-Package': "!org.mockito.internal.*,org.mockito.*;version=${version}",
+        'Export-Package': "org.mockito.internal.*;status=INTERNAL;mandatory:=status;version=${version},org.mockito.*;version=${version}",
         'Import-Package': [
                 'net.bytebuddy.*;version="[1.6.0,2.0)"',
                 'junit.*;resolution:=optional',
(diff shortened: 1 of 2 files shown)


gradle/mockito-core/osgi.gradle
```diff
@@ -27,88 +18,8 @@ jar {
...
- * It cannot be pinned or removed, and is a source of non-determinism in the build.
- *
- * This class addresses the problem by rewriting the JAR and removing that entry from the Manifest. A side-effect of this
- * is having to set `lastModifiedTime` for each entry to a fixed value, else this would also introduce non-determinism.
- *
- * The fixed value is the same as the default value for Gradle's fixed timestamps, and the rationale behind it is detailed
- * in https://github.com/gradle/gradle/blob/58538513a3bff3b7015718976fe1304e23f40694/subprojects/core/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java#L42-L57
- */
-
-class RemoveOsgiLastModifiedHeader {
```
TimvdLippe, line 57: It's not clear to me why this part was removed and what changed it. Could you please add a comment to the functionality that replaces it to explain why we need it? (E.g. reproducible jars) And could you update your commit message to explain the change that was made.

subprojects/junit-jupiter/junit-jupiter.gradle
```diff
@@ -13,3 +18,43 @@ dependencies {
...
+// metadata.
+tasks.register('osgiProperties', WriteProperties) {
+  outputFile = osgiPropertiesFile
+  property('-standalone', true)
+  property('-runee', "JavaSE-${targetCompatibility}")
+  property('-runrequires', 'osgi.identity;filter:="(osgi.identity=org.mockito.junit-jupiter)"')
+}
+
+// Bnd's Resolve task is what verifies that a jar can be used in OSGi and
+// that its metadata is valid. If the metadata is invalid this task will
```
TimvdLippe, line 50: Could you add an example in the comment what would be invalid?

TimvdLippe (commented):
I understand the usage of the apiguardian, but I am not thrilled of adding another compile dependency to Mockito. I am going to ponder on it a bit to understand why it is used and whether there are alternative solutions available.

gradle/mockito-core/osgi.gradle
rotty3000: bnd has this feature builtin! I enabled it by adding `-noextraheaders: true` rendering this code obsolete.

rotty3000: if you diff the results before and after (which I did to double check) you will see what I mean :)

subprojects/junit-jupiter/junit-jupiter.gradle
rotty3000: An error will cause the build to fail with something like this:
```
> Task :junit-jupiter:verifyOSGi FAILED
Resolution failed. Capabilities satisfying the following requirements could not be found:
    [<<INITIAL>>]
      ⇒ osgi.identity: (osgi.identity=org.mockito.junit-jupiter)
          ⇒ [org.mockito.junit-jupiter version=3.5.11]
              ⇒ osgi.wiring.package: (osgi.wiring.package=com.foo)
```
I triggered this failure by manually importing a fake package `com.foo` (to replicate some package which cannot be satisfied by the available runtime dependencies) which is the same effect for any change that would inadvertently break the OSGi support.
If such a failure ever happens, you can ping me and I would be more than happy to help track down the reason (though I think it might become obvious to you pretty quickly :) )

rotty3000:
@TimvdLippe we _could_ avoid using apiguardian if you prefer.

There are three main options:
- OSGi fragment
- manual definition of "friend" packages
- use apiguardian

Let's consider the differences that would be involved:
**OSGi fragments**
-- simplest solution
-- worst OSGi experience (because there is an ordering problem, most likely to produce issues that people don't understand)
-- all you need to do is mark the mockito-junit-jupiter bundle as being a fragment of mockito-core (one liner)
-- all the packages that are imported by mockito-junit-jupiter
**manual definition of "friend" packages**
-- hardest solution (the idea is a parallel of [_qualified exports_ in JPMS](https://docs.oracle.com/javase/specs/jls/se11/html/jls-7.html#jls-7.7.2))
-- really good OSGi experience with predictable behaviours
-- need to manage packages you want to make "qualified exports" to `mockito-junit-bundle` in a manual way which means adding package attributes to exports/imports of both bundles which will be error prone, won't have any tooling support and painful for even the most experienced OSGi dev.
**apiguardian**
-- middle of the road difficulty
-- same solution as **manual definition of "friend" packages** but managed with a bnd plugin that does the heavy lifting on both sides
-- really good OSGi experience with predictable behaviours
-- requires some java class level metadata (@API) to reflect the intent of certain classes/packages

HTH a little.

rotty3000:
whichever option you choose I can refactor the PR into... it's not a problem. I immediately opted to go with the most ideal solution from the OSGi side... but I understand if you opt to chose something less invasive! :)

TimvdLippe:
Thanks for the explanations! I will think a bit about it and then get back to you 😄

TimvdLippe:
I have filed https://github.com/apiguardian-team/apiguardian/issues/19 to explore adding the annotation on a package statement. Then we would only need to add it to the `package-info.java` of `org.mockito.internal`. Let's see how that discussion develops.

timothyjward:
@rotty3000 Do you think it would be worth reconsidering the fragment approach?

>OSGi fragments
> -- simplest solution
> -- worst OSGi experience (because there is an ordering problem, most likely to produce issues that people don't understand)
> -- all you need to do is mark the mockito-junit-jupiter bundle as being a fragment of mockito-core (one liner)

I'm not sure that I agree that being a fragment would give a poor OSGi experience. The Mockito JUnit 5 "bundle" contains types that are explicitly referenced and used in test classes in an "API" package, and then those make use of internal packages from Mockito. It's not like JUnit 5, where I need a bunch of types that I don't have a direct dependency chain to (e.g. the JUnit engine and launcher).

Essentially if I make use of `@MockitoSettings` or `@ExtendWith(MockitoExtension.class)` then my test class file has a hard link to `org.mockito.junit.jupiter`, then as that package comes from a fragment it would have a hard link to mockito-core (which I probably also have a hard link to by using `@Mock` or `Mockito`. I'm not sure where the possible ordering problem comes in for this case.

While I'm sure the enhanced API Guardian support will be useful in the future I am keen to see this issue fixed in Mockito sooner if possible.

TimvdLippe:
@rotty3000 Could you give an update on this PR?

rotty3000:
I'm doing an updated check on this. Sorry for the delay.

rotty3000:
I'm made small changes that I think simplify the solution and don't require API guardian to be updated or even used directly, but still leverage it's support in bnd (without any changes being required).

i.e. this solution is complete and mergeable as is.

TimvdLippe (approved):
Thanks for all the work that was involved into this!

wborn:
Thanks a lot! It's great to see a fix for the issue merged. :+1:

PR #2051, opened 2020-09-18
Update to OpenJDK 15.

raphw:
Use Java 15 in builds for most recent Java version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,20 +15,20 @@ on:
     branches: ['**']
 
 jobs:
-  
-  # 
-  # Main build job  
+
+  #
+  # Main build job
   #
   build:
     runs-on: ubuntu-latest
     if: "! contains(toJSON(github.event.commits.*.message), '[skip ci]')"
-    
+
     # Definition of the build matrix
     strategy:
       matrix:
-        java: [8, 11, 14]
+        java: [8, 11, 15]
         mock-maker: ['mock-maker-default', 'mock-maker-inline']
-    
+
     # All build steps
     # SINGLE-MATRIX-JOB means that the step does not need to be executed on every job in the matrix
     steps:
@@ -45,7 +45,7 @@ jobs:
 
     - name: 3. Validate Gradle wrapper
       if: matrix.java == 8 && matrix.mock-maker == 'mock-maker-default' # SINGLE-MATRIX-JOB
-      uses: gradle/wrapper-validation-action@v1 # https://github.com/gradle/wrapper-validation-action    
+      uses: gradle/wrapper-validation-action@v1 # https://github.com/gradle/wrapper-validation-action
 
     - name: 4. Build and check reproducibility of artifacts (single job only)
       if: matrix.java == 8 && matrix.mock-maker == 'mock-maker-default' # SINGLE-MATRIX-JOB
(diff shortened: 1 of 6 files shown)


TimvdLippe (approved):

TimvdLippe:
Seems like we got some test failures :(

raphw:
Needed to update Jacoco. Works now.

raphw:
Strangely enough, it builds locally. Will investigate some more!

raphw:
It's weird, if I build locally with the same command and Java version (build and major), I cannot reproduce this error...

TimvdLippe:
I don't think I have Java 15 installed on any of my machines and not sure if I can get a hold of them. Looking at the test failures, many of them seem related to how we handle `null` with primitive matchers. Is there some way to check the changelog to see if we see any suspicious changes?

raphw:
I ran the exact build version locally and it works, on a fresh clone even, also on Linux.

Marcono1234:
I am seeing these failures for JDK 15 (build 15+36-1562) as well, but on Windows.
It looks like the cause for this is [JDK-8233014](https://bugs.openjdk.java.net/browse/JDK-8233014) which enables "Helpful NullPointerExceptions" by default.

The tests which are failing apparently make assumptions about how NullPointerExceptions look like, which are now not true anymore:
* https://github.com/mockito/mockito/blob/4d44781331fd57dda6c31cf1d70151ad04883e2a/src/test/java/org/mockitousage/stubbing/StubbingWithDelegateTest.java#L107
NullPointerException now mentions the name of the method which returned `null`:
  > java.lang.NullPointerException: Cannot invoke "java.lang.Byte.byteValue()" because the return value of "org.mockitousage.IMethods.byteObjectReturningMethod()" is null
* https://github.com/mockito/mockito/blob/4d44781331fd57dda6c31cf1d70151ad04883e2a/src/test/java/org/mockitousage/matchers/InvalidUseOfMatchersTest.java#L109
Assumes that NullPointerException has no message.
* https://github.com/mockito/mockito/blob/4d44781331fd57dda6c31cf1d70151ad04883e2a/src/test/java/org/mockitousage/serialization/DeepStubsSerializableTest.java#L85
Assumes that NullPointerExceptions has no message.

TimvdLippe:
Thanks for the great investigation @Marcono1234 Do you mind creating a PR with the necessary fixes to make it work on JDK 15?

Marcono1234:
I fear that I am not familiar enough with Mockito to perform these changes while still keeping the intended functionality for the tests.
For `StubbingWithDelegateTest` it looks like this could be simplified to checking whether the returned `Byte` is `null` instead of provoking the `NullPointerException`, is that correct? I assume for `DeepStubsSerializableTest` it is similar, it might already suffice to check whether `next()` returned `null`.
But for `InvalidUseOfMatchersTest` maybe the only solution would be to drop the requirement that the message must be `null`.

raphw:
The requirement does not seem meaningful to me, at least. I just fixed the tests to not require an empty message.

Marcono1234 (commented):
I might not be familiar enough with Mockito, but wouldn't it be possible for `StubbingWithDelegateTest` and `DeepStubsSerializableTest` to not provoke the `NullPointerException` at all (as outlined above), but instead test for a `null` return value?  That would probably make the tests more robust since they might otherwise catch unrelated NullPointerExceptions which are caused by bugs.

raphw:
Good point, I just adjusted the tests.

TimvdLippe (approved):
Test changes LGTM.

PR #395, opened 2016-04-19
Fixed method name to "verifyNoMoreInteractions"

vanta:
> Hey, 
> 
> Thanks for the contribution, this is awesome.
> As you may have read, project members have somehow an opinionated view on what and how should be
> Mockito, e.g. we don't want mockito to be a feature bloat.
> There may be a thorough review, with feedback -> code change loop.
> 
> _This block can be removed_
> _Something wrong in the template fix it here `.github/PULL_REQUEST_TEMPLATE.md`

check list

 - [ ] PR should be motivated, i.e. what does it fix, why, and if relevant how
 - [ ] If possible / relevant include an example in the description, that could help all readers
       including project members to get a better picture of the change
 - [ ] Avoid other runtime dependencies
 - [ ] Meaningful commit history ; intention is important please rebase your commit history so that each
       commit is meaningful and help the people that will explore a change in 2 years
 - [ ] Read the [contributing guide](https://github.com/mockito/mockito/blob/master/.github/CONTRIBUTING.md)
 - [ ] Mention `Fixes #<issue number>` in the description _if relevant_
 - [ ] At least one commit should mention `Fixes #<issue number>` _if relevant_

PR #2069, opened 2020-10-16
Migrated to new Shipkit libraries

mockitoguy:
Migrated our build automation to new Shipkit libraries so that:
 - we no longer generate "version bump" commit with each release
 - we no longer change the release notes file in the repo (instead we push release notes to GH releases)
 - we now support "concurrent" PR merges (more info: #395)
 - the build configuration is explicit in Mockito's build files instead of being hidden in the original Shipkit plugin. This way, the build automation is easier to "work with" (fix, debug, tune). The trade-off is a little more Build code in Mockito repo (we can live with that).

This PR does not change other behavior/features of our automation. We are still able to publish to Bintray on every PR and to publish to Maven Central on demand in the typical way.

### Testing
- assembled Maven publications/artifacts for this branch and our main dev branch. Compared contents of the artifacts and poms - we're good.
- ran build locally
- published to Bintray from my local using the new tasks - all worked


mockitoguy:
@TimvdLippe, are you OK if I merge it?

TimvdLippe (commented):
While I am in favor of build simplification (the changes are looking good), we are currently experiencing problems with publishing Mockito 3.6.0. See my failed attempt at https://github.com/mockito/mockito/commit/eba5964a4b051bf566aaf46ca749cde2b36d96ba which broke with Bintray.

Can we get a successful release of Mockito 3.6.0 before we land this PR?

mockitoguy:
> Can we get a successful release of Mockito 3.6.0 before we land this PR?

Absolutely!

mockitoguy:
@TimvdLippe, OK to merge? (will fix the merge conflict beforehand)

TimvdLippe:
Yes let's merge this. To make sure everything is working in order, let's figure out a small patch release to Bintray first that we can piggyback to verify our release mechanism is working. Once we have confident it works with Bintray, we can at some point release a version to Maven Central and monitor how that works.

mockitoguy:
> Yes let's merge this

Cool. I'll be monitoring, I'm also on slack. I've also tested this PR by publishing to Bintray, so we should be good there ;)

version.properties
```diff
@@ -1,5 +1,2 @@
-# Currently building Mockito version
-version=3.6.1
-
-# Previous version used to generate release notes delta
-previousVersion=3.6.0
+//Used by Shipkit Auto Version Gradle plugin: https://github.com/shipkit/shipkit-auto-version
+version=3.5.*
```
TimvdLippe, line 2: This needs to be `3.6.*`

mockitoguy: Great catch, thank you!

build.gradle
```diff
@@ -9,8 +9,10 @@ buildscript {
         classpath 'gradle.plugin.nl.javadude.gradle.plugins:license-gradle-plugin:0.14.0'
         classpath 'net.ltgt.gradle:gradle-errorprone-plugin:1.2.1'
 
-        //Using buildscript.classpath so that we can resolve shipkit from maven local, during local testing
-        classpath 'org.shipkit:shipkit:2.1.6'
+        //Using buildscript.classpath so that we can resolve plugins from maven local, during local testing
+        classpath "org.shipkit:shipkit-auto-version:0.0.+"
```
TimvdLippe, line 13: I would prefer if we lock the version down here, to make sure we can control when we update Shipkit. WDYT?

TimvdLippe (approved):

build.gradle
mockitoguy: Sure! OK if I lock it in one of the next PRs? I'd like to move off Travis because the builds are taking forever.

PR #2071, opened 2020-10-17
Fixes #2061: ArgumentMatcher error messages use lambda class names

### LINKED ISSUE #2061: Lambda used as ArgumentMatcher causes decamelized lambda name to appear in error message
### Version
Mockito version: 3.5.13

### Description
Relates to #1932

When using a lambda or method reference expression as `ArgumentMatcher` and the matcher fails, the error message contains the lambda / method reference class name in "decamelized" form which is rather irritating, e.g.:
> myInterface.doSomething(
    <Mockito test$$ lambda$ 4 8/ 0x 0 0 0 0 0 0 0 8 0 0c 8f 4 4 0>
);

Expected would be that similar to anonymous classes the error output says "\<custom argument matcher\>". This could probably be achieved by testing for `Class.isSynthetic()` in `org.mockito.internal.matchers.text.MatcherToString.toString(ArgumentMatcher<?>)` before calling `decamelizeMatcher` there.

### Test case
```java
public class MockitoTest {
    interface MyInterface {
        void doSomething(String s);
    }
    
    private static boolean matches(String s) {
        return false;
    }
    
    public static void main(String[] args) {
        MyInterface mock = mock(MyInterface.class);
        try {
            verify(mock).doSomething(argThat(arg -> false));
        } catch (Error e) {
            e.printStackTrace();
        }
        try {
            verify(mock).doSomething(argThat(MockitoTest::matches));
        } catch (Error e) {
            e.printStackTrace();
        }
        
        // When an anonymous class is used the output is '<custom argument matcher>'
        // (as expected)
        verify(mock).doSomething(argThat(new ArgumentMatcher<>() {
            @Override
            public boolean matches(String argument) {
                return false;
            }
        }));
    }
}
```

Marcono1234:
Fixes #2061

### DIFF
--- a/src/main/java/org/mockito/internal/matchers/text/MatcherToString.java
+++ b/src/main/java/org/mockito/internal/matchers/text/MatcherToString.java
@@ -5,7 +5,7 @@
 package org.mockito.internal.matchers.text;
 
 import static org.mockito.internal.util.ObjectMethodsGuru.isToStringMethod;
-import static org.mockito.internal.util.StringUtil.decamelizeMatcher;
+import static org.mockito.internal.util.StringUtil.decamelizeMatcherName;
 
 import java.lang.reflect.Method;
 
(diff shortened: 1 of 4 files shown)


TimvdLippe (approved):
Nice fix!

PR #2084, opened 2020-10-28
Fixes #2083 : Modify casing of Bundle-Version manifest header

### LINKED ISSUE #2083: Issue with Bundle-Version manifest header when used in Eclipse PDE development
The MANIFEST.MF currently contains e.g.: `Bundle-version: 3.6.0`

Due to the lowercase 'v' this appears to not be properly recognised by Eclipse PDE so is giving me errors on min versioned require bundle directives.

The standard headers at  https://www.osgi.org/bundle-headers-reference/ define it as `Bundle-Version` (capital V).

After a small amount of research it looks like OSGi specification possibly does accept case insensitive headers but haven't found definitive source.

In any case, changing it would keep it consistant with the other headers.

joeshannon:
OSGi specification may accept case insensitive headers but this is
causing problems for Eclipse PDE development with version constriants.

Switch to match casing of other headers and agree with name provided
at: https://www.osgi.org/bundle-headers-reference/

### DIFF
--- a/gradle/mockito-core/osgi.gradle
+++ b/gradle/mockito-core/osgi.gradle
@@ -14,7 +14,7 @@ jar {
     bnd(
         'Bundle-Name': 'Mockito Mock Library for Java. Core bundle requires Byte Buddy and Objenesis.',
         'Bundle-SymbolicName': 'org.mockito.mockito-core',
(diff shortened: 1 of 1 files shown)


TimvdLippe (approved):
Thanks for the quick fix!

PR #2091, opened 2020-11-08
Remove duplicated Gradle property setting

shestee:
There is no need of specifying _--parallel_ explicitly as it is Gradle's default (the less code, the better).


mockitoguy (approved):
Thanks!

TimvdLippe (approved):
Thanks

PR #2158, opened 2020-12-30
Add field type in @Mock example

pzygielo:
javadoc-only change
<sup>(IMO not worth separate issue nor RN mentioning.)</sup>


mockitoguy:
Nice, thanks for fixing!

Issue #2159, opened 2020-12-30
Mockito published maven artifact is missing .module files


checketts:
Around Gradle 5.3 new Gradle metadata was introduced (see https://docs.gradle.org/current/userguide/publishing_gradle_module_metadata.html) This allows Gradle builds to publish more detail about dependencies than the lagacy maven pom.

Mockito is generating those files, but they aren't getting uploaded to Maven Central see: https://repo1.maven.org/maven2/org/mockito/mockito-junit-jupiter/3.6.28/

However in the [pom.xml](https://repo1.maven.org/maven2/org/mockito/mockito-junit-jupiter/3.6.28/mockito-junit-jupiter-3.6.28.pom) it contains the comment that tells Gradle to check for the `.module` files:

```
<!--
 This module was also published with a richer model, Gradle metadata,  
-->
<!--
 which should be used instead. Do not delete the following line which  
-->
<!--
 is to indicate to Gradle or any Gradle module metadata file consumer  
-->
<!-- that they should prefer consuming it instead. -->
<!-- do_not_remove: published-with-gradle-metadata -->
```

## Solution 1 [Disable metadata](https://docs.gradle.org/current/userguide/publishing_gradle_module_metadata.html#sub:disabling-gmm-publication ): 

This will remove the comment from the pom.xml and so Gradle won't continue to check for the `.module` file.

```
    tasks.withType<GenerateModuleMetadata> {
        enabled = false
    }
```

## Solution 2- Update the maven publishing to include the missing files

This is the better long-term solution. However I don't have a direct recommendation since I don't have experience in that regard and don't know what is mis-configured.

## Why hasn't this been a problem before?

When using `mavenCentral()` as a repo directly, it serves up a 404 for the missing `.module` file and Gradle recovers fine. However when using an internal repository like Artifactory, it is common to configure it to return `403` for missing files to try to obscure data that shouldn't be accessible.

So to reproduce this you would need to use Gradle > 5.3, Using Mockito as a dependency, when using a locked down Maven repository that is configured to serve 403 instead of exposing 404 for missing artifacts.

PR #2160, opened 2020-12-30
Avoid Gradle module files

mockitoguy:
Avoid Gradle module files until there is a use case for them. The less we publish the less complexity.

Fixes: #2159

Tested by running `./gradlew publishToMavenLocal' and inspecting files / poms.


TimvdLippe (approved):

mockitoguy:
I'm going to merge it. I believe that the failure will be fixed by my earlier PR.

PR #2175, opened 2021-01-15
ArgumentMatches: fix Javadoc typo/grammar

perlun:
I started out with noticing the "safer that it was" part which felt like a typo. Then I realized there were perhaps more errors in this sentence (which was duplicated in multiple places in this file). Please verify that I understood the intent of this comment correctly, but my hope is that this is now the more correct way to express this.

Again, thanks for a great framework! :bow:


TimvdLippe (approved):
Thanks!

PR #2180, opened 2021-01-17
Fix typo in Javadocs of MockedConstruction

hotire:
Hi 🖐

I noted this while reading the Javadocs of MockedConstruction.

They were referring to static method and out of scope variable. It isn't quite easy to grasp in this context.
 
I think it's just a copy-paste mistake.


TimvdLippe (approved):
Good catch!

PR #2188, opened 2021-01-22
"But invoked here" lists the invocation parameters

### LINKED ISSUE #2058: "But invoked here" should list the invocation parameters
Thanks to Mockito, I just had an assertion failure that I just had a failure thanks to Mockito told me of an issue in my code that I need to fix. The error message looked like this:

```
cLogger.saveError("Error", <any string>);
Never wanted here:
-> at au.org.greekwelfaresa.idempiere.bpmodels.test.MBPartnerLocationTest.lambda$1(MBPartnerLocationTest.java:118)
But invoked here:
-> at org.compiere.model.PO.save(PO.java:2155)
-> at org.compiere.model.PO.save(PO.java:2155)
```
It occurred to me that it might be useful debugging to include not only the *place* where the invocation happened, but the arguments that were provided for each invocation. Because of the `anyString()` matcher, it is not immediately obvious what they are - if you're using exact match semantics, then you can infer the parameters from the first line, but that's not possible when you're using a matcher.

This feature could be added on all failure messages, or else only on failure messages when matchers are being used (as it is redundant when you're using exact match).

EugeneLesnov:
Fixes #2058

### DIFF
--- a/src/main/java/org/mockito/internal/exceptions/Reporter.java
+++ b/src/main/java/org/mockito/internal/exceptions/Reporter.java
@@ -10,9 +10,7 @@ import static org.mockito.internal.util.StringUtil.join;
 
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.List;
+import java.util.*;
 
 import org.mockito.exceptions.base.MockitoAssertionError;
 import org.mockito.exceptions.base.MockitoException;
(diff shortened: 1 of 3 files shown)


EugeneLesnov:
It might be a good idea to add similar code for `Reporter#tooManyActualInvocations`

TimvdLippe (commented):
This PR roughly looks okay. I will do some polish tomorrow to clean some things and improve test coverage and then I will merge it. Thanks!

TimvdLippe:
@EugeneLesnov I don't have permission to push to your PR. Could you please pull in https://github.com/mockito/mockito/tree/but-invoked-here ?

PR #2192, opened 2021-02-01
Fixed 2154 : java.lang.Object is now instrumented again

JanMosigItemis:
Seems like 

```
// The object type does not ever need instrumentation.
targets.remove(Object.class);
```

in `org.mockito.internal.creation.bytebuddy.InlineBytecodeGenerator.triggerRetransformation(Set<Class<?>>, boolean)` was a bit over the top, i. e. it does seem to break mocking of toString when inherited from `java.lang.Object`.

I added a test for this which should have been there in the first place. Due to this, I am still not sure if removing the line is safe, even if the whole test suite is green. Please review carefully.


JanMosigItemis:
Got the commit message wrong. Will open a new one.

PR #2193, opened 2021-02-01
Fixes #2154 : java.lang.Object is now instrumented again

### LINKED ISSUE #2154: `toString()` seems to be broken on 3.6.28
Hi guys, I think the `toString()` method was broken recently, I have a test for mockito-scala that checks a mock `toString` invocation returns its name, but if I try to update to said version it returns the standard `Object` implementation of `toString`.

I tried with the "pure" Java API of Mockito and the same thing happens

Thanks!

JanMosigItemis:
Seems like 

```
// The object type does not ever need instrumentation.
targets.remove(Object.class);
```

in `org.mockito.internal.creation.bytebuddy.InlineBytecodeGenerator.triggerRetransformation(Set<Class<?>>, boolean)` was a bit over the top, i. e. it does seem to break mocking of `toString` when inherited from `java.lang.Object`.

I added a test for this which should have been there in the first place. Due to this, I am still not sure if removing the line is safe, even if the whole test suite is green. Please review carefully.

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/InlineBytecodeGenerator.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/InlineBytecodeGenerator.java
@@ -269,9 +269,6 @@ public class InlineBytecodeGenerator implements BytecodeGenerator, ClassFileTran
             throw t;
         }
 
-        // The object type does not ever need instrumentation.
-        targets.remove(Object.class);
-
         if (!targets.isEmpty()) {
             try {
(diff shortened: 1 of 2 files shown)


TimvdLippe (approved):
This is great, thanks for fixing! Added @raphw for a quick double-check, but given that all tests are passing I am assuming this is WAI.

raphw:
Yeah, good point. I don't know why I ever added that.

raphw (approved):

Issue #2178, opened 2021-01-15
Performance degression on Upgrade from 1.10.19 to current 3.6.28


keiki85:
Hello Mockito team,

Quick stats:
Java 11
Mockito 1.10.19 to 3.6.28
1:30 minutes to 9 minutes

in our company we still use Mockito 1.10.19, which includes cglib.

In our unit tests we kind of have our own testing framework that encapsulates e.g. the persistence layer. And in the actual unit tests we don't need to actual mock. With that in mind the framework is already a lot of preparations which is not be needed for every test.
We use mainly mock creations and methodinterceptors as far as I can tell.

At the moment an upgrade wouldn't make sense for us. As the duration of the tests would increase from 1 minute and 30 sec to roughly 9 minutes so 6 times worse.

I had a first look what's taking so long. In my opinion there are two things:
1.
org/mockito/internal/creation/bytebuddy/MockMethodInterceptor.java:55
In ` return doIntercept(mock, invokedMethod, arguments, realMethod, new LocationImpl())` you construct and initliaise a new LocationImpl, which is quite expensive according to my CPU profiling (IntelliJ profiler)

After replacing that part with a dummy Locationimpl, which does nothing and is a static instance I got down to something like 5 min from 9.

I dunno exactly what the stacktrace information is being used for. It  seems in 1.x this is not being done.

2. the mock creation itself

Here it seems to me that bytebuddy is doing way more complicated and of course as I understand more feature rich type traversal, which is making things go slow, but on the other hand work .

Here I'm wondering, if there is a global config switch to do just simple mocking like cglib has done previously.

I know that in general the architecture has changed significantly from version 1 to 2/3 to have features/bugfixes that were impossible beforehand. So I'm not sure what performance wise is feasible without going completely a different way.

Thank you in advance for your help ;)

TimvdLippe:
Thanks for filing this issue! We are aware that Mockito 2 is slower at creating mocks, but this is mostly due to the fact that the old CGLib mockmaker was not spec-compliant and was creating classes that violated numerous JVM semantics. As such, CGLib was sometimes flat out wrong and was producing invalid results in tests. We made the trade-off to value correctness over performance in this specific instance, given that tests are used to ensure correctness of production code.

That said, if there are opportunities to improve the situation, we are definitely open for that. Could you let us know how many mocks you create as part of your test suite and maybe how many interactions you perform with these mocks?

[...]

PR #2194, opened 2021-02-01
Add API for clearing mocks.

raphw:
By clearing mocks, caches are emptied and instrumentations are reversed.


TimvdLippe (approved):
Approved pending feedback of #2178

PR #2207, opened 2021-02-15
Fixes #2204: Ignore Groovy methods annotated with Internal

### LINKED ISSUE #2204: Groovy 3.0 internal methods are not ignored as they should be
Hello,

The following test passes with Groovy 2.4 and 2.5, and fails with Groovy 3.0 (with Mockito 3.7.7 and JUnit 4.13.1 in all cases):

```groovy
import org.junit.Test
import org.junit.runner.RunWith
import org.mockito.InjectMocks
import org.mockito.Mock
import org.mockito.junit.MockitoJUnitRunner

import static org.mockito.Mockito.verify

@RunWith(MockitoJUnitRunner)
class GroovyMockitoTest {
    static class Helper {
        void helper() { }
    }

    static class Service {
        private final Helper helper

        Service(Helper helper) {
            this.helper = helper
        }

        void service() {
            helper.helper()
        }
    }

    @Mock Helper helper
    @InjectMocks Service service

    @Test
    void testService() {
        service.service()
        verify(helper).helper()
    }
}
```

The actual error:
```
	at org.codehaus.groovy.runtime.callsite.PogoMetaClassSite.call(PogoMetaClassSite.java:38)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:47)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:125)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:130)
	at GroovyMockitoTest$Service.service(GroovyMockitoTest.groovy:24)
	at GroovyMockitoTest$Service$service.call(Unknown Source)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:47)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:125)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:130)
	at GroovyMockitoTest.testService(GroovyMockitoTest.groovy:33)
	[... junit ...]
```

I guess this is like #72, where `getMetaClass()` is not ignored and returns `null`.

I think the difference is that in Groovy 3.0, internal methods such as `getMetaClass()` are no longer marked `synthetic`: [GROOVY-8495](https://issues.apache.org/jira/browse/GROOVY-8495). Now they are `@groovy.transform.Generated` and `@groovy.transform.Internal`.

As far as I can tell, Mockito works with Groovy 2.4/2.5 only because ByteBuddy ignores synthetic methods. I see the existing `ignoreAlso(isGroovyMethod())` but as far as I can tell, it has no effect with current versions of Groovy. (I assume it did in the past.)

I think that the fix is to ignore methods annotated with `@groovy.transform.Internal`. I don't know whether that should be done in Mockito or in ByteBuddy. Please let me know if I should report it somewhere else, or if you'd like me to ask the Groovy developers for their input.

Thank you for working on Mockito!

rtandy:
Starting from Groovy 3.0, compiler-generated methods such as `getMetaClass()` [are no longer marked synthetic](https://issues.apache.org/jira/browse/GROOVY-8495), therefore ByteBuddy stopped automatically ignoring them. Instead they are now annotated with `@groovy.transform.Internal`. Update `isGroovyMethod()` to also check this new annotation.

#2204 has a test case, and this change fixes it.

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -35,6 +35,8 @@ libraries.osgi = 'org.osgi:osgi.core:8.0.0'
 libraries.equinox = 'org.eclipse.platform:org.eclipse.osgi:3.16.100'
 libraries.bndGradle =  'biz.aQute.bnd:biz.aQute.bnd.gradle:5.2.0'
 
+libraries.groovy = 'org.codehaus.groovy:groovy:3.0.7'
+
 def kotlinVersion = '1.4.30'
 libraries.kotlin = [
     version: kotlinVersion,
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -5,6 +5,7 @@ plugins {
 include("deprecatedPluginsTest",
     "inline",
     "extTest",
+    "groovyTest",
     "kotlinTest",
     "kotlinReleaseCoroutinesTest",
     "android",
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
@@ -226,7 +226,8 @@ class SubclassBytecodeGenerator implements BytecodeGenerator {
     }
 
     private static ElementMatcher<MethodDescription> isGroovyMethod() {
-        return isDeclaredBy(named("groovy.lang.GroovyObjectSupport"));
+        return isDeclaredBy(named("groovy.lang.GroovyObjectSupport"))
+                .or(isAnnotatedWith(named("groovy.transform.Internal")));
     }
 
     private boolean isComingFromJDK(Class<?> type) {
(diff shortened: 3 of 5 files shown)


TimvdLippe (commented):
Thanks for the PR! Just to check: how difficult would it be to add the regression test from #2204 to our test suite? I can imagine us adding a new `subproject` like we did for `kotlinReleaseCoroutinesTest` to verify the Groovy integration works. Let me know if that is feasible.

rtandy:
Thanks for the subproject suggestion. I hadn't attempted a test because I thought I'd have to invent an entire test class for SubclassBytecodeGenerator. This was much easier.

TimvdLippe (approved):
Awesome, thank you so much for the fix and the regression test!

PR #2222, opened 2021-03-06
Primitives: make internal lookup maps unmodifiable

arteam:
Make sure they are used only for lookups.


TimvdLippe (commented):
These maps are private and are not exposed outside of this class. Is there a particular motivation you want to make them unmodifiable?

TimvdLippe:
Closing per the above comment.

PR #2225, opened 2021-03-09
Use java.lang.StackWalker instead of sun.misc.*

TimvdLippe:
`java.lang.StackWalker` is the public API shipped in Java 9. Tentatively opening this PR to see whether the memory improvements are still there.


TimvdLippe:
I didn't get to this and the memory test has been disabled (#2478). Let's revisit this as part of resolving #2478

Issue #2235, opened 2021-03-22
verify() on a wrapped mock fails with UnfinishedVerificationException


mschieder:
I am currently using the Mock resolver plugin feature (#2042) in my projects, but verify() on a wrapped mock does not work as expected and throws a UnfinishedVerificationException.

A junit5 test case to reproduce this bug is already in the extTest module: [MockResolverTest](https://github.com/mockito/mockito/blob/3fe4a18ba90462e8f6fa5677dfb6d0dc5ab4df17/subprojects/extTest/src/test/java/org/mockitousage/plugins/resolver/MockResolverTest.java,) but it is not executed by default.

https://github.com/mockito/mockito/blob/3fe4a18ba90462e8f6fa5677dfb6d0dc5ab4df17/subprojects/extTest/src/test/java/org/mockitousage/plugins/resolver/MockResolverTest.java#L20-L27

After adding the junitJupiterEngine as test runtime, the MockResolverTest is executed, but fails with following message:
<pre>Missing method call for verify(mock) here:
-> at org.mockitousage.plugins.resolver.MockResolverTest$MockWrapper.doIt(MockResolverTest.java:47)

Example of correct verification:
    verify(mock).doSomething()

Also, this error might show up because you verify either of: final/private/equals()/hashCode() methods.
Those methods *cannot* be stubbed/verified.
Mocking methods declared on non-public parent classes is not supported.

org.mockito.exceptions.misusing.UnfinishedVerificationException: 
Missing method call for verify(mock) here:
-> at org.mockitousage.plugins.resolver.MockResolverTest$MockWrapper.doIt(MockResolverTest.java:47)

Example of correct verification:
    verify(mock).doSomething()

Also, this error might show up because you verify either of: final/private/equals()/hashCode() methods.
Those methods *cannot* be stubbed/verified.
Mocking methods declared on non-public parent classes is not supported.

	at org.mockito.internal.exceptions.Reporter.unfinishedVerificationException(Reporter.java:115)
	at org.mockito.internal.progress.MockingProgressImpl.validateMostStuff(MockingProgressImpl.java:124)
	at org.mockito.internal.progress.MockingProgressImpl.validateState(MockingProgressImpl.java:105)
	at org.mockito.internal.MockitoCore.validateMockitoUsage(MockitoCore.java:248)
	at org.mockito.Mockito.validateMockitoUsage(Mockito.java:3248)
[...]

PR #2236, opened 2021-03-22
Bugfix/2235 verify on a wrapped mock fails

mschieder:
PR for #2235: verify() on a wrapped mock fails with UnfinishedVerificationException


TimvdLippe (approved):
Initially I wanted to request a regression test, until I read the comment in #2235 which stated that the test already exists, but isn't run because of a Gradle misconfiguration. Therefore, this all LGTM! Thanks for the great issue description and quick PR 😄

PR #2240, opened 2021-03-24
Fix verifyZeroInteractions replacement suggestion

Ekrekr:
Particularly helpful for stopping intellisense suggesting `verifyMoreInteractions` instead of the functionally identical `verifyNoMoreInteractions`

check list

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -2539,7 +2539,7 @@ public class Mockito extends ArgumentMatchers {
      * This method has the same behavior as {@link #verifyNoMoreInteractions(Object...)}.
      *
      * @param mocks to be verified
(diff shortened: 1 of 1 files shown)


TimvdLippe (commented):
I think this replacement is intentional. The semantic intention of `verifyZeroInteractions` is the behavior of `verifyNoInteractions`. However, Mockito didn't do so and was using the behavior of `verifyNoMoreInteractions`. So while you are correct to conclude that the functionally equivalent method is `verifyNoMoreInteractions`, the semantically appropriate replacement would be `verifyNoInteractions`. Therefore, I think we should keep it as-is,  unless I am missing something?

Ekrekr:
Ah interesting! I agree that it makes sense from that context, I will close this PR.

PR #2241, opened 2021-03-25
Fixes #2238: Do not omit stack frames of causing class

### LINKED ISSUE #2238: Stacktraces of exceptions are trimmed too much
Hi,

I'm not sure if this is a bug or an intended behaviour.

When testing a method which throws an exception (checked or unchecked), the causing line is suppressed in the stack coming from Mockito.

Simple example:

```java
package test;

import java.io.IOException;

import org.mockito.Mockito;

public class TestSpy {
    public static void main(String[] args) throws IOException {
        Foo spy = Mockito.spy(new Foo());
        spy.message();
    }

    public static class Foo {
        public void message() throws IOException {
            System.out.println("Hello 1!"); // line 15
            System.out.println("Hello 2!");
            throw new IOException("fatal"); // line 17
        }
    }
}
```

When calling 'message()' on the spy (same happens on a full mock as well), I would expect to see line 17 in the exception cause.
Instead I see this:
```
Exception in thread "main" java.io.IOException: fatal
	at test.TestSpy$Foo.message(TestSpy.java:15)
	at test.TestSpy.main(TestSpy.java:10)
```

Line 15 contains 'System.out.println', which is not causing the exception, but is the first expression inside of the method 'message()'.

I debugged this behaviour and came to `org.mockito.internal.creation.bytebuddy.MockMethodAdvice.hideRecursiveCall(Throwable, int, Class<?>)` 
which seems to be the root cause for this. 
The stack trace given to that method includes the correct line (line 17 in this case). 

My stack trace in this case has 18 entries in total:
```
java.io.IOException: fatal
	at test.TestSpy$Foo.message(TestSpy.java:17)
	at java.base/java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:710)
	at org.mockito.internal.util.reflection.InstrumentationMemberAccessor$Dispatcher$ByteBuddy$2kmhnFkd.invokeWithArguments(Unknown Source)
	at org.mockito.internal.util.reflection.InstrumentationMemberAccessor.invoke(InstrumentationMemberAccessor.java:230)
	at org.mockito.internal.util.reflection.ModuleMemberAccessor.invoke(ModuleMemberAccessor.java:43)
	at org.mockito.internal.creation.bytebuddy.MockMethodAdvice.tryInvoke(MockMethodAdvice.java:329)
	at org.mockito.internal.creation.bytebuddy.MockMethodAdvice.access$500(MockMethodAdvice.java:56)
	at org.mockito.internal.creation.bytebuddy.MockMethodAdvice$RealMethodCall.invoke(MockMethodAdvice.java:249)
	at org.mockito.internal.invocation.InterceptedInvocation.callRealMethod(InterceptedInvocation.java:141)
	at org.mockito.internal.stubbing.answers.CallsRealMethods.answer(CallsRealMethods.java:44)
	at org.mockito.Answers.answer(Answers.java:98)
	at org.mockito.internal.handler.MockHandlerImpl.handle(MockHandlerImpl.java:106)
	at org.mockito.internal.handler.InvocationNotifierHandler.handle(InvocationNotifierHandler.java:33)
	at org.mockito.internal.creation.bytebuddy.MockMethodAdvice.handle(MockMethodAdvice.java:147)
	at test.TestSpy$Foo.message(TestSpy.java:15)
	at test.TestSpy.main(TestSpy.java:10)
```

When the method is finished, 13 entries are left and the most important entry is missing:
```
java.io.IOException: fatal
	at org.mockito.internal.creation.bytebuddy.MockMethodAdvice.tryInvoke(MockMethodAdvice.java:329)
	at org.mockito.internal.creation.bytebuddy.MockMethodAdvice.access$500(MockMethodAdvice.java:56)
	at org.mockito.internal.creation.bytebuddy.MockMethodAdvice$RealMethodCall.invoke(MockMethodAdvice.java:249)
	at org.mockito.internal.invocation.InterceptedInvocation.callRealMethod(InterceptedInvocation.java:141)
	at org.mockito.internal.stubbing.answers.CallsRealMethods.answer(CallsRealMethods.java:44)
	at org.mockito.Answers.answer(Answers.java:98)
	at org.mockito.internal.handler.MockHandlerImpl.handle(MockHandlerImpl.java:106)
	at org.mockito.internal.handler.InvocationNotifierHandler.handle(InvocationNotifierHandler.java:33)
	at org.mockito.internal.creation.bytebuddy.MockMethodAdvice.handle(MockMethodAdvice.java:147)
	at test.TestSpy$Foo.message(TestSpy.java:15)
	at test.TestSpy.main(TestSpy.java:10)
```

Disabling "cleansStackTrace" using MockitoConfiguration does not help - hideRecursiveCall is always called.

This bug/feature is very annoying when testing methods which may throw the same type of exception at different locations.
Eg. in my case I know that the exception was thrown in that very method, but I don't know if it happens in e.g. line 20 or 85.

Same thing happens to `RuntimeException`s, which is also annoying... Seeing a NullPointerException without knowing where it happens is not helpful.

Is there any (configuration) option to disable the 'cleaning' of the stack trace completely?
Any chance of getting this fixed?

**Tested Mockito versions:** 3.6.0 and 3.8.0
**OS:** Linux/Ubuntu 20.04 64-Bit
**JDK:** OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.9+11)

hypfvieh:
This PR will fix this problems described in #2238. 

The main cause wasn't the `hideRecursiveCall` method but `tryInvoke` method.
In `tryInvoke` the original stacktrace will be passed through the configured filter after it has been modified by `hideRecursiveCall`.

When calling `hideRecursiveCall` the second argument (`current`) is set the the length of the stack trace of the causing exception.
This will omit too much stack frames, ultimately skipping the lines containing of the root exception cause.

In this patch, `tryInvoke` will try to find a better starting frame when an exception is caught.
It will do this by investigating the created stack trace and look for the first entry matching the class name of the `instance` object given to `tryInvoke`. This offset will then be given to `hideRecursiveCall` as `current` argument instead of using the stack trace length.
If there is no `instance` object (`null`), the length of the stack trace is used (like before).

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodAdvice.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodAdvice.java
@@ -4,6 +4,11 @@
  */
 package org.mockito.internal.creation.bytebuddy;
 
+import static net.bytebuddy.matcher.ElementMatchers.isConstructor;
+import static net.bytebuddy.matcher.ElementMatchers.isPrivate;
+import static net.bytebuddy.matcher.ElementMatchers.isStatic;
+import static net.bytebuddy.matcher.ElementMatchers.not;
+
 import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.Serializable;
@@ -38,6 +43,7 @@ import net.bytebuddy.jar.asm.Opcodes;
 import net.bytebuddy.jar.asm.Type;
 import net.bytebuddy.pool.TypePool;
 import net.bytebuddy.utility.OpenedClassReader;
+
 import org.mockito.exceptions.base.MockitoException;
 import org.mockito.internal.configuration.plugins.Plugins;
 import org.mockito.internal.creation.bytebuddy.inject.MockMethodDispatcher;
@@ -51,8 +57,6 @@ import org.mockito.internal.util.concurrent.DetachedThreadLocal;
 import org.mockito.internal.util.concurrent.WeakConcurrentMap;
 import org.mockito.plugins.MemberAccessor;
 
-import static net.bytebuddy.matcher.ElementMatchers.*;
-
 public class MockMethodAdvice extends MockMethodDispatcher {
 
     private final WeakConcurrentMap<Object, MockMethodInterceptor> interceptors;
@@ -63,7 +67,7 @@ public class MockMethodAdvice extends MockMethodDispatcher {
     private final SelfCallInfo selfCallInfo = new SelfCallInfo();
     private final MethodGraph.Compiler compiler = MethodGraph.Compiler.Default.forJavaHierarchy();
     private final WeakConcurrentMap<Class<?>, SoftReference<MethodGraph>> graphs =
-            new WeakConcurrentMap.WithInlinedExpunction<Class<?>, SoftReference<MethodGraph>>();
+            new WeakConcurrentMap.WithInlinedExpunction<>();
 
     private final Predicate<Class<?>> isMockConstruction;
     private final ConstructionCallback onConstruction;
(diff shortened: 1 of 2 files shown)


TimvdLippe (approved):
Great fix, thanks for contributing!

PR #2243, opened 2021-03-26
CI overhaul, migrated from Bintray

mockitoguy:
We need to remove the dependency on Bintray/JCenter


TimvdLippe (changes requested):
Oh actually, I think we need a similar fix to https://github.com/mockito/mockito-kotlin/pull/425 where we have to drop the `tags-ignore` from line 13 and match on tags of `3.*`

gradle/shipkit.gradle
```diff
@@ -17,28 +16,45 @@ tasks.named("githubRelease") {
...
-    boolean centralRelease = centralReleaseByCommit || centralReleaseByProjectProperty
-    def message = """Bintray publish was using following settings:
- - commit message contains '[ci maven-central-release]': $centralReleaseByCommit
- - project property 'maven-central-release' exists: $centralReleaseByProjectProperty
- - Maven Central release is enabled: $centralRelease"""
-    project.logger.info(message)
-    project.afterEvaluate {
-        project.tasks.bintrayPublish.doLast { logger.lifecycle(message) }
+                username = System.getenv('NEXUS_TOKEN_USER')
+                password = System.getenv('NEXUS_TOKEN_PWD')
```
szpak, line 29: Cool that you use the [user token](https://blog.solidsoft.pl/2015/09/08/deploy-to-maven-central-using-api-key-aka-auth-token/) instead of the regular credentials!

gradle/shipkit.gradle
```diff
@@ -17,28 +16,45 @@ tasks.named("githubRelease") {
...
+tasks.register("releaseSummary") {
+    doLast {
+        if (isSnapshot) {
+            println "RELEASE SUMMARY\n" +
+                "  SNAPSHOTS released to: https://s01.oss.sonatype.org/content/repositories/snapshots/org/mockito/mockito-core\n" +
+                "  Release to Maven Central: SKIPPED FOR SNAPSHOTS\n" +
+                "  Github releases: SKIPPED FOR SNAPSHOTS"
+        } else {
+            println "RELEASE SUMMARY\n" +
+                "  Release to Maven Central (available in few hours): https://repo1.maven.org/maven2/org/mockito/mockito-core/\n" +
```
szpak, line 56: Did they degrade so much recently? It's true that the search index is updated after a few hours, but when I checked it a year ago (or so), the artifacts were available in `https://repo1.maven.org/maven2/...` (and ready for usage in Gradle/Maven/...) after a few (maybe several) minutes.

You can check that with the next Mockito release and update that statement if needed (e.g. to change the hours to minutes :) ).

mockitoguy: Sounds good!

PR #2247, opened 2021-03-31
Replace usage of Function with custom FunctionalInterface

TimvdLippe:
`java.util.function.Function` is not available in older versions of
Android. We can replace it with a custom `@FunctionalInterface`
interface that serves the same purpose.

### DIFF
--- a/src/main/java/org/mockito/MockSettings.java
+++ b/src/main/java/org/mockito/MockSettings.java
@@ -368,4 +368,9 @@ public interface MockSettings extends Serializable {
      */
     @Incubating
     MockSettings lenient();
+
+    @FunctionalInterface
+    interface MockSettingsFactory {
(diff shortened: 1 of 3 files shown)


TimvdLippe:
There appear to be more problems, but this might be an issue internally at Google. I am consulting Android engineers atm to figure out the best way forward.

raphw:
It's a binary incompatible change, but a source compatible one for lambda application, I think this can be defended. We use `Predicate` in the mock makers. Should we redo those, too?

raphw (approved):

TimvdLippe:
Yeah these are problematic, but also the `Optional` return types. So I am still consulting with our Android experts as to what options we have here. Will keep you posted.

TimvdLippe:
Since this is a breaking change, we can't land this. We mitigated the impact of this change downstream.

PR #2248, opened 2021-03-31
Fix thread race condition

TimvdLippe:
Discovered internally at Google. The culprit was a thread race condition
around the usage of usedAt, which could race between setting the field
and retrieving it.


TimvdLippe:
Reviewed internally at Google. Assuming no objections on this PR, so merging.

PR #2263, opened 2021-04-09
Fixes #2242 : Make exceptions for Inorder.verify more standard / spec…

### LINKED ISSUE #2242: InOrder.verify with non-mock should perhaps give standard non-mock warning
Minor, but I noticed today when I accidentally passed in a non-mock to an `InOrder.verify` call that it warned me that it can only verify mocks that were passed in during creation of `InOrder`.  While correct, it seems like perhaps using the standard "non-mock" warning would be more helpful as it has better hinting for this situation.  In particular, this seems like it would be useful to newer users of the API who might have a hard time figuring out what they did wrong.

## Mockito.verify

```java
List<?> mock = Mockito.mock(List.class);
Mockito.verify(Mockito.mock.get(5));
```

### Output

```text
Argument passed to verify() should be a mock but is null!
Examples of correct verifications:
    verify(mock).someMethod();
    verify(mock, times(10)).someMethod();
    verify(mock, atLeastOnce()).someMethod();
    not: verify(mock.someMethod());
Also, if you use @Mock annotation don't miss initMocks()
```

## inOrder.verify

```java
List<?> mock = Mockito.mock(List.class);
InOrder inOrder = Mockito.inOrder(mock);
inOrder.verify(mock.get(5));
```

### Output

```text
InOrder can only verify mocks that were passed in during creation of InOrder.
For example:
    InOrder inOrder = inOrder(mockOne);
    inOrder.verify(mockOne).doStuff();
```

## Desired behavior

Note that the current `InOrder` message does not give the hint of how to fix the non-mock that was passed in.  Using the standard non-mock warning in this case seems like a more helpful response.  Alternatively, that the hinting could be included in the `InOrder` output when a non-mock is used.

devin040:
…ific

*Adds NullInsteadOfMockException and NotAMockException checks to Inorder.verify() *before* checking if mocks were passed in during creation of Inorder

See https://github.com/mockito/mockito/issues/2242 for a description of the issue.

Inorder.verify(mock) now first checks if the passed argument is null, and if the mock is in fact a mock, and gives appropriate Exception messages, before checking if the passed in argument was passed during creation of Inorder object, and therefore giving a less specific error message.

## Checklist

 - [x ] Read the [contributing guide](https://github.com/mockito/mockito/blob/release/3.x/.github/CONTRIBUTING.md)
 - [ x] PR should be motivated, i.e. what does it fix, why, and if relevant how
 - [ x] If possible / relevant include an example in the description, that could help all readers
 - [x ] Avoid other runtime dependencies
 - [x ] Meaningful commit history ; intention is important please rebase your commit history so that each
 - [x ] The pull request follows coding style
 - [ x] Mention `Fixes #<issue number>` in the description _if relevant_
 - [ x] At least one commit should mention `Fixes #<issue number>` _if relevant_

### DIFF
--- a/src/main/java/org/mockito/internal/InOrderImpl.java
+++ b/src/main/java/org/mockito/internal/InOrderImpl.java
@@ -4,12 +4,11 @@
  */
 package org.mockito.internal;
 
-import static org.mockito.internal.exceptions.Reporter.inOrderRequiresFamiliarMock;
-
 import java.util.LinkedList;
 import java.util.List;
 
 import org.mockito.InOrder;
+import org.mockito.MockingDetails;
 import org.mockito.exceptions.base.MockitoException;
 import org.mockito.internal.verification.InOrderContextImpl;
 import org.mockito.internal.verification.InOrderWrapper;
@@ -20,6 +19,9 @@ import org.mockito.internal.verification.api.InOrderContext;
 import org.mockito.internal.verification.api.VerificationInOrderMode;
 import org.mockito.invocation.Invocation;
 import org.mockito.verification.VerificationMode;
+
+import static org.mockito.Mockito.mockingDetails;
+import static org.mockito.internal.exceptions.Reporter.*;
 
 /**
  * Allows verifying in order. This class should not be exposed, hence default access.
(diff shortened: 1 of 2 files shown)


TimvdLippe (approved):
Thanks!

mjustin:
Indeed, thanks!

PR #2267, opened 2021-04-14
Include cause when FieldReader fails to access field

### LINKED ISSUE #2266: Swallowed illegal access exceptions make it harder to debug modularity-related issues
### Issue Summary

In FieldReader and ReflectionMemberAccessor, if an IllegalAccessException is thrown, it will be suppressed.

At the very least, Mockito should include the original exception. When testing a modular application, it is common to have packages encapsulated by default.

### Stacktrace

In FieldReader, the IllegalAccessException is caught. The exception is re-thrown but the cause is lost.

https://github.com/mockito/mockito/blob/ea000f72444ed4c7e409441aedc3021951b89660/src/main/java/org/mockito/internal/util/reflection/FieldReader.java#L28-L35

Using a debugger, the IllegalAccessException can be observed: https://pastebin.com/GsnU29QV

### Problematic Code

1. Use `module-info` in the project and don't open the package under test to mockito.

2. Use the `MockitoExtension` for JUnit:

```java
public class MyClassTest {

  private final MyType mock;

  public MyClassTest(@Mock MyType mock) {
    this.mock = mock;
  }
}
```

### Environment
Mockito-Junit-Jupiter 3.9.0
Zulu16.28+11-CA (build 16+36)

A248:
Fixes #2266

Simply passes the exception cause to the thrown MockitoException

### DIFF
--- a/src/main/java/org/mockito/internal/util/reflection/FieldReader.java
+++ b/src/main/java/org/mockito/internal/util/reflection/FieldReader.java
@@ -30,7 +30,7 @@ public class FieldReader {
             return accessor.get(field, target);
         } catch (Exception e) {
             throw new MockitoException(
-                    "Cannot read state from field: " + field + ", on instance: " + target);
+                    "Cannot read state from field: " + field + ", on instance: " + target, e);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


TimvdLippe (changes requested):
Thanks for the PR! Could you add a regression test to our test suite as well? You can use the snippet that you posted in the issue and put it in the moduleTest subproject.

A248:
I can confirm that the regression test fails when the change to FieldReader is reverted.

TimvdLippe (approved):
Great unit test, thank you so much for your contribution!

PR #2277, opened 2021-04-27
Apply automated ErrorProne fixes

TimvdLippe:
This is the result of running ErrorProne on the codebase. Most notably,
it adds missing override annotations, adds private constructors for
classes that should not be subclassed, removes redundant generic type
specified and fixes handling of casts/instance checks.


TimvdLippe:
@raphw Not sure if it is worth to review, as it is a mechanical change that passes all tests. So feel free to rubberstamp.

TimvdLippe:
I am assuming no objections on this mechanical PR and will therefore merge it. Let me know if you do have any objections and am happy to resolve them.

PR #2279, opened 2021-04-28
Remove ASM dependency

iemejia:
Mockito depends on byte-buddy to make most of its bytecode fiddling but ASM is still needed for the tests. We can remove this dependency by relying on the already shaded version included on ASM, that we we won't get unaligned between both dependencies and we will remove the need for the explicit ASM library.

R: @raphw

## Checklist

 - [ ] Read the [contributing guide](https://github.com/mockito/mockito/blob/release/3.x/.github/CONTRIBUTING.md)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -86,8 +86,6 @@ dependencies {
     compileOnly libraries.junit4, libraries.hamcrest, libraries.opentest4j
     compile libraries.objenesis
 
-    testCompile libraries.asm
-
     testCompile libraries.assertj
 
     //putting 'provided' dependencies on test compile and runtime classpath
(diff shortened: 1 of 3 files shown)


raphw:
I'd go as far as to say we should rather eliminate ASM altogether and use Byte Buddy's API to create the interface. That's a one-liner more or less:

```java
new ByteBuddy()
            .makeInterface(qualifiedName)
            .make()
```

This avoids handling the technicalities.

iemejia:
Oh I only took a look at the imports :) thanks for the suggestion @raphw, PTAL again

raphw (approved):
LGTM

Issue #1930, opened 2020-05-10
Kotlin Result seems to behave weirdly when trying to be returned with Mockito


adrianhescalante:
I am facing an issue when trying to use the Kotlin Result in my project. I am trying to mock the return value as shown here:
```kotlin
`when`(useCase.getA()).thenReturn(Result.success(ObjectA("testName")))
```
The exception when trying to use that returned Result is this one: `java.lang.ClassCastException: kotlin.Result cannot be cast to ObjectA`

An interesting thing I found when debugging is that if I check the returned value of that `useCase.getA()` function, it returns an object with description `Success(Success(ObjectA(name=testName)))`. A regular Result returns this description `Success(ObjectA(name=name))`. I guess that's the reason it's failing, it seems there is like a nested Result object, so I am assuming there is something wrong with Mockito when handling the Kotlin Result class.

This is a [gist](https://gist.github.com/adrianhescalante/30449c876608fb7283e52c4d6acc5ad6)  I created to test it. There are two tests, one with the Kotlin Result and another one with a quick Result implementation (OwnResult) to check there is not something wrong I am doing myself.

I am using Mockito 3.3.3 on an Android project with Java 8.

ChristianSchwarz:
@adrianhescalante You can try mockk instead if you use kotlin

adrianhescalante:
Yes, I migrated the project to Mockk in the end. Thank you.

TimvdLippe:
For Kotlin projects, we now recommend `mockito-kotlin`: https://github.com/mockito/mockito-kotlin If you can still reproduce this issue, please open an issue over there and we can investigate further.

PR #2280, opened 2021-04-30
Support mocking of function with Kotlin inline classes

ilmirus:
Despite functions with inline class parameters being already supported,
there were not tests for it. Thus, add them.

The other part is functions, returning inline classes. There can be several
kinds of inline classes - with primitive underlying type (nullable or not-null)
and reference types (nullable or not-null). Also, a function can return either
nullable of not-null inline class. If a function returns nullable inline class,
we do not do anything special - the function's signature contains inline class.
However, if the function returns not-null inline class, its signature contains
underlying type, thus, we need to unbox the inline class (i.e. get underlying
type's value) to mock the return value.

For example, if we have the following inline class

  inline class IC(val i: Int)

and we want to mock return value of a function, returning the inline class

  fun returnsIC(): IC

like

  `when`(mock.returnsIC()).thenReturn(IC(1))

then if we do not unbox the argument of `thenReturn` function, we get error
about incompatible types.

The only issue, however, is that we cannot check, whether the function indeed
returns `IC`, unless we use kotlin.reflect. So, the following code will work.

  fun returnsInt(): Int

  `when`(mock.returnsInt()).thenReturn(IC(1))

Fixes issue #1930

## Checklist

 - [X] Read the [contributing guide](https://github.com/mockito/mockito/blob/release/3.x/.github/CONTRIBUTING.md)


src/main/java/org/mockito/internal/util/KotlinInlineClassUtil.java
```diff
@@ -0,0 +1,108 @@
+/*
+ * Copyright (c) 2021 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.internal.util;
+
+import com.sun.istack.internal.NotNull;
+import com.sun.istack.internal.Nullable;
```
TimvdLippe, line 8: These appear to be unavailable. Can we remove these and the annotations?

TimvdLippe (commented):
Implementation looks good. My only concern is that this code is in a hot path. Looking up methods on a class is potentially expensive. Is there an O(1) way of identifying a class as a boxing class. Rather than having to look for the box-impl method through all methods? E.g. could we use `getDeclaredMethod` at the least, if we know the parameter types?

ilmirus:
@TimvdLippe Actually, there is a way. Since 1.5.0 we generate `@kotlin.jvm.JvmInline` annotation for each inline class. So, I see three possibilities

1. I pushed the use of `getDeclaredMethod` with known parameter type, as you suggested.
2. We bump Kotlin version to 1.5.0-RC and check for the annotation, or
3. We wait for Kotlin 1.5.0 release and then bump the version and check for the annotation.

src/main/java/org/mockito/internal/util/KotlinInlineClassUtil.java
```diff
@@ -0,0 +1,77 @@
...
+        } catch (NoSuchMethodException e) {
+            return false;
+        }
+    }
+
+    private static Object unboxInlineClass(Object boxedValue)
+            throws InvocationTargetException, IllegalAccessException {
+        Class<?> inlineClass = boxedValue.getClass();
+        Method unboxImpl = null;
+        for (Method declaredMethod : inlineClass.getDeclaredMethods()) {
```
TimvdLippe, line 34: The same would be required here, assuming that `unbox-impl` also has a pre-known method signature?

src/main/java/org/mockito/internal/util/KotlinInlineClassUtil.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private static Class<?> getReturnTypeOfInvocation(BaseStubbing<?> stubbing) {
+        if (!(stubbing instanceof OngoingStubbingImpl<?>)) {
+            return null;
+        }
+        // TODO: Hack to get return type of method
+        OngoingStubbingImpl<?> ongoingStubbing = (OngoingStubbingImpl<?>) stubbing;
+        if (ongoingStubbing.getRegisteredInvocations().size() == 0) return null;
+
+        MatchableInvocation invocationForStubbing = ongoingStubbing.getInvocationForStubbing();
```
TimvdLippe, line 54: I think we can use `org.mockito.internal.stubbing.answers.InvocationInfo` here, in a similar fashion to how `Returns` uses it: https://github.com/mockito/mockito/blob/80374488162380d006b8ae6da915d940cbf68c7c/src/main/java/org/mockito/internal/stubbing/answers/Returns.java#L32

In that case, we can augment `Returns.answer` to use the `InvocationInfo` API and we don't need to update the `OngoingStubbingImpl` interface. E.g. we move `unboxUnderlyingValueIfNeeded` into `Returns.answer`. We can then lazily compute `value` when it is invoked, to be minimally computational impactful.

src/main/java/org/mockito/internal/util/KotlinInlineClassUtil.java
```diff
@@ -0,0 +1,77 @@
...
+
+public class KotlinInlineClassUtil {
+    // When mocking function, returning inline class, its return type is
+    // underlying type.
+    // So, `thenReturn` calls fails, because of non-compatible types.
+    public static boolean isInlineClassWithAssignableUnderlyingType(
+            Class<?> inlineClass, Class<?> underlyingType) {
+        try {
+            // All inline classes have 'box-impl' method, which accepts
+            // underlying type and returns inline class.
```
TimvdLippe, line 22: In this case, yes I think we should ideally wait for Kotlin 1.5 to ship. Of course we need to be compatible with previous Kotlin versions. Therefore, we could check if the annotation exists on the class and return that?

Since `Class.getAnnotation` takes a `Class`, the annotation would be on the classpath. Therefore, we can optimize `unboxUnderlyingValueIfNeeded` to only run when the annotation is available. E.g. we load the class based on its string name in a `static {}` block into a field. If that throws a `ClassNotFoundException`, the `Class<?>` field remains `null` and we can short-circuit `unboxUnderlyingValueIfNeeded` to bail if the class is `null`. We use a similar function for some JVM-internals in https://github.com/mockito/mockito/blob/80374488162380d006b8ae6da915d940cbf68c7c/src/main/java/org/mockito/internal/exceptions/stacktrace/StackTraceFilter.java#L23-L39

TimvdLippe (changes requested):

ilmirus:
> In this case, yes I think we should ideally wait for Kotlin 1.5 to ship.

All right then. After the release I will update the PR. For now, I'll make it a draft.

TimvdLippe:
@ilmirus I just realized we had #2275 which was supposed to update us to Kotlin 1.5, wouldn't it? Sadly that PR seems to be hitting some strange test failures, which is a bit confusing and thus stalled. But am I assuming correctly that if we would be able to merge #2275 then we would be able to unblock this PR?

ilmirus:
@TimvdLippe You are right. I was waiting for the official release (which is when the blogpost is posted). I fixed the issue with metadata version and updated the change to use @JvmInline annotation, alongside with your proposed changes.

src/main/java/org/mockito/internal/util/KotlinInlineClassUtil.java
```diff
@@ -0,0 +1,72 @@
...
+    public static boolean isInlineClassWithAssignableUnderlyingType(
+            Class<?> inlineClass, Class<?> underlyingType) {
+        try {
+            // Since 1.5.0 inline classes have @JvmInline annotation
+            if (jvmInlineAnnotation == null
+                    || !inlineClass.isAnnotationPresent(jvmInlineAnnotation)) return false;
+
+            // All inline classes have 'box-impl' method, which accepts
+            // underlying type and returns inline class.
+            inlineClass.getDeclaredMethod("box-impl", underlyingType);
```
TimvdLippe, line 37: If all inline classes have the annotation, is this still necessary? If I am not mistaken, we can update this method to the following:

```java
public static boolean isInlineClassWithAssignableUnderlyingType(
            Class<?> inlineClass, Class<?> underlyingType) {
    return jvmInlineAnnotation != null
                     && inlineClass.isAnnotationPresent(jvmInlineAnnotation);
}
```

ilmirus: Yes, otherwise, it would be possible to mock a function with different inline class:
```
inline class ICAny(val a: Any)
inline class ICInt(val i: Int)

fun foo(): ICInt

`when`(foo()).thenReturn(ICAny)
```
will try to unbox and then cast the result, failing.

subprojects/kotlinTest/src/test/kotlin/org/mockito/kotlin/InlineClassTest.kt
```diff
@@ -0,0 +1,342 @@
+package org.mockito.kotlin
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.mockito.Mockito.*
+
+@JvmInline
```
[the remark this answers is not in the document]
ilmirus: Well, long story short:
On JVM, they are required to have the annotation
On other backends (Native and JS) they do not.

So, for our use-case, we can safely assume, that they do have the annotation.

src/main/java/org/mockito/internal/util/KotlinInlineClassUtil.java
TimvdLippe: Oooh I see. Yes that makes sense now. Could we then update the comment to make that a bit clearer?

```java
// Make sure that the current inline class is also the class that is compatible with the underlying type.
// If we don't make this check, then we would potentially pass a mock of inline type A into a method
// that accepts inline type B.
```

TimvdLippe (approved):
Thank you!

fvermeulen:
Do anyone know the plan, when it will be released?
Thank you

TimvdLippe:
3.10.0 should be available on Maven Central in a bit.

lbilger:
This works for `thenReturn` now - is there a chance to fix it also for `thenAnswer`?

PR #2283, opened 2021-05-04
Improve error messages in case of missing dependency on Byte Buddy and avoid hard dependency on Byte Buddy in member accessor.

raphw:
This improves error messages in case the dependency is not set up correctly. This should rarely happen but especially with the module system gaining traction, this might trigger a few bugs that we can better explain this way.


src/main/java/org/mockito/internal/creation/bytebuddy/ByteBuddyMockMaker.java
```diff
@@ -22,44 +25,65 @@
...
 public class ByteBuddyMockMaker implements ClassCreatingMockMaker {
-    private ClassCreatingMockMaker defaultByteBuddyMockMaker = new SubclassByteBuddyMockMaker();
+    private final SubclassByteBuddyMockMaker subclassByteBuddyMockMaker;
+
+    public ByteBuddyMockMaker() {
+        try {
+            subclassByteBuddyMockMaker = new SubclassByteBuddyMockMaker();
+        } catch (NoClassDefFoundError e) {
+            if (e.getMessage() != null && e.getMessage().startsWith("net/bytebuddy/")) {
+                throw new MockitoInitializationException(
```
TimvdLippe, line 35: Can we put this exception in Reporter.java so that we dont have to include the same string message twice in both the mockmaker implementations?

TimvdLippe (approved):
Reviewing on mobile and it is quite a large diff, but I think it was mostly a refactoring to extract a separate class. Only one suggestion with regards to the duplication of the exception message.

raphw:
I do not quite understand why the CI is failing. This does not seem related to my changes at all.

TimvdLippe:
@raphw Since we moved the classes to a new name, I think we need to update these two string references to the new class name: https://github.com/mockito/mockito/blob/5a0cb27b7064588ae33536f0d997945c3246892c/src/test/java/org/mockitointegration/NoJUnitDependenciesTest.java#L47-L51

raphw:
Good catch about the initialization. I'll give it another try.

PR #2292, opened 2021-05-11
Fix stackoverflow in ArgumentsAreDifferent reporting

TimvdLippe:
When using Mockito without opentest4j, reporting an
ArgumentsAreDifferent exception would throw a StackOverflowError when
attempting to obtain the message from the exception.

The root problem was that super.toString() would call its own
getMessage(). Instead, we should obtain the message from the super, to
avoid the circular call.

### DIFF
--- a/src/main/java/org/mockito/exceptions/verification/ArgumentsAreDifferent.java
+++ b/src/main/java/org/mockito/exceptions/verification/ArgumentsAreDifferent.java
@@ -31,6 +31,6 @@ public class ArgumentsAreDifferent extends MockitoAssertionError {
 
     @Override
     public String getMessage() {
-        return removeFirstLine(super.toString());
+        return removeFirstLine(super.getMessage());
     }
 }


TimvdLippe:
Note that I tried to create a regression test for this, but sadly was unable to. To make the integration test work, we would need to exclude both opentest4j and junit4. While that would technically be possible, Gradle wouldn't be able to pick up a test that is *not* annotated with `@Test`. I also tried using JUnit5, but they have a transitive dependency on opentest4j. Any attempt to exclude the transitive dependency causes the Gradle task to crash. I have manually verified that removing https://github.com/mockito/mockito/blob/067ff2446ad08d49a634374d04dd0e6913e1259c/src/main/java/org/mockito/internal/junit/ExceptionFactory.java#L22-L31 and running `BasicVerificationTest` would reproduce the issue and that the fix is making the test pass as expected.

TimvdLippe:
Are all of our implementations are doing this: https://github.com/mockito/mockito/blob/067ff2446ad08d49a634374d04dd0e6913e1259c/src/main/java/org/mockito/exceptions/verification/opentest4j/ArgumentsAreDifferent.java#L38 and https://github.com/mockito/mockito/blob/067ff2446ad08d49a634374d04dd0e6913e1259c/src/main/java/org/mockito/exceptions/verification/junit/ArgumentsAreDifferent.java#L38

I am not sure what the effect of the removal is. Can we separate that change out, to unblock non-opentest4j users and figure out what the impact of removing `removeFirstLine` will be?

PR #2295, opened 2021-05-13
Renamed main dev branch

mockitoguy:
Renamed the main dev branch to a simpler and more intuitive name, GitHub default `main`. Before we merge this, we need to change the main dev branch name in the project settings (happy to do it when this PR is approved).


TimvdLippe (approved):
In the last year or so, we haven't had a need for making changes to past versions and I think this change makes sense. Let's do it 👍

PR #2306, opened 2021-05-25
Use the parent classloader if the context classloader is a child of it.

### LINKED ISSUE #2303: "The type is not public and its mock class is loaded by a different class loader" with a context classloader that delegates
Repro case:
```java
package com.google.clm.mockitobug;

import org.junit.Test;
import org.junit.runner.RunWith;
import org.junit.runners.JUnit4;

/**
 * My goal is to create a context classloader that is identical in every way to the existing one,
 * except that it has a different object identity and possibly some extra fields. However, even a
 * basic classloader that always delegates causes problems for mockito.
 */
@RunWith(JUnit4.class)
public final class ClassLoaderTest {

  @Test
  public void mockPackagePrivateInterface() {
    ClassLoader old = Thread.currentThread().getContextClassLoader();
    Thread.currentThread().setContextClassLoader(new ClassLoader(old) {});
    // fails
    Object unused = mock(PackagePrivate.class);
  }

  @Test
  public void mockPublicInterface() {
    ClassLoader old = Thread.currentThread().getContextClassLoader();
    Thread.currentThread().setContextClassLoader(new ClassLoader(old) {});
    // succeeds
    Object unused = mock(Public.class);
  }

  interface PackagePrivate {}

  public interface Public {}
}
```
Error:
```
Mockito cannot mock this class: interface com.google.clm.mockitobug.ClassLoaderTest$PackagePrivate.

Mockito can only mock non-private & non-final classes.
If you're not sure why you're getting this error, please report to the mailing list.


Java               : 11
JVM vendor name    : Google Inc.
JVM vendor version : 11.0.10+9-google-release-371350251
JVM version        : 11.0.10+9-google-release-371350251
JVM info           : mixed mode, sharing
OS name            : Linux
OS version         : 4.15.0-smp-912.23.0.0


Underlying exception : org.mockito.exceptions.base.MockitoException: 
Cannot create mock for interface com.google.clm.mockitobug.ClassLoaderTest$PackagePrivate

The type is not public and its mock class is loaded by a different class loader.
This can have multiple reasons:
 - You are mocking a class with additional interfaces of another class loader
 - Mockito is loaded by a different class loader than the mocked type (e.g. with OSGi)
 - The thread's context class loader is different than the mock's class loader
	at com.google.clm.mockitobug.ClassLoaderTest.mockPackagePrivateInterface(ClassLoaderTest.java:21)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:57)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:59)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:81)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:327)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:84)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:292)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:73)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:290)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:60)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:270)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:370)
	at com.google.testing.junit.runner.internal.junit4.CancellableRequestFactory$CancellableRunner.run(CancellableRequestFactory.java:108)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:115)
	at com.google.testing.junit.runner.junit4.JUnit4Runner.run(JUnit4Runner.java:104)
	at com.google.testing.junit.runner.RunnerShell$2.run(RunnerShell.java:34)
	at com.google.testing.junit.runner.GoogleTestRunner.runTestsInSuite(GoogleTestRunner.java:200)
	at com.google.testing.junit.runner.GoogleTestRunner.runTestsInSuite(GoogleTestRunner.java:184)
	at com.google.testing.junit.runner.GoogleTestRunner.main(GoogleTestRunner.java:137)
Caused by: org.mockito.exceptions.base.MockitoException: 
Cannot create mock for interface com.google.clm.mockitobug.ClassLoaderTest$PackagePrivate

The type is not public and its mock class is loaded by a different class loader.
This can have multiple reasons:
 - You are mocking a class with additional interfaces of another class loader
 - Mockito is loaded by a different class loader than the mocked type (e.g. with OSGi)
 - The thread's context class loader is different than the mock's class loader
	at net.bytebuddy.TypeCache.findOrInsert(TypeCache.java:153)
	at net.bytebuddy.TypeCache$WithInlineExpunction.findOrInsert(TypeCache.java:366)
	at net.bytebuddy.TypeCache.findOrInsert(TypeCache.java:175)
	at net.bytebuddy.TypeCache$WithInlineExpunction.findOrInsert(TypeCache.java:377)
	... 27 more

```

If mockito actually tried to use the classloader, I think it would work. Since we don't see an interesting exception deep in the stack, my guess is that mockito is doing some extra validation to avoid generating bytecode it thinks would fail.

charlesmunger:
This should only affect cases that previously would have thrown an
exception.
Fixes #2303


TimvdLippe:
There are some formatting issues that should be automatically fixed if you run `./gradlew spotlessApply` locally

charlesmunger:
Formatting issues fixed.

TimvdLippe (approved):
LGTM from my side. @raphw could you double-check as well?

src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
```diff
@@ -90,15 +90,53 @@ protected SubclassBytecodeGenerator(
...
-                        .appendMostSpecific(MockAccess.class)
-                        .build();
+                        .append(features.mockedType)
+                        .append(features.interfaces)
+                        .append(MockAccess.class, DispatcherDefaultingToRealMethod.class)
+                        .append(
+                                MockMethodInterceptor.class,
+                                MockMethodInterceptor.ForHashCode.class,
+                                MockMethodInterceptor.ForEquals.class);
+        ClassLoader contextLoader = currentThread().getContextClassLoader();
```
raphw, line 120: I'd prefer to avoid the static import here as it's the only use and we tend to not use static imports.

src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
```diff
@@ -90,15 +90,53 @@ protected SubclassBytecodeGenerator(
...
     public <T> Class<? extends T> mockClass(MockFeatures<T> features) {
-        ClassLoader classLoader =
+        MultipleParentClassLoader.Builder loaderBuilder =
                 new MultipleParentClassLoader.Builder()
-                        .appendMostSpecific(getAllTypes(features.mockedType))
-                        .appendMostSpecific(features.interfaces)
-                        .appendMostSpecific(currentThread().getContextClassLoader())
-                        .appendMostSpecific(MockAccess.class)
-                        .build();
+                        .append(features.mockedType)
```
raphw, line 113: Not quite sure, why not most specific anymore?

src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
```diff
@@ -90,15 +90,53 @@ protected SubclassBytecodeGenerator(
         handler = ModuleHandler.make(byteBuddy, loader, random);
     }
 
+    private static boolean needsSamePackageClassLoader(MockFeatures<?> features) {
+        if (!Modifier.isPublic(features.mockedType.getModifiers())
+                || !features.mockedType.isInterface()) {
+            // The mocked type is package private or is not an interface and thus may contain
+            // package
```
raphw, line 97: Move comment to one line.

src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
```diff
@@ -90,15 +90,53 @@ protected SubclassBytecodeGenerator(
...
+            // private methods.
+            return true;
+        }
+        for (Class<?> iface : features.interfaces) {
```
raphw, line 101: Interfaces might declare interfaces of their own.

src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
TimvdLippe: We already did this on line 99 in the current implementation, so we don't change this here. Can we clean these up in a follow-up?

src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
charlesmunger: That's OK, as all interface methods are public, and a public interface that extends a package-private interface doesn't require the implementation to be in the same package. All that matters for the generated type is that the directly implemented interfaces are accessible.

src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
charlesmunger: Fixed in new commit, I originally based off a different branch I think

src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
TimvdLippe: Let's figure this out in a follow-up. I think we need to reconfigure the formatter to do this.

src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
raphw: Of course, my bad. I was thinking that you iterated over the interfaces of the class but it's of course the directly implemented ones that matter.

src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
raphw: Agreed.

raphw (approved):

TimvdLippe:
Thanks @charlesmunger for working on this and the regression test!

charlesmunger:
I have since encountered two cases where this PR introduces breakage. 

1. Mocking a private abstract class while using https://github.com/google/gwtmockito. This worked before because gwtmockito was stripping final and other modifiers off of loaded classes, but now the changed loader behavior causes it to fail.
2. A custom classloader that instruments some but not all of the packages it loads, delegating others to its parent.

It turns out that the root of these problems is 
`loaderBuilder.build(MockMethodInterceptor.class.getClassLoader())`. I didn't actually intend to include that, it came from the 2.x release branch when I initially forked the repo, and was removed in 9bc9be6d0fb1809346dcd02b22ced477816e0fc1 originally. 

Sorry for the error! I'll send a PR to fix.

PR #2312, opened 2021-06-02
Undo parent for MultipleParentsClassLoader

charlesmunger:
66998ea7f81c6adf49983cea081586e19760b7a0 inadvertently undid a change from 9bc9be6d0fb1809346dcd02b22ced477816e0fc1, which breaks behavior under some classloaders that special-case mockito classes. Use of an explicit parent with `MultipleParentsClassLoader` appears to be discouraged in the documentation also.

Followup for #2303

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
@@ -136,7 +136,7 @@ class SubclassBytecodeGenerator implements BytecodeGenerator {
         if (shouldIncludeContextLoader) {
             loaderBuilder = loaderBuilder.appendMostSpecific(contextLoader);
         }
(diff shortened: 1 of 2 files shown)


TimvdLippe (changes requested):
Thanks for the quick fix! Can we please add a regression test for this scenario? Preferably without introducing any new dependencies. E.g. let's not add GWTMockito as dependency to our test suite, but instead construct a ClassLoader ourselves that would behave as if it was the ClassLoader in GWTMockito.

charlesmunger:
I spent some time thinking about it and reduced the test case to something understandable.

TimvdLippe (approved):
I slightly refactored the test code to make it clear which exact classes will be loaded by the child ClassLoader and which are not. I have verified that reverting the change in SubclassBytecodeGenerator makes the test fail again.

Thanks Charles for fixing this!

PR #2318, opened 2021-06-08
Check package-privacy of method params

charlesmunger:
This patch handles methods where an interface uses a package-private type as a param or return value - without it, the generated mock class will produce AbstractMethodError, since its methods do not actually override the methods of the interfaces.

Followup on #2303

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
@@ -98,7 +98,7 @@ class SubclassBytecodeGenerator implements BytecodeGenerator {
             // package private methods.
             return true;
         }
-        if (hasPackagePrivateParam(features.mockedType)) {
+        if (hasNonPublicTypeReference(features.mockedType)) {
             return true;
         }
 
@@ -106,14 +106,14 @@ class SubclassBytecodeGenerator implements BytecodeGenerator {
             if (!Modifier.isPublic(iface.getModifiers())) {
                 return true;
             }
-            if (hasPackagePrivateParam(iface)) {
+            if (hasNonPublicTypeReference(iface)) {
                 return true;
             }
         }
         return false;
     }
 
-    private static boolean hasPackagePrivateParam(Class<?> iface) {
+    private static boolean hasNonPublicTypeReference(Class<?> iface) {
         for (Method method : iface.getMethods()) {
             if (!Modifier.isPublic(method.getReturnType().getModifiers())) {
                 return true;
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
@@ -25,6 +25,7 @@ import static org.mockito.internal.util.StringUtil.join;
 import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
```diff
@@ -97,10 +98,31 @@ private static boolean needsSamePackageClassLoader(MockFeatures<?> features) {
...
                 return true;
             }
+            if (hasPackagePrivateParam(iface)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private static boolean hasPackagePrivateParam(Class<?> iface) {
```
TimvdLippe, line 116: Nit: `hasPackagePrivateTypeReference`, since we are also checking the return type.

src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
```diff
@@ -97,10 +98,31 @@ private static boolean needsSamePackageClassLoader(MockFeatures<?> features) {
             // package private methods.
             return true;
         }
+        if (hasPackagePrivateParam(features.mockedType)) {
```
TimvdLippe, line 101: I wonder what the performance implications of this change are. Depending on the amount of methods declared by the class or any of its parents, this could potentially lead to a considerable amount of checks. Could you do a quick check if we are regressing in terms of performance (I think we can mock a class with loads of methods and see what the characteristics are)?

If the performance impact is larger than ideal, we could consider caching the results with a `org.mockito.internal.util.concurrent.WeakConcurrentSet`

src/test/java/org/mockitousage/bugs/creation/PackagePrivateWithContextClassLoaderTest.java
```diff
@@ -55,6 +69,28 @@ public void should_be_able_to_mock_package_private_method() throws Exception {
         assertThat(publicClass.packagePrivateMethod()).isEqualTo(3);
     }
 
+    @Test
```
TimvdLippe, line 72: I love these tests.

src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
charlesmunger: I thought about that, but checking modifiers is a JVM intrinsic backed by a single word memory fetch. If I was recursively checking the class tree you'd be right to worry about pathological interface DAGs causing exponential blowup, but that's not happening here.

In addition, we short-circuit on per-type checks before we check methods, and we are only checking public methods. So this path only gets hit if the mocked type is a public interface, and it only checks public methods. By definition the methods are unique for a single class, so there's no benefit to memoizing at that level. And the per-param modifier checks are much cheaper than doing a hashmap store/lookup.

And all in all, every single method and param we process this way has to get parsed and generated in the actual mock class anyway, which is inherently more expensive. So I'm not concerned about performance here - if you mock an interface with tens of thousands of methods, we have to generate a class with tens of thousands of methods, verify it, and load it, which will always be more expensive.

src/test/java/org/mockitousage/bugs/creation/PackagePrivateWithContextClassLoaderTest.java
charlesmunger: :-)

src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
charlesmunger: Renamed to hasNonPublicTypeReference

src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
TimvdLippe: Thanks for the explanation! Yes that makes sense to me 😄

charlesmunger:
I have run this through google's internal tests and it doesn't break anything.

TimvdLippe (approved):
Thanks!

PR #2320, opened 2021-06-09
Fixes mockito#2311

saurabh7248:
Prints fully qualified class name when the simple names of arguments match.

## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/matchers/Equals.java
+++ b/src/main/java/org/mockito/internal/matchers/Equals.java
@@ -56,7 +56,17 @@ public class Equals implements ArgumentMatcher<Object>, ContainsExtraTypeInfo, S
     }
 
     @Override
+    public String toStringWithFullName() {
+        return "(" + wanted.getClass().getCanonicalName() + ") " + describe(wanted);
+    }
+
+    @Override
     public boolean typeMatches(Object target) {
         return wanted != null && target != null && target.getClass() == wanted.getClass();
+    }
+
+    @Override
+    public boolean sameName(Object target) {
+        return wanted != null && target != null && target.getClass().getSimpleName().equals(wanted.getClass().getSimpleName());
     }
 }
--- a/src/main/java/org/mockito/internal/matchers/Equals.java
+++ b/src/main/java/org/mockito/internal/matchers/Equals.java
@@ -31,7 +31,8 @@ public class Equals implements ArgumentMatcher<Object>, ContainsExtraTypeInfo, S
         return ValuePrinter.print(object);
     }
 
-    protected final Object getWanted() {
+    @Override
+    public final Object getWanted() {
         return wanted;
     }
 
@@ -51,18 +52,13 @@ public class Equals implements ArgumentMatcher<Object>, ContainsExtraTypeInfo, S
     }
 
     @Override
-    public String toStringWithType(boolean useFullyQualifiedClassName) {
-        return "(" + (useFullyQualifiedClassName ? wanted.getClass().getCanonicalName() : wanted.getClass().getSimpleName()) + ") " + describe(wanted);
+    public String toStringWithType(String className) {
+        return "(" + className + ") " + describe(wanted);
     }
 
     @Override
     public boolean typeMatches(Object target) {
         return wanted != null && target != null && target.getClass() == wanted.getClass();
-    }
-
-    @Override
-    public Class getWantedClass() {
-        return wanted.getClass();
     }
 
 }
--- a/src/main/java/org/mockito/internal/matchers/Equals.java
+++ b/src/main/java/org/mockito/internal/matchers/Equals.java
@@ -60,5 +60,4 @@ public class Equals implements ArgumentMatcher<Object>, ContainsExtraTypeInfo, S
     public boolean typeMatches(Object target) {
         return wanted != null && target != null && target.getClass() == wanted.getClass();
     }
-
 }
--- a/src/main/java/org/mockito/internal/matchers/Equals.java
+++ b/src/main/java/org/mockito/internal/matchers/Equals.java
@@ -51,13 +51,8 @@ public class Equals implements ArgumentMatcher<Object>, ContainsExtraTypeInfo, S
     }
 
     @Override
-    public String toStringWithType() {
-        return "(" + wanted.getClass().getSimpleName() + ") " + describe(wanted);
-    }
-
-    @Override
-    public String toStringWithFullName() {
-        return "(" + wanted.getClass().getCanonicalName() + ") " + describe(wanted);
+    public String toStringWithType(boolean useFullyQualifiedClassName) {
+        return "(" + (useFullyQualifiedClassName ? wanted.getClass().getCanonicalName() : wanted.getClass().getSimpleName()) + ") " + describe(wanted);
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


src/main/java/org/mockito/internal/matchers/Equals.java
```diff
@@ -55,8 +55,18 @@ public String toStringWithType() {
         return "(" + wanted.getClass().getSimpleName() + ") " + describe(wanted);
     }
 
+    @Override
+    public String toStringWithFullName() {
```
TimvdLippe, line 59: I am not the biggest fan of adding another method here. Especially as it is very similar to the `toStringWithType()`, but the only difference is the name retrieval.

What if instead we get the wanted object and perform the computation in the `MatchersPrinter`? E.g. avoid introductions of new methods and pass in the desired name into `toStringWithType(String className)`

src/main/java/org/mockito/internal/verification/argumentmatching/ArgumentMatchingTool.java
```diff
@@ -48,4 +48,28 @@ private static boolean safelyMatches(ArgumentMatcher m, Object arg) {
     private static boolean toStringEquals(ArgumentMatcher m, Object arg) {
         return m.toString().equals(String.valueOf(arg));
     }
+
+    /**
+     * Suspiciously not matching arguments are those that don't match, and the classes have same simple name.
+     */
+    public static Integer[] getNotMatchingArgsWithSameNameIndexes(
```
TimvdLippe, line 55: Rather than returning Integers, can we instead return all names that have duplicates? Then we can do the lookup on the name. E.g. return a `Set<String>` here, where the set contains all names that occur at least twice in `arguments`

TimvdLippe (changes requested):

src/main/java/org/mockito/internal/matchers/Equals.java
```diff
@@ -51,12 +51,18 @@ public int hashCode() {
...
+        return "(" + (useFullyQualifiedClassName ? wanted.getClass().getCanonicalName() : wanted.getClass().getSimpleName()) + ") " + describe(wanted);
     }
 
     @Override
     public boolean typeMatches(Object target) {
         return wanted != null && target != null && target.getClass() == wanted.getClass();
     }
+
+    @Override
+    public Class getWantedClass() {
```
TimvdLippe, line 64: We already define `getWanted` above, but it is `protected`. I think we should make that `public` instead and then remove this function. WDYT?

src/main/java/org/mockito/internal/matchers/Equals.java
```diff
@@ -51,12 +51,18 @@ public int hashCode() {
     }
 
     @Override
-    public String toStringWithType() {
-        return "(" + wanted.getClass().getSimpleName() + ") " + describe(wanted);
+    public String toStringWithType(boolean useFullyQualifiedClassName) {
```
TimvdLippe, line 54: Rather than a `boolean`, can we pass in the `String` here?

src/main/java/org/mockito/internal/matchers/text/MatchersPrinter.java
```diff
@@ -31,7 +31,10 @@ public String getArgumentsBlock(List<ArgumentMatcher> matchers, PrintSettings pr
         int i = 0;
         for (final ArgumentMatcher matcher : matchers) {
             if (matcher instanceof ContainsExtraTypeInfo && printSettings.extraTypeInfoFor(i)) {
-                out.add(new FormattedText(((ContainsExtraTypeInfo) matcher).toStringWithType()));
+                out.add(new FormattedText(((ContainsExtraTypeInfo) matcher).toStringWithType(false)));
+            } else if(matcher instanceof ContainsExtraTypeInfo
+                      && printSettings.fullyQualifiedNameFor(((ContainsExtraTypeInfo) matcher).getWantedClass().getSimpleName())){
+                out.add(new FormattedText(((ContainsExtraTypeInfo) matcher).toStringWithType(true)));
```
TimvdLippe, line 37: I think if we refactor it a bit, it is a bit more readable:

```java
if (matcher instanceof ContainsExtraTypeInfo) {
    ContainsExtraTypeInfo typeInfoMatcher = (ContainsExtraTypeInfo) matcher;
    Class<?> wantedClass = typeInfoMatcher.getWanted().getClass();
    String simpleNameOfArgument = wantedClass.getSimpleName();

    if (printSettings.extraTypeInfoFor(i)) {
        out.add(new FormattedText(typeInfoMatcher.toStringWithType(simpleNameOfArgument)));
    } else if (printSettings.fullyQualifiedNameFor(simpleNameOfArgument)) {
        out.add(new FormattedText(typeInfoMatcher.toStringWithType(wantedClass.getCanonicalName())));
    }
}
```

TimvdLippe (commented):
We are almost there!

Lastly, can we add another test to https://github.com/mockito/mockito/blob/6f9108b833323345bd926f10914224d3e0a11b80/src/test/java/org/mockitousage/verification/DescriptiveMessagesWhenVerificationFailsTest.java which verifies that if a method takes two of the same arguments the output is correct? I think if you take `should_print_method_when_matcher_used` as a starting point, it should be relatively straightforward to add more tests with the following cases:
1. 2 arguments with the same simple type name
2. 2 arguments with different simple type names
3. 3 arguments of which the first and last have the same simple type name, but the middle one has a different one.

src/main/java/org/mockito/internal/matchers/text/MatchersPrinter.java
saurabh7248: Have refactored this, but also had to add the else to this block of instance of as well, because for the case when matcher is of type ContainsExtraTypeInfo but extra info and fully qualified name is also not needed. So if that else block is not included it would take them as empty string.

src/main/java/org/mockito/internal/matchers/Equals.java
saurabh7248: Had kept boolean thinking, we either need fully name of simple name, but keeping it as String gives flexibility to the caller to choose the name.

src/main/java/org/mockito/internal/matchers/Equals.java
saurabh7248: Had thought the same but was not sure whether to make it public . Also have added 3 test cases. I tried finding classes with same name through java library, but if two classes had same name, they didn't have a toString implemented. 
So i have made a dummy class named Date in IMethods. 
Also let me know if the method names are too long.

src/test/java/org/mockitousage/verification/DescriptiveMessagesWhenVerificationFailsTest.java
```diff
@@ -386,6 +388,118 @@ public void should_never_break_method_string_when_no_args_in_method() throws Exc
         }
     }
 
+    @Test
+    public void should_print_fully_qualified_name_when_arguments_classes_have_same_simple_name() {
```
TimvdLippe, line 392: I am loving these tests, thanks!

TimvdLippe (approved):
Nice work and nice tests! Thanks for getting through the review comments and applying them 😄

TimvdLippe:
Oh, you need to run `./gradlew :spotlessApply` locally to fix the formatter issues. Do you mind running that?

saurabh7248:
Sorry, missed to run the gradle check before pushing.

TimvdLippe:
100% test coverage on the diff. Nice job! Merging this 🎉

saurabh7248:
Nice!!!
Thanks @TimvdLippe for the reviews.

PR #2323, opened 2021-06-11
Introduce a BOM for Mockito's artifacts (closes #2321)

### LINKED ISSUE #2321: Provide a bill of materials (BOM)
I figured that the number of Mockito artifacts grows, so it might be useful to provide a BOM (bill of material) that lists all versions of all artifacts in a central place.

Such a thing could then be consumed in a [Gradle multi project build](https://docs.gradle.org/current/userguide/multi_project_builds.html) through

```kotlin
// my.platform.project:test
plugins {
    id("java-platform")
}

dependencies {
    api(platform("org.mockito:mockito-bom:$someVersion"))
}
```

and this could be referenced in individual module files like so:

```kotlin
dependencies {
    testImplementation(platform("my.platform.project:test"))

    testImplementation("org.mockito:mockito-core")
    testImplementation("org.mockito:mockito-junit-jupiter")
}
```
Creating and publishing such a BOM should be fairly easy. Thanks!

realdadfish:
See discussion in #2321

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -16,7 +16,8 @@ include("deprecatedPluginsTest",
     "memory-test",
     "errorprone",
     "junitJupiterParallelTest",
-    "osgi-test")
+    "osgi-test",
+    "bom")
 
 rootProject.name = "mockito"
 
--- a/gradle/java-library.gradle
+++ b/gradle/java-library.gradle
@@ -1,7 +1,5 @@
 apply plugin: "java"
 
-group = 'org.mockito'
-
 if (!archivesBaseName.startsWith("mockito-")) {
     archivesBaseName = "mockito-" + project.name
 }
--- a/gradle/java-publication.gradle
+++ b/gradle/java-publication.gradle
@@ -1,29 +1,34 @@
-//Sources/javadoc artifacts required by Maven module publications
-def licenseSpec = copySpec {
-    from project.rootDir
-    include "LICENSE"
+group = 'org.mockito'
+
+plugins.withId("java") {
+    //Sources/javadoc artifacts required by Maven module publications
+    def licenseSpec = copySpec {
+        from project.rootDir
+        include "LICENSE"
+    }
+
+    task sourcesJar(type: Jar, dependsOn: classes) {
+        classifier = 'sources'
+        from sourceSets.main.allSource
+        with licenseSpec
+    }
+
+    task javadocJar(type: Jar, dependsOn: javadoc) {
+        classifier = 'javadoc'
+        from tasks.javadoc
+        with licenseSpec
+    }
+
+    artifacts {
+        archives sourcesJar
+        archives javadocJar
+    }
+
+    jar {
+        with licenseSpec
+    }
 }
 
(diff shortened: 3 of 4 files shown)


settings.gradle.kts
```diff
@@ -16,7 +16,8 @@ include("deprecatedPluginsTest",
     "memory-test",
     "errorprone",
     "junitJupiterParallelTest",
-    "osgi-test")
+    "osgi-test",
+    "bom")
```
TimvdLippe, line 20: I would have expected some files to be added in `subprojects/bom/`, but I don't see these in this PR. Are they missing or is this working automagically?

TimvdLippe (changes requested):
The BOM in https://github.com/mockito/mockito/issues/2321#issuecomment-859573476 lists all versions with `SNAPSHOT`. Instead, I think we should list the explicit versions. We currently do that with Shipkit on tags: https://github.com/shipkit/shipkit-auto-version#usage-based-on-annotated-tag

Can we make sure that the explicit versions work with our GitHub actions integration? I am not actually sure how that would look like, but I think you should be able to run `./gradlew publishToMavenLocal -Pversion=3.12.0` and see what the result is.

settings.gradle.kts
realdadfish: My bad, had to reset my previous commit because of a temporary change in some file and forgot to add the new file again. TGIF...

realdadfish: This works automagically, you only need 

> The BOM in [#2321 (comment)](https://github.com/mockito/mockito/issues/2321#issuecomment-859573476) lists all versions with `SNAPSHOT`. Instead, I think we should list the explicit versions. We currently do that with Shipkit on tags: https://github.com/shipkit/shipkit-auto-version#usage-based-on-annotated-tag
> 
> Can we make sure that the explicit versions work with our GitHub actions integration? I am not actually sure how that would look like, but I think you should be able to run `./gradlew publishToMavenLocal -Pversion=3.12.0` and see what the result is.

Yes, this is exactly what happens. The other modules are versioned by the same mechanism, so the BOM retrieves their fixed version set by shipkit through Gradle. Just verified again that this works locally. Try it yourself!

realdadfish: Sorry, not my day, this was supposed to go as reply to your other comment, but the window was still open. Really TGIF

TimvdLippe:
@mockitoguy This looks fine to me, but do we need to create this new artifact in Maven Central or does that work automatically?

realdadfish:
Hey @TimvdLippe, what's the status of this? :-)

TimvdLippe:
This is still pending a review of @mockitoguy per https://github.com/mockito/mockito/pull/2323#issuecomment-859736405

TimvdLippe:
I am inclined to merge this PR as-is and see if we run into any issues when we release to Maven Central. We might have to revert the PR then or do a follow-up to get things to work properly. Therefore, merging it 😄

TimvdLippe (approved):
Thanks! Let's see whether this breaks on our next Maven Central release.

Issue #2340, opened 2021-06-22
Mockito 3.11.2 (?) fails with "can't load this type of class file"


robpridham-bbc:
**Broken version**: 3.11.2?
**Working versions:** 3.10.0

**Description**

All was working OK with 3.8.0 and we upgraded to 3.11.2 but this presented the problem below. I have sequentially downgraded and it seems 3.10.0 doesn't present me with any problems.

I have not been able to confidently isolate it to this version because I have also experienced #2316 and #2327 in the interim versions.

Our context is Android and a `com.android.test` instrumented test module. We are using `org.mockito:mockito-core`, plus `org.mockito:mockito-android` with the same version, and `org.mockito.kotlin:mockito-kotlin:3.2.0`.

```
2021-06-22 21:35:57.029 9913-9913/ourapp.package.debug E/MonitoringInstr: Exception encountered by: ourapp.package.testing.integrations.TestApplication@b4bbfaf. Dumping thread state to outputs and pining for the fjords.
    org.mockito.exceptions.base.MockitoException: 
    Failed to release mocks
    
    This should not happen unless you are using a third-part mock maker
        at ourapp.package.testing.integrations.TestApplication.onCreate(OurAppTestRunner.kt:44)
        at android.app.Instrumentation.callApplicationOnCreate(Instrumentation.java:1192)
        at androidx.test.runner.MonitoringInstrumentation.callApplicationOnCreate(MonitoringInstrumentation.java:428)
        at android.app.ActivityThread.handleMakeApplication(ActivityThread.java:7507)
        at android.app.ActivityThread.handleBindApplication(ActivityThread.java:7446)
        at android.app.ActivityThread.access$1500(ActivityThread.java:301)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2148)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:246)
        at android.app.ActivityThread.main(ActivityThread.java:8512)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:602)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1130)
     Caused by: java.lang.IllegalStateException: Could not initialize plugin: interface org.mockito.plugins.MockMaker (alternate: null)
[...]

Issue #2327, opened 2021-06-15
Mockito Android 3.11.1 fails to mock class with lambda with generic


fondesa:
**Broken version**: 3.11.1
**Working versions**: <= 3.10.0

#### Description
This Kotlin class can't be mocked with Mockito Android 3.11.1, but it can be mocked with previous versions:

```kotlin
open class Foo {
  open fun <T : Any> bar(action: () -> T): T = action()
}
```

Instead, this one can be mocked successfully even with Mockito Android 3.11.1:
```kotlin
open class Foo {
  open fun bar(action: () -> Any): Any = action()
}
```

#### Sample
```kotlin
class DummyTest {
  @Test
  fun test() {
    Mockito.mock(Foo::class.java)
  }
}
```

#### Stacktrace
```
Underlying exception : java.lang.IllegalStateException: Could not invoke proxy: Unexpected error: Cannot assign self type: interface java.lang.reflect.TypeVariable on public abstract java.lang.reflect.AnnotatedElement[] net.bytebuddy.description.type.TypeDescription$Generic$AnnotationReader$ForTypeVariableBoundType$OfFormalTypeVariable$FormalTypeVariable.getAnnotatedBounds(java.lang.reflect.AnnotatedElement)
at com.example.DummyTest.test(DummyTest.kt:29)
at java.lang.reflect.Method.invoke(Native Method)
at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
[...]

Issue #2316, opened 2021-06-04
Mockito 3.11.0 seems to fail in Android instrumentation tests


tmtrademarked:
After upgrading from 3.10.0 to 3.11.0, our instrumentation tests started failing with a strange error about Mockito plugins:
```
java.lang.IllegalStateException: Could not initialize plugin: interface org.mockito.plugins.MockMaker (alternate: null)
	at org.mockito.internal.configuration.plugins.PluginLoader$1.invoke(PluginLoader.java:84)
	at java.lang.reflect.Proxy.invoke(Proxy.java:1006)
	at $Proxy12.isTypeMockable(Unknown Source)
	at org.mockito.internal.util.MockUtil.typeMockabilityOf(MockUtil.java:33)
	at org.mockito.internal.util.MockCreationValidator.validateType(MockCreationValidator.java:22)
	at org.mockito.internal.creation.MockSettingsImpl.validatedSettings(MockSettingsImpl.java:250)
	at org.mockito.internal.creation.MockSettingsImpl.build(MockSettingsImpl.java:232)
	at org.mockito.internal.MockitoCore.mock(MockitoCore.java:83)
	at org.mockito.Mockito.mock(Mockito.java:1954)
	at org.mockito.Mockito.mock(Mockito.java:1865)
	at com.blueapron.service.mock.ServiceTestApplication$1.mock(ServiceTestApplication.java:29)
	at com.blueapron.testcommon.StaticMock.provide(StaticMock.java:42)
	at com.blueapron.service.mock.InstrumentAppModule.provideClock_impl(InstrumentAppModule.java:55)
	at com.blueapron.service.dagger.ApplicationModule.provideClock(ApplicationModule.java:230)
	at com.blueapron.service.dagger.ApplicationModule_ProvideClockFactory.provideClock(ApplicationModule_ProvideClockFactory.java:35)
	at com.blueapron.service.dagger.ApplicationModule_ProvideClockFactory.get(ApplicationModule_ProvideClockFactory.java:27)
	at com.blueapron.service.dagger.ApplicationModule_ProvideClockFactory.get(ApplicationModule_ProvideClockFactory.java:9)
	at dagger.internal.DoubleCheck.get(DoubleCheck.java:47)
	at com.blueapron.service.mock.DaggerInstrumentApplicationComponent.injectTimeUtils(DaggerInstrumentApplicationComponent.java:653)
[...]

PR #2360, opened 2021-07-16
Basic Android instrumented and unit tests (closes #2341)

### LINKED ISSUE #2341: Introduce regression test suite for Android ecosystem
Per https://github.com/mockito/mockito/issues/2340#issuecomment-866347287

We should work on a test suite to prevent issues such as #2316 #2327 and #2340. Ultimately this should then also catch issues that arise in ByteBuddy with the bytecode magic that Mockito implicitly depends on.

CC @robpridham-bbc

robpridham-bbc:
See https://github.com/mockito/mockito/issues/2341

Most of this PR is project overhead associated with a new Android project.

The primary benefit is providing basic instrumented tests which replicate failures seen in #2316, #2327 and #2340. Currently it only performs the most rudimentary establishment of some mocks; however, this is enough to produce these failures on the problematic versions.

I have included basic JUnit 4/5 non-instrumented tests which is perhaps unnecessary, given that more extensive tests exist in the non-Android subprojects, and there is nothing Android-specific about them. However I thought it might be valuable to have them exist within the context of the Android toolchain to provide basic assurance.

There is a README included with this, which may need further development to inform non-Android developers.


TimvdLippe:
@robpridham-bbc Thank you so much for getting this of the ground! I refactored the PR slightly to make sure it ingrates with `./gradlew build`. Now, if you run the build locally, it should run the unit tests and use the locally built version of Mockito.

Unfortunately, I wasn't able to figure out how to run the instrumentation themselves from the command line. As I understand the, the instrumentation tests were the actual tests that were failing. I tried downgrading ByteBuddy to check that the tests are failing as expected, but none of them were.

Based on https://spin.atomicobject.com/2016/03/10/android-test-script/ I understand that we would need to have a script that can boot up an emulator on CI and then run the corresponding scripts? At least that's what it told me when I ran `./gradlew :androidTest:connectedCheck` locally.

I am fine if we land this as-is if we need more infrastructure changes to support these emulators. We probably want to add a separate GitHub action task to run these. However, it would be good to first verify that approach will work before we land this PR. WDYT?

TimvdLippe:
Ah. Seems like the memory increase worked and we got a green build 🎉

robpridham-bbc:
> @robpridham-bbc Thank you so much for getting this of the ground! I refactored the PR slightly to make sure it ingrates with `./gradlew build`. Now, if you run the build locally, it should run the unit tests and use the locally built version of Mockito.
> 
> Unfortunately, I wasn't able to figure out how to run the instrumentation themselves from the command line. As I understand the, the instrumentation tests were the actual tests that were failing. I tried downgrading ByteBuddy to check that the tests are failing as expected, but none of them were.
> 
> Based on https://spin.atomicobject.com/2016/03/10/android-test-script/ I understand that we would need to have a script that can boot up an emulator on CI and then run the corresponding scripts? At least that's what it told me when I ran `./gradlew :androidTest:connectedCheck` locally.
> 
> I am fine if we land this as-is if we need more infrastructure changes to support these emulators. We probably want to add a separate GitHub action task to run these. However, it would be good to first verify that approach will work before we land this PR. WDYT?

Subject to confirmation, `./gradlew :androidTest:connectedCheck` sounds right. I'm a little bit out of touch with the scripting requirements because (a) our normal dev work is IDE-driven where it handles this and (b) we do our device/emulator automation testing on Firebase Test Lab, which involves a Gradle plugin (fladle) uploading the output APK there and invoking the tests.

Firebase might actually be a good option for this project. For us in our day jobs it costs money, but a very small amount as there's a free quota of daily execution minutes that we rarely exceed.

TimvdLippe:
Hm, I would prefer not to add any dependencies on external services such as Firebase. I will look into making a simple bash script that can do not. Not sure if that will work, but based on the blog post that maybe is feasible?

robpridham-bbc:
The bash script article looks OK although it seems to assume an already created emulator image. You can do that at the command line too but it's a bit more involved.

By the way, I feel I should probably point out that the migration into the top level project probably works fine from a Gradle point of view but it does mean it's not maintainable as a normal Android project any more, e.g. in Android Studio or as far as I'm aware IntelliJ. I'm not averse to what you've done; this is a balance between convenient integration into the main project (rather than being a weird Android-specific outlier) and making authoring any future extensions more difficult.

However it might also have unwanted consequences. I _think_ this is why you were seeing no exception messages for Lint. Similarly, in its current form it doesn't pass for me, and its output is totally opaque, although I haven't dug into why.

```
Starting 6 tests on SM-G998B - 11

org.mockitousage.androidtest.BasicInstrumentedTests > mockAndUseBasicClassUsingAnnotatedMock[SM-G998B - 11] FAILED .21.jar
        java.lang.Exception: java

```

This is not normal - when I go back to my original commit and test against 3.11.2, I get proper output for the failure:

```
> Task :app:connectedDebugAndroidTest
Starting 6 tests on SM-G998B - 11

org.mockitousage.androidtest.BasicInstrumentedTests > mockAndUseBasicClassWithVerify[SM-G998B - 11] FAILED 
        at org.mockito.internal.configuration.plugins.PluginLoader$1.invoke(PluginLoader.java:84)

org.mockitousage.androidtest.BasicInstrumentedTests > mockAndUseBasicClassUsingLocalMock[SM-G998B - 11] FAILED 
        at org.mockito.internal.configuration.plugins.PluginLoader$1.invoke(PluginLoader.java:84)

org.mockitousage.androidtest.BasicInstrumentedTests > mockAndUseBasicInterfaceUsingAnnotatedMock[SM-G998B - 11] FAILED 
        at org.mockito.internal.configuration.plugins.PluginLoader$1.invoke(PluginLoader.java:84)

org.mockitousage.androidtest.BasicInstrumentedTests > mockAndUseBasicClassUsingAnnotatedMock[SM-G998B - 11] FAILED 
        at org.mockito.internal.configuration.plugins.PluginLoader$1.invoke(PluginLoader.java:84)

org.mockitousage.androidtest.BasicInstrumentedTests > mockAndUseBasicInterfaceAndVerify[SM-G998B - 11] FAILED 
        at org.mockito.internal.configuration.plugins.PluginLoader$1.invoke(PluginLoader.java:84)

org.mockitousage.androidtest.BasicInstrumentedTests > mockAndUseBasicInterfaceUsingLocalMock[SM-G998B - 11] FAILED 
        at org.mockito.internal.configuration.plugins.PluginLoader$1.invoke(PluginLoader.java:84)

> Task :app:connectedDebugAndroidTest FAILED

```

TimvdLippe:
I wasn't able to work on this in the past two weeks. That said, I am inclined to merge this PR as-is and keep iterating on the bash script in a follow-up. Preferably we get at least *some* coverage for Android, although the real value will be in the instrumentation tests that will require an emulator to run.

PR #2369, opened 2021-07-30
Fixes #2331

### LINKED ISSUE #2331: Spy doesn't forward hashcode/equals to actual object
### Mockito 3.11.1

```
public class MockitoTest {

    private static class ValueHolder {
        private final String value;

        public ValueHolder( String value ) {
            this.value = value;
        }

        public String getValue() {
            return value;
        }

        @Override public boolean equals( Object o ) {
            if (this == o) return true;
            if (!(o instanceof ValueHolder)) return false;
            ValueHolder that = (ValueHolder) o;
            return getValue().equals( that.getValue() );
        }

        @Override
        public int hashCode() {
            return Objects.hash( getValue() );
        }
    }

    @Test
    public void test() {
        ValueHolder value = new ValueHolder( "test" );
        ValueHolder spy = Mockito.spy( value );
        assert value.equals( spy );
        assert spy.equals( value ); //fails
        assert value.hashCode() == spy.hashCode(); //fails
    }
}
```

saurabh7248:
Has conditionally on basis of whether default answer is pass to actual methods made hash code and equals pass to actual methods.

## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/invocation/InvocationMatcher.java
+++ b/src/main/java/org/mockito/internal/invocation/InvocationMatcher.java
@@ -77,7 +77,7 @@ public class InvocationMatcher implements MatchableInvocation, DescribedInvocati
 
     @Override
     public boolean matches(Invocation candidate) {
-        return invocation.getMock().equals(candidate.getMock())
+        return invocation.getMock() == candidate.getMock()
                 && hasSameMethod(candidate)
                 && argumentsMatch(candidate);
     }
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/InlineDelegateByteBuddyMockMaker.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/InlineDelegateByteBuddyMockMaker.java
@@ -393,7 +393,8 @@ class InlineDelegateByteBuddyMockMaker
                             settings.getTypeToMock(),
                             settings.getExtraInterfaces(),
                             settings.getSerializableMode(),
-                            settings.isStripAnnotations()));
+                            settings.isStripAnnotations(),
+                            settings.getDefaultAnswer()));
         } catch (Exception bytecodeGenerationFailed) {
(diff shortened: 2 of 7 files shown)


src/main/java/org/mockito/internal/invocation/InvocationMatcher.java
```diff
@@ -77,7 +77,7 @@ public String toString() {
 
     @Override
     public boolean matches(Invocation candidate) {
-        return invocation.getMock().equals(candidate.getMock())
+        return invocation.getMock() == candidate.getMock()
```
TimvdLippe, line 80: I am not sure if I follow why this line was changed. Could you elaborate as to why?

TimvdLippe (approved):
It's not ideal to have another branch, but I think it is the best way forward. Only 1 question, but the rest LGTM

src/main/java/org/mockito/internal/invocation/InvocationMatcher.java
saurabh7248: Consider the test PartialMockingWithSpiesTest.shouldAllowStubbingOfMethodsThatDelegateToOtherMethods.
Now the call to getName is handled by MockMethodInterceptor.DispatcherDefaultingToRealMethod.
interceptSuperCallable here the interceptor will not be null and it will go to interceptor.doIntercept. This finally leads to InvocationMatcher.matches

Now with changes in this pull request the call to equals is again delegated to  MockMethodInterceptor.DispatcherDefaultingToRealMethod and it will come at same place. And this way it leads to stackoverflow. Earlier equals of mock was this==that. So i just replced equals with ==.

TimvdLippe: Awesome, thanks for the explanation! That makes sense to me 😄

saurabh7248: Haha Thanks. Had missed to put links to the classes.

PR #2379, opened 2021-08-02
Applied @CheckReturnValue to some classes

### LINKED ISSUE #2363: Apply @CheckReturnValue more aggressively
Mockito has a local copy of [`@CheckReturnValue`](https://github.com/mockito/mockito/blob/main/src/main/java/org/mockito/CheckReturnValue.java) (presumably to avoid a dependency on [errorprone](https://github.com/google/error-prone), but it's only used on a few APIs.  There's a _lot_ of additional APIs that it could be applied to, an in fact the best solution is often to annotate entire classes (and then opt-out specific APIs) or entire packages (via `package-info.java`).

Top APIs to add `@CheckReturnValue` to:
* `org.mockito.ArgumentMatchers`
* `org.mockito.ArgumentCaptor`
* `org.mockito.Mockito`
* `org.mockito.stubbing.Stubber`

EugeneLesnov:
Should close #2363 

As I understood, we can remove @CheckReturnValue from methods in `org.mockito.Mockito` after adding annotation on class

### DIFF
--- a/src/main/java/org/mockito/ArgumentCaptor.java
+++ b/src/main/java/org/mockito/ArgumentCaptor.java
@@ -59,6 +59,7 @@ import org.mockito.internal.matchers.CapturingMatcher;
  * @see Captor
  * @since 1.8.0
  */
+@CheckReturnValue
 public class ArgumentCaptor<T> {
 
     private final CapturingMatcher<T> capturingMatcher = new CapturingMatcher<T>();
@@ -79,7 +80,7 @@ public class ArgumentCaptor<T> {
      * @return null or default values
      */
     public T capture() {
-        Mockito.argThat(capturingMatcher);
+        T ignored = Mockito.argThat(capturingMatcher);
         return defaultValue(clazz);
     }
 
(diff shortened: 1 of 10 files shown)


TimvdLippe:
It seems like we are going to need to update our own test suite to accommodate for these additional errors. Other than that, yes this solution LGTM!

EugeneLesnov:
@TimvdLippe How do you think, how is it better to process errors in tests..Do I have to add local variables for situations like:

```
error: [CheckReturnValue] Ignored return value of method that is annotated with @CheckReturnValue
        anyObject();
```

Or should I add `errorprone` dependency to use `@CanIgnoreReturnValue` annotation?

TimvdLippe:
We can assign it to an ignored local variable, like I did in https://github.com/mockito/mockito/pull/2365/commits/182fd7f8d86c21ae6e3ee0ca79088fca13d4565b

TimvdLippe (approved):
Yes this is great, thanks!

PR #2397, opened 2021-08-22
Add a limited mock maker that is based only on the java.lang.reflect.Proxy utility

raphw:
This could be an interesting addition for puristic projects that want to use Mockito's API but not rely on byte code generation which naturally requires an update of Byte Buddy for any new JVM version. The advantage of this mock maker is that it will work with any future release for this limited scope.

### DIFF
--- a/src/main/java/org/mockito/internal/creation/proxy/ProxyMockMaker.java
+++ b/src/main/java/org/mockito/internal/creation/proxy/ProxyMockMaker.java
@@ -1,0 +1,172 @@
+/*
+ * Copyright (c) 2021 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.internal.creation.proxy;
+
+import org.mockito.exceptions.base.MockitoException;
+import org.mockito.internal.debugging.LocationImpl;
+import org.mockito.internal.invocation.RealMethod;
+import org.mockito.internal.invocation.SerializableMethod;
+import org.mockito.internal.util.Platform;
+import org.mockito.invocation.MockHandler;
+import org.mockito.mock.MockCreationSettings;
+import org.mockito.plugins.MockMaker;
+
+import java.lang.reflect.*;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static org.mockito.internal.invocation.DefaultInvocationFactory.createInvocation;
+import static org.mockito.internal.util.StringUtil.join;
+
+/**
+ * A mock maker that is using the {@link Proxy} utility and is therefore only capable of mocking interfaces but
+ * does not rely on manual byte code generation but only uses official and public Java API.
+ */
+public class ProxyMockMaker implements MockMaker {
+
+    private final Method invokeDefault;
+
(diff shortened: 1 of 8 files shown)


src/test/java/org/mockito/internal/creation/proxy/ProxyMockMakerTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.junit.Test;
+import org.mockito.internal.creation.AbstractMockMakerTest;
+
+import java.io.Serializable;
+import java.lang.reflect.Proxy;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ProxyMockMakerTest
```
TimvdLippe, line 15: Can we maybe also have some tests that verify that attempting to mock non-interfaces correctly fails with a descriptive error message?

TimvdLippe: Based on https://app.codecov.io/gh/mockito/mockito/compare/2397/diff#D2 I think it would also be good to add some tests that method invocations and all are working as expected.

TimvdLippe (approved):
I like the idea, thanks for suggesting! I would like to see some tighter test coverage, but other than that this LGTM.

src/main/java/org/mockito/internal/creation/proxy/ProxyMockMaker.java
```diff
@@ -0,0 +1,172 @@
...
+    private final Method invokeDefault;
+
+    public ProxyMockMaker() {
+        Method m;
+        try {
+            m =
+                    InvocationHandler.class.getMethod(
+                            "invokeDefault", Object.class, Method.class, Object[].class);
+        } catch (NoSuchMethodException ignored) {
+            m = null;
```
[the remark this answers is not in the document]
raphw: It can be used. The method is only added in Java 16, so the functionality is offered only if it exists.

src/main/java/org/mockito/internal/creation/proxy/ProxyMockMaker.java
```diff
@@ -0,0 +1,172 @@
...
+        @Override
+        public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+            if (method.getDeclaringClass() == Object.class) {
+                switch (method.getName()) {
+                    case "hashCode":
+                        return System.identityHashCode(proxy);
+                    case "equals":
+                        return proxy == args[0];
+                    case "toString":
+                        return "";
```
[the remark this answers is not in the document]
raphw: You are right of course, darn, I overlooked your comments, I will add another PR.

src/test/java/org/mockito/internal/creation/proxy/ProxyMockMakerTest.java
raphw: We test for the `mockable` state. To avoid the overhead, those methods will never be invoked with non-mockable types and the error messages are rather undefined.

PR #2400, opened 2021-08-24
Fixes #2399 : Adds defaultAnswer to the MockitoMockKey to distinguish the mock types, i.e. to separate mocks from spies otherwise spy type is reused for a mock or vice versa.

### LINKED ISSUE #2399: Sporadic mock verification failures related to hashCode/equals on 3.12.1
Since updating from `3.11.2` to `3.12.0` (and getting similar errors on `3.12.1`), I have seen some strange sometimes-reproducible verification failures on Mockito for tests which previously passed consistently.

I can't yet establish a pattern or simple reproduction; it feels like it is somehow dependent on the order of test evaluation/usage with respect to other tests within the same JVM instance, and I have been unable to replicate locally.

There are two main classes I am seeing

1. mocks with interactions being tracked (and thus failing `verifyNoInteractions(mock)` after being put into `HashSet`s (these did not seem to previously count as an interaction, and usually do not)
    ```
    com.thoughtworks.go.server.materials.postcommit.git.GitPostCommitHookImplementerTest > shouldReturnEmptyListIfParamHasNoValueForRepoURL() FAILED
        org.mockito.exceptions.verification.NoInteractionsWanted: 
        No interactions wanted here:
        -> at com.thoughtworks.go.server.materials.postcommit.git.GitPostCommitHookImplementerTest.shouldReturnEmptyListIfParamHasNoValueForRepoURL(GitPostCommitHookImplementerTest.java:93)
        But found this interaction on mock 'gitMaterial':
        -> at java.base/java.util.HashMap.hash(HashMap.java:340)
        Actually, above is the only interaction with this mock.
            at com.thoughtworks.go.server.materials.postcommit.git.GitPostCommitHookImplementerTest.shouldReturnEmptyListIfParamHasNoValueForRepoURL(GitPostCommitHookImplementerTest.java:93)
    ```
    https://github.com/gocd/gocd/blob/b7bba59201c857efc010c3493664802f218d2bae/server/src/test-fast/java/com/thoughtworks/go/server/materials/postcommit/git/GitPostCommitHookImplementerTest.java#L81-L94

2. mocks that are not considered equal to themselves when put inside a `Collection` and then searched for 
    ```
    com.thoughtworks.go.server.materials.postcommit.git.GitPostCommitHookImplementerTest > shouldReturnListOfMaterialMatchingThePayloadURL() FAILED
        java.lang.AssertionError: 
        Expected: a collection containing <Mock for GitMaterial, hashCode: 0>
             but: was <Mock for GitMaterial, hashCode: 0>, was <Mock for GitMaterial, hashCode: 0>
            at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
            at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:8)
            at com.thoughtworks.go.server.materials.postcommit.git.GitPostCommitHookImplementerTest.shouldReturnListOfMaterialMatchingThePayloadURL(GitPostCommitHookImplementerTest.java:58)
    ```
    https://github.com/gocd/gocd/blob/b7bba59201c857efc010c3493664802f218d2bae/server/src/test-fast/java/com/thoughtworks/go/server/materials/postcommit/git/GitPostCommitHookImplementerTest.java#L41-L65

I have a handful of other examples too in different classes.

This could be confirmation bias, however since I saw some stuff was done with respect to hashCodes in #2331 I wondered if it could be related. Logging it here to see if anyone else has seen any similar weirdness or might know what is going on.

**Environment**
- Currently I _cannot_ replicate locally (MacOS Big Sur w/AdoptOpenJDK `15.0.2`) even where it is replicable on build/CI automation environment.
- The failures seem to be (somewhat) reproducible on AL2 Linux w/ Oracle JDK `15.0.1`) but I cannot distinguish a pattern, and sometimes a different combination/ordering of tests does not seem to yield the same failure on exactly the same test code.

vyazelenko:
## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/TypeCachingBytecodeGenerator.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/TypeCachingBytecodeGenerator.java
@@ -89,11 +89,11 @@ class TypeCachingBytecodeGenerator extends ReferenceQueue<ClassLoader>
         private final Answer defaultAnswer;
 
         private MockitoMockKey(
-            Class<?> type,
-            Set<Class<?>> additionalType,
-            SerializableMode serializableMode,
-            boolean stripAnnotations,
-            Answer defaultAnswer) {
+                Class<?> type,
+                Set<Class<?>> additionalType,
+                SerializableMode serializableMode,
+                boolean stripAnnotations,
+                Answer defaultAnswer) {
             super(type, additionalType);
             this.serializableMode = serializableMode;
             this.stripAnnotations = stripAnnotations;
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/TypeCachingBytecodeGenerator.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/TypeCachingBytecodeGenerator.java
@@ -5,12 +5,14 @@
 package org.mockito.internal.creation.bytebuddy;
 
 import java.lang.ref.ReferenceQueue;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import net.bytebuddy.TypeCache;
 import org.mockito.mock.SerializableMode;
+import org.mockito.stubbing.Answer;
 
 class TypeCachingBytecodeGenerator extends ReferenceQueue<ClassLoader>
         implements BytecodeGenerator {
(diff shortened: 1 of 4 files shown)


TimvdLippe (approved):
Nice find, fix and tests! 🎉

TimvdLippe:
Ah, you will need to run `./gradlew spotlessApply` locally to fix the formatting issues.

vyazelenko:
@TimvdLippe Pushed formatting fixes.

raphw:
That's a bad patch, I'm afraid. `Answer` is an interface and most instances do not implement hashCode/equals, this effectively breaks the caching mechanism what can cause significant overhead. The problem is the fix for #2331 which we must roll back. Mocks and spies classes must not consider the default answer when created classes. This must be solved on the interceptor level. I see what I can do but generally, we should not have allowed for the exception on the object methods.

TimvdLippe:
Oof, that's definitely on me for not fully understanding the domain space. Thanks Rafael for the additional knowledge, that was not something I was aware of. Reverting all the fixes seems like the right course of action at this point in time, to avoid further snowballing.

I still think the intent of #2331 is good (it caught numerous bad tests at Google), so I would like to figure out an alternative solution that does not have the downsides of the existing implementation.

PR #2402, opened 2021-08-24
Avoid cache breakage

raphw:
Removes a bad patch that introduces special treatment for equals/hashCode based on a particular default answer. This would require a more general overhaul as it is a major behavioral change and breaks the type cache.


src/test/java/org/mockitousage/spies/MockCreationShouldNotAffectSpyTest.java
```diff
@@ -1,48 +0,0 @@
...
-package org.mockitousage.spies;
-
-import org.junit.Test;
-import org.mockitoutil.TestBase;
-
-import static org.junit.Assert.assertEquals;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.spy;
-
-public class MockCreationShouldNotAffectSpyTest extends TestBase {
```
TimvdLippe, line 14: Can we still keep these tests? I think they are valuable, especially when we attempt to reland the fix at a later point in time.

src/test/java/org/mockitousage/spies/SpyCreationShouldNotAffectMockTest.java
```diff
@@ -1,48 +0,0 @@
...
-package org.mockitousage.spies;
-
-import org.junit.Test;
-import org.mockitoutil.TestBase;
-
-import static org.junit.Assert.assertEquals;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.spy;
-
-public class SpyCreationShouldNotAffectMockTest extends TestBase {
```
TimvdLippe, line 14: Same here

TimvdLippe (approved):
My bad for not fully understanding the domain space. I mostly based my reviews on the passing tests and did not consider the fact that the cache would be broken altogether. It would be great if we can have some sort of test that verifies the cache is working as expected, but I suppose that is quite difficult to do?

In any case, thanks for catching this!

src/test/java/org/mockitousage/spies/MockCreationShouldNotAffectSpyTest.java
```diff
@@ -1,48 +0,0 @@
...
-    @Test
-    public void test() {
-        TestClass instance = new TestClass(42);
-
-        TestClass mock = mock(TestClass.class);
-        assertEquals(mock.hashCode(), mock.hashCode());
-        assertEquals(mock, mock);
-
-        TestClass spy = spy(instance);
-        assertEquals(instance.hashCode(), spy.hashCode());
```
TimvdLippe, line 25: I guess we can't reinstate the test as-is, but I think we can if we remove this particular line?

src/test/java/org/mockitousage/spies/SpyCreationShouldNotAffectMockTest.java
raphw: I do not think these tests make sense as things currently are in Mockito. They basically test for the feature that is removed. I think there's a better way for test for this once we choose to enable hashCode/equals enablement.

PR #2411, opened 2021-08-26
Upgrade to Gradle 7.2

TimvdLippe:
Per https://github.com/mockito/mockito/pull/2410#issuecomment-906325812


TimvdLippe:
@raphw Not sure if you want to review actually, since it is not really an interesting PR. So if you don't mind, I will be merging this one tomorrow to continue working on Java 16 CI.

raphw (approved):
Looks good!

PR #2414, opened 2021-08-29
Adds additional strategies for invoking super methods

raphw:
When using the proxy mock maker different approaches are available when using older JVMs that do not yet offer a standard API. This should cover all use cases but older Android versions that do not offer this functionality until method handles are introduced.


TimvdLippe (approved):
I suppose the tests pass 😄

PR #2417, opened 2021-09-01
Run CI on Java 17

TimvdLippe:
Let's see how far we can get.

### DIFF
--- a/src/test/java/org/mockito/internal/creation/bytebuddy/SubclassByteBuddyMockMakerTest.java
+++ b/src/test/java/org/mockito/internal/creation/bytebuddy/SubclassByteBuddyMockMakerTest.java
@@ -4,24 +4,23 @@
  */
 package org.mockito.internal.creation.bytebuddy;
 
-import static org.assertj.core.api.Assertions.assertThat;
+import net.bytebuddy.ByteBuddy;
+import net.bytebuddy.ClassFileVersion;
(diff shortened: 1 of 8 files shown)


src/test/java/org/mockito/internal/creation/bytebuddy/SubclassByteBuddyMockMakerTest.java
```diff
@@ -34,6 +35,7 @@ public void is_type_mockable_excludes_primitive_wrapper_classes() {
     }
 
     @Test
+    @Ignore("Broken on JDK 17")
```
TimvdLippe, line 38: We were only running this on 17, but it seems to be failing anyways 😢

TimvdLippe:
The CodeCov diff now shows we are correctly using the Java 16+ codepath for invoking the relevant proxy method in the proxy mockmaker. So everything is working as expected and we have more code coverage 🎉

TimvdLippe:
@raphw Do you mind taking a look at the disabled test and let me know if that is intended?

src/test/java/org/mockito/internal/creation/bytebuddy/SubclassByteBuddyMockMakerTest.java
raphw: Fixed the test!

raphw (approved):

PR #2418, opened 2021-09-01
Remove all deprecated APIs for Mockito 4

### LINKED ISSUE #1517: Mockito 4
Suggested scope of Mockito 4

- [x] ~Move Java version compatibility to 8+~ - implemented in Mockito 3
- [ ] Enable strict stubbing by default in JUnit4 test runners (#769)
- [x] Remove all deprecated Mockito 2 classes (https://github.com/mockito/mockito/search?q=deprecated&unscoped_q=deprecated)
- [ ] Un-incubate (or promote) incubating features
- [x] Write migration guide for existing users
  - [ ] Write ErrorProne refactoring check that automatically moves all runners to `Silent`, so that companies can gradually upgrade.
- [ ] Move Junit5 extension from subproject to core project.

Any comments or feedback?

TimvdLippe:
All of these APIs have been marked as deprecated and have been present
in Mockito for quite a while. To cleanup our API surface and reduce
technical debt, let's remove these long-deprecated methods/classes.

An overview of now-deleted classes/methods:
- org.mockito.Matchers which was an alias for
org.mockito.ArgumentMatchers
- org.mockito.ArgumentMatchers#{anyObject,anyVararg} both which were
aliases for org.mockito.ArgumentMatchers#any
- org.mockito.ArgumentMatchers#any*Of, which were aliases for the same
method name without the Of and the generic parameters (which were
ignored)
- org.mockito.ArgumentMatchers#{is}{Not}Null(Class) which took a class
which was ignored. Aliases for the same methods without the parameter
- org.mockito.MockedStatic#verify which had the parameter types reversed
- org.mockito.Mockito#verifyZeroInteractions an alias of
verifyNoMoreInteractions
- org.mockito.Mockito#debug framework integration API that we later
refactored
- org.mockito.configuration.AnnotationEngine which was leaking internal
APIs and instead users should use org.mockito.plugins.AnnotationEngine
- org.mockito.exceptions.verification.TooLittleActualInvocations fixed
the grammar from "Little" to "Few"
- Numerous internal APIs that we never officially supported and can now
remove
- org.mockito.plugins.InstantiatorProvider which was leaking internal
APIs and instead users should use InstantiatorProvider2 (we should
probably rename back to remove the number in a future major release)
- org.mockito.runners a package that hosted several old JUnit runners
which were no longer supported. Users should instead use
org.mockito.junit.MockitoJUnitRunner which is our official JUnit4
runner.

Since these APIs are removed, this change effectively defines Mockito 4.

Closes #1517


TimvdLippe:
@mockito/core Today I was working on Mockito again and realized we are still shipping loads of deprecated APIs that have been part of Mockito for a long time. All of these APIs have valid replacements, sometimes they are even aliases of other methods.

Therefore, I propose we remove these deprecated APIs and publish that as Mockito 4. This should cleanup our codebase considerably, remove loads of tech debt and should guide our users to the officially supported classes/methods.

This PR is large, as I wasn't sure what was the best way to work on this. I can create separate PRs for each sub-item, but I wasn't sure how we do that with release. Therefore, this PR is large and can largely serve as discussion opportunity. If you want to, I can later split up the PRs and land them separately if you prefer.

Let me know what you think.

raphw (approved):
Good effort. Those have been there forever. I think reducing the surface is a good call especially since the API should be easy to learn.

bric3 (approved):
I would have split the removal of these APIs by _domain_, ie test usage, plugin loading. But maybe I'm too nit picky.

The change looks good, but since there's a long  list of removal, maybe have a document that list the removal, this could be done in the wiki.

Thank you @TimvdLippe for the effort, you rock !

TimvdLippe:
@bric3 Yeah I initially wanted to do that, but then I would have to open a couple of PRs and that would hamper the discussion. I think this PR is on the edge of manageable, so I guess it is easiest to continue with it, but yeah it is not ideal.

I will make sure that the GitHub release section will have the relevant APIs listed and their replacement.

TimvdLippe:
This PR is now pending a final approval of @mockitoguy and I think then we can merge this 🎉  Thanks for the speedy reviews!

bric3 (approved):
@TimvdLippe That makes sense, plus you're the one doing the effort.

TimvdLippe:
After a quick discussion, we decided to go ahead with this change. At the moment, we don't have any other breaking changes in mind that we want to land in Mockito 4 as well. To keep churn low for our users, we will remove these APIs in a single release, to hopefully keep the cost of upgrading to Mockito 4 manageable.

PR #2434, opened 2021-09-24
Javadoc: monospace font fallback

slauth:
I'm on Ubuntu and have none of the listed fonts installed. All code blocks are therefore rendered with a serif font which looks rather ugly. Adding the generic `monospace` family as fallback fixes this.


TimvdLippe (approved):
Nice fix!

PR #2437, opened 2021-10-01
Remove all Incubating annotations

### LINKED ISSUE #2433: Roadmap for Mockito.MockedStatic<T>
Hello.

What is the current roadmap for [`Mockito.MockedStatic<T>`](https://github.com/mockito/mockito/blob/caf35b24e2764df0498469526ecb3e7ec68a0430/src/main/java/org/mockito/MockedStatic.java#L27)?

Is it feature complete, or there are new features planned for it? Any other changes expected, before graduating?

How stable is the current implementation?

When is it expected for it to graduate out of incubation? Will it ever?

What work should be done and in general what else needs to happen to reach that? Maybe the community can help with that work. 

Thanks a lot for your open-source contributions!
--Vasile

---

_P.S. Don't know if these questions should be answered by @raphw, who seems to be the author of `Mockito.MockedStatic<T>` based on github history, or somebody else._

TimvdLippe:
For all of these APIs, we have been shipping them for at least 1 major
release. We consider all of these APIs part of the public API contract
now and therefore we should mark them as such.

This PR removes all usages of the annotation. If we were to later remove
an API, it will first have to be deprecated and then removed in a
subsequent major release. However, we have no plans to do so with any of
these APIs.

Closes #2433

### DIFF
--- a/src/main/java/org/mockito/AdditionalAnswers.java
+++ b/src/main/java/org/mockito/AdditionalAnswers.java
@@ -330,7 +330,6 @@ public final class AdditionalAnswers {
      *
      * @since 2.8.44
      */
-    @Incubating
     public static <T> Answer<T> answersWithDelay(long sleepyTime, Answer<T> answer) {
         return (Answer<T>) new AnswersWithDelay(sleepyTime, (Answer<Object>) answer);
     }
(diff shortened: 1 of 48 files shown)


raphw (approved):
I think this is right to do, those APIs have partially been stable for years and I do not assume them to change with any higher probability than any of the others.

mockitoguy:
Looks good, thanks!

PR #2494, opened 2021-11-30
Remove the recommendation to import ArgumentMatchers methods using Mockito

cushon:
The change removes the recommendation in javadoc to import methods in ArgumentMatchers by their non-canonical name, e.g. using `import static org.mockito.ArgumentMatchers.any;` to import `org.mockito.ArgumentMatchers.any` may be clearer than `import org.mockito.Mockito.any;`.

There's some related discussion in http://errorprone.info/bugpattern/NonCanonicalStaticImport


TimvdLippe (approved):
Good catch!

Issue #2499, opened 2021-12-07
Obscure issue with mockito-inline:  mock object in a test seems to be corrupting mock in later test


smas321:
This is a very obscure issue, happens only when the planets are aligned. Can you possibly tell me if this is an issue with mockito-inline or if I'm doing something wrong. If its an issue, is there a workaround?

I've simplified what I'm seeing as much as possible. I have 2 tests: Test1 and Test2. I mock MyClass in Test1 (calling clearInlineMocks in @AfterClass).  In Test2, I mock TestSubInterface which extends TestInterface, which creates an instance of the class I mocked in Test1. The issue is that the constructor for MyClass is not being called when mocking the TestSubInterface in Test2  (I guess because mockito-inline still sees MyClass as a mock from the previous test?). I initialize a variable in the MyClass constructor and later throw an error if it hasn't been initialized, resulting in a java.lang.NoClassDefFoundError when mocking.

HOWEVER, the constructor for MyClass IS called when:
- I mock TestInterface directly instead of TestSubInterface OR
- I do NOT have a default method defined in TestInterface OR
- I use mockito-core instead of mockito-inline

Here are the details:

```
java.lang.NoClassDefFoundError: Could not initialize class TestSubInterface$MockitoMock$1328129486
        at jdk.internal.reflect.GeneratedSerializationConstructorAccessor2.newInstance(Unknown Source)
        at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:500)
        at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:481)
        at org.objenesis.instantiator.sun.SunReflectionFactoryInstantiator.newInstance(SunReflectionFactoryInstantiator.java:48)
        at org.objenesis.ObjenesisBase.newInstance(ObjenesisBase.java:73)
        at org.mockito.internal.creation.instance.ObjenesisInstantiator.newInstance(ObjenesisInstantiator.java:21)
        at org.mockito.internal.creation.bytebuddy.InlineByteBuddyMockMaker.doCreateMock(InlineByteBuddyMockMaker.java:360)
        at org.mockito.internal.creation.bytebuddy.InlineByteBuddyMockMaker.createMock(InlineByteBuddyMockMaker.java:318)
        at org.mockito.internal.util.MockUtil.createMock(MockUtil.java:53)
        at org.mockito.internal.MockitoCore.mock(MockitoCore.java:62)
[...]

PR #2505, opened 2021-12-10
Make sure interface types are initialized before inline mocking to avoid blocking code of static initializers.

raphw:
Fixes https://github.com/mockito/mockito/issues/2499

There was no explicit initialization of subinterfaces which can of course also declare initializers. Those were not executed at the right time and broke the mocking.


TimvdLippe (changes requested):
Can we have a small regression test for this? The bug contains a small test case that we can hopefully copy?

raphw:
Good point, I stripped it to the minimum and merged it into a single test to avoid that the tests are run out of order.

TimvdLippe (approved):
Awesome thanks!

PR #2506, opened 2021-12-12
Fixes #2497: Throw exception on invalid matchers for `mockStatic`

### LINKED ISSUE #2497: ArgumentMatchers not working for Mockito.mockStatic
When I do a mockStatic with argument matchers like this:

```
var mockedRequestClass = Mockito.mockStatic(RequestClass.class);
mockedRequestClass.when(() -> RequestClass.sendRequest(mockedUrl, any(), any())).thenReturn(null);
```

Mockito throws the following exception:
`org.mockito.exceptions.misusing.MissingMethodInvocationException: 
when() requires an argument which has to be 'a method call on a mock'.`

How can I use argument matchers with Mockito.mockStatic? Shouldn't that be possible?

temp-droid:
I'm not sure if the `Throwable` catch block can be removed so I decided to keep it for now. Feel free to mention if it can be deleted.

I didn't follow the "snake case convention" for the test method to be consistent with the other test methods of the file, but I'm fine changing it.

## Checklist

### DIFF
--- a/subprojects/inline/src/test/java/org/mockitoinline/StaticMockTest.java
+++ b/subprojects/inline/src/test/java/org/mockitoinline/StaticMockTest.java
@@ -7,11 +7,14 @@ package org.mockitoinline;
 import static junit.framework.TestCase.assertEquals;
 import static junit.framework.TestCase.assertNull;
 import static junit.framework.TestCase.fail;
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.times;
 
 import java.util.concurrent.atomic.AtomicReference;
 
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 import org.mockito.MockedStatic;
 import org.mockito.Mockito;
 import org.mockito.exceptions.base.MockitoException;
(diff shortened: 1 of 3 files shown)


subprojects/inline/src/test/java/org/mockitoinline/StaticMockTest.java
```diff
@@ -20,6 +23,9 @@
 
 public final class StaticMockTest {
 
+    @Rule
+    public ExpectedException exception = ExpectedException.none();
```
TimvdLippe, line 27: Rather than this rule, let's keep it consistent with the rest of the codebase and use `assertj`. Then we can wrap the statement that we expect to fail, without letting it fail the whole method.

TimvdLippe (changes requested):
Nice and easy fix! I am not actually sure why we swallow exceptions here, but this is probably the safest change we can make. Only 1 nit with using assertj for more precise assertions.

subprojects/inline/src/test/java/org/mockitoinline/StaticMockTest.java
TimvdLippe: Ah yes makes sense. I filed #2510 for us to clean up `ExpectedException` and will clean some up today. Thanks for the fix!

TimvdLippe (approved):

PR #2513, opened 2021-12-14
Fix JavaDoc warnings and enforce zero errors in Gradle

TimvdLippe:
Now, we ensure that our JavaDoc doesn't produce any warnings. The most
common problems were using < instead of &lt; and & instead of the word
"and".

### DIFF
--- a/src/main/java/org/mockito/exceptions/verification/opentest4j/package-info.java
+++ b/src/main/java/org/mockito/exceptions/verification/opentest4j/package-info.java
@@ -6,7 +6,6 @@
 /**
  * Integration to provide better support for IDEs that support OpenTest4J.
  *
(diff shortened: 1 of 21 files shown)


src/main/java/org/mockito/exceptions/verification/opentest4j/package-info.java
```diff
@@ -6,7 +6,6 @@
 /**
  * Integration to provide better support for IDEs that support OpenTest4J.
  *
- * @see org.mocktio.exceptions.verification.junit
```
TimvdLippe, line 9: We misspelled our own package name here 😄

TimvdLippe:
I am merging this one as low-risk and mostly mechanical.

PR #2518, opened 2021-12-16
Fixes #2510: Remove ExpectedException from internal test suite

### LINKED ISSUE #2510: Remove `ExpectedException` from internal test suite
The rule is not ideal, as it will trigger even if the test fails for a completely different reason. Instead, we should use `assertj` with `assertThatThrownBy` to denote which specific line should trigger the exception.

temp-droid:
## Checklist

### DIFF
--- a/src/test/java/org/concurrentmockito/ThreadsRunAllTestsHalfManualTest.java
+++ b/src/test/java/org/concurrentmockito/ThreadsRunAllTestsHalfManualTest.java
@@ -115,7 +115,7 @@ public class ThreadsRunAllTestsHalfManualTest extends TestBase {
                             DescriptiveMessagesWhenVerificationFailsTest.class,
                             DescriptiveMessagesWhenTimesXVerificationFailsTest.class,
                             BasicVerificationInOrderTest.class,
-                            VerificationInOrderMixedWithOrdiraryVerificationTest.class,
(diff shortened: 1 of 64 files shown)


temp-droid:
Hey @TimvdLippe!

Outside one `ExpectedException` case in `SpyAnnotationTest` that I'm unsure of (not used in the past?) and one `@Test(expected…` in the `osgi-test` I'm not confortable touching, the rest should be replaced.

Before continuing replacing the `try fail finally` cases, I wanted to check with you if it wouldn't be best to create another PR for those. We are already at 72 commits for this one and I counted more than 380 occurrences in ~120 files for `fail()`.

I feel like this would be too much to review in a single PR, but you tell me.

TimvdLippe:
Yes. This PR is already massive, so let's cut it here. I hope to have time on Friday to review. Thanks for your effort!

Note that in this case I am okay with so many commits, since it is mechanical. However, for other PRs let's try to keep them small 😄

temp-droid:
Do you have suggestions how to approach the new PR? First creating an issue to link it to? Or referencing the same one?

I'm afraid it might create confusion to have 2 PRs for the same issue qua closing.

TimvdLippe:
Let's file a new issue for that. Thanks!

TimvdLippe (approved):
Thanks for the cleanup!

PR #2545, opened 2022-01-18
Issue 2544

jspyeatt:
Fixed issue 2544.

Mockito.mockConstructionWithAnswer() has a bug when there are more than 2 invocations possible.

If you run the ConstructionMockTest.testConstructionMocDefaultAnswerMultiple() as this:
```
    public void testConstructionMockDefaultAnswerMultiple() {
        try (MockedConstruction<Dummy> ignored = Mockito.mockConstructionWithAnswer(Dummy.class, 
              invocation -> "bar", 
              invocation -> "qux")) {
            assertEquals("bar", new Dummy().foo());
            assertEquals("qux", new Dummy().foo());
            assertEquals("qux", new Dummy().foo());
        }
    }
```
It works fine.
But if you change it to this:

```
    public void testConstructionMockDefaultAnswerMultiple() {
        try (MockedConstruction<Dummy> ignored = Mockito.mockConstructionWithAnswer(Dummy.class, 
               invocation -> "bar", 
               invocation -> "qux", 
               invocation -> "baz")) {
            assertEquals("bar", new Dummy().foo());
            assertEquals("qux", new Dummy().foo());
            assertEquals("baz", new Dummy().foo());
            assertEquals("baz", new Dummy().foo());
        }
    }
```
It will fail when trying to assert `qux`. It's because there is an error in the conditional logic in the `else if (context.getCount() >= additionalAnswers.length)`. It should be `>` not `>=`.

### DIFF
--- a/subprojects/inline/src/test/java/org/mockitoinline/ConstructionMockTest.java
+++ b/subprojects/inline/src/test/java/org/mockitoinline/ConstructionMockTest.java
@@ -50,10 +50,11 @@ public final class ConstructionMockTest {
 
     @Test
     public void testConstructionMockDefaultAnswerMultiple() {
-        try (MockedConstruction<Dummy> ignored = Mockito.mockConstructionWithAnswer(Dummy.class, invocation -> "bar", invocation -> "qux")) {
+        try (MockedConstruction<Dummy> ignored = Mockito.mockConstructionWithAnswer(Dummy.class, invocation -> "bar", invocation -> "qux", invocation -> "baz")) {
             assertEquals("bar", new Dummy().foo());
             assertEquals("qux", new Dummy().foo());
-            assertEquals("qux", new Dummy().foo());
+            assertEquals("baz", new Dummy().foo());
+            assertEquals("baz", new Dummy().foo());
         }
     }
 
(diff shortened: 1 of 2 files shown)


subprojects/inline/src/test/java/org/mockitoinline/ConstructionMockTest.java
```diff
@@ -50,10 +50,11 @@ public void testConstructionMockDefaultAnswer() {
 
     @Test
     public void testConstructionMockDefaultAnswerMultiple() {
```
TimvdLippe, line 52: Can we retain the original test (e.g. 2 invocation definitions and invoking it 3 times) and instead add a new regression test which is this new format? That way we ensure that both scenarios still work, thanks!

TimvdLippe (changes requested):
Nice and easy fix! Only a tiny nit with regards to the test.

subprojects/inline/src/test/java/org/mockitoinline/ConstructionMockTest.java
jspyeatt: Done

TimvdLippe (approved):
Thanks!

PR #2547, opened 2022-01-23
Fixes #2201 : Fixed checking of declared exceptions.

### LINKED ISSUE #2201: Calling getExceptionTypes() on concrete object that is used as interface doesn't return exception types from interface
org.mockito.internal.stubbing.answers.InvocationInfo#isValidException is doing validation on the exception types that can be thrown from this method.

In my codebase, I'm programming against an interface, but in fact, a spy is created from an instance that is implementing the interface, that drops the exception types that that implementation does not throw. However, the code that is calling the interface is of course unaware of that fact. Also, that implementation is a mock implementation (a real mock implementation, not a Mockito mock()/spy() instance). 

A testcase here:
```
import org.mockito.exceptions.base.MockitoException;

import java.io.IOException;

import static org.junit.jupiter.api.Assertions.assertThrows;
import static org.mockito.Mockito.doThrow;
import static org.mockito.Mockito.spy;

public class MockitoExceptionTypesIssue {
    @Test
    void exceptionTypeIoExceptionShouldBeMockableForObjectBeingSpyedOn() {
        final I x = spy(new C());
        final IOException ioException = new IOException("Oh my god!");
        final MockitoException e = assertThrows(
                MockitoException.class,
                () -> doThrow(ioException)
                        .when(x)
                        .m());
        assertEquals("\n" +
                "Checked exception is invalid for this method!\n" +
                "Invalid: " + ioException,
                e.getMessage()
        );
    }

    interface I {
        void m() throws IOException;
    }

    static class C implements I {
        public void m() {
            // no-op
        }
    }
}
```

andrey-kozel:
Fixed checking of declared exceptions. In case when parent contains throws keyword on its method and child overrides this method removing throws, it should be possible to mock throwing exception from child

## Description.
Lets assume that we have the following interface
```
public interface Foo {
        void bar() throws IOException; 
    }
```
And we have class that extends the interface, removing `throws` exception
```
public class Baz implements Foo {
        @Override
        public void bar() {
            
        }
    }
```

When we trying to mock exception throw, it wasn't possible
```
Foo mock = Mockito.spy(new Baz());
Mockito.doThrow(new IOException())
    .when(mock)
    .bar();
```

Now this behavior fixed

## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/stubbing/answers/InvocationInfo.java
+++ b/src/main/java/org/mockito/internal/stubbing/answers/InvocationInfo.java
@@ -6,6 +6,7 @@ package org.mockito.internal.stubbing.answers;
 
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.util.Arrays;
 
 import org.mockito.internal.invocation.AbstractAwareMethod;
 import org.mockito.internal.util.MockUtil;
@@ -25,6 +26,26 @@ public class InvocationInfo implements AbstractAwareMethod {
     }
 
     public boolean isValidException(Throwable throwable) {
+        if (isValidException(method, throwable)) {
+            return true;
+        }
+
+        return Arrays.stream(method.getDeclaringClass().getInterfaces())
+                .anyMatch(parent -> isValidExceptionForParent(parent, throwable));
+    }
+
+    private boolean isValidExceptionForParent(Class<?> parent, Throwable throwable) {
+        try {
+            Method parentMethod =
+                    parent.getMethod(this.method.getName(), this.method.getParameterTypes());
+            return isValidException(parentMethod, throwable);
+        } catch (NoSuchMethodException e) {
+            // ignore interfaces that doesn't have such a method
+            return false;
+        }
+    }
+
+    private boolean isValidException(Method method, Throwable throwable) {
         Class<?>[] exceptions = method.getExceptionTypes();
         Class<?> throwableClass = throwable.getClass();
         for (Class<?> exception : exceptions) {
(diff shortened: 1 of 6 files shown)


ruslanzianevich (approved):

aglumova:
WOW Amazing!!

src/main/java/org/mockito/internal/stubbing/answers/InvocationInfo.java
```diff
@@ -25,14 +26,33 @@ public InvocationInfo(InvocationOnMock theInvocation) {
     }
 
     public boolean isValidException(Throwable throwable) {
+        if (isValidException(method, throwable)) {
+            return true;
+        }
+
+        return Arrays.stream(method.getDeclaringClass().getInterfaces())
+                .anyMatch(parent -> isValidExceptionForParent(parent, throwable));
```
TimvdLippe, line 34: I am not 100% sure, but I think we should also recursively walk up the parent class hierarchy in case the direct parent doesn't define the method. E.g. I think the current regression tests still fails:

```java
public interface Foo {
    void bar() throws IOException; 
}
public abstract class Baz implements Foo {}
public class SubClass extends Baz {
    @Override
    public void bar() {
    }
}
```

TimvdLippe (changes requested):
Thanks for the PR! I think we are nearly there, but there are some more edge cases we need to handle.

andrey-kozel:
> Thanks for the PR! I think we are nearly there, but there are some more edge cases we need to handle.

Thank you for the comment. I've extended checking for exceptions and added tests for all the cases you've mentioned

TimvdLippe (approved):
Nice work and regression tests, thanks! 🎉

PR #2558, opened 2022-02-02
Fixes #2557 : fix rare mock self equals

### LINKED ISSUE #2557: Mock equals self not working
Mock generated class, when calls equals, it uses MockMethodInterceptor.ForEquals interceptor that checks reference equality.
During my tests, in some unknown case, a mock equals method instead calls DispatcherDefaultingToRealMethod that resolves to return default answer for boolean return method, that is false... resulting in invalid self equals result.

I don't have a reproductible test case, because I don't know how the generated mock class actually manages to call the dispatcher instead of ForEquals, only that if I debug this is what happens.

mockito 3.12.0

check that

cdalexndr:
Closes https://github.com/mockito/mockito/issues/2557

### DIFF
--- a/src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValues.java
+++ b/src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValues.java
@@ -4,9 +4,6 @@
  */
 package org.mockito.internal.stubbing.defaultanswers;
 
-import static org.mockito.internal.util.ObjectMethodsGuru.isCompareToMethod;
-import static org.mockito.internal.util.ObjectMethodsGuru.isToStringMethod;
-
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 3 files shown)


TimvdLippe (changes requested):
For us to better understand the problem you are running into, can you please share a standalone reproduction case similar to the existing cases such as https://github.com/mockito/mockito/blob/main/src/test/java/org/mockitousage/bugs/CovariantOverrideTest.java

I know you wrote in your issue description that you weren't able to do so yet, but it would greatly help us understand the root cause, so that we can evaluate whether this is indeed the appropriate fix. At the moment, I am not certain it is, but having a full use case would assist in that process. Thanks!

cdalexndr:
I'm running a suite of thousands of tests, when it reproduces.
If I run the bad test in isolation, it does not reproduce.

It seems that this is a bytebuddy related issue, because the call on mock is different in these two cases, so the generated bytecode must be different.

The code I added will act as a guard and will not be reached under normal workflow.
When this abberant workflow happens, it will prevent invalid behavior of equals.

cdalexndr:
fixed by **https://github.com/mockito/mockito/commit/4f81d4f1de676f845d6ee111eb6285dbdd5935c6**

Issue #2522, opened 2021-12-20
NullPointerException from Groovy metaclass methods when using mockito-inline (but not mockito-core)


stevegreenn:
Hey Mockito team,

I'm working on a project where we've dropped the usage of PowerMock in favour of `mockStatic` from `mockito-inline` but unfortunately this has led to an issue.

The issue seems to be the same as #2204. An outcome of that issue was the following test: https://github.com/mockito/mockito/blob/main/subprojects/groovyTest/src/test/groovy/org/mockito/groovy/GroovyMockitoTest.groovy

I've copied this into my project and this test is actually failing. I can make the test pass by replacing the dependency of `mockito-inline` with `mockito-core` but then I'd lose the ability to mock static/finals.

Versions:
- Groovy 3.0.9
- Mockito 4.2.0
- JUnit 4.13.1

The output of the failing test:

```
java.lang.NullPointerException
	at org.codehaus.groovy.runtime.callsite.PogoMetaClassSite.call(PogoMetaClassSite.java:38)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:47)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:125)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:130)
	at demo.GroovyMockitoTest$ClassUnderTest.methodUnderTest(test.groovy:36)
	at demo.GroovyMockitoTest$ClassUnderTest$methodUnderTest.call(Unknown Source)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:47)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:125)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:130)
```
 
Here is the failing test isolated in a small Gradle project: https://github.com/stevegreenn/mockito-inline-issue

Any ideas?

Many thanks,
Steve

chadlwilson:
Did you happen to find any correlation or more clues here @stevegreenn ?

Adding `mockito-inline` to the classpath via a transitive dependency of another framework seemed to break basically all of our Groovy tests (which tend to use Traits for common behaviour).

temp-droid:
[...]

PR #2566, opened 2022-02-16
#2522: mockito-inline & Groovy NullPointerException 

stevegreenn:
Hi!

It was requested by @TimvdLippe that I create a pull request with a new subproject that tests Groovy and mockito-inline. See here: https://github.com/mockito/mockito/issues/2522#issuecomment-1011376058

Unfortunately this test is currently failing as part of the `./gradlew clean build`, still throwing a NullPointerException. Hopefully the new subproject can be used to track the issue down though.

Steve


raphw:
Thanks, I'll see what we can do about this!

raphw:
I do not get an error when running this test, it passes.

You can extract the class files using `-Dnet.bytebuddy.dump=/some/folder`. Could you compare the instrumented files and suggest what methods should not be instrumented?

TimvdLippe:
Per https://github.com/mockito/mockito/issues/2522#issuecomment-1069615814 I think we need to update this PR to include https://github.com/mockito/mockito/commit/81bc5bd2ef05ffc16b3a76823fa5e4d15ee66dba, would we not? That said, we do need the regression test to fail without any modifications and then it would need to pass with the fix. Otherwise, it is not catching the correct error.

Lastly, @stevegreenn can you run `./gradlew spotlessApply` on your PR so that it gets formatted correctly?

raphw:
@stevegreenn Could you rebase your PR on my change and see if the test is still failing?

TimvdLippe:
@nick-ernie confirmed in https://github.com/mockito/mockito/issues/2522#issuecomment-1070921396 that https://github.com/mockito/mockito/commit/61279b27869f081551c5ae287216967013ee5e3b should fix the issue.

@stevegreenn can you incorporate the above commit or would you prefer us to take in your PR and get it ready-for-merge?

stevegreenn:
Thanks for the further work on this guys! I'll get this PR updated later today.

TimvdLippe (approved):
Nice fix, thanks for bearing with us!

TimvdLippe:
@stevegreenn Seems like there was a compilation failure on the commit. Can you check it out and verify that `./gradlew build` runs locally?

stevegreenn:
Looks like there is indeed a compilation error. From a quick glance, it looks like this is the problem:

https://github.com/mockito/mockito/pull/2566/files#diff-749d9baac17dc6e18e7db9ee686a840d65ac746dc291b591b37834ba4c8261d1R394

The method definition (found [here](https://github.com/mockito/mockito/pull/2566/files#diff-ae35471e0471321b026f5b500c35bb52d82365b1a6a19e1e00e06de930d31416R23)) requires a boolean parameter, which isn't being provided in the above snippet.

@raphw - can you take a look please? I'm on holiday at the minute but if it's as simple as adding true/false there I can make that change, or I can bring in another commit again.

TimvdLippe:
It's probably easier if we (as maintainers) push to your branch to fix it up to get it submitted. That way, we can ensure credit on the commits and get this merged.

raphw:
Was this due to a merge conflict? My branch is building fine.

HonorKnight:
@TimvdLippe @stevegreenn is anyone working on this? it seems to have stagnated a bit, and this fix would be invaluable to me if it can move forward.

raphw:
I cherry-picked the test project onto my branch here: https://github.com/mockito/mockito/pull/2618

Once @TimvdLippe reviews the PR, I can merge it and create a new release.

TimvdLippe:
This landed in #2618

PR #2571, opened 2022-02-18
Fixes #348 In stub mode even the last invocation is irrelevant

### LINKED ISSUE #348: Why does a stubOnly mock keeps reference of the last invocation?
I was looking at a memory leak caused by a mock in our tests and I noticed that creating a mock this way 

```java
Foo foo = mock(Foo.class, withSettings().stubOnly());
```
will still keep the last invocation in memory:
https://github.com/mockito/mockito/blob/c303e49b60d10960b4155bc1202483b08fa4ec9d/src/main/java/org/mockito/internal/stubbing/InvocationContainerImpl.java#L141
https://github.com/mockito/mockito/blob/c303e49b60d10960b4155bc1202483b08fa4ec9d/src/main/java/org/mockito/internal/verification/SingleRegisteredInvocation.java#L16

I was wondering if that was on purpose? I thought the whole point of `stubOnly()` was to prevent allocating too much memory for nothing by preventing `verify()` on a mock. I think something like that would make more sense:

```java
public class EmptyRegisteredInvocation implements RegisteredInvocations, Serializable {

    public void add(Invocation invocation) {
        // do nothing
    }

    public void removeLast() {
        // do nothing
    }

    public List<Invocation> getAll() {
        return Collections.emptyList();
    }

    public boolean isEmpty() {
        return true;
    }
}
```

bric3:
Hi there,

After re-reading #348, I think the invocation reservoir, doesn't have to keep the last invocation.

This PR changes the semantics of the invocation reservoir, so no invocations are stored.

## Checklist


davidburstromspotify:
Hi! Will this PR be worked on?

eduardbosch-jt:
Hi!
What is the status of this PR?
I think that it makes sense to not store any invocation when using configuring a mock with `stubOnly`.

I'm actually having come leaks due to this.

I'm trying to remove them with `clearInvocations`, but anyway, it looks like `stubOnly` should not store any invocation reference.

PR #2583, opened 2022-03-05
Restored spy equals/hashcode fix

### LINKED ISSUE #2331: Spy doesn't forward hashcode/equals to actual object

cdalexndr:
Fixes #2331
 
## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassBytecodeGenerator.java
@@ -44,6 +44,7 @@ import net.bytebuddy.implementation.FieldAccessor;
 import net.bytebuddy.implementation.Implementation;
 import net.bytebuddy.implementation.attribute.MethodAttributeAppender;
 import net.bytebuddy.matcher.ElementMatcher;
+import org.mockito.Answers;
 import org.mockito.codegen.InjectionBase;
 import org.mockito.exceptions.base.MockitoException;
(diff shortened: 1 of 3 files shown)


TimvdLippe (changes requested):
We can't revert this patch, as it will break caching of classes. I thought we added a regression test for that, but I would have to double check.

In any case, we will have to fix this in a different way, although unfortunately it is unclear what that entails.

cdalexndr:
> We can't revert this patch, as it will break caching of classes

Can you provide more details about this issue?

The cache key was fixed by https://github.com/mockito/mockito/commit/4f81d4f1de676f845d6ee111eb6285dbdd5935c6 and this revert includes that fix.

TimvdLippe:
#2400 was reverted per https://github.com/mockito/mockito/pull/2400#issuecomment-904663703 in #2402 The fix was not working as intended and broke the class cache.

PR #2593, opened 2022-03-14
Deprecate ListUtil and Fields classes

andreisilviudragnea:
- Deprecated ListUtil class in favor of Java 8 Streams
- Deprecated unused Fields class

Relates to #1926.


TimvdLippe (commented):
With "Android users", I meant those that are using Mockito on older versions of Android that does not support lambda's and all. However, these projects can desugar, which might solve the problem.

Therefore, if we are still compatible with Android SDK 24 (I believe that's the version we use), I think we can merge this as-is. Let's see what CI says.

src/main/java/org/mockito/internal/invocation/InvocationsFinder.java
```diff
@@ -128,7 +127,9 @@ public static Location getLastLocation(List<Invocation> invocations) {
     public static Invocation findPreviousVerifiedInOrder(
             List<Invocation> invocations, InOrderContext context) {
         LinkedList<Invocation> verifiedOnly =
```
TimvdLippe, line 129: Let's refactor this to an `List<Invocation>` so that we can use `Collectors.toList()` like the rest of the usages.

src/main/java/org/mockito/internal/util/collections/ListUtil.java
```diff
@@ -12,8 +12,10 @@
  * use cases neatly. However, we want to keep Mockito dependencies minimal. In Java8 we should be
  * able to get rid of this class.
  */
+@Deprecated
 public final class ListUtil {
```
TimvdLippe, line 16: For my understanding, are there any remaining usages of `ListUtil`? If not, then we can remove them. This is an internal class and we don't allow users on relying on it. The only reason I wanted to keep all of this was to remain compatible with our users who have to run on older versions of Android, but it seems like our CI is happy with this PR. Therefore, I think we can remove this altogether, if we removed all of our own usages.

TimvdLippe (changes requested):
Seems like CI is happy, so we can make this change after all. Thanks for submitting this again! Only a couple of small comments.

src/main/java/org/mockito/internal/util/collections/ListUtil.java
andreisilviudragnea: I suggest maybe separating the last commit of this PR into a separate release, so if there is a big number of users of `ListUtil` or `Fields` classes, they will have more time to migrate away from them.
But if you think it's ok to delete them now, then let's just do it.

TimvdLippe: Both of these classes reside in `org.mockito.internal`, so we don't consider them for SemVer. Therefore, we can safely delete them in any new version of Mockito.

TimvdLippe (approved):
Thanks for getting back to us! Yes, this is a great cleanup and the Stream usages read a lot cleaner than we had before. Merging 🎉

PR #2603, opened 2022-03-24
Only run injection once

### LINKED ISSUE #2602: mockito-core MockitoAnnotations::openMocks initializes class twice if said class has super class
## Short, Self Contained, Correct (Compilable), Example:
Have a look [here ](https://github.com/1-alex98/mre-mockito/tree/master/src/test/java). 
You will find a `Works` class and a `DoesNotWork` class. Both are essentially the same with a small difference being the `DoesNotWork` class does fail and extends and empty class.

### DoesNotWork.java:
```
@ExtendWith({MockitoExtension.class})
public class DoesNotWork extends ToBeExtended{
    @Mock
    ToBeMocked toBeMocked;
    @InjectMocks
    ToBeInitialized toBeInitialized;

    @Test
    public void foo(){
        assert toBeInitialized.toBeMocked == null;
    }
}
```
### ToBeInitialized.java:
```
public class ToBeInitialized {
    public ToBeMocked toBeMocked;

    public ToBeInitialized(ToBeMocked toBeMocked) {
        this.toBeMocked = null;
    }
}
```
Reading the docs that should work flawlessly. But it does not. As can be seen in `Works.java` the reason is the ` extends ToBeExtended` if you remove it everything works.

### Why does it not work

In the  `DoesNotWork` the `@InjectMocks` annotated instances are initialized twice. It does it twice, one time more for every super class there is. First the `@InjectMocks` annotated instances are initialized via constructor injection. In a second round Mockito realizes that DoesNotWork.toBeInitialized is already initialized and therefore chooses Field injection over constructor injection. That makes sense if somebody does something like:

```
...
    @InjectMocks
    Foo foo= new Foo();
...
```

### The code responsible
InjectingAnnotationEngine::processInjectMocks https://github.com/mockito/mockito/blob/caf35b24e2764df0498469526ecb3e7ec68a0430/src/main/java/org/mockito/internal/configuration/InjectingAnnotationEngine.java#L56

The method is called by the MockitoExtension:
```
processInjectMocks:61, InjectingAnnotationEngine (org.mockito.internal.configuration)
process:48, InjectingAnnotationEngine (org.mockito.internal.configuration)
openMocks:82, MockitoAnnotations (org.mockito)
<init>:43, DefaultMockitoSession (org.mockito.internal.framework)
startMocking:83, DefaultMockitoSessionBuilder (org.mockito.internal.session)
beforeEach:153, MockitoExtension (org.mockito.junit.jupiter)
```

```
    private List<AutoCloseable> processInjectMocks(
            final Class<?> clazz, final Object testInstance) {
        List<AutoCloseable> closeables = new ArrayList<>();
        Class<?> classContext = clazz;
        while (classContext != Object.class) {
            closeables.add(injectCloseableMocks(testInstance));
            classContext = classContext.getSuperclass();
        }
        return closeables;
    }
```
In this method for every super class the Test class has(`classContext.getSuperclass()`) it triggers the injection(` injectCloseableMocks`) yet another time(`while (classContext != Object.class)`). That might be clever if the super class has  `@InjectMocks` annotated instances its-self. But it is odd to call `closeables.add(injectCloseableMocks(testInstance));` instead of `closeables.add(injectCloseableMocks(classContext));` then. For me it would make more sense to call it on the super class then instead of repeating the initialization yet another time. But I have not tested this change. I can not foresee what implications this would have. Fixing this concrete bug is probably better of in the hands of somebody that is more familiar with Mockito's internal workings. 

## Why even care about this bug
Seems like it still initializes the instance right? Who does 
```
public ToBeInitialized(ToBeMocked toBeMocked) {
    this.toBeMocked = null;
 }
``` 
anyways?

Since newer Java versions get stricter in enforcing reflections over module borders this will increasingly become an issue. In my case Mockito does access a field via reflections it is not allowed to making it fail. But it should never have tried field injection since the class had a working constructor. Due to the bug that is descripted here it did nevertheless.  

## Versions:
JDK 17, Windows 11, Mockito Version 4.4.0 (Even though this all should be reproducible elsewhere and with different versions)

Thank you for working on Mockito. Great Framework ❤️ 
Any more questions? I am happy to help.

1-alex98:
Add regression test that makers sure constructor injection is used when possible
Fixes #2602

## Checklist


1-alex98:
See issue #2602  for context

src/test/java/org/mockito/internal/configuration/InjectingAnnotationEngineTest.java
```diff
@@ -1,55 +0,0 @@
...
-import org.junit.runner.RunWith;
-import org.mockito.InjectMocks;
-import org.mockito.Mock;
-import org.mockito.Spy;
-import org.mockito.junit.MockitoJUnitRunner;
-
-class I {}
-
-@RunWith(MockitoJUnitRunner.class)
-public class InjectingAnnotationEngineTest extends I {
```
TimvdLippe, line 18: As a user, I think the behavior as described in this test makes sense. Therefore, I would like to keep it. That said, it would be ideal if we can keep this behavior as-is, while also fixing your issue. Do you think that is feasible or are they mutually exclusive?

1-alex98: 1. This test tests a behaviour that was never there. It tests a bug. The proper behaviour for this would need to be implemented.
2. While this behaviour might make sense in some cases, it might be highly problematic in other cases. So I strongly advise against this feature. I could very well imagine this breaking quite some use cases of other's. The only reason nobody complained yet about this "feature"/tested behaviour that is does not exist yet(only in edge cases and even there does the wrong thing). Proper way of doing this would probably be to annotate setters used for DI explicitly. But I think that is a little to much to code into Mockito. 
3. The docs explicitly state the opposite of what this test is doing.

For more details/examples please read the comments in the issue.

1-alex98: Maybe there is a clever way to combine all use cases here but it would need to be implemented. There is nothing there yet as one might assume by this test

1-alex98: Commonly Beans contain fields that are not supposed to be injected, maybe an object field even. How would one know if a field is supposed to be injected? Well only way to let the user make annotations right?

A lot of strange cases can arise here, I would not go down that road. That's why DI Frameworks have annotations to clarify all this. And commonly constructor injection is recommended.

TimvdLippe: Injection in Mockito has always been a difficult topic. What one considers a bug, others consider a feature. In this case, the current behavior is what some of our users are relying on, as you saw with the reverts that happened. Therefore, if we were to make changes again, we are breaking those users.

This means that the Mockito maintainers now have to choose between two scenarios, without being able to satisfy all users. In this case, I lean towards keeping the current behavior to be conservative, rather than breaking these users and introduce regressions in their test suites.

If you want fully deterministic dependency injection, I advise you to use a DI framework as you mentioned. Mockito injection is a best-effort attempt.

1-alex98: Yes but current state is just a mess

1-alex98: Whatever you do leaving it like it is know is super broken

1-alex98: If I were you I would not support cases of mixed injection, where both constructor and setter injection is being used

subprojects/junit-jupiter/src/test/java/org/mockitousage/InjectMocksTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+class ToBeMockedInTestSuperClass{
+    int identifier;
+
+    public ToBeMockedInTestSuperClass(int identifier) {
+        this.identifier = identifier;
+    }
+}
+
+class ToBeMocked{
```
ZNhatAnhZ, line 23: dsadda

ZNhatAnhZ (commented):
asd

1-alex98:
@ZNhatAnhZ sadly nobody ever looked at it :D 

The current state is a mess. My Pull Request cleans up the mess and aliens the code with what is promised in the docs. It therefore has a remove a strange integration test that tests a bug(or a feature that was never there).

src/test/java/org/mockito/internal/configuration/InjectingAnnotationEngineTest.java
TimvdLippe: I am still not sure whether this behavior makes sense or not. @tokuhirom since you contributed the original regression test, can you please review this PR?

TimvdLippe (approved):
Based on user feedback, this is the appropriate fix after all. Running the tests one more time and then will merge it. Sorry for the long wait, it was unclear to me at the time what the appropriate behaviour of Mockito is.

subprojects/junit-jupiter/src/test/java/org/mockitousage/InjectMocksTest.java
```diff
@@ -0,0 +1,98 @@
...
+    public ToBeMocked(int identifier) {
+        this.identifier = identifier;
+    }
+}
+
+class TestClassToBeInitiatedViaConstructorInSuperClass {
+    ToBeMockedInTestSuperClass toBeMockedInTestSuperClass;
+
+    public TestClassToBeInitiatedViaConstructorInSuperClass(ToBeMockedInTestSuperClass toBeMockedInTestSuperClass) {
+        assert toBeMockedInTestSuperClass != null;
```
TimvdLippe, line 35: These asserts need to be changed to `AssertJ`, as it fails the build.

1-alex98:
I will update the PR tonight if I find the time

subprojects/junit-jupiter/src/test/java/org/mockitousage/InjectMocksTest.java
TimvdLippe: @1-alex98 Did you see this comment? We need to change the code, not just a rebase.

PR #2618, opened 2022-04-13
Groovy inline

raphw:
Fixes inline-mock-maker to support Groovy classes.


TimvdLippe (approved):
Woohoo!

TimvdLippe:
@HonorKnight We release about once a month, which will happen once we finish up #2613

TimvdLippe:
4.5.0 is now available on Maven Central: https://repo1.maven.org/maven2/org/mockito/mockito-core/4.5.0/

HonorKnight:
@TimvdLippe Doesn't there need to be a 4.5.0 release of mockito-inline too?

TimvdLippe:
Yes that is published as well: https://repo1.maven.org/maven2/org/mockito/mockito-inline/4.5.0/

PR #2642, opened 2022-05-15
fix Reproducible Build issue

hboutemy:
see https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/mockito/mockito-core/README.md

release 4.5.1 has only 1 cause of non-reproducible bits: unix group permission on files

with that little change, next release should be 100% reproducible as reproduced by independant people

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -72,6 +72,8 @@ allprojects { proj ->
     tasks.withType(AbstractArchiveTask) {
         preserveFileTimestamps = false
         reproducibleFileOrder = true
+        dirMode = Integer.parseInt("0755", 8)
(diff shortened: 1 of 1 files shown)


TimvdLippe (approved):
Thanks for the fix!

hboutemy:
FYI, to day I rebuilt 4.6.0 release and Mockito and got a perfect reproducibility!
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/mockito/mockito-core/mockito-core-4.6.0.buildcompare

nice work, it's the first Gradle rebuild of an OSS release that was fully confirmed!
I hope it will be followed by a long list of other projects

PR #2650, opened 2022-05-23
Fixes #2648 : Add support for customising strictness via @Mock annotation and MockSettings

### LINKED ISSUE #2648: Any way to enable Strict Stubbing when using Mockito.mock() without using @Mock?
Hi,
I am using mockito-core 4.5.1 and JUnit 5. I can use `MockitoExtension` and `@Mock` annotation and by default I get Strict Stubbing behaviour. However, if I use `Mockito.Mock(MyClass.class)` to create mock it seems Strictness is not enabled. Even if I stub unnecessary methods it is not giving any error. 

I know I can work around this by using MockitoExtension to create mocks using @Mock, but is there any way to enable Strict Stubbing while using `Mockito.mock(MyClass.class);`

```
class ProductServiceTest {

    private ProductRepository productRepository;
    private ProductService productService;

    @BeforeEach
    void setUp() {
        productRepository = Mockito.mock(ProductRepository.class);
        productService = new ProductService(productRepository);
        // this is unnecessary method stubbing but it is not giving any error
        Mockito.when(productRepository.findAllActiveProducts()).thenReturn(List.of());
    }

    @Test
    void shouldReturnOnlyActiveProducts() {
        BDDMockito.given(productRepository.findAll()).willReturn(List.of(p1, p2));
        productService.findAll(); // which calls productRepository.findAll()
    }
}
```

sivaprasadreddy:
This PR fixes the issue https://github.com/mockito/mockito/issues/2648.

This is to add support for customising strictness level for mock using @Mock annotation or MockSettings.

Example 1:

```
class MyTest {
     @Mock(strictness = Strictness.LENIENT)
     Foo foo;
     ....
}
```

Example 2:

```
class MyTest {
     Foo mock = Mockito.mock(Foo.class, withSettings().strictness(Strictness.WARN));
     ....
}
```

### DIFF
--- a/src/main/java/org/mockito/Mock.java
+++ b/src/main/java/org/mockito/Mock.java
@@ -13,6 +13,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
 import org.mockito.junit.MockitoJUnitRunner;
+import org.mockito.quality.Strictness;
 import org.mockito.stubbing.Answer;
 
 /**
@@ -111,4 +112,12 @@ public @interface Mock {
      * @since 2.23.3
      */
     boolean lenient() default false;
+
+    /**
+     * Mock will have custom strictness, see {@link MockSettings#strictness(Strictness)}.
+     * For examples how to use 'Mock' annotation and parameters see {@link Mock}.
+     *
+     * @since 4.5.2
+     */
+    Strictness strictness() default Strictness.STRICT_STUBS;
 }
--- a/src/main/java/org/mockito/MockSettings.java
+++ b/src/main/java/org/mockito/MockSettings.java
@@ -361,4 +361,20 @@ public interface MockSettings extends Serializable {
      * For more information and an elaborate example, see {@link Mockito#lenient()}.
      */
     MockSettings lenient();
+
+    /**
+     * Specifies strictness level for the mock.
+     * The default strictness level is STRICT_STUBS
+     *
+     * <pre class="code"><code class="java">
+     *   Foo defaultStrictMock = mock(Foo.class);
+     *   Foo explicitStrictMock = mock(Foo.class, withSettings().strictness(Strictness.STRICT_STUBS));
+     *   Foo lenientMock = mock(Foo.class, withSettings().strictness(Strictness.LENIENT));
+     * </code></pre>
+     *
+     * @param strictness the strictness level to set on mock
+     * @return settings instance so that you can fluently specify other settings
+     * @since 4.5.2
+     */
+    MockSettings strictness(Strictness strictness);
 }
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -105,6 +105,7 @@ import java.util.function.Function;
  *      <a href="#49">49. New API for mocking object construction (Since 3.5.0)</a><br/>
  *      <a href="#50">50. Avoiding code generation when restricting mocks to interfaces (Since 3.12.2)</a><br/>
  *      <a href="#51">51. New API for marking classes as unmockable (Since 4.1.0)</a><br/>
+ *      <a href="#51">52. New strictness attribute for @Mock annotation and <code>MockSettings.strictness()</code> methods (Since 4.5.2)</a><br/>
  * </b>
  *
  * <h3 id="0">0. <a class="meaningful_link" href="#mockito2" name="mockito2">Migrating to Mockito 2</a></h3>
@@ -1606,6 +1607,20 @@ import java.util.function.Function;
  * For any class/interface you own that is problematic to mock, you can now mark the class with {@link org.mockito.DoNotMock @DoNotMock}. For usage
  * of the annotation and how to ship your own (to avoid a compile time dependency on a test artifact), please see its JavaDoc.
  * <p>
+ *
+ * <h3 id="52">52. <a class="meaningful_link" href="#mockito_strictness" name="mockito_strictness">
+ *  New strictness attribute for @Mock annotation and <code>MockSettings.strictness()</code> methods (Since 4.5.2)</a></h3>
+ *
+ * By default Mocks use Strict stubbing and now you can customize the strictness level
+ * either using @Mock annotation strictness attribute or using MockSettings.strictness().
+ *
+ * <pre class="code"><code class="java">
+ *   @Mock(strictness = Strictness.LENIENT)
+ *   Foo mock;
+ *   // using MockSettings.withSettings()
+ *   Foo mock = Mockito.mock(Foo.class, withSettings().strictness(Strictness.WARN));
+ * </code></pre>
+ *
  */
 @CheckReturnValue
 @SuppressWarnings("unchecked")
--- a/src/main/java/org/mockito/internal/creation/MockSettingsImpl.java
+++ b/src/main/java/org/mockito/internal/creation/MockSettingsImpl.java
@@ -12,6 +12,7 @@ import static org.mockito.internal.exceptions.Reporter.extraInterfacesDoesNotAcc
 import static org.mockito.internal.exceptions.Reporter.extraInterfacesRequiresAtLeastOneInterface;
 import static org.mockito.internal.exceptions.Reporter.methodDoesNotAcceptParameter;
 import static org.mockito.internal.exceptions.Reporter.requiresAtLeastOneListener;
+import static org.mockito.internal.exceptions.Reporter.strictnessDoesNotAcceptNullParameter;
 import static org.mockito.internal.util.collections.Sets.newSet;
 
 import java.io.Serializable;
@@ -33,6 +34,7 @@ import org.mockito.listeners.VerificationStartedListener;
 import org.mockito.mock.MockCreationSettings;
 import org.mockito.mock.MockName;
 import org.mockito.mock.SerializableMode;
+import org.mockito.quality.Strictness;
 import org.mockito.stubbing.Answer;
 
 @SuppressWarnings("unchecked")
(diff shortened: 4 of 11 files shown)


src/main/java/org/mockito/MockSettings.java
```diff
@@ -361,4 +361,20 @@ public interface MockSettings extends Serializable {
      * For more information and an elaborate example, see {@link Mockito#lenient()}.
      */
     MockSettings lenient();
+
+    /**
+     * Specifies strictness level for the mock.
+     * The default strictness level is STRICT_STUBS
```
TimvdLippe, line 367: Nit: the default is the default of the runner. If you use the lenient runner, it will be lenient. I suggest to formulate it the following way:

```
The default strictness level is determined by the rule/runner used. If you are using no rule/runner, the default strictness level is LENIENT.
```

src/main/java/org/mockito/internal/creation/settings/CreationSettings.java
```diff
@@ -170,6 +171,11 @@ public boolean isStubOnly() {
 
     @Override
     public boolean isLenient() {
```
TimvdLippe, line 173: Do we still use this method somewhere else? If yes, we should refactor these to call `getStrictness` and deprecate this method (since we ship it in the `MockSettings` interface, we can't remove it just yet).

src/main/java/org/mockito/mock/MockCreationSettings.java
```diff
@@ -124,4 +124,12 @@
      * @since 2.20.0
      */
     boolean isLenient();
+
+    /**
+     * Sets strictness level for the mock, e.g. having {@link Strictness#STRICT_STUBS} characteristic.
+     * For more information about using mocks with custom strictness, see {@link MockSettings#strictness(Strictness)}.
+     *
+     * @since 4.5.2
```
TimvdLippe, line 132: Nit: 4.6.0

src/main/java/org/mockito/internal/creation/MockSettingsImpl.java
```diff
@@ -239,7 +241,16 @@ public <T2> MockCreationSettings<T2> buildStatic(Class<T2> classToMock) {
...
-        this.lenient = true;
+        this.strictness = Strictness.LENIENT;
+        return this;
+    }
+
+    @Override
+    public MockSettings strictness(Strictness strictness) {
+        this.strictness = strictness;
+        if (strictness == null) {
+            throw strictnessDoesNotAcceptNullParameter();
```
TimvdLippe, line 252: This line isn't covered yet by the tests. Therefore, let's add a test that attempts to set the `strictness` to `null` and ensures that an Exception is thrown.

src/main/java/org/mockito/MockSettings.java
```diff
@@ -361,4 +361,20 @@ public interface MockSettings extends Serializable {
      * For more information and an elaborate example, see {@link Mockito#lenient()}.
      */
     MockSettings lenient();
```
TimvdLippe, line 363: We should deprecate this method and point users to the `strictness` method.

src/main/java/org/mockito/Mock.java
```diff
@@ -111,4 +112,12 @@
      * @since 2.23.3
      */
     boolean lenient() default false;
```
TimvdLippe, line 114: We should deprecate this method as well.

src/main/java/org/mockito/Mockito.java
```diff
@@ -1606,6 +1607,20 @@
  * For any class/interface you own that is problematic to mock, you can now mark the class with {@link org.mockito.DoNotMock @DoNotMock}. For usage
  * of the annotation and how to ship your own (to avoid a compile time dependency on a test artifact), please see its JavaDoc.
  * <p>
+ *
+ * <h3 id="52">52. <a class="meaningful_link" href="#mockito_strictness" name="mockito_strictness">
+ *  New strictness attribute for @Mock annotation and <code>MockSettings.strictness()</code> methods (Since 4.5.2)</a></h3>
+ *
+ * By default Mocks use Strict stubbing and now you can customize the strictness level
```
TimvdLippe, line 1614: Nit: mocks by default are lenient. I would omit that part altogether and reword it to:

```
You can now customize the strictness level for a single mock, either using `@Mock` annotation strictness attribute or using `MockSettings.strictness()`. This can be useful if you want all of your mocks to be strict, but one of the mocks to be lenient.
```

src/test/java/org/mockitousage/strictness/StrictnessMockAnnotationTest.java
```diff
@@ -0,0 +1,46 @@
...
+    IMethods lenientMock;
+
+    @Mock IMethods regularMock;
+
+    @Test
+    public void mock_is_lenient() {
+        when(lenientMock.simpleMethod("1")).thenReturn("1");
+        when(regularMock.simpleMethod("2")).thenReturn("2");
+
+        // then lenient mock does not throw:
```
TimvdLippe, line 34: Nit: let's extract separate test methods for these cases. E.g. we have 2 tests here, one which ensures the `lenientMock` is allowed and one where the `regularMock` throws.

src/test/java/org/mockitousage/strictness/StrictnessMockAnnotationTest.java
```diff
@@ -0,0 +1,46 @@
...
+        ProductionCode.simpleMethod(lenientMock, "3");
+
+        // but regular mock throws:
+        Assertions.assertThatThrownBy(
+                        new ThrowableAssert.ThrowingCallable() {
```
TimvdLippe, line 39: Nit: I think we can use a lambda here, can we not?

```java
() -> ProductionCode.simpleMethod(regularMock, "4")
```

src/test/java/org/mockitousage/strictness/StrictnessWithSettingsTest.java
```diff
@@ -0,0 +1,62 @@
...
+        strictMock = mock(IMethods.class, withSettings().strictness(Strictness.STRICT_STUBS));
+    }
+
+    @Test
+    public void mock_is_lenient() {
+        when(lenientMock.simpleMethod("1")).thenReturn("1");
+        when(regularMock.simpleMethod("3")).thenReturn("3");
+        when(strictMock.simpleMethod("2")).thenReturn("2");
+
+        // then lenient mock does not throw:
```
TimvdLippe, line 41: Same here, let's have 3 separate 3 test cases here.

src/main/java/org/mockito/Mock.java
```diff
@@ -111,4 +112,12 @@
...
+
+    /**
+     * Mock will have custom strictness, see {@link MockSettings#strictness(Strictness)}.
+     * For examples how to use 'Mock' annotation and parameters see {@link Mock}.
+     *
+     * @since 4.5.2
```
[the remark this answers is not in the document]
sivaprasadreddy: Updated.

src/main/java/org/mockito/MockSettings.java
sivaprasadreddy: Updated.

src/main/java/org/mockito/internal/creation/settings/CreationSettings.java
sivaprasadreddy: Yes, this isLenient() is being used in one place in main code. I have refactored that to use getStrictness().

But this isLenient() is also being used in few tests, I think we can remove those reference while removing this deprecated method.

src/main/java/org/mockito/internal/creation/MockSettingsImpl.java
sivaprasadreddy: Added test to verify this.

src/main/java/org/mockito/Mock.java
sivaprasadreddy: Added deprecation marker and suggestion to use `strictness()`.

src/main/java/org/mockito/Mockito.java
sivaprasadreddy: Updated.

src/test/java/org/mockitousage/strictness/StrictnessMockAnnotationTest.java
sivaprasadreddy: Updated.

src/test/java/org/mockitousage/strictness/StrictnessWithSettingsTest.java
sivaprasadreddy: Separated this into multiple tests.

TimvdLippe (approved):
Thanks for the PR and great implementation 🎉

src/test/java/org/mockitousage/strictness/StrictnessWithSettingsTest.java
```diff
@@ -0,0 +1,57 @@
...
+    @Test
+    public void mock_is_lenient() {
+        when(lenientMock.simpleMethod("1")).thenReturn("1");
+
+        // lenient mock does not throw
+        ProductionCode.simpleMethod(lenientMock, "3");
+    }
+
+    @Test
+    public void mock_is_strict_with_default_settings() {
```
chadlwilson, line 43: Quick Q - was it intentional for this to change mocks defined by `mock(Blah.class)`  i.e with no rule/Junit `MockitoSettings` to strict by default?

Quick Q - was it intentional for this to change mocks defined by `mock(Blah.class)`  i.e with no rule/Junit `MockitoSettings` to strict by default?

I believe the previous behaviour was lenient by default?

TimvdLippe: We have not changed the behavior here. This test is using a Strict rule explicitly (https://github.com/mockito/mockito/pull/2650/files/d0907808e9fce24f6dcddc94d1ea2c3ebfa15676#diff-3040268f6b18ef562dd495ac7fc13465723311889d5f326bfb615cc3aed37415R21) to verify that we handle that case correctly.

In case you are hitting any regressions, please file a new issue. You are most likely running into a different problem.

chadlwilson: OK, thanks. We definitely have regressions (previously lenient mocks now behaving as strict) but I may have misidentified the possible cause. Will need to dig further if I get time.

chadlwilson: Another person has beat me to raising it at #2656 but this seems to be the same issue we are having.

Issue #2656, opened 2022-05-31
Regression? Strictness set in `@MockitoSettings` ignored after upgrade from 4.5.1 to 4.6.0


big-andy-coates:
Upgrading from Mockito 4.5.1 to 4.6.0 and it looks to me as though https://github.com/mockito/mockito/pull/2650 or something around there has introduced a regression.

I'm seeing `PotentialStubbingProblem` exceptions where previously there were none, because the test class is annotated with `@MockitoSettings(strictness = Strictness.LENIENT)`.

The issue seems to be that `StrictnessSelector` prefers the strictness set in the mock to the test level strictness, and the mock _defaults_ to `STRICT_STUBS`, so always overrides the strictness set in `@MockitoSettings`.

I've added a failing example test that demonstrates the issue here:  https://github.com/big-andy-coates/mockito/pull/1

The above test fails with:

```
expected: 
  Optional.empty
 but was: 
  Optional[org.mockito.exceptions.misusing.PotentialStubbingProblem: 
  Strict stubbing argument mismatch. Please check:
   - this invocation of 'test' method:
      rootMock.test("Bar");
      -> at org.mockitousage.ProductionCode.simpleMethod(ProductionCode.java:9)
   - has following stubbing(s) with different arguments:
      1. rootMock.test("Foo");
        -> at org.mockitousage.StrictnessTest$LenientMockitoSettings.should_not_throw_on_potential_stubbing_issue(StrictnessTest.java:176)
  Typically, stubbing argument mismatch indicates user mistake when writing tests.
  Mockito fails early so that you can debug potential problem easily.
  However, there are legit scenarios when this exception generates false negative signal:
    - stubbing the same method multiple times using 'given().will()' or 'when().then()' API
      Please use 'will().given()' or 'doReturn().when()' API for stubbing.
    - stubbed method is intentionally invoked with different arguments by code under test
      Please use default or 'silent' JUnit Rule (equivalent of Strictness.LENIENT).
  For more information see javadoc for PotentialStubbingProblem class.]
```

 - [x] The mockito message in the stacktrace have useful information, but it didn't help
 - [x] The problematic code (if that's possible) is copied here;
       Note that some configuration are impossible to mock via Mockito
 - [x] Provide versions (mockito / jdk / os / any other relevant information)
 - [x] Provide a [Short, Self Contained, Correct (Compilable), Example](http://sscce.org) of the issue
       (same as any question on stackoverflow.com)
[...]

PR #2659, opened 2022-05-31
Fix for Issue2656

big-andy-coates:
fix: https://github.com/mockito/mockito/issues/2656

Release 4.6.0 introduced a new feature https://github.com/mockito/mockito/pull/2650 to allow strictness to be set via the `@Mock` annotation. Unfortunately, the feature has a bug that results in any test-level strictness setting being ignored, e.g. 

```java
@ExtendWith(MockitorExtension.class)
@MockitoSettings(strictness = Strictness.LENIENT)
class ThingTest {

  @Mock
  private Thing lenientMock;
}
```

In the above code the `lenientMock` would actually be strict, as the `@Mock`'s `strictness` defaults to `STRICT_STUBS`. This default was overriding the test level `LENIENT`.

To fix the issue the default of the `Mock.strictness` method needs to be something meaning `not set`, allowing the mock's strictness to be ignored and the test level strictness to be used.

Unfortunately, no `no set` value exists in the `Strictness` enum and it is illegal to default to `null`. Fixing this in a backwards compatible way would have meant polluting the public API, (e.g. adding a `DEFAULT` value to the existing `Strictness` enum, which doesn't make sense in a lot of other places the enum is used).  As this feature has only just been released, and is essentially broken, a breaking change is being introduced to fix the issue.  

BREAKING CHANGE: This PR changes the return value of `Mock.strictness()` to a local `Strictness` enum value.  This will require users to change any code written against 4.6.0 that sets the `strictness` via the `@Mock` annotation to change their code.

Example code migration:

```java
class ThingTest {

    @Mock(strictness = Strictness.LENIENT)
    private Thing thing;

    ...
}
```

To:

```java
class ThingTest {

    @Mock(strictness = Mock.Strictness.LENIENT)
    private Thing thing;

    ...
}
```

## Checklist

### DIFF
--- a/src/main/java/org/mockito/Mock.java
+++ b/src/main/java/org/mockito/Mock.java
@@ -122,7 +122,7 @@ public @interface Mock {
      *
      * @since 4.6.1
      */
-    Strictness strictness() default Strictness.NOT_SET;
+    Strictness strictness() default Strictness.TEST_LEVEL_DEFAULT;
 
     enum Strictness {
 
@@ -131,7 +131,7 @@ public @interface Mock {
          *
          * @since 4.6.1
          */
-        NOT_SET,
+        TEST_LEVEL_DEFAULT,
 
         /**
          * See {@link org.mockito.quality.Strictness#LENIENT}
--- a/src/main/java/org/mockito/Mock.java
+++ b/src/main/java/org/mockito/Mock.java
@@ -13,7 +13,6 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
 import org.mockito.junit.MockitoJUnitRunner;
-import org.mockito.quality.Strictness;
 import org.mockito.stubbing.Answer;
 
 /**
@@ -34,12 +33,13 @@ import org.mockito.stubbing.Answer;
  *       &#064;Mock(name = "database") private ArticleDatabase dbMock;
  *       &#064;Mock(answer = RETURNS_MOCKS) private UserProvider userProvider;
  *       &#064;Mock(extraInterfaces = {Queue.class, Observer.class}) private ArticleMonitor articleMonitor;
+ *       &#064;Mock(strictness = Mock.Strictness.LENIENT) private ArticleConsumer articleConsumer;
  *       &#064;Mock(stubOnly = true) private Logger logger;
  *
  *       private ArticleManager manager;
  *
  *       &#064;Before public void setup() {
- *           manager = new ArticleManager(userProvider, database, calculator, articleMonitor, logger);
+ *           manager = new ArticleManager(userProvider, database, calculator, articleMonitor, articleConsumer, logger);
  *       }
  *   }
  *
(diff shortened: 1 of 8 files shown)


src/main/java/org/mockito/internal/configuration/MockAnnotationProcessor.java
```diff
@@ -45,10 +46,12 @@ public static Object processAnnotationForMock(
         if (annotation.stubOnly()) {
             mockSettings.stubOnly();
         }
-        mockSettings.strictness(annotation.strictness());
         if (annotation.lenient()) {
             mockSettings.lenient();
         }
+        if (annotation.strictness() != Mock.Strictness.NOT_SET) {
+            mockSettings.strictness(annotation.strictness().outer());
```
TimvdLippe, line 53: Rather than defining an `outer` method, can we instead use the dynamic enum valueOf here?

src/main/java/org/mockito/Mock.java
```diff
@@ -117,10 +117,45 @@
...
-    Strictness strictness() default Strictness.STRICT_STUBS;
+    Strictness strictness() default Strictness.NOT_SET;
+
+    enum Strictness {
+
+        /**
+         * Default value used to indicate the mock does not override the test level strictness.
+         *
+         * @since 4.6.1
+         */
```
TimvdLippe, line 133: I think naming it `TEST_LEVEL_DEFAULT` is a bit more descriptive and easier to understand for our users.

src/test/java/org/mockito/MockTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Test
+    public void should_have_matching_enum_in_mock_strictness_enum() {
+        final Mock.Strictness mockStrictness = Mock.Strictness.valueOf(this.strictness.name());
+        assertThat(mockStrictness.outer()).isEqualTo(strictness);
+    }
+
+    @Parameterized.Parameters
+    public static org.mockito.quality.Strictness[] data() {
+        return org.mockito.quality.Strictness.values();
```
TimvdLippe, line 30: Should we also test it the other way around? E.g. verify that all enum values in Mock.Strictness also match the Strictness enum values?

TimvdLippe (changes requested):
Some minor changes, but overall looking good!

src/main/java/org/mockito/internal/configuration/MockAnnotationProcessor.java
big-andy-coates: sure

src/main/java/org/mockito/Mock.java
big-andy-coates: Not sure I agree.  This value is explicitly saying that the `strictness` at the mock level is not set, hence `NOT_SET`.  IMHO, `NOT_SET` is more descriptive than `TEST_LEVEL_DEFAULT`.

However, I don't feel strongly about it.  So if you do I'm happy to change.

TimvdLippe: Yeah I think `TEST_LEVEL_DEFAULT` is more descriptive for our users, so I would prefer to name it that.

big-andy-coates: OK, done.

TimvdLippe (approved):
Thanks for the fix!

Issue #2644, opened 2022-05-17
Varargs methods cause `ClassCastException` in `AnswerFunctionalInterfaces`


perlun:
Hi,

The other day, I discovered that this code does not work:

```java
        doAnswer(answerVoid((VoidAnswer2<String, Object[]>)
                logger::info
        )).when(mock)
                .info(any(), (Object[]) any());

        // This call will throw an exception
        mock.info("Some message with {} {} {}", "three", "parameters", "");
```

The exception details looks as follows:

```
class java.lang.String cannot be cast to class [Ljava.lang.Object; (java.lang.String and [Ljava.lang.Object; are in module java.base of loader 'bootstrap')
java.lang.ClassCastException: class java.lang.String cannot be cast to class [Ljava.lang.Object; (java.lang.String and [Ljava.lang.Object; are in module java.base of loader 'bootstrap')
	at org.mockito.internal.stubbing.answers.AnswerFunctionalInterfaces$4.answer(AnswerFunctionalInterfaces.java:97)
	at org.mockito.internal.stubbing.answers.AnswerFunctionalInterfaces$4.answer(AnswerFunctionalInterfaces.java:93)
	at org.mockito.internal.stubbing.StubbedInvocationMatcher.answer(StubbedInvocationMatcher.java:42)
	at org.mockito.internal.handler.MockHandlerImpl.handle(MockHandlerImpl.java:103)
	at org.mockito.internal.handler.NullResultGuardian.handle(NullResultGuardian.java:29)
	at org.mockito.internal.handler.InvocationNotifierHandler.handle(InvocationNotifierHandler.java:34)
	at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.doIntercept(MockMethodInterceptor.java:82)
	at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.doIntercept(MockMethodInterceptor.java:56)
	at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor$DispatcherDefaultingToRealMethod.interceptAbstract(MockMethodInterceptor.java:161)
	at org.slf4j.Logger$MockitoMock$f4Org7xs.info(Unknown Source)
	at LoggerMockingTest.logger_info_can_be_called_without_exceptions(LoggerMockingTest.java:34)
        [...]
```

The following _works_, but feels much less elegant:

```java
        doAnswer( invocation -> {
            String format = invocation.getArgument( 0 );
            Object[] allArguments = invocation.getArguments();
[...]

PR #2664, opened 2022-06-02
Improve Varargs handling in AdditionalAnswers

big-andy-coates:
Fixes: #2644

Fixes issues around vararg handling for the following methods in `AdditionalAnswers`:
 * `returnsFirstArg`
 * `returnsSecondArg`
 * `returnsLastArg`
 * `returnsArgAt`
 * `answer`
 * `answerVoid`

These methods were not correctly handling varargs. For example,

```java
doAnswer(answerVoid(
      (VoidAnswer2<String, Object[]>) logger::info
   )).when(mock)
      .info(any(), (Object[]) any());

mock.info("Some message with {} {} {}", "three", "parameters", "");
```

Would previously have resulted in a `ClassCastException` being thrown from the `mock.info` call.  This was because the `answerVoid` method was not taking into account that the second parameter was a varargs parameter and was attempting to pass the second actual argument `"three"`, rather than the second _raw_ argument `["three", "parameters", ""]`.

The fix is to add checks for varargs methods and, when encountered, check to see if the required parameter type is the raw varargs type or the its component type.

## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/stubbing/answers/ReturnsArgumentAt.java
+++ b/src/main/java/org/mockito/internal/stubbing/answers/ReturnsArgumentAt.java
@@ -51,9 +51,7 @@ public class ReturnsArgumentAt implements Answer<Object>, ValidableAnswer, Seria
     }
 
     @Override
-    public Object answer(InvocationOnMock invocationOnMock) throws Throwable {
-        Invocation invocation = (Invocation) invocationOnMock;
-
+    public Object answer(InvocationOnMock invocation) throws Throwable {
         if (wantedArgIndexIsVarargAndSameTypeAsReturnType(invocation)) {
             // answer raw vararg array argument
             return invocation.getRawArguments()[invocation.getRawArguments().length - 1];
@@ -82,7 +80,7 @@ public class ReturnsArgumentAt implements Answer<Object>, ValidableAnswer, Seria
         return wantedArgumentPosition;
     }
 
-    private int inferWantedRawArgumentPosition(Invocation invocation) {
+    private int inferWantedRawArgumentPosition(InvocationOnMock invocation) {
         if (wantedArgumentPosition == LAST_ARGUMENT) {
             return invocation.getRawArguments().length - 1;
         }
@@ -121,7 +119,7 @@ public class ReturnsArgumentAt implements Answer<Object>, ValidableAnswer, Seria
         }
     }
 
-    private boolean wantedArgIndexIsVarargAndSameTypeAsReturnType(Invocation invocation) {
+    private boolean wantedArgIndexIsVarargAndSameTypeAsReturnType(InvocationOnMock invocation) {
         int rawArgumentPosition = inferWantedRawArgumentPosition(invocation);
         Method method = invocation.getMethod();
         Class<?>[] parameterTypes = method.getParameterTypes();
--- a/src/main/java/org/mockito/internal/stubbing/answers/ReturnsArgumentAt.java
+++ b/src/main/java/org/mockito/internal/stubbing/answers/ReturnsArgumentAt.java
@@ -51,25 +51,27 @@ public class ReturnsArgumentAt implements Answer<Object>, ValidableAnswer, Seria
     }
 
     @Override
-    public Object answer(InvocationOnMock invocation) throws Throwable {
+    public Object answer(InvocationOnMock invocationOnMock) throws Throwable {
+        Invocation invocation = (Invocation) invocationOnMock;
+
+        if (wantedArgIndexIsVarargAndSameTypeAsReturnType(invocation)) {
+            // answer raw vararg array argument
+            return invocation.getRawArguments()[invocation.getRawArguments().length - 1];
+        }
+
         int argumentPosition = inferWantedArgumentPosition(invocation);
         validateIndexWithinInvocationRange(invocation, argumentPosition);
-
-        if (wantedArgIndexIsVarargAndSameTypeAsReturnType(
-                invocation.getMethod(), argumentPosition)) {
-            // answer raw vararg array argument
-            return ((Invocation) invocation).getRawArguments()[argumentPosition];
-        }
 
         // answer expanded argument at wanted position
         return invocation.getArgument(argumentPosition);
     }
 
     @Override
-    public void validateFor(InvocationOnMock invocation) {
+    public void validateFor(InvocationOnMock invocationOnMock) {
+        Invocation invocation = (Invocation) invocationOnMock;
         int argumentPosition = inferWantedArgumentPosition(invocation);
-        validateIndexWithinInvocationRange(invocation, argumentPosition);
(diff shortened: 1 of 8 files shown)


big-andy-coates:
@TimvdLippe any chance of a quick review to check I'm going in the right direction with this before I flesh out all the testing and clean up the code?

src/main/java/org/mockito/internal/stubbing/answers/ReturnsArgumentAt.java
```diff
@@ -51,24 +51,26 @@ public ReturnsArgumentAt(int wantedArgumentPosition) {
...
-        int argumentPosition = inferWantedArgumentPosition(invocation);
-        validateIndexWithinInvocationRange(invocation, argumentPosition);
+    public Object answer(InvocationOnMock invocationOnMock) throws Throwable {
+        Invocation invocation = (Invocation) invocationOnMock;
 
-        if (wantedArgIndexIsVarargAndSameTypeAsReturnType(
-                invocation.getMethod(), argumentPosition)) {
+        if (wantedArgIndexIsVarargAndSameTypeAsReturnType(invocation)) {
             // answer raw vararg array argument
-            return ((Invocation) invocation).getRawArguments()[argumentPosition];
```
big-andy-coates, line 61: Note the existing cast of `InvocationOnMock` to `Invocation` to access the `getRawArguments` method. 

What do you think to exposing `getRawArguments` on the `InvocationOnMock` interface, and thereby removing the need for  lots of casts to get to the `Invocation`?

src/main/java/org/mockito/internal/stubbing/answers/ReturnsArgumentAt.java
```diff
@@ -51,24 +51,26 @@ public ReturnsArgumentAt(int wantedArgumentPosition) {
...
+
         // answer expanded argument at wanted position
         return invocation.getArgument(argumentPosition);
     }
 
     @Override
     public void validateFor(InvocationOnMock invocation) {
         int argumentPosition = inferWantedArgumentPosition(invocation);
-        validateIndexWithinInvocationRange(invocation, argumentPosition);
+        validateIndexWithinTheoreticalInvocationRange(invocation, argumentPosition);
```
big-andy-coates, line 73: When this is invoked, both `getArguments()` and `getRawArguments()` return the same, which is the number of _raw_ arguments.

I've renamed to `theoretical` because at this point in the code the actual parameters are not known, so we're only asking is its theoretically possible that the argument position is valid.

src/main/java/org/mockito/internal/stubbing/answers/ReturnsArgumentAt.java
TimvdLippe: If I remember correctly, we do these casts in quite a few place. If that's indeed the case, then yes let's introduce that method. If this is the only place, let's keep the cast here.

big-andy-coates: Agreed.  When making it public would you also want me to make the method return a copy of any array parameters, so that we're not exposing mutable internal state?

TimvdLippe: Yes makes sense.

big-andy-coates: Argh, I've noticed that we don't make copies when returning arrays from `getArgument`. So I'll stay consistent and not create a copy for `getRawArgument`.

big-andy-coates: Actually, I notice that `getArguments` does not do an array copy, so I'll stick with the same pattern when exposing `getRawArguments`.

big-andy-coates:
@TimvdLippe any chance of a review please?

TimvdLippe:
I have a very busy week this week. I will likely get to this early next week.

src/main/java/org/mockito/invocation/InvocationOnMock.java
```diff
@@ -32,6 +32,14 @@ public interface InvocationOnMock extends Serializable {
      */
     Method getMethod();
 
+    /**
+     * Returns unprocessed arguments whereas {@link #getArguments()} returns
+     * arguments already processed (e.g. varargs expended, etc.).
+     *
+     * @return unprocessed arguments, exactly as provided to this invocation.
```
TimvdLippe, line 39: Nit: add `@since 4.7.0` here

src/test/java/org/mockitousage/stubbing/StubbingWithAdditionalAnswersTest.java
```diff
@@ -52,10 +56,39 @@ public void can_return_arguments_of_invocation() throws Exception {
...
+    public void returns_arg_at_throws_on_out_of_range_var_args() throws Exception {
+        given(iMethods.mixedVarargsReturningString(eq(1), any())).will(returnsArgAt(3));
+
+        assertThatThrownBy(() -> iMethods.mixedVarargsReturningString(1, "a", "b"))
+                .isInstanceOf(MockitoException.class)
+                .hasMessageContaining("Invalid argument index");
+
+        assertThatThrownBy(
+                        () ->
+                                given(iMethods.mixedVarargsReturningStringArray(eq(1), any()))
```
TimvdLippe, line 88: Nit: put this in a separate test case. Since we are stubbing the same method twice, this could lead to strange error messages/unintended behavior when the first stubbing isn't properly cleaned up by Mockito. In that case, I think this test can pass, even though the underlying logic is wrong.

TimvdLippe (changes requested):
Solid work and great test coverage! Only two small nits, but other than this it is ready to go 🎉

TimvdLippe (approved):
Thanks!

big-andy-coates:
No worries. Good to give something back. I've used Mockito for YEARS!

perlun:
@big-andy-coates I found a similar semi-related issue I think... With 4.7.0, this now works correctly:

```java
        doAnswer(answerVoid((VoidAnswer2<String, Object[]>)
                logger::info
        )).when(mock)
                .info(any(), (Object[]) any());
```

_However_, if I try to use this form which is slightly more ergonomic:

```java
        doAnswer( answerVoid( (VoidAnswer2<String, Object[]>)
                logger::info
        ) ).when( mock )
                .info( any(), any( Object[].class ) );
```

...the mocking will fail to intercept these method calls. :thinking: No error when creating the mock, but it just won't detect the method calls as expected.

I _presume_ this has again something to do with the varargs parsing. `javac` resolves those two calls to the correct/same `logger.info()` overload, so so far so good.

Do you want me to create a separate issue/repro repo for this? I'm not even sure if it's something that can be (easily) fixed, I just happened to see it now while trying to "clean up our code"... :sweat_smile:

big-andy-coates:
Hey @perlun,  

If you could pull this related issue out into its own issue, then I'll take a look. Probably tomorrow.

PR #2682, opened 2022-06-11
Avoids starting mocks "half-way" if a superclass constructor is mocked but an unmocked subclass is initiated.

### LINKED ISSUE #2680: MockedConstruction has side effects even after close() (with Sub-/Superclass)
I am not 100% sure, but I think this is a bug. (Might be me just doing something wrong, but I don't see anything)

I prepared a small example [here](https://github.com/eNBeWe/mockito-bug) with most recent versions of mockito-inline (4.6.1) and junit5 (5.8.2). I prepared it on Ubuntu, using OpenJDK 11, but I also experienced this issue on other systems.

In short, I have two tests, each mocking constructors. The two constructors are related to each other, via inheritance. Both constructor mocks are auto-closed through try-with-resource.
If I first mock the subclass constructor, close it again and then mock the superclass constructor and create a subclass object, my test fails. If I do it the other way around, my tests work.

```
class SuperClass {}

class SubClass extends SuperClass {
	public boolean isOtherValue() {
		return true;
	}
}
```
```
@TestMethodOrder(OrderAnnotation.class)
class FailingTest {

	@Test
	@Order(1)
	void testThingsFirst() {
		try (MockedConstruction<SubClass> mocked = Mockito.mockConstruction(SubClass.class)) {
			System.out.println("Running testCreateStuff");
		}
	}

	@Test
	@Order(2)
	void testThingsSecond() {
		try (MockedConstruction<SuperClass> constructionMock = Mockito.mockConstruction(SuperClass.class)) {
			System.out.println("Running testOtherValue");
			SubClass testee = new SubClass();
			assertTrue(testee.isOtherValue());
		}
	}
}
```
```
[ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.793 s <<< FAILURE! - in de.enbewe.bug.mockito.FailingTest
[ERROR] de.enbewe.bug.mockito.FailingTest.testThingsSecond  Time elapsed: 0.039 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
```

So it seems the old, closed subclass-constructor is becoming active again, when I mock the superclass. 

I tried to find some more info / debug more, but I don't really understand enough of the internals of Mockito.

raphw:
Fixes #2680. Offers two implementations, one using a regular stack trace processor, one using StackWalker which is more accurate but requires a VM using Java 9 or newer.

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/StackWalkerChecker.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/StackWalkerChecker.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright (c) 2022 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.internal.creation.bytebuddy;
+
+import java.lang.reflect.Method;
+import java.util.Iterator;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+class StackWalkerChecker implements Predicate<Class<?>> {
+
+    private final Method stackWalkerGetInstance;
+    private final Method stackWalkerWalk;
+    private final Method stackWalkerStackFrameGetDeclaringClass;
+    private final Enum<?> stackWalkerOptionRetainClassReference;
+
+    StackWalkerChecker() throws Exception {
+        Class<?> stackWalker = Class.forName("java.lang.StackWalker");
+        @SuppressWarnings({"unchecked", "rawtypes"})
+        Class<? extends Enum> stackWalkerOption =
+                (Class<? extends Enum<?>>) Class.forName("java.lang.StackWalker$Option");
+        stackWalkerGetInstance = stackWalker.getMethod("getInstance", stackWalkerOption);
+        stackWalkerWalk = stackWalker.getMethod("walk", Function.class);
+        Class<?> stackWalkerStackFrame = Class.forName("java.lang.StackWalker$StackFrame");
+        stackWalkerStackFrameGetDeclaringClass =
+                stackWalkerStackFrame.getMethod("getDeclaringClass");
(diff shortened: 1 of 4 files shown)


src/main/java/org/mockito/internal/creation/bytebuddy/StackWalkerChecker.java
```diff
@@ -0,0 +1,91 @@
...
+        try {
+            Object walker =
+                    stackWalkerGetInstance.invoke(null, stackWalkerOptionRetainClassReference);
+            return (Boolean)
+                    stackWalkerWalk.invoke(
+                            walker,
+                            (Function<?, ?>)
+                                    stream -> {
+                                        Iterator<?> iterator = ((Stream<?>) stream).iterator();
+                                        while (iterator.hasNext()) {
```
TimvdLippe, line 55: Nit: this code is quite complex and the indentation is showing. Should we extract this into a method to make it more readable?

TimvdLippe (commented):
Nice fix! Do you know what the performance impact on this is? Can we cache the result (I am assuming not)?

Also, can we cover a bit more of the code or is this mostly it?

src/main/java/org/mockito/internal/creation/bytebuddy/StackWalkerChecker.java
raphw: I thought its already a small class that's why I did not factor it out. Unfortunately, the stack trace cannot be cached as we do not know the call site.

TimvdLippe: Ack

TimvdLippe:
@raphw shall we merge this one?

raphw:
Yes!

Issue #2692, opened 2022-06-24
Misleading error message when mocking and a class (of a parameter) is not found


dstango:
Situation:
we're using a dependency that contains the following code (abbreviated):

```
import javax.ws.rs.core.MultivaluedMap;

public interface RestClient {
	RestResponse getRequest(String path, String body, MultivaluedMap<String, String> queryParams);
}
```

When trying to mock this interface like this:
```
		Mockito.mock(RestClient.class);
```

the following misleading error message is created:

```
org.mockito.exceptions.base.MockitoException: 
Mockito cannot mock this class: interface de.persona.dstools.network.rest.RestClient.

Mockito can only mock non-private & non-final classes.
If you're not sure why you're getting this error, please report to the mailing list.


Java               : 11
JVM vendor name    : Azul Systems, Inc.
JVM vendor version : 11.0.14.1+1-LTS
JVM name           : OpenJDK 64-Bit Server VM
JVM version        : 11.0.14.1+1-LTS
JVM info           : mixed mode
OS name            : Windows 10
OS version         : 10.0


Underlying exception : java.lang.IllegalArgumentException: Could not create type
	at de.persona.salesforce.fetcher.SalesforceToAzureTransferTest.setUp(SalesforceToAzureTransferTest.java:41)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:725)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:149)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptLifecycleMethod(TimeoutExtension.java:126)
[...]

PR #2693, opened 2022-06-25
Makes error message less misleading and points to github for help. Issue #2692

dstango:
Small changes to some error messages, that might help users not to search in the wrong place for the cause of an issue.


TimvdLippe (approved):
Nice!

PR #2700, opened 2022-07-02
Allow @InjectMock into static/final fields (fixes #1417)

### LINKED ISSUE #1417: Allow @InjectMock into static/final fields

HonoluluHenk:
Allow @InjectMock into static/final fields (fixes #1417)

This pr is a rebase on current main and partial reimplementation of the old PR: #1418

### DIFF
--- a/src/main/java/org/mockito/InjectUnsafe.java
+++ b/src/main/java/org/mockito/InjectUnsafe.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright (c) 2016 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito;
+
+import java.lang.annotation.*;
+
+/**
+ * Change the behavior of {@link InjectMocks}: allow injection of static/final fields that are normally skipped.
+ *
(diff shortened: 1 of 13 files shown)


HonoluluHenk:
In case this PR is still relevant for the Mockito team:

I noticed, that the injection into static fields would require resetting the static field to the old value before the test.
Is is to ensure proper behavior of other, unrelated tests.

TimvdLippe:
Unfortunately, I think this PR falls in the same bucket as https://github.com/mockito/mockito/pull/1936#issuecomment-636465679

HonoluluHenk:
Thanks for your reply.

Keep up the good work, I really love Mockito!

Issue #2589, opened 2022-03-08
Switch the default mockmaker to the inline mockmaker on JDK 17+


TimvdLippe:
All right, we have been talking about this for a while so let's consolidate discussions. TLDR: more and more use cases are broken (by default) with Mockito and JDK 17. That's because the subclass mockmaker runs into fundamental limitations on JDK 17, but the inline mockmaker works as expected.

There are a lot of issues related to this:
- https://github.com/mockito/mockito/issues/2587
- https://github.com/mockito/mockito/issues/2560
- https://github.com/mockito/mockito/issues/2573
- https://github.com/mockito/mockito/issues/2561 (suspected)
- https://github.com/mockito/mockito/issues/2521 (suspected)
- https://github.com/mockito/mockito/issues/2449 (suspected)

At this point, the rate of issues created is increasing, as more and more of our users are adopting JDK 17+

We have historically talked about this as well: https://github.com/mockito/mockito/issues/1728

When chatting with @raphw about this issue, there were a couple of options that we have available:
1. Keep on using the subclass mockmaker by default on all JDK's and suggest users to use `mockito-inline` whenever appropriate
2. Switch the default to the inline mockmaker for all JDK's and put the subclass mockmaker in an artifact such as `mockito-subclass` (legacy)
3. Detect the current JDK version and make the inline mockmaker the default on JDK 17+, but keep the subclass mockmaker the default for older JDK versions

In my opinion, we should opt for least amount of surprise. E.g. Mockito should "just" work out-of-the-box for the majority of use cases. Therefore, if Mockito (by default) does not work nicely with JDK 17 in more and more cases, I consider that a bug.

That said, I don't think it is a good choice to force all existing users that are on JDK 16 and below to update their codebase wholesale to the new mockmaker. Inheritently, the new mockmaker works slightly different and forcing the update onto users doesn't feel like a nice solution. In the end, it works just fine on JDK 16 on below, so why change it?

Therefore, my proposed way forward is:
1. We publish Mockito 5 which switches the default mockmaker to the inline mockmaker on JDK 17+
2. We update the JavaDoc of the subclass mockmaker to make users aware it is only intended to be used for specific use cases

In terms of timeline, I think we should collective user feedback from the community first. This issue seems like the best place to do so. Once we are confident that we would be able to switch the default, we could publish Mockito 5. For now, what if we take June 2022 as a temporary date to make such a change, unless we receive strong/valid user feedback that should block such a change?

CC @mockito/developers

[...]

PR #2701, opened 2022-07-04
Fixes #2626 : Introduce MockSettings.mockMaker

### LINKED ISSUE #2626: Introduce option to disable inline-mock-maker for a specific instance
In order to ease the migration to the inline mock maker in big projects, it would nice if the mock make can be changed for individual mocks. This would allow projects to change the default mock maker to `inline-mock-maker` without immediately fixing all edge cases where the new mock maker doesn't work.

Such option could be provided over a new method on the `MockSettings` class.

JojOatXGME:
I implemented a prototype to fix #2626 as suggested in https://github.com/mockito/mockito/issues/2589#issuecomment-1105163928. The basics doesn't seem that difficult, but I could have missed some edge cases. I mostly skipped documentation and tests for now. Furthermore, I ignored that `ByteBuddyMockMaker` and `InlineByteBuddyMockMaker` are not exposed by the API, so we still have to adjust the API to allow setting the mock maker without accessing these classes.

Personally, I don't need this functionality right now. However, if you think it would be a worthwhile addition, I can still try to help with that topic.

This pull request allows users of Mockito to choose different MockMakers per mock. This feature could ease the migration from one MockMaker to another in a big project.

```
SomeClass subclassMock = Mockito.mock(SomeClass.class, Mockito.withSettings()
    .mockMaker(AwesomeMockMaker.class));
```

~~PS: The Gradle task `spotlessApply` failed on my system with *“java.lang.NoClassDefFoundError: Could not initialize class com.google.googlejavaformat.java.java14.Java14InputAstVisitor”*. Probably because it couldn't find `BindingPatternTree`. Not sure if it is a local problem or something with the project configuration.~~

## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/configuration/SpyAnnotationEngine.java
+++ b/src/main/java/org/mockito/internal/configuration/SpyAnnotationEngine.java
@@ -83,6 +83,7 @@ public class SpyAnnotationEngine implements AnnotationEngine {
     }
 
     private static Object spyInstance(Field field, Object instance) {
+        // TODO: Add mockMaker option for @Spy annotation (#2740)
         return Mockito.mock(
                 instance.getClass(),
                 withSettings()
@@ -93,6 +94,7 @@ public class SpyAnnotationEngine implements AnnotationEngine {
 
     private static Object spyNewInstance(Object testInstance, Field field)
             throws InstantiationException, IllegalAccessException, InvocationTargetException {
+        // TODO: Add mockMaker option for @Spy annotation (#2740)
         MockSettings settings =
                 withSettings().defaultAnswer(CALLS_REAL_METHODS).name(field.getName());
         Class<?> type = field.getType();
--- a/src/main/java/org/mockito/internal/configuration/plugins/DefaultMockitoPlugins.java
+++ b/src/main/java/org/mockito/internal/configuration/plugins/DefaultMockitoPlugins.java
@@ -5,7 +5,11 @@
 package org.mockito.internal.configuration.plugins;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
+
+import org.mockito.MockMakers;
 import org.mockito.plugins.AnnotationEngine;
 import org.mockito.plugins.DoNotMockEnforcer;
 import org.mockito.plugins.InstantiatorProvider2;
@@ -16,11 +20,13 @@ import org.mockito.plugins.MockitoPlugins;
 import org.mockito.plugins.PluginSwitch;
 import org.mockito.plugins.StackTraceCleanerProvider;
 
-class DefaultMockitoPlugins implements MockitoPlugins {
+public class DefaultMockitoPlugins implements MockitoPlugins {
 
     private static final Map<String, String> DEFAULT_PLUGINS = new HashMap<>();
-    static final String INLINE_ALIAS = "mock-maker-inline";
-    static final String PROXY_ALIAS = "mock-maker-proxy";
+    static final String INLINE_ALIAS = MockMakers.INLINE;
+    static final String PROXY_ALIAS = MockMakers.PROXY;
+    static final String SUBCLASS_ALIAS = MockMakers.SUBCLASS;
+    public static final Set<String> MOCK_MAKER_ALIASES = new HashSet<>();
     static final String MODULE_ALIAS = "member-accessor-module";
 
     static {
@@ -42,6 +48,8 @@ class DefaultMockitoPlugins implements MockitoPlugins {
                 INLINE_ALIAS, "org.mockito.internal.creation.bytebuddy.InlineByteBuddyMockMaker");
         DEFAULT_PLUGINS.put(PROXY_ALIAS, "org.mockito.internal.creation.proxy.ProxyMockMaker");
         DEFAULT_PLUGINS.put(
+                SUBCLASS_ALIAS, "org.mockito.internal.creation.bytebuddy.ByteBuddyMockMaker");
+        DEFAULT_PLUGINS.put(
                 MockitoLogger.class.getName(), "org.mockito.internal.util.ConsoleMockitoLogger");
         DEFAULT_PLUGINS.put(
                 MemberAccessor.class.getName(),
@@ -51,6 +59,10 @@ class DefaultMockitoPlugins implements MockitoPlugins {
         DEFAULT_PLUGINS.put(
                 DoNotMockEnforcer.class.getName(),
                 "org.mockito.internal.configuration.DefaultDoNotMockEnforcer");
+
+        MOCK_MAKER_ALIASES.add(INLINE_ALIAS);
+        MOCK_MAKER_ALIASES.add(PROXY_ALIAS);
+        MOCK_MAKER_ALIASES.add(SUBCLASS_ALIAS);
     }
 
     @Override
@@ -59,7 +71,7 @@ class DefaultMockitoPlugins implements MockitoPlugins {
         return create(pluginType, className);
     }
 
-    String getDefaultPluginClass(String classOrAlias) {
+    public static String getDefaultPluginClass(String classOrAlias) {
         return DEFAULT_PLUGINS.get(classOrAlias);
     }
 
--- a/src/main/java/org/mockito/internal/stubbing/defaultanswers/RetrieveGenericsForDefaultAnswers.java
+++ b/src/main/java/org/mockito/internal/stubbing/defaultanswers/RetrieveGenericsForDefaultAnswers.java
@@ -8,15 +8,12 @@ import java.lang.reflect.GenericArrayType;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
 
-import org.mockito.internal.MockitoCore;
 import org.mockito.internal.util.MockUtil;
 import org.mockito.internal.util.reflection.GenericMetadataSupport;
 import org.mockito.invocation.InvocationOnMock;
 import org.mockito.mock.MockCreationSettings;
 
 final class RetrieveGenericsForDefaultAnswers {
-
-    private static final MockitoCore MOCKITO_CORE = new MockitoCore();
 
     static Object returnTypeForMockWithCorrectGenerics(
             InvocationOnMock invocation, AnswerCallback answerCallback) {
@@ -38,7 +35,9 @@ final class RetrieveGenericsForDefaultAnswers {
         }
 
         if (type != null) {
-            if (!MOCKITO_CORE.isTypeMockable(type)) {
+            final MockCreationSettings<?> mockSettings =
+                    MockUtil.getMockSettings(invocation.getMock());
+            if (!MockUtil.typeMockabilityOf(type, mockSettings.getMockMaker()).mockable()) {
                 return null;
             }
 
--- a/src/main/java/org/mockito/internal/util/MockUtil.java
+++ b/src/main/java/org/mockito/internal/util/MockUtil.java
@@ -7,6 +7,7 @@ package org.mockito.internal.util;
 import org.mockito.MockedConstruction;
 import org.mockito.Mockito;
 import org.mockito.exceptions.misusing.NotAMockException;
+import org.mockito.internal.configuration.plugins.DefaultMockitoPlugins;
 import org.mockito.internal.configuration.plugins.Plugins;
 import org.mockito.internal.creation.settings.CreationSettings;
 import org.mockito.internal.stubbing.InvocationContainerImpl;
@@ -18,6 +19,9 @@ import org.mockito.plugins.MockMaker;
 import org.mockito.plugins.MockMaker.TypeMockability;
 import org.mockito.plugins.MockResolver;
 
+import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 
 import static org.mockito.internal.handler.MockHandlerFactory.createMockHandler;
@@ -25,15 +29,57 @@ import static org.mockito.internal.handler.MockHandlerFactory.createMockHandler;
 @SuppressWarnings("unchecked")
 public class MockUtil {
 
-    private static final MockMaker mockMaker = Plugins.getMockMaker();
+    private static final MockMaker defaultMockMaker = Plugins.getMockMaker();
+    private static final Map<Class<? extends MockMaker>, MockMaker> mockMakers =
+            new ConcurrentHashMap<>(
+                    Collections.singletonMap(defaultMockMaker.getClass(), defaultMockMaker));
 
     private MockUtil() {}
 
-    public static TypeMockability typeMockabilityOf(Class<?> type) {
-        return mockMaker.isTypeMockable(type);
+    private static MockMaker getMockMaker(String mockMaker) {
+        if (mockMaker == null) {
+            return defaultMockMaker;
+        }
+
+        String typeName;
+        if (DefaultMockitoPlugins.MOCK_MAKER_ALIASES.contains(mockMaker)) {
+            typeName = DefaultMockitoPlugins.getDefaultPluginClass(mockMaker);
+        } else {
+            typeName = mockMaker;
+        }
+
+        Class<? extends MockMaker> type;
+        // Using the context class loader because PluginInitializer.loadImpl is using it as well.
+        // Personally, I am suspicious whether the context class loader is a good choice in either
+        // of these cases.
+        ClassLoader loader = Thread.currentThread().getContextClassLoader();
+        if (loader == null) {
+            loader = ClassLoader.getSystemClassLoader();
+        }
+        try {
+            type = loader.loadClass(typeName).asSubclass(MockMaker.class);
+        } catch (Exception e) {
+            throw new IllegalStateException("Failed to load MockMaker: " + mockMaker, e);
+        }
+
+        return mockMakers.computeIfAbsent(
+                type,
+                t -> {
+                    try {
+                        return t.getDeclaredConstructor().newInstance();
+                    } catch (Exception e) {
+                        throw new IllegalStateException(
+                                "Failed to construct MockMaker: " + t.getName(), e);
+                    }
+                });
+    }
+
+    public static TypeMockability typeMockabilityOf(Class<?> type, String mockMaker) {
+        return getMockMaker(mockMaker).isTypeMockable(type);
     }
 
     public static <T> T createMock(MockCreationSettings<T> settings) {
+        MockMaker mockMaker = getMockMaker(settings.getMockMaker());
         MockHandler mockHandler = createMockHandler(settings);
 
         Object spiedInstance = settings.getSpiedInstance();
@@ -62,17 +108,11 @@ public class MockUtil {
         MockHandler newHandler = createMockHandler(settings);
 
         mock = resolve(mock);
-        mockMaker.resetMock(mock, newHandler, settings);
+        getMockMaker(settings.getMockMaker()).resetMock(mock, newHandler, settings);
     }
 
     public static MockHandler<?> getMockHandler(Object mock) {
-        if (mock == null) {
-            throw new NotAMockException("Argument should be a mock, but is null!");
-        }
-
-        mock = resolve(mock);
-
-        MockHandler handler = mockMaker.getHandler(mock);
+        MockHandler handler = getMockHandlerOrNull(mock);
         if (handler != null) {
             return handler;
         } else {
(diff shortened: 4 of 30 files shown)


JojOatXGME:
I was able to run `spotlessApply` from the terminal (PowerShell on Windows). I previously tried to run the task from IntelliJ which for some reason was using a development build of JDK 14. 😅

src/main/java/org/mockito/internal/util/MockUtil.java
```diff
@@ -18,22 +18,55 @@
...
+        mockMakers = Collections.synchronizedMap(new WeakHashMap<>());
+        mockMakers.put(defaultMockMaker.getClass(), defaultMockMaker);
+    }
 
     private MockUtil() {}
 
+    private static MockMaker getMockMaker(Class<? extends MockMaker> type) {
+        if (type == null) {
+            return defaultMockMaker;
+        } else {
```
TimvdLippe, line 44: Nit: remove the else so that the return is on the same level as the if

src/main/java/org/mockito/internal/util/MockUtil.java
```diff
@@ -104,10 +131,24 @@ public static boolean isMock(Object mock) {
         if (mock == null) {
             return false;
         }
+        return getMockHandler0(mock) != null;
+    }
+
+    private static MockHandler<?> getMockHandler0(Object mock) {
```
TimvdLippe, line 137: Can we use a more descriptive name here? What about `maybeGetMockHandler`?

src/test/java/org/mockito/CustomMockMakerTest.java
```diff
@@ -0,0 +1,50 @@
...
+package org.mockito;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.mockito.internal.creation.bytebuddy.ByteBuddyMockMaker;
+import org.mockito.internal.creation.bytebuddy.InlineByteBuddyMockMaker;
+
+import java.util.Arrays;
+
+public final class CustomMockMakerTest {
```
TimvdLippe, line 14: Can we also add support for `@Mock` and add a corresponding test as well?

Additionally, we are missing a bit of code coverage. Would it be possible to write some additional regression tests for these?

src/main/java/org/mockito/internal/stubbing/defaultanswers/RetrieveGenericsForDefaultAnswers.java
```diff
@@ -38,6 +38,7 @@ static Object returnTypeForMockWithCorrectGenerics(
         }
 
         if (type != null) {
+            // TODO: Should we use the mockMaker of the mock?
             if (!MOCKITO_CORE.isTypeMockable(type)) {
```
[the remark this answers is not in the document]
JojOatXGME: Done.

src/main/java/org/mockito/internal/util/MockUtil.java
JojOatXGME: Done. For me, it feels easier to read with the else, but I guess that might just be because I am used to that style.

src/main/java/org/mockito/internal/util/MockUtil.java
JojOatXGME: I renamed it to `getMockHandlerOrNull`.

src/test/java/org/mockito/CustomMockMakerTest.java
JojOatXGME: I added an option for `@Mock` and a corresponding test. I did not add an option for `@Spy`, through.

src/main/java/org/mockito/internal/configuration/SpyAnnotationEngine.java
```diff
@@ -83,6 +83,7 @@ public AutoCloseable process(Class<?> context, Object testInstance) {
     }
 
     private static Object spyInstance(Field field, Object instance) {
+        // TODO: Should we also add an option to @Spy?
```
JojOatXGME, line 86: I did not add an option for now, because it would be the first option on that annotation.

src/main/java/org/mockito/internal/util/MockUtil.java
```diff
@@ -29,35 +29,26 @@
...
     private static final MockMaker defaultMockMaker = Plugins.getMockMaker();
-    private static final Map<Class<? extends MockMaker>, MockMaker> mockMakers;
-
-    static {
-        mockMakers = Collections.synchronizedMap(new WeakHashMap<>());
-        mockMakers.put(defaultMockMaker.getClass(), defaultMockMaker);
-    }
+    private static final Map<Class<? extends MockMaker>, MockMaker> mockMakers =
+            new ConcurrentHashMap<>(
+                    Collections.singletonMap(defaultMockMaker.getClass(), defaultMockMaker));
```
JojOatXGME, line 34: I removed that `WeakHashMap` because I realized that it wouldn't remove anything anyway. (The class cannot be garbage collected when we are still holding the instance as the value.) However, I think not being able to unload a class loader because we have a reference to the mock maker is an acceptable limitation. I could not think of a realistic solution. (We could dynamically generate classes in the class loader which belongs to the mock maker to hold the reference, but I think that would be overkill. 😄)

src/main/java/org/mockito/internal/util/reflection/FieldInitializer.java
```diff
@@ -259,8 +259,11 @@ public int compare(Constructor<?> constructorA, Constructor<?> constructorB) {
                     private int countMockableParams(Constructor<?> constructor) {
                         int constructorMockableParamsSize = 0;
                         for (Class<?> aClass : constructor.getParameterTypes()) {
-                            // TODO: Should we somehow use the mockMaker from the context?
-                            if (MockUtil.typeMockabilityOf(aClass).mockable()) {
+                            // TODO: I don't really understand the reason for this check here.
+                            //  It looks like we are not actually creating any mocks for these
+                            //  parameters, but might only use mocks which are already created
+                            //  and stored in some fields.
+                            if (MockUtil.typeMockabilityOf(aClass, null).mockable()) {
```
JojOatXGME, line 266: Not sure how to handle this.

JojOatXGME (commented):
Here are my changes. I hope I am able to respond faster the next time. We also have to think about the API. Should the user be able to provide arbitrary mock makers (as right now in the PR), or should we make it more specific to the upcoming migration.

**Generic**

```
withSettings().mockMaker(SomeMockMaker.class)
```

In this case, we need to expose the internal mock makers somehow, or provide separate methods in the settings interface.

**Specific**

```
withSettings().useSubclassMockMaker()
```

Keeping the interface more specific to the upcoming migration would make it easier to deprecate and remove the feature in the future if it causes maintenance problems for some reason.

src/test/java/org/mockito/CustomMockMakerTest.java
```diff
@@ -0,0 +1,189 @@
...
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThrows;
+import static org.mockito.Mockito.withSettings;
+
+import org.junit.Test;
+import org.mockito.exceptions.base.MockitoException;
+import org.mockito.exceptions.verification.SmartNullPointerException;
+import org.mockito.internal.creation.bytebuddy.ByteBuddyMockMaker;
+import org.mockito.internal.creation.bytebuddy.InlineByteBuddyMockMaker;
+import org.mockito.internal.creation.bytebuddy.SubclassByteBuddyMockMaker;
```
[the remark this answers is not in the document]
JojOatXGME: > Unfortunately, we can't expose these classes in Mockito.java, since we don't know which mockmaker we should load.

Maybe we could use static methods to make the class objects available without loading or exposing the class?

```java
public final class MockMakers {
  private MockMakers() {} // Cannot be instantiated

  public Class<? extends MockMaker> inline() { return InlineByteBuddyMockMaker.class; };
  public Class<? extends MockMaker> subclass() { return SubclassByteBuddyMockMaker.class; };
  // ...
}
```

In contrast to constants, I think this should only load the classes for which you call the method. I am unsure why it is important to not load the classes, though.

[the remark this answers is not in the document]
TimvdLippe: I wonder if we should instead use the String aliases for the mockmakers. E.g. you can set an enum here, of which the string value will correspond with the alias defined in https://github.com/mockito/mockito/blob/de63c24352d035dff77e9ef8e17481d9d85e87d9/src/main/java/org/mockito/internal/configuration/plugins/DefaultMockitoPlugins.java#L22

Then we would need to introduce an alias for the subclass mockmaker as well, such as `mockito-subclass-maker`?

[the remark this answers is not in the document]
JojOatXGME: I guess strings or providing two separate options are our only possibility. At least I cannot think of anything else. I am not sure where you want me to use an enum. I would just use a class with string constants instead.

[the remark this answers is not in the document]
TimvdLippe: Yeah agreed. The easiest solution here is to allow users to specify the String fully qualified name of the mockmaker. Since we don't want to expose the internal mockmaker locations as public API, for Mockito itself we would need to support the alias for the `mock-maker-inline` and `mock-maker-subclass`.

And disregard the comment regarding enums. That wouldn't work for community mockmakers, which would be a major benefit for them. If we lock the possible values to just the enum, that wouldn't work for them. Good point 👍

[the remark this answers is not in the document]
TimvdLippe: @JojOatXGME Shall we go for the String solution including the default alias for `mock-maker-inline` and a new alias `mock-maker-subclass`?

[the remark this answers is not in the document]
JojOatXGME: At least, I don't have a good idea for an alternative. I adjusted the implementation. However, I duplicated the code for the aliases for the time being. I was not sure how to share the code between `MockUtil.getMockMaker` and `PluginInitializer.loadImpl` without making bigger changes.

src/main/java/org/mockito/internal/util/MockUtil.java
```diff
@@ -4,10 +4,14 @@
...
 package org.mockito.internal.util;
 
+import org.mockito.MockMakers;
 import org.mockito.MockedConstruction;
 import org.mockito.Mockito;
 import org.mockito.exceptions.misusing.NotAMockException;
 import org.mockito.internal.configuration.plugins.Plugins;
+import org.mockito.internal.creation.bytebuddy.ByteBuddyMockMaker;
+import org.mockito.internal.creation.bytebuddy.InlineByteBuddyMockMaker;
+import org.mockito.internal.creation.proxy.ProxyMockMaker;
```
TimvdLippe, line 14: Since we do special loading in these mockmakers, we can't load their classes until we are certain that the user wants them. Therefore, we will need to dynamically load these classes using reflection (based on the names). In other words, similar to what `DefaultMockitoPlugins` does, where we also refer to them as fully qualified names rather than the type class.

The `getMockMaker` implementation down below would then becomes something along the lines of:

```java
        Class<? extends MockMaker> type;
        switch (mockMaker) {
            case MockMakers.INLINE:
            case MockMakers.PROXY:
            case MockMakers.SUBCLASS:
                type = defaultMockitoPlugins.getDefaultPluginClass(mockMaker);
                break;
            default:
                {
                    ClassLoader loader = Thread.currentThread().getContextClassLoader();
                    if (loader == null) {
                        loader = ClassLoader.getSystemClassLoader();
                    }
                    try {
                        type = loader.loadClass(mockMaker).asSubclass(MockMaker.class);
                    } catch (Exception e) {
                        throw new IllegalStateException(
                                "Failed to load MockMaker: " + mockMaker, e);
                    }
                    break;
                }
        }
```

(Store the `defaultMockitoPlugins` in a field as a constant and make `getDefaultPluginClass` public, it's an internal class anyways so not part of our API)

If you feel adventurous, it would be good to add an integration test to ensure this remains so. I think it would be good to copy https://github.com/mockito/mockito/blob/main/src/test/java/org/mockitointegration/NoByteCodeDependenciesTest.java and adjust it to ensure that we do not load the mockmakers.

E.g. load both `Mockito.class` and `Matchers.class` and assert that it does **not** load any of the mockmakers that we ship (`.without("org.mockito.internal.creation.bytebuddy", "org.mockito.internal.creation.proxy")` on the ` ClassLoaders.excludingClassLoader()`)

src/main/java/org/mockito/internal/configuration/SpyAnnotationEngine.java
TimvdLippe, line 86: Yes, but let's file a separate GitHub issue for that and link it here: https://github.com/mockito/mockito/issues/2740

src/main/java/org/mockito/plugins/MockMaker.java
```diff
@@ -45,6 +46,15 @@
  * <p>Note that if several <code>mockito-extensions/org.mockito.plugins.MockMaker</code> files exists in the classpath
  * Mockito will only use the first returned by the standard {@link ClassLoader#getResource} mechanism.
  *
+ * <h3>Using the MockSettings of individual mocks</h3>
+ *
+ * <p>If you want to use your {@code AwesomeMockMaker} only for a specific mock,
+ * you can specify it using {@link MockSettings#mockMaker(String)}.</p>
+ * <pre>
+ *     Object mock = Mockito.mock(Object.class, Mockito.withSettings()
+ *             .mockMaker(AwesomeMockMaker.class));
```
TimvdLippe, line 55: Nit: this should be updated to use the String. Let's include two examples here:

```java
 * <pre>
 *     // Use a mockmaker shipping with Mockito
 *     Object mock = Mockito.mock(Object.class, Mockito.withSettings()
 *             .mockMaker(MockMakers.INLINE));
 *     // Or load a mockmaker using a fully qualified class name
 *     Object mock = Mockito.mock(Object.class, Mockito.withSettings()
 *             .mockMaker("com.company.mockito.MyAwesomeMockMaker"));
 * </pre>
```

TimvdLippe (changes requested):
Nice, yes this is looking a lot better. Very happy about this design choice, this looks natural for loading custom mockmakers as well. I like your solution of `MockMakers.java`, since these strings were kinda magic at first and now we have a proper dynamic API for them as well.

Some last minor changes, but the overall design LGTM now.

We should also include a nice mention in our main Javadoc. Can you add section 53 to `Mockito.java`? Something along the lines of:

```
 * <h3 id="53">53. <a class="meaningful_link" href="#mock_annotation_mockmaker" name="mock_annotation_mockmaker">
 *  New mockmaker attribute for @Mock annotation and <code>MockSettings.mockMaker(String)</code> methods (Since 4.6.0)</a></h3>
 *
 * You can now customize the mockmaker for a single mock, either using `@Mock` annotation mockmaker attribute or
 * using `MockSettings.mockMaker(String)`. This can be useful for migrating to a different mockmakers
 * (which we are planning to do in Mockito 5 with https://github.com/mockito/mockito/issues/2589).
 *
 * <pre class="code"><code class="java">
 *   &#064;Mock(mockMaker = MockMakers.INLINE)
 *   Foo mock;
 *   // using MockSettings.withSettings()
 *   Foo mock = Mockito.mock(Foo.class, withSettings().mockMaker(MockMakers.SUBCLASS));
 * </code></pre>
```

Also, make sure you add a link to section 53 at the top of the Javadoc section.

src/main/java/org/mockito/internal/configuration/SpyAnnotationEngine.java
JojOatXGME: Ok. I have added the issue number to the comment.

src/main/java/org/mockito/internal/util/MockUtil.java
JojOatXGME: I adjusted the implementation, but I have not yet implemented a test.

I think the test you have suggested might not be sufficient. At least the JVMs which I am using do not load classes used by methods before the method is called. This means the test you have suggested succeeds with my previous implementation, at least on the JVM I am using. I think I would actually have to create a mock with another mock maker, while the built-in mock makers are not accessible. Maybe I will look at this later this week.

src/main/java/org/mockito/plugins/MockMaker.java
JojOatXGME: Fixed.

src/main/java/org/mockito/internal/configuration/plugins/DefaultMockitoPlugins.java
```diff
@@ -59,7 +71,8 @@ public <T> T getDefaultPlugin(Class<T> pluginType) {
         return create(pluginType, className);
     }
 
-    String getDefaultPluginClass(String classOrAlias) {
+    public String getDefaultPluginClass(String classOrAlias) {
+        // TODO: #2701 We could make this method static?
         return DEFAULT_PLUGINS.get(classOrAlias);
     }
```
JojOatXGME, line 77: @TimvdLippe Are there any objections for making this method static?

src/main/java/org/mockito/internal/util/MockUtil.java
```diff
@@ -18,22 +19,68 @@
...
+            typeName = defaultMockitoPlugins.getDefaultPluginClass(mockMaker);
+        } else {
+            typeName = mockMaker;
+        }
+
+        Class<? extends MockMaker> type;
+        // TODO: #2701 I doubt that it is a good idea to use the context class loader here,
+        //  but PluginInitializer.loadImpl is using it as well,
+        //  but DefaultMockitoPlugins.create is explicitly not using it.
+        ClassLoader loader = Thread.currentThread().getContextClassLoader();
```
JojOatXGME, line 56: Using `Thread.currentThread().getContextClassLoader()` could cause problems with our built-in mock makers (see pre-existing comment in [`DefaultMockitoPlugins.create`](https://github.com/mockito/mockito/blob/530558ae3c721cb310f3410a66ff4061414cf836/src/main/java/org/mockito/internal/configuration/plugins/DefaultMockitoPlugins.java#L80-L83)). It might also cause a memory leak if someone is loading a lot of mock makers from temporary class loaders.

However, `PluginInitializer.loadImpl` is also using the context class loader, which is even more problematic in my opinion. The result of `PluginInitializer.loadImpl` is stored in static fields (i.e. `MockUtil.defaultMockMaker` and `Plugins.registry.mockMaker`), so Mockito considers the context class loader only when these static fields are first initialized. This probably means that the test which is executed first determines the mock maker which is used as the default, which seems quite arbitrary to me.

@TimvdLippe, do you know if there is any history about the usage of context class loaders with Mockito? Or do you have any opinion on that?

JojOatXGME (commented):
I made some adjustments. I want to the Javadoc later this week. There are also a few remaining TODOs I would like to discuss before merging this pull request.

src/main/java/org/mockito/internal/util/MockUtil.java
TimvdLippe: No I am not ware of any history. Usually, if there are issues with ClassLoaders, we write specific integration tests and then we fix them. In this case, since I am not aware of any problems, I think it's fine to keep it as-is.

JojOatXGME:
@TimvdLippe, I have finished my work. If you have no further suggestions, I would replace my `${NEXT_VERSION}` placeholder with `4.8.0` (or should it be `4.7.1`?) and rebase everything into one commit. Or what is the process?

**FYI:** I noticed that the following test is flaky, even on the `main` branch.

https://github.com/mockito/mockito/blob/530558ae3c721cb310f3410a66ff4061414cf836/src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsDeepStubsConcurrentTest.java#L17-L24

It sometimes throws an `ArrayIndexOutOfBoundsException` in `InvocationContainerImpl.getStubbingsAscending`. The exception seems to be caused by a race condition. While this thread is reading `InvocationContainerImpl.stubbed`, another thread seems to add or remove items from the list.

src/main/java/org/mockito/internal/util/MockUtil.java
TimvdLippe: Hm good point. In that case, let's punt it for now.

TimvdLippe:
@JojOatXGME Great work! The next version will be 4.8.0, so let's use that in the documentation 😄 

Regarding the flaky test: if you have time to fix it, please go ahead. But for now, since we retry our tests, I think we can keep it as-is for now. Not ideal, I know...

JojOatXGME:
@TimvdLippe I replaced my version placeholder and rebased and squashed my commits. I completely forgot [your suggestion](https://github.com/mockito/mockito/pull/2701#pullrequestreview-1087903980) while writing the Javadoc, so if you have any final remarks or prefer another wording, I am happy to make adjustments. Otherwise, I think there is nothing left to be done for this PR.

src/test/java/org/mockitointegration/DeferMockMakersClassLoadingTest.java
```diff
@@ -0,0 +1,96 @@
...
+import org.mockito.internal.creation.bytebuddy.ByteBuddyMockMaker;
+import org.mockito.internal.creation.bytebuddy.InlineByteBuddyMockMaker;
+import org.mockito.internal.creation.bytebuddy.SubclassByteBuddyMockMaker;
+import org.mockito.internal.creation.proxy.ProxyMockMaker;
+import org.mockito.invocation.MockHandler;
+import org.mockito.mock.MockCreationSettings;
+import org.mockito.plugins.MockMaker;
+import org.mockitoutil.ClassLoaders;
+
+public class DeferMockMakersClassLoadingTest {
```
TimvdLippe, line 25: Nice test!

TimvdLippe (approved):
Woohoo! Great feature 😄 

Running the final checks on CI and then I will merge.

TimvdLippe:
Released as part of 4.8.0 🎉

PR #2723, opened 2022-08-05
Fixes #2720: Use StackWalker on Java 9+ to create Locations

### LINKED ISSUE #2720: LocationImpl adds performance overheads due to instantiating a stack trace
I'm trying to use Mockito in a test of a Java class I didn't write. It takes a dependency which I am stubbing out using Mockito. Skipping all of the business logic, the code ends up being rather similar to this:

```java
SomeClass mocked = mock(SomeClass.class, Mockito.withSettings().stubOnly());
when(mocked.someParam()).thenReturn(0);

for (int i = 0; i < bigNumber; i++) {
    mocked.someParam();
}
```

This is in Mockito 4.6.1. This is extremely slow, and looking at Mockito this seems to be a problem for two reasons.

Firstly, when a mock invocation occurs, LocationImpl is instantiated. This instantiates a Throwable, which constructs a stack trace, error message, etc (which is then discarded). In my test, Mockito was generating 70k exceptions per second.

The second slowdown that seemed to be affecting me comes from TypeSafeMatching.getArgumentType. This iterates through all of the methods on the argument matcher class to find the appropriate matches method, seemingly also for each method call. By performing this operation, a lot of JVM internal datastructures get copied - it calls Class.getMethods, which copies all the methods on the class. This method could probably be cached on a per-class basis.

This dramatically affected my application's allocation rate; Mockito was leading to approx 3.4GiB/sec allocations. In the end, I rewrote my mock using ByteBuddy proxies directly. It took the test runtime from 30 seconds down to 1, with the number of exceptions being created per second going down from 70k to ~0.

I know that Mockito isn't meant to be a Netty-like high performance library, but wanted to flag these two codepaths as affecting every mock call.

j-baker:
Fixes #2720.

In terms of memory allocations, this reduces the overall memory
allocations of creating a Location by an order of magnitude in
Java 9, and as compared to Java 8.

The implementation is somewhat involved due to these desires:

- Minimize the amount of work done if the Location is never used. This
  is done by not converting the StackFrame into a StackTraceElement,
  instead wrapping in an intermediate type. The StackTraceElement conversion
  will only occur (internally) if the .getFileName() method is called, which it usually isn't.
- Ensure the implementation is still Serializable. This is ensured with
  a .writeReplace method.
- Minimize the number of allocations, which is basically an exercise in
  lambda caching.
- Ensuring the old mechanism still works on Java 8.
- MethodHandles are even lower overhead than the reflection, but AnimalSniffer does
  not like them since they don't exist in Android.

Presently on Java 9+, on a stack depth of 1000 the old mechanism will allocate
40kB of RAM per call. The new one will allocate 1.5kB of RAM per call,
which is a huge improvement. Even in a very basic JUnit test with no stack depth,
the difference is about 1.5kB of allocation vs 8kB.

This is still sadly not the 'close-to-no-overhead' solution I was
looking for. I therefore also added a system property that can be used
to fully disable Location creation. I'm aware that this is likely not
the right approach given Mockito has plugins and settings - mostly
looking for guidance here given I'm not sure what would be idiomatic
here.

## Checklist


src/main/java/org/mockito/internal/debugging/Java9PlusLocationImpl.java
```diff
@@ -0,0 +1,262 @@
...
+                    return metadata.indexOf(Java9PlusLocationImpl.class.getName());
+                });
+    }
+
+    @SuppressWarnings("unchecked")
+    private static <T> T stackWalk(Function<Stream<Object>, T> function) {
+        try {
+            return (T) walk.invoke(stackWalker, function);
+        } catch (InvocationTargetException | IllegalAccessException e) {
+            throw new RuntimeException(e);
```
TimvdLippe, line 125: Rather than `RuntimeException`, let's use `MockitoException` with a descriptive error message that we don't intend to ever reach this codepath. https://github.com/mockito/mockito/blob/main/src/main/java/org/mockito/internal/creation/bytebuddy/SubclassByteBuddyMockMaker.java has numerous examples of descriptive exceptions that would be helpful for the user. If the message ends up being the same for all of these, we can also define a custom subclass of `MockitoException` in this class and use that.

src/main/java/org/mockito/internal/debugging/Java9PlusLocationImpl.java
```diff
@@ -0,0 +1,262 @@
...
+        return stackWalk(
+                stream ->
+                        stream.map(toStackFrameMetadata)
+                                .skip(FRAMES_TO_SKIP)
+                                .filter(cleanerIsIn)
+                                .skip(isInline ? 1 : 0)
+                                .findFirst()
+                                .orElseThrow(
+                                        () ->
+                                                new IllegalArgumentException(
```
TimvdLippe, line 101: Same here. Let's make this a `MockitoException`

src/main/java/org/mockito/internal/debugging/Java9PlusLocationImpl.java
```diff
@@ -0,0 +1,262 @@
...
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    private static Object stackWalker() {
+        try {
+            Set options =
+                    Collections.singleton(Enum.valueOf((Class) optionClazz, SHOW_REFLECT_FRAMES));
+            Method getInstance =
+                    stackWalkerClazz.getDeclaredMethod("getInstance", Set.class, int.class);
+            return getInstance.invoke(null, options, BUFFER_SIZE);
+        } catch (InvocationTargetException | NoSuchMethodException | IllegalAccessException e) {
+            throw new RuntimeException(e);
```
TimvdLippe, line 154: Same here

src/main/java/org/mockito/internal/debugging/LocationFactory.java
```diff
@@ -0,0 +1,75 @@
...
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.internal.debugging;
+
+import org.mockito.invocation.Location;
+
+import java.io.Serializable;
+
+public final class LocationFactory {
+    private static final String PROPERTY = "mockito.locationFactory.disable";
```
TimvdLippe, line 12: Not sure if I like that we can disable it now. Mockito is trying to be helpful here and we always want to be as descriptive as possible when failing so that users can debug their issue. Therefore, I would prefer not to have this toggle.

If we were want to expose this as a plugin, we would need to follow the plugin-mechanism as defined in https://github.com/mockito/mockito/blob/main/src/main/java/org/mockito/internal/configuration/plugins/DefaultMockitoPlugins.java which users can also override using resources. However, I would prefer to ship this PR without a toggle and then based on user feedback further tweak it in a follow-up. For now, the improvements are already great, so I think that has benefits on its own.

TimvdLippe (changes requested):
The CI build is failing because of auto-formatting issues. You can run `./gradlew spotlessApply` to fix these automatically.

Overall a nice improvement! Mostly some minor comments, but overall looking solid already. Good work 😄

src/main/java/org/mockito/internal/debugging/Java9PlusLocationImpl.java
```diff
@@ -0,0 +1,262 @@
...
+
+    private static final String PREFIX = "-> at ";
+
+    private static final StackTraceCleaner CLEANER =
+            Plugins.getStackTraceCleanerProvider()
+                    .getStackTraceCleaner(new DefaultStackTraceCleaner());
+
+    private static final Function<Object, StackFrameMetadata> toStackFrameMetadata =
+            MetadataShim::new;
+    private static final Predicate<StackFrameMetadata> cleanerIsIn = CLEANER::isIn;
```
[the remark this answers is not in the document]
j-baker: If this is inlined, each time the method is run a new Predicate is allocated (the method handle is converted into a Predicate object at this time). If it's not inlined, the allocation can be avoided. I'll add a comment to the effect?

I went through this with Java flight recorder attempting to minimise overheads.

src/main/java/org/mockito/internal/debugging/LocationFactory.java
```diff
@@ -0,0 +1,75 @@
...
+    private static final Factory factory;
+
+    static {
+        Factory lFactory;
```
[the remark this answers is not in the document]
j-baker: I assigned to lFactory because the compiler was unable to determine in the catch case that the variable wasn't already assigned to before the throw. I've moved it into a method from the `static {}` which can return and so avoid the boilerplate.

src/main/java/org/mockito/internal/debugging/Java9PlusLocationImpl.java
j-baker: I've made some error message changes which are hopefully helpful. If you would like the wording tweaked, please feel free to change the PR directly? These kinds of communications are usually some of the hardest bits to fit with the 'style' of the project.

src/main/java/org/mockito/internal/debugging/LocationFactory.java
j-baker: I've removed this behaviour for now.

src/main/java/org/mockito/internal/debugging/Java9PlusLocationImpl.java
TimvdLippe: The new error messages are fine by me, thanks!

src/test/java/org/mockitousage/internal/debugging/LocationFactoryTest.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.IntStream;
+
+import com.sun.management.ThreadMXBean;
+import org.junit.Test;
+import org.mockito.internal.debugging.LocationFactory;
+import org.mockitoutil.TestBase;
+
+public class LocationFactoryTest extends TestBase {
```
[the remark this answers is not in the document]
j-baker: I've re-enabled the test. I've changed the condition for enablement - that StackWalker should exist. I didn't improve the old codepath.

j-baker:
Latest commit is green but the status checks are red. Not sure why, but I assume a click of retry will do the trick.

src/test/java/org/mockitointegration/ClassLoadabilityChecker.java
```diff
@@ -0,0 +1,70 @@
...
+
+    public ClassLoadabilityChecker(ClassLoader classLoader, String purpose) {
+        this.classLoader = classLoader;
+        this.purpose = purpose;
+    }
+
+    public void checkLoadability(String className) {
+        try {
+            Class.forName(className, INITIALIZE_CLASSES, classLoader);
+        } catch (ClassNotFoundException | NoClassDefFoundError e) {
```
TimvdLippe, line 35: I think we need to catch `Exception` here, since the actual exception that is thrown is `ExceptionInInitializerError`: https://github.com/mockito/mockito/runs/7820014375?check_suite_focus=true#step:7:183

TimvdLippe:
@j-baker The latest commit referenced in this PR is the one from #2729. The JUnit test uses `Class.forName`, which breaks on Java 8. See my comment for a suggestion on how to fix it.

j-baker:
Sigh, you were right. My bad. Anyway, it looks like it'll go green now.

src/main/java/org/mockito/internal/debugging/LocationFactory.java
```diff
@@ -0,0 +1,49 @@
...
+            return new Java9PlusLocationFactory();
+        } catch (ClassNotFoundException e) {
+            return new Java8LocationFactory();
+        }
+    }
+
+    private static final class Java8LocationFactory implements Factory {
+        @Override
+        public Location create(boolean inline) {
+            return new Java8LocationImpl(new Throwable(), inline);
```
TimvdLippe, line 38: Nit: based on the coverage report, we are only using 1 constructor of `Java8LocationImpl`. Since the others are unused, let's clean it up and remove them.

TimvdLippe (changes requested):
Last nits and then this is ready-to-go 🎉

TimvdLippe (approved):
Thanks for the hard work and investigation into performance! Much appreciated 😄

GThinesh:
Is there any specific reason to remove the system variable to disable location tracking altogether? My teams is still using Java 8 and this is a major problem for us as well. Locally testing without location reduces the test time from 10+ min to 5 min. (Yes we have a huge code base with large entity graphs :) )

TimvdLippe:
I prefer to not ship a flag just for Java 8, which is EOL. Instead, upgrading to Java 9 and above should yield similar benefits. Shipping flags is expensive in the longterm and in this case, I don't think it is worth carrying that cost.

j-baker:
It’s probably worth saying:
Before my changes, a simple mock call takes 3us.
After, it takes 2us.
If you remove all call site recording, it takes 200ns.
In my case I was considering making it so that it stopped recording call
sites on hotter paths (E.g. after the thousandth call to a mock or
whatever) but it wasn’t a big enough blocker for me.

GThinesh:
@TimvdLippe That is true. Unfortunately we are still stuck with Java 8.

GThinesh:
@j-baker The solution you refer the flag a counter that stops recording after certain invocations?

j-baker:
I was considering proposing making the location tracking code always stop
recording after enough invocations, the logic being that if you’ve done
thousands of them, knowing the precise location where a single call happens
from is probably less interesting.
There’s possibly also something where if mock invocation recording is
disabled, it might be reasonable to not record stub invocation locations.

PR #2729, opened 2022-08-13
TypeSafeMatching no longer iterates over class methods inefficiently

j-baker:
Fixes #2723.

Class.getMethods is an inefficient method call which is being called on
each mock invocation. It ends up constructing new Method objects for
each method on the class, and this can dominate the overall performance
of Mockito mocks. This commit caches the result of the computation.

Once concern is that this change uses some static state. I considered:

- Instance state - based on where this type is constructed it seemed
  like it'd be a big imposition on code readability elsewhere.
- Weakly referenced map. Mockito has a type for this, but the
  constructor of that type produces a Thread with which to clean up.

Both of these seemed like overkill compared to the overhead expected
in the real world (which should be on the order of a few kilobytes of
RAM at most).

## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/invocation/TypeSafeMatching.java
+++ b/src/main/java/org/mockito/internal/invocation/TypeSafeMatching.java
@@ -4,14 +4,25 @@
  */
 package org.mockito.internal.invocation;
 
-import java.lang.reflect.Method;
-
 import org.mockito.ArgumentMatcher;
+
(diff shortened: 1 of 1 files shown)


j-baker:
I saw you triggered CI and it failed. I've edited the PR subsequently and it certainly passes locally...

TimvdLippe (approved):
Nice fix, thanks!

PR #2742, opened 2022-08-30
Assign GlobalConfiguration initializer to unused variable

kluever:
(This silences some errorprone static analysis about calling constructors w/o using the new object.)

## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/configuration/GlobalConfiguration.java
+++ b/src/main/java/org/mockito/internal/configuration/GlobalConfiguration.java
@@ -43,7 +43,7 @@ public class GlobalConfiguration implements IMockitoConfiguration, Serializable 
     }
 
     public static void validate() {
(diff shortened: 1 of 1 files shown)


kluever:
Inside of google we've made CheckReturnValue the default (!), and also added coverage for constructors (calling a constructor for side effects is _super_ weird).

We haven't released those checks yet, but I imagine we will at some point. If you'd rather that we carry a local patch for this edit, that's something we can do as well...but then again, this PR is a no-op, so landing it would obviously be my preferred solution to carrying a local patch.

TimvdLippe:
No I am okay with landing this patch, since it is more descriptive what is going on. I was mostly confused as to why ErrorProne didn't catch this error for us.

TimvdLippe (approved):
Thanks! Looking forward to the ErrorProne release as well.

PR #2759, opened 2022-09-25
Update Javadoc deeps link to Java 11

TimvdLippe:
We currently use Java 6 to deeplink to the JDK and the URL no longer shows the necessary element-list, which is why it currently fails the build: https://github.com/mockito/mockito/actions/runs/3100073977/jobs/5019939573#step:7:229

### DIFF
--- a/gradle/mockito-core/javadoc.gradle
+++ b/gradle/mockito-core/javadoc.gradle
@@ -38,7 +38,7 @@ javadoc {
     options.noIndex = false
     options.noNavBar = false
     options.noTree = false
-    options.links = ['https://docs.oracle.com/javase/6/docs/api/',
+    options.links = ['https://docs.oracle.com/javase/11/docs/api/',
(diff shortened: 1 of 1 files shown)


marcphilipp:
Why not deep link against JDK 8 since that's the baseline?

TimvdLippe:
@marcphilipp The problem is not the Oracle version of the JDK, but the JDK version that we use for the javadoc tool. https://docs.oracle.com/javase/8/docs/api/element-list is missing as well, but https://docs.oracle.com/javase/8/docs/api/package-list does exist. Therefore, we have to use JDK 11+ for the generation of our Javadoc if we want to fix this issue, as that one is aware of `package-list` and understands how to use it.

marcphilipp:
FWIW we have this workaround in the JUnit build:
https://github.com/junit-team/junit5/blob/517bfd56ad8ba8d22609b5055f4bb46a3d38c953/documentation/documentation.gradle.kts#L307-L320

PR #2766, opened 2022-10-16
Possible fix #2765: Add task to download package-list file from java as element-list

### LINKED ISSUE #2765: JavaDoc warning is blocking all pull requests
Starting with #2758, our JavaDoc build has been erroring because of a change on Oracle's side. There's more information in #2759, but the gist of it is that Oracle removed the `element-list` URL and replaced it with the `package-list` URL. However, the latter can only be consumed by JavaDoc from JDK 9 and above, whereas we are running JDK 8 (since our minimum supported version is still Java 8).

We have a couple of options to remedy the situation:
1. We upgrade to Java 11 and publish Mockito 5, as we update our minimum supported version
2. We employ JUnit's workaround where we download the `package-list` and write it to disk as `element-list`, after which we use `-linkoffline` instead of `-link` to consume the file
3. We use Java 11 to publish our JavaDoc, but we use Java 8 for the rest (e.g. artifact publication)
4. We use Java 11, but we still publish with minimum version Java 8

Of all these options, I think 2 is the best we have. Upgrading to Java 11 runtime (regardless of what version we decide to publish for) is still a risk, as we want our artifacts to work on Java 8. And the best way to confirm that's all working is to run on Java 8 ourselves. I don't think we should upgrade to Java 11 yet, but it's one more drop in the bucket for us dropping Java 8 support.

andrepaschoal:
I think these changes could a fix, or at least part of, for the issue https://github.com/mockito/mockito/issues/2765.

## Checklist


andrepaschoal:
hey @TimvdLippe, please check if this could a solution for https://github.com/mockito/mockito/issues/2765

andrepaschoal:
Per [this](https://github.com/mockito/mockito/actions/runs/3259663624/jobs/5355440652#step:7:91) Java 8 build is going after `package-list`... let me do something very dumb and hava both `element-list` and `package-list` in that dir...

andrepaschoal:
🥳

gradle/mockito-core/javadoc.gradle
```diff
@@ -38,8 +38,8 @@ javadoc {
     options.noIndex = false
     options.noNavBar = false
     options.noTree = false
-    options.links = ['https://docs.oracle.com/javase/6/docs/api/',
-                     'https://junit.org/junit4/javadoc/4.12/']
+    options.links('https://junit.org/junit4/javadoc/4.12/')
+    options.linksOffline('https://docs.oracle.com/javase/6/docs/api/', "$buildDir/javase-6-docs/")
```
TimvdLippe, line 42: Nit: while we are at it, let's update our Javadoc version to 8 as well, since that's the version we support

gradle/mockito-core/javadoc.gradle
```diff
@@ -52,6 +52,23 @@ javadoc {
     options.addStringOption('Xwerror', '-quiet')
     options.addBooleanOption('-allow-script-in-comments', true)
 
+    // attempt to fix https://github.com/mockito/mockito/issues/2765
+    doFirst {
```
TimvdLippe, line 56: Let's put this in a separate task and add it as a dependency to the `javadoc` task here. Then, let's also add a verify task per https://github.com/michel-kraemer/gradle-download-task#verify-task to make sure we don't accidentally download malicious content from the internet.

gradle/mockito-core/javadoc.gradle
```diff
@@ -52,6 +52,23 @@ javadoc {
...
+        download.run {
+            src 'https://docs.oracle.com/javase/6/docs/api/package-list'
+            // somehow downloading src/javadoc or buildDir/javadoc didn't work, probably something related to the folder
+            // lifecycle
+            dest "$buildDir/javase-6-docs/element-list"
+            overwrite true
+        }
+        // dumbly downloading again just to see if we can make all jdks happy
```
TimvdLippe, line 64: Nit: replace comment with the following:

```java
// For Java 9+, download the `package-list` as-is
```

gradle/mockito-core/javadoc.gradle
```diff
@@ -52,6 +52,23 @@ javadoc {
     options.addStringOption('Xwerror', '-quiet')
     options.addBooleanOption('-allow-script-in-comments', true)
 
+    // attempt to fix https://github.com/mockito/mockito/issues/2765
```
TimvdLippe, line 55: Nit: Rewrite to

```java
// Temporary workaround for https://github.com/mockito/mockito/issues/2765
```

gradle/mockito-core/javadoc.gradle
```diff
@@ -52,6 +52,23 @@ javadoc {
...
+    doFirst {
+        download.run {
+            src 'https://docs.oracle.com/javase/6/docs/api/package-list'
+            // somehow downloading src/javadoc or buildDir/javadoc didn't work, probably something related to the folder
```
TimvdLippe, line 59: Nit: replace comment with the following:

```java
// Since the JavaDoc output is generated in `$buildDir/javadoc`, specify a different location for this task
```

TimvdLippe (changes requested):
Nice, glad that works! A couple of nits and let's make sure the artifact is verified with a shasum to prevent malicious content to be downloaded.

gradle/mockito-core/javadoc.gradle
TimvdLippe: We can manually download the file, generate a shasum and then add that number as the verification sha. Then we know for sure it has the same file content as we expect and have read.

gradle/mockito-core/javadoc.gradle
```diff
@@ -1,7 +1,31 @@
...
+task downloadJava8ElementList(type: Download) {
+    src packageListUrl
+    // Since the JavaDoc output is generated in `$buildDir/javadoc`, specify a different location for this task
+    dest "$java8DocsDir/element-list"
+    overwrite true
+}
+
+// Temporary workaround for https://github.com/mockito/mockito/issues/2765
+// For Java 9+, download the `package-list` as-is
+task downloadJava9PlusPackageList(type: Download, dependsOn: downloadJava8ElementList) {
```
TimvdLippe, line 19: Nit: rather than having the `dependsOn` here, let's add the `dependsOn` for both in the javadoc task. since the download of this task is independent of the previous task.

gradle/mockito-core/javadoc.gradle
andrepaschoal: > We can manually download the file, generate a shasum and then add that number as the verification sha.
true! 🤦‍♂️ 

thinking here... do we expect this `package-list` file to change? only if we move away from 8, right? why not just download it now and add it to the project?

TimvdLippe: We don't expect it to change I think. In that case, yeah we can download it as well. Good point 😂

andrepaschoal: BTW, what's the suggested place for these files? I was considering the below:

```
- gradle/mockito-core/java-8-docs/element-list
- gradle/mockito-core/java-8-docs/package-list
```

TimvdLippe: Yeah that sounds good to me.

build.gradle
```diff
@@ -25,6 +25,7 @@ plugins {
     id 'com.github.ben-manes.versions' version '0.42.0'
     id 'biz.aQute.bnd.builder' version '6.3.1'
     id 'ru.vyarus.animalsniffer' version '1.5.2'
+    id 'de.undercouch.download' version '5.3.0'
```
TimvdLippe, line 28: Nit: we can drop this one as well then 🎉

gradle/mockito-core/javadoc.gradle
```diff
@@ -1,7 +1,19 @@
 //It seems the gradle javadoc task works file by file and as such disable some features of javadoc tool
 //such as link to packages, https://groups.google.com/d/msg/gradle-dev/R83dy_6PHMc/bgw0cUTMFAAJ
+
+// In this directory we keep two copies of the `package-list` file for java 8 api docs:
+// - as `package-list` for Java 9+
+// - as `element-list` for java 8
+def java8DocsDir = "$buildDir/java-8-docs"
+
+task copyJava8Docs(type: Copy) {
```
TimvdLippe, line 9: Rather than copying, can we use the source directory for `linksoffline`? That should be `$projectDir/gradle/mockito-core`

gradle/mockito-core/javadoc.gradle
```diff
@@ -1,5 +1,11 @@
 //It seems the gradle javadoc task works file by file and as such disable some features of javadoc tool
 //such as link to packages, https://groups.google.com/d/msg/gradle-dev/R83dy_6PHMc/bgw0cUTMFAAJ
+
+// In this directory we keep two copies of the `package-list` file for java 8 api docs:
+// - as `package-list` for Java 9+
+// - as `element-list` for java 8
+def java8DocsDir = 'gradle/mockito-core/java-8-docs'
```
andrepaschoal, line 7: keeping this var just for the sake of documentation

TimvdLippe (approved):
The smallest fixes are most of the times the best ones 👍 Thanks for the work! Greatly appreciated 😄

PR #2778, opened 2022-10-28
[experiment] break gradle checksum: will CI show the mismatch?

asolntsev:


asolntsev:
Closing PR.
We got the answer: gradle checksum is verified only if's present in gradle-wrapper.properties

PR #2779, opened 2022-10-28
automatically detect class to mock

asolntsev:
this commit adds method `Mockito.mock()` as a shorter alternative for `Mockito.mock(Class)`. When result of this call is assigned to a variable/field, java will detect the needed class automatically.

P.S. I had to ignore errorprone checks "DoNotMock" and "DoNotMockAutoValue" because they fail with IndexOutOfBoundException trying to get the first argument of `Mockito.mock()`. :)

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -1902,6 +1902,23 @@ public class Mockito extends ArgumentMatchers {
     public static final Answer<Object> RETURNS_SELF = Answers.RETURNS_SELF;
 
     /**
+     * Creates mock object of requested class or interface.
+     * <p>
+     * See examples in javadoc for {@link Mockito} class
+     *
+     * @param reified don't pass any values to it. It's a trick to detect the class/interface you want to mock.
+     * @return mock object
+     */
+    public static <T> T mock(T... reified) {
+        if (reified.length > 0) {
+            throw new IllegalArgumentException(
+                    "Please don't pass any values here. Java will detect class automagically.");
+        }
+        Class<T> classToMock = (Class<T>) reified.getClass().getComponentType();
+        return mock(classToMock, withSettings());
+    }
+
+    /**
      * Creates mock object of given class or interface.
      * <p>
      * See examples in javadoc for {@link Mockito} class
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -1647,10 +1647,6 @@ import java.util.function.Function;
  * <h3 id="54">54. <a class="meaningful_link" href="#mock_without_class" name="mock_without_class">
  *  Mocking/spying without specifying class</a> (Since 4.9.0)</h3>
  *
- * <p>
- * Starting from 4.9.0, there is a shorter form of creating a mock or spy.
- * </p>
- *
  * Instead of calling method {@link Mockito#mock(Class)} or {@link Mockito#spy(Class)} with Class parameter, you can now
  * now call method {@code mock()} or {@code spy()} <strong>without parameters</strong>:
  *
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -1657,8 +1657,9 @@ import java.util.function.Function;
  *
  * Mockito will automatically detect the needed class.
  * <p>
- * P.S. Works only if you assign result of {@code mock()} or {@code spy()} to a variable or field.
- * Only then Java compiler can know the type of that variable.
+ * It works only if you assign result of {@code mock()} or {@code spy()} to a variable or field with an explicit type.
+ * With an implicit type, the Java compiler is unable to automatically determine the type of a mock and you need
+ * to pass in the {@code Class} explicitly.
  * </p>
  */
 @CheckReturnValue
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -33,7 +33,11 @@ import org.mockito.stubbing.LenientStubber;
 import org.mockito.stubbing.OngoingStubbing;
 import org.mockito.stubbing.Stubber;
 import org.mockito.stubbing.VoidAnswer1;
-import org.mockito.verification.*;
+import org.mockito.verification.After;
+import org.mockito.verification.Timeout;
+import org.mockito.verification.VerificationAfterDelay;
+import org.mockito.verification.VerificationMode;
+import org.mockito.verification.VerificationWithTimeout;
 
 import java.util.function.Function;
 
@@ -107,7 +111,7 @@ import java.util.function.Function;
  *      <a href="#51">51. New API for marking classes as unmockable (Since 4.1.0)</a><br/>
  *      <a href="#52">52. New strictness attribute for @Mock annotation and <code>MockSettings.strictness()</code> methods (Since 4.6.0)</a><br/>
  *      <a href="#53">53. Specifying mock maker for individual mocks (Since 4.8.0)</a><br/>
- *      <a href="#54">54. Mocking without specifying class (Since 4.9.0)</a><br/>
+ *      <a href="#54">54. Mocking/spying without specifying class (Since 4.9.0)</a><br/>
  * </b>
  *
  * <h3 id="0">0. <a class="meaningful_link" href="#mockito2" name="mockito2">Migrating to Mockito 2</a></h3>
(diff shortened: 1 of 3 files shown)


TimvdLippe:
This is a super interesting implementation and I am not sure what to think of it. What I like about `@Mock` is that we don't have to deal with the unchecked warnings. This method would fix that exact problem, but now for `mock()`, which is awesome. At the same time, it also feels like a bit of a hack and I wonder what the readability of this will be. The expressiveness of the `.class` argument makes it at least clear what we are attempting to mock and what you will get.

That said, I am not sure if there is any harm in shipping this method, given that it does resolve the unchecked warning problem for `Mockito.mock()`, which is valuable on its own. For that, I think we do need to fix these 2 ErrorProne checks, as they will be crashing otherwise. However, I think that needs to be done after this PR is merged.

@mockito/developers WDYT?

raphw:
I do like the concept! I am however not sure if overloading might create some issues here and there, but as our tests compile, it does not seem to be a problem. We are already doing a lot with the Java language that is not intuitive at first glance, so I think this would be a good API addition.

TimvdLippe (changes requested):
Please add a new section to the JavaDoc (section 54) that showcases the new feature. Don't forget to also link to the new section in the Contents section at the top.

szpak:
This API is similar to the Spock one where it has been in common use [for years](https://github.com/spockframework/spock/pull/560). If there are no warnings in IDE, why not to leverage it (assuming no technical glitches are found).

```
var mock = mock(MyClass.class)
```
is a (current) alternative, but the new one is even more robust (and shorter).

src/main/java/org/mockito/Mockito.java
```diff
@@ -1901,6 +1901,23 @@ public class Mockito extends ArgumentMatchers {
      */
     public static final Answer<Object> RETURNS_SELF = Answers.RETURNS_SELF;
 
+    /**
+     * Creates mock object of requested class or interface.
+     * <p>
+     * See examples in javadoc for {@link Mockito} class
+     *
+     * @param reified don't pass any values to it. It's a trick to detect the class/interface you want to mock.
+     * @return mock object
```
szpak, line 1910: Maybe it could be worth to add `@since 4.9.0`?

src/main/java/org/mockito/Mockito.java
```diff
@@ -1901,6 +1901,23 @@ public class Mockito extends ArgumentMatchers {
...
+     */
+    public static <T> T mock(T... reified) {
+        if (reified.length > 0) {
+            throw new IllegalArgumentException(
+                    "Please don't pass any values here. Java will detect class automagically.");
```
szpak, line 1915: And maybe a test for the corner case (to keep the coverage high ;-) )?

src/main/java/org/mockito/Mockito.java
asolntsev: thanks, added `@since` to javadoc.

src/main/java/org/mockito/Mockito.java
asolntsev: @szpak thanks, added negative test

asolntsev:
> Btw, Groovy allows to do that kind of tricks, but I didn't know about that one in Java, good idea @asolntsev!

I also didn't know that Java can detect the class name.
I found this trick in [Tagir Valeev post](https://twitter.com/tagir_valeev/status/1262763570904719361)

asolntsev:
@szpak @TimvdLippe Similarly to `Mockito.mock()`, I have also added method `Mockito.spy()`. 
Also implemented all suggestions from code reviews.

src/main/java/org/mockito/Mockito.java
```diff
@@ -1639,6 +1644,26 @@
  *   Foo mock = Mockito.mock(Foo.class, withSettings().mockMaker(MockMakers.SUBCLASS));
  * </code></pre>
  *
+ * <h3 id="54">54. <a class="meaningful_link" href="#mock_without_class" name="mock_without_class">
+ *  Mocking/spying without specifying class</a> (Since 4.9.0)</h3>
+ *
+ * <p>
+ * Starting from 4.9.0, there is a shorter form of creating a mock or spy.
+ * </p>
+ *
```
TimvdLippe, line 1653:
```suggestion
```

src/main/java/org/mockito/Mockito.java
```diff
@@ -1639,6 +1644,26 @@
...
+ * <pre class="code"><code class="java">
+ *   Foo foo = Mockito.mock();
+ *   Bar bar = Mockito.spy();
+ * </code></pre>
+ *
+ * Mockito will automatically detect the needed class.
+ * <p>
+ * P.S. Works only if you assign result of {@code mock()} or {@code spy()} to a variable or field.
+ * Only then Java compiler can know the type of that variable.
```
TimvdLippe, line 1665:
```suggestion
 * It works only if you assign result of {@code mock()} or {@code spy()} to a variable or field with an explicit type.
 * With an implicit type, the Java compiler is unable to automatically determine the type of a mock and you need
 * to pass in the {@code Class} explicitly.
```

TimvdLippe (changes requested):
2 small nits. I wanted to fix them myself, but I didn't have permission to push to your fork. After that, this LGTM!

asolntsev:
> 2 small nits. I wanted to fix them myself, but I didn't have permission to push to your fork. After that, this LGTM!

Thank you, merged.

TimvdLippe (approved):
Thanks!

rdesgroppes:
> P.S. I had to ignore errorprone checks "DoNotMock" and "DoNotMockAutoValue" because they fail with IndexOutOfBoundException trying to get the first argument of `Mockito.mock()`. :)

I filed https://github.com/google/error-prone/issues/3609: feel free to adjust as per your own observations.

ankurbajaj9:
This change has broken my Mockito.spy(object_to_mock) calls in a scala project. Basically Scala cant differenciate between the two spy methods.

asolntsev:
@ankurbajaj9 Can you share some reproducible example? I have just tried with Scala 2 and Scala 3 - both `Mockito.mock(Class)` and `Mockito.mock()` work as expected.

ankurbajaj9:
can you try the Mokcito.spy(Object) and see if you can reproduce the same

asolntsev:
@ankurbajaj9 Yes, works for me:

```java
  @Test def spyWithObject(): Unit = {
    val names: util.ArrayList[String] = Mockito.spy(util.ArrayList())
    assert(names.size == 0)
    names.add("tere")
    assert(names.size == 1)
    doReturn(42).when(names).size
    assert(names.size == 42)
  }
```

@ankurbajaj9 Once again: it would be much simpler if you shared some reproducible example.

sijtsche:
We upgraded to a newer version of Mockito (4.11.0, works with 4,9.0) and it breaks our integration test, because I think it takes this new mock method instead of the original one, while it should have executed the original one with this signature public static <T> T mock(Class<T> classToMock).   

This mock
<bean id="conversionService" name="conversionService" class="org.mockito.Mockito" factory-method="mock" >
        <constructor-arg value="org.springframework.core.convert.ConversionService"/>
</bean>
Causes the following error when creating the applicationcontext for integration tests:
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'conversionService' defined in URL [file:/C:/sometestproject/target/test-classes/MyTest-context.xml]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [java.lang.Object]: Factory method 'mock' threw exception; nested exception is java.lang.IllegalArgumentException: Please don't pass any values here. Java will detect class automagically.

This is described as a working way of using Mockito by Spring in their docs:
https://docs.spring.io/spring-framework/docs/current/reference/html/testing.html
See snippet on that page:
 <bean id="accountService" class="org.mockito.Mockito" factory-method="mock">
    <constructor-arg value="org.example.AccountService"/>
</bean>

asolntsev:
@sijtsche Thank you for reporting the problem.
1. I would say this problem should be solved on Spring side. 
2. I would recommend to avoid using such XMLs etc. This is very old-school way of using Spring. Are you aware of annotations, autowired dependencies etc? It's much more convenient, and it existed in Spring for 10+ years already. 

Anyway, could you please share some simple project to reproduce the problem? Some trivial "hello world" project?

sijtsche:
@asolntsev we found a solution which may be worth adding to the releasenotes, so others don't struggle with it anymore. This way the right method within the Mockito class is invoked and no error occurs.
Yes I am fully aware of the possibilities of Spring configuration, but XML configuration is still valid Spring configuration in use by lots of companies (since it exists as long as you say) and XML configuration is not deprecated.
This is the working solution: 
```
    <bean id="conversionService" name="conversionService" class="org.mockito.Mockito" factory-method="mock">
        <constructor-arg>
            <value type="java.lang.Class">org.springframework.core.convert.ConversionService</value>
        </constructor-arg>
    </bean>
```
Another work around could be moving all mocks to a Spring configuration class and including that configuration class as a bean in the xml config file.

sijtsche:
Yes, I am aware of all types of configuration in Spring. But not all companies use that and it is not even a deprecated way of configuring a Spring application. So IMHO we should be able to use XML configuration in Spring. I think large projects that exists for a long time can not rewrite their whole application wiring because of a Mockito update. So it would be great if we could at least find a workaround.

[mockitodemo.zip](https://github.com/mockito/mockito/files/10415653/mockitodemo.zip)

I have created a demo Spring Boot project that reproduces this issue. With a partial solution in XML config too. You will find two testclasses, one that gives the error I mentioned, the other with this workaround that now invokes the right mock method in the Mockito class by specifying the type java.lang.Class:
```
    <bean id="conversionService" name="conversionService" class="org.mockito.Mockito" factory-method="mock">
        <constructor-arg>
            <value type="java.lang.Class">org.springframework.core.convert.ConversionService</value>
        </constructor-arg>
    </bean>
```

But, if I add a class to the application context that uses this conversionService mock as an autowired dependency (see TestClass in the testproject) it can not instantiate the TestClass class and fails with this error: Error creating bean with name 'testclass': Unsatisfied dependency expressed through field 'conversionService'; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.springframework.core.convert.ConversionService' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=true)}

For that I do not have a workaround (yet).

As soon as I set the dependency of mockito-core back to 4.9.0, this works again without other changes. Anyway, you can try it now yourself with this testproject.

asolntsev:
@sijtsche Thank you for providing the working example!
Now I see that this declaration is not really correct:
```xml
        <constructor-arg value="org.springframework.core.convert.ConversionService"/>
```

Since Spring doesn't know the type of parameter "ConversionService", it just tries to find a static method with one parameter _of any type_ in class `Mockito`. It occasionally worked before `Mockito` had only one static method `mock` with a single parameter. 

Now `Mockito` has at least two public static methods `mock` with a single parameter. 
So it's anyway a good idea to declare the type of the parameter explicitly:
```xml
        <constructor-arg>
            <value type="java.lang.Class">org.springframework.core.convert.ConversionService</value>
        </constructor-arg>
```

P.S. About your second problem: I think it's not related to the static method `Mockito.mock`. It's really caused by the order of dependencies in your xml: you need to declare bean `testclass` AFTER bean `conversionService`. It seems reasonable to me because `testclass` depends on `conversionService`.

sijtsche:
Hi @asolntsev the typing issue is then solved. But the missing dependency works if I just set mockito back to 4.9.0 it works and if I set it to a higher version it does not work anymore. It indeed probably is not related to this particular change, but I think it is somehow related to Mockito.

Issue #2789, opened 2022-11-13
Documentation describes a broken way to mock final classes


asolntsev:
Documentation stays that for mocking final classes, one needs to add file `mockito-extensions/org.mockito.plugins.MockMaker`:

https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#39

BUT
**It doesn't always work.**

This sample shows that spying enums this way doesn't really work:
https://github.com/asolntsev/mockito-enum-test

P.S. Seems that the proper way to mock final classes/enums is adding dependency `org.mockito:mockito-inline:4.8.1`.

TimvdLippe:
While `mockito-inline` works, that is a convenience artifact that provides this resource file for you: https://github.com/mockito/mockito/blob/main/subprojects/inline/src/main/resources/mockito-extensions/org.mockito.plugins.MockMaker It is using the structure as is described in the JavaDoc. Therefore, I think it is best to keep the documentation here, since it is what `mockito-inline` does for you.

Are you sure that in your project, it is picking up the `src/test/resources` as expected?

asolntsev:
That's the point: artifact `mockito-inline` provides two files, not one.
Only `org.mockito.plugins.MockMaker` is not enough (as my example shows).

<img width="433" alt="image" src="https://user-images.githubusercontent.com/279773/201761098-863921fc-8f70-4ab3-af60-783ca39332b3.png">



> Are you sure that in your project, it is picking up the src/test/resources as expected?

Yes, I am sure.
1. When I delete my folder `src/test/resources`, Mockito fails with the expected error `Cannot mock/spy class io.github.asolntsev.Hobby: Mockito cannot mock/spy because : - final class`
2. When I put an invalid content `noooope` in file `src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker`, Mockito fails with the expected error `Could not initialize plugin: interface org.mockito.plugins.MockMaker.... Caused by: java.lang.ClassNotFoundException: noooope`

TimvdLippe:
In that case, we should document that you also need `org.mockito.plugins.MemberAccessor` to make things work correctly. I don't think we should advertise `mockito-inline` in that section, as it is targeted to implementation. We already list `mockito-inline` in section 0 as a convenience artifact.

asolntsev:
[...]

PR #2790, opened 2022-11-13
#2789 fix javadoc about mocking final classes

asolntsev:
the described way with `mockito-extensions/org.mockito.plugins.MockMaker` file didn't really work (when spying enums). Now I left the only way working in all cases (dependency `mockito-inline`).

fix for https://github.com/mockito/mockito/issues/2789


TimvdLippe:
We are switching the default mockmaker to `mockito-inline` (#2589), so I think this became obsolete now. Apologies for the delay in reply.

PR #2794, opened 2022-11-21
Upgrade errorprone from 2.14.0 to 2.16

XN137:
we need to fix the following violation:

StubbingWithCustomAnswerTest.java:113: error: [DoNotCall] Calling getClass on Method returns the Class object for Method, you probably meant to retrieve the class containing the method represented by this Method using getDeclaringClass

Supersedes and closes https://github.com/mockito/mockito/pull/2771


TimvdLippe:
Awesome, thanks!

Issue #1593, opened 2019-01-17
Mockito confuses varargs and single-param methods (regression)


lewisd32:
In Mockito 2.2.5 and later, when setting expectations for mocking a varargs method, using an array doesn't work:
```Java
Mockito.when(good.call(Mockito.any(String[].class))).thenReturn(1L);
```
And instead the contained type must be used:
```Java
Mockito.when(good.call(Mockito.any(String.class))).thenReturn(1L);
```
However in the face of an overload of the varargs method that takes a single argument (of the same type):
```Java
    Long call(String value);
    Long call(String... values);
```
This results in being unable to set an expectation for the varargs call.
This worked previously in Mockito 2.2.4 and earlier.

Reproducer: [mockito-bug.tar.gz](https://github.com/mockito/mockito/files/2769668/mockito-bug.tar.gz)

jjDevPL:
I can fix it.

jjDevPL:
try this as workaround:
`Mockito.when(good.call((String[])Mockito.any())).thenReturn(1L);`

jjDevPL:
the problem is matches method from internal class VarArgAware in InstanceOf. It need to be override.

bric3:
Hi @jjDevPL 

Thanks for tracking the version that changed the behavior ! This was done as part of #635.

For other here's the code in the archive : 

```java
interface Good {
    Long call(String... values);
}

interface Bad {
    Long call(String value);
    Long call(String... values);
}

private Good good = Mockito.mock(Good.class);
private Bad bad = Mockito.mock(Bad.class);

private String value = "abc";

// Expectation with `String[]` and call vararg method with a single-element outer-array.
// This passes in Mockito 2.2.4 and earlier
// This fails in Mockito 2.2.5 and later
@Test
public void testMockitoVarArgs1() {
    Mockito.when(good.call(Mockito.any(String[].class))).thenReturn(1L);

    Long count = good.call(new String[]{value});
    assertEquals(1, (long) count);
}

// The failure above can be fixed by changing the expectation to `String` (with the exception of the case below)
// This passes all Mockito versions
@Test
public void testMockitoVarArgs2() {
    Mockito.when(good.call(Mockito.any(String.class))).thenReturn(1L);

    Long count = good.call(new String[]{value});
    assertEquals(1, (long) count);
}

// However, with `bad` having the single-arg method, the fix doesn't work, and there is no way I can find to make
// this pass.
// This fails in all Mockito versions
@Test
[...]

PR #2797, opened 2022-11-23
Hack about to fix #2796

### LINKED ISSUE #2796: Mocking varargs method with `any(String[].class)` doesn't work as expected
This issue is related to a resolved issue  https://github.com/mockito/mockito/issues/2644, which was resolved in 4.7.0 by https://github.com/mockito/mockito/pull/2664.

As per the comment by @perlun in the above PR: https://github.com/mockito/mockito/pull/2664#issuecomment-1319816402

@big-andy-coates I found a similar semi-related issue I think... With 4.7.0, this now works correctly:

```java
        doAnswer(answerVoid((VoidAnswer2<String, Object[]>)
                logger::info
        )).when(mock)
                .info(any(), (Object[]) any());
```

_However_, if I try to use this form which is slightly more ergonomic:

```java
        doAnswer( answerVoid( (VoidAnswer2<String, Object[]>)
                logger::info
        ) ).when( mock )
                .info( any(), any( Object[].class ) );
```

...the mocking will fail to intercept these method calls. :thinking: No error when creating the mock, but it just won't detect the method calls as expected.

I _presume_ this has again something to do with the varargs parsing. `javac` resolves those two calls to the correct/same `logger.info()` overload, so so far so good.

Do you want me to create a separate issue/repro repo for this? I'm not even sure if it's something that can be (easily) fixed, I just happened to see it now while trying to "clean up our code"... :sweat_smile:

_Originally posted by @perlun in https://github.com/mockito/mockito/issues/2664#issuecomment-1319816402_

big-andy-coates:
This PR contains changes NOT intended to be committed 'as-is', but as a showcase for a potential solution to:

* https://github.com/mockito/mockito/issues/2796
* https://github.com/mockito/mockito/issues/1593

And potentially other vararg related issues.

The crux of the issue is that Mockito needs to handle the last matcher passed to a method, when that matcher aligns with a vararg parameter and has the same type as the vararg parameter.

For example,

```java
public interface Foo {
 String m1(String... args);  // Vararg param at index 0 and type String[]
}

@Test
public void shouldWork2() throws Exception  {
  // Last matcher at index 0, and with type String[]: needs special handling!
  given(foo.m1(any(String[].class))).willReturn("var arg method");
  ...
}
```

In such situations that code needs to match the raw argument, _not_ the current functionality, which is to use the last matcher to match the last _non raw_ argument.

Unfortunately, I'm not aware of a way to get at the type of the matcher without adding a method to `VarargMatcher` to get this information. This is the downside of this approach.

The PR includes the tests suggested in https://github.com/mockito/mockito/pull/1236, though one still fails (and is commented out with explanation).

Crux of the change is in `MatcherApplicationStrategy`. Obviously, it needs some refactoring. Not sure if you'd want the refactor as a separate PR or as the first commit of the main PR????

Alternatively, I could resurrect https://github.com/mockito/mockito/pull/1224 and take it through to merging.

### DIFF
--- a/src/test/java/org/mockitousage/stubbing/StubbingWithAdditionalAnswersTest.java
+++ b/src/test/java/org/mockitousage/stubbing/StubbingWithAdditionalAnswersTest.java
@@ -28,6 +28,8 @@ import java.util.Date;
 
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.ArgumentMatchers;
 import org.mockito.Mock;
 import org.mockito.exceptions.base.MockitoException;
 import org.mockito.junit.MockitoJUnitRunner;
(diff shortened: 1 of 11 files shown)


src/test/java/org/mockitousage/stubbing/StubbingWithAdditionalAnswersTest.java
```diff
@@ -56,12 +58,114 @@ public void can_return_arguments_of_invocation() throws Exception {
         given(iMethods.objectArgMethod(any())).will(returnsFirstArg());
         given(iMethods.threeArgumentMethod(eq(0), any(), anyString())).will(returnsSecondArg());
         given(iMethods.threeArgumentMethod(eq(1), any(), anyString())).will(returnsLastArg());
-        given(iMethods.mixedVarargsReturningString(eq(1), any())).will(returnsArgAt(2));
 
         assertThat(iMethods.objectArgMethod("first")).isEqualTo("first");
         assertThat(iMethods.threeArgumentMethod(0, "second", "whatever")).isEqualTo("second");
         assertThat(iMethods.threeArgumentMethod(1, "whatever", "last")).isEqualTo("last");
-        assertThat(iMethods.mixedVarargsReturningString(1, "a", "b")).isEqualTo("b");
```
TimvdLippe, line 64: Is this the only stub that would now start failing?

big-andy-coates:
That removal was not intentional. 99.99% sure it'll pass if I put it back.

src/test/java/org/mockitousage/stubbing/StubbingWithAdditionalAnswersTest.java
big-andy-coates: Unintentional removal - I'll add it back in.

big-andy-coates:
Questions for me are:
1. Are we happy with the API change to `VarargMatcher`? This could be either:
  1.  with a default implementation that returns `Object.class`, which would NOT be a breaking change for types implementing the interface, but the default return value is a bit ... meh!
  2. without a default implementation, which IS a breaking change...
2.  Are we happy with the API change to `CapturingMatcher`?  i.e. the addition of the constructor taking the type.
     If this type is considered part of the public API, then probably best to add a (deprecated?) default constructor as well, which sets `clazz` to `Object.class`.


Everything else seems internal implementation details.

Would be great if anyone out there with more context / experience can raise any edge cases that may have been missed. I can see @ChristianSchwarz, @mockitoguy, @paulduffin have all worked in this area...)

src/test/java/org/mockitousage/stubbing/StubbingWithAdditionalAnswersTest.java
TimvdLippe: Ah great! In that case, if all existing tests pass and these additional cases also pass, I am up for this! I have not reviewed your implementation yet, but as long as we can keep all existing regression tests passing I am happy 😄

TimvdLippe:
> 1. Are we happy with the API change to `VarargMatcher`? This could be either:
>    
>    1. with a default implementation that returns `Object.class`, which would NOT be a breaking change for existing implementations, but the default return value is a bit ... meh!
>    2. without a default implementation, which IS a breaking change...

These matchers are in `org.mockito.internal`, so we don't consider them part of our public API. Therefore, we can technically make any breaking change we want in them. However, if we can avoid large breakages (because I assume some of our users will use them, especially `VarargMatcher` which is known on StackOverflow to be used https://stackoverflow.com/a/7988918). Therefore, I think if we avoid any breakage by return `Object.class`, I would say let's go for that.

Looking at your implementation, I think the overall design looks fine. There are probably a couple of nits, but I like the general direction.

big-andy-coates:
I'll probably create a new PR for the actual work, starting afresh.  I'm going to refactor `MatcherApplicationStrategy` first, as its weirdly overly complicated.  It could just be a utility method.

Would you prefer I do that in a separate PR, or just a separate commit in main PR?

TimvdLippe:
@big-andy-coates Let's do separate PRs, but please keep refactoring to a minimum. Refactoring `MatcherApplicationStrategy` seems good to me, but let's keep it limited to that. I would prefer not to overhaul the system too much here.

PR #2805, opened 2022-11-28
Improve vararg handling

big-andy-coates:
Fixes: https://github.com/mockito/mockito/issues/2796

An alternative approach is available in https://github.com/mockito/mockito/pull/2807.

Add an optional method to `VarargMatcher`, which implementations can choose to override to return the type of object the matcher is matching.

This is used by `MatcherApplicationStrategy` to determine if the type of matcher used to match a vararg parameter is of a type compatible with the vararg parameter.

Where a vararg compatible matcher is found, the matcher is used to match the _raw_ parameters.

## Known limitations

The main limitation I'm aware of, is not a new limitation. It is that it is not possible to assert only a single parameter is passed to the vararg parameter, when using a `VarargMatcher`. (ref: https://github.com/mockito/mockito/issues/1593). For example:

```java
// Given method:
int vararg(String... args);

// I want to mock this invocation:
mock.vararag("one param");

// ...but not these:
mock.vararg();
mock.vararg("more than", "one param");
```

There is no current way to do this.  This is because in the following intuitive mocking:

```java
given(mock.vararg(any(String.class))).willReturn(1);
```

... matches zero or more vararg parameters, as the `any()` method is using `VarargMatcher`.

It seems to me that `VarargMatcher` is... a little broken.  This PR leverages `VarargMatcher`... which, if its broken, might not be something we want to do.

## Checklist

### DIFF
--- a/src/main/java/org/mockito/ArgumentCaptor.java
+++ b/src/main/java/org/mockito/ArgumentCaptor.java
@@ -62,11 +62,12 @@ import org.mockito.internal.matchers.CapturingMatcher;
 @CheckReturnValue
 public class ArgumentCaptor<T> {
 
-    private final CapturingMatcher<T> capturingMatcher = new CapturingMatcher<T>();
+    private final CapturingMatcher<T> capturingMatcher;
     private final Class<? extends T> clazz;
 
     private ArgumentCaptor(Class<? extends T> clazz) {
         this.clazz = clazz;
+        this.capturingMatcher = new CapturingMatcher<T>(clazz);
     }
 
     /**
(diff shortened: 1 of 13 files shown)


big-andy-coates:
@TimvdLippe here's one possible solution. Please take a look when you get time.  However, please don't merge if you're happy just yet, as I'd like to try out another possible solution.  One that removes the need for `VarargMatcher` all together....

big-andy-coates:
@TimvdLippe alternative approach raised in https://github.com/mockito/mockito/pull/2807.

Issue #1498, opened 2018-09-24
(Argument)Matchers regression from 1.10.19 to 2.18.3 for varargs


ViToni:
While updating from 1.10.19 to 2.18.3 I have a regression regarding Matchers and varargs.
Passing tests now fail because the Matchers don't work anymore in the verify step.

Below is a reduced test case to reprocude the issue. Version 2.22.0 and ArgumentMatcher were also tested with the same (failing) result.

Error:

	Argument(s) are different! Wanted:
	spyableLogger.debug(
		<any string>,
		<any java.lang.Object[]>
	);
	-> at tests.MockitoVarargsTest.testLoggerGetsCalled_Matchers(MockitoVarargsTest.java:29)
	Actual invocation has different arguments:
	spyableLogger.debug(
		"Something happened",
		java.lang.RuntimeException: Oops
	);
	-> at tests.MockitoVarargsTest.callLoggerDebug(MockitoVarargsTest.java:59)

		at tests.MockitoVarargsTest.testLoggerGetsCalled_Matchers(MockitoVarargsTest.java:29)

Test case:

    package tests;
    
    import java.util.Objects;
    
    import org.junit.Test;
    //import org.mockito.ArgumentMatchers;
    import org.mockito.Matchers;
    import org.mockito.Mockito;
    import org.slf4j.Logger;
    import org.slf4j.LoggerFactory;
    import org.slf4j.Marker;
    
    public class MockitoVarargsTest {
        
        private final Logger logger = LoggerFactory.getLogger(getClass());
    
        @Test
        public void testLoggerGetsCalled_Matchers() {
            final Logger spyableLogger = new SpyableLogger(logger);
            final Logger loggerSpy = Mockito.spy(spyableLogger);
    
            final RuntimeException e = new RuntimeException("Oops");
    
            final String message = "Something happened";
            callLoggerDebug(loggerSpy, message, e);
    
            // Mockito 1.10.19 works
            // Mockito 2.18.3 fails
            Mockito.verify(loggerSpy, Mockito.times(1)).debug(Matchers.anyString(), Matchers.any(Object[].class));
            Mockito.verify(loggerSpy, Mockito.never()).trace(Matchers.anyString(), Matchers.any(Object[].class));
            Mockito.verify(loggerSpy, Mockito.never()).warn(Matchers.anyString(), Matchers.any(Object[].class));
[...]

Issue #567, opened 2016-08-19
Verification of an empty varargs call fails when isNotNull() is used


ChristianSchwarz:
The follwing tests fail unexpected with : "Argument(s) are different"

```
@Test
public void shouldMatchVarArgs_isNotNull()   {
    mock.varargs();
    verify(mock).varargs(isNotNull());
}

@Test
public void shouldMatchVarArgs_isNotNullArray()   {
    mock.varargs();

    verify(mock).varargs((String[])isNotNull());
}
```

Expected is that the tests pass cause an empty vararg array is not null!

bric3:
Thanks for the report

big-andy-coates:
With https://github.com/mockito/mockito/pull/2807 merged, the second test above can be achieved, not through casts, but through using the variant of `isNotNull` that takes a type:

```java
@Test
public void shouldMatchVarArgs_isNotNullArray()   {
    mock.varargs();

    verify(mock).varargs((String[])isNotNull(String[].class));
}
```

However, the first test is invalid:

```java
@Test
public void shouldMatchVarArgs_isNotNull()   {
    mock.varargs();
    verify(mock).varargs(isNotNull());
}
```

This previously would have matched a `varargs()` call with 1 or more parameters, and now matches a call with exactly 1 parameter. Hence the `verify` call would fail in both cases as the method is called with no parameters.

PR #2807, opened 2022-11-28
Improve vararg handling: approach 2

big-andy-coates:
Fixes: https://github.com/mockito/mockito/issues/2796
Fixes: https://github.com/mockito/mockito/issues/567
Fixes: https://github.com/mockito/mockito/issues/584
Fixes: https://github.com/mockito/mockito/issues/1222
Fixes: https://github.com/mockito/mockito/issues/1498

An alternative approach to https://github.com/mockito/mockito/pull/2805. 

## Key difference:

This approach adds the new `type()` method to the main `ArgumentMatcher` interface and leaves the, IMHO broken, `VarargsMatcher` interface alone.  

## Benefit:

Because this approach leaves `VarargsMatcher` untouched, it does not require additional existing matchers to implement `VarargsMatcher` to fix issues such as https://github.com/mockito/mockito/issues/567. Where as the first PR would require `Null` and `NotNull` to be marked `VarargsMatcher`.

This PR creates new variants of `isNotNull` and `isNull` to address https://github.com/mockito/mockito/issues/567. 
Having `InstanceOf` override `type()` provides a workable solution to https://github.com/mockito/mockito/issues/1593.
Having `equals` override `type` addresses https://github.com/mockito/mockito/issues/1222.

If we want to go with this approach I'll implement `type()` for all the other existing matchers.

## Downside

The obvious downside is that this changes the public `ArgumentMatcher` interface, though in a backwards compatible way.

The default implementation _could_ be dropped in the next major version bump.

## Known limitation

The main limitation I'm aware of, is not a new limitation. It is that it is not possible to assert only a single parameter is passed to the vararg parameter, when using a `VarargMatcher`, e.g. `any()`. (ref: https://github.com/mockito/mockito/issues/1593). For example:

```java
// Given method:
int vararg(String... args);

// I want to mock this invocation:
mock.vararag("one param");

// ...but not these:
mock.vararg();
mock.vararg("more than", "one param");
```

There is no current way to do this.  This is because in the following intuitive mocking:

```java
given(mock.vararg(any(String.class))).willReturn(1);
```

... matches zero or more vararg parameters, as the `any()` method is using `VarargMatcher`. It seems to me that `VarargMatcher` is... a little broken!  This is maybe something that should be consider a candiate for fixing in the next major version bump.  

While it is not possible to fix any `VarargMatcher` based matchers in a backwards compatible way, this the approach in this PR it is possible to mock/verify exactly one vararg param using `isA`, rather than `any`:

```java
    @Test
    public void shouldMatchExactlyOnParam() {
        mock.varargs("one param");

        verify(mock).varargs(isA(String.class));
    }

    @Test
    public void shouldNotMatchMoreParams() {
        mock.varargs("two", "params");

        verify(mock, never()).varargs(isA(String.class));
    }

    @Test
    public void shouldMatchAnyNumberOfParams() {
        mock.varargs("two", "params");

        verify(mock).varargs(isA(String[].class));
    }

```

... because `isA` does not implement `VarargsMatcher`, and so can work as expected once it implements `type()`.

## Checklist


big-andy-coates:
@TimvdLippe Personally, I prefer this approach.

src/main/java/org/mockito/ArgumentMatcher.java
```diff
@@ -125,4 +125,48 @@
...
+     *  // (as there are more matchers than raw arguments)
+     *  mock.someVarargMethod(eq(1), any(), any());
+     * </code></pre>
+     *
+     * @return the type this matcher handles.
+     * @since 4.10.0
+     */
+    default Class<?> type() {
+        return Void.class;
+    }
```
big-andy-coates, line 171: Public API change...

Default implementation is only really there to maintain backwards compatibility,.

src/main/java/org/mockito/internal/invocation/MatcherApplicationStrategy.java
```diff
@@ -58,17 +58,24 @@ public static MatcherApplicationStrategy getMatcherApplicationStrategyFor(
...
+                        && invocation.getRawArguments().length == matchers.size();
 
         if (isVararg) {
-            int times = varargLength(invocation);
+            final Class<?> matcherType = lastMatcher().type();
+            final Class<?> paramType = lastParameterType();
+            if (paramType.isAssignableFrom(matcherType)) {
+                return argsMatch(invocation.getRawArguments(), matchers, action);
+            }
+        }
```
big-andy-coates, line 71: Main matching logic change.

Similar to the other PR, but this time the code doesn't require the matcher to implement `VarargAware` matcher.

src/main/java/org/mockito/internal/matchers/InstanceOf.java
```diff
@@ -30,6 +30,11 @@ public boolean matches(Object actual) {
                         || clazz.isAssignableFrom(actual.getClass()));
     }
 
+    @Override
+    public Class<?> type() {
+        return clazz;
+    }
```
big-andy-coates, line 36: Exposing this provides solution to https://github.com/mockito/mockito/issues/1593

src/main/java/org/mockito/internal/matchers/NotNull.java
```diff
@@ -8,17 +8,26 @@
...
 
     @Override
     public boolean matches(Object actual) {
         return actual != null;
     }
 
+    @Override
+    public Class<T> type() {
+        return type;
+    }
```
big-andy-coates, line 29: Exposing this fixes https://github.com/mockito/mockito/issues/567

src/main/java/org/mockito/internal/matchers/Null.java
```diff
@@ -8,17 +8,25 @@
...
 
     @Override
     public boolean matches(Object actual) {
         return actual == null;
     }
 
+    @Override
+    public Class<T> type() {
+        return type;
+    }
```
big-andy-coates, line 28: Exposing this fixes https://github.com/mockito/mockito/issues/567

big-andy-coates (commented):
Added some review notes...

big-andy-coates:
OK @TimvdLippe , I think that's all the implementations of `ArgumentMatcher` it makes sense to add the `type()` method to.

Look forward to a review when you have time.

src/main/java/org/mockito/ArgumentMatcher.java
```diff
@@ -125,4 +125,44 @@
...
+     *  // The following will match invocations with two parameters, i.e. two strings in the raw array
+     *  mock.someVarargMethod(isA(String.class), isA(String.class));
+     * </code></pre>
+     *
+     * <p>Only matcher implementations that can conceptually match a raw vararg parameter should override this method.
+     *
+     * @return the type this matcher handles. The default value of {@link Void} means the type is not known.
+     * @since 4.10.0
+     */
+    default Class<?> type() {
```
TimvdLippe, line 165: With just 1 method, `ArgumentMatcher` automatically qualifies as a `FunctionalInterface`. E.g. it can be written as a clean lambda. With the addition of this method, can users still write an `ArgumentMatcher` as a lambda? In case we remove the default implementation, is that also true?

If that is not possible, we can consider splitting up the argument matchers into two. We can keep `ArgumentMatcher` as is, but we define a new interface `ArgumentMatcherWithType` that extends `ArgumentMatcher`, but does not define a default implementation. Then we can implement this interface in all our internal matchers, but external users can keep on using `ArgumentMatcher` as-is.

In the logic below, we would then still check if the `ArgumentMatcher` we have is also `ArgumentMatcherWithType` and apply the logic there. WDYT?

TimvdLippe:
I haven't done a thorough review yet, as I mostly want to discuss the change to `ArgumentMatcher` first. Ideally, I want to keep that interface as it is, so that the experience for our users remains the same and they can keep on using lambdas. We can then make our internal matchers smarter for varargs, but avoid too many breakages. Does that sound feasible to you or would you like me to look at other parts of the PR first?

src/main/java/org/mockito/ArgumentMatcher.java
TimvdLippe: Now that I realize it, this comment would basically boil down to: add a new method to `VarargMatcher` that defines its type. That sounds quite natural to me and would be a minimally impactful change, yet still provide significant value by solving all the issues that you linked.

big-andy-coates: Having the new defaulted method doesn't change things. In fact I can mark it as `@FunctionalInterface` if you want. Users can still use lambdas.

Yes, removing the default would mean it's not longer functional and hence users could not use lambdas. However, my latest thinking is that we should not remove the default implementation.  The default implementation's return value means the type doesn't conceptually have a base type its comparing to. 

Yes, we could achieve similar functionality by splitting into the two interfaces you mention. But, adding a defaulted method is cleaner: it doesn't require casts in the code and is more intuitive / more descriptive / more discoverable, IMHO.

big-andy-coates: > Now that I realize it, this comment would basically boil down to: add a new method to `VarargMatcher` that defines its type.

This was my original approach. 

Currently, is a VarargMatcher is last argument, it is used to match against each additional element in a vararg parameter. Unfortunately, this leaves edgecases that's can't be handled.

We could repurpose `VarargMatcher` to be the second interface we need, fixing the implementation. But I'm still in favour of the defaulted method on `ArgumentMatcher`.

big-andy-coates: See #2805 for an approach adding `type()` to `VarargMatcher`.

Though, if we go this way, we should remove the default impl.

TimvdLippe: Okay, given the amount of issues this approach fixes and the fact that this can still be written as a lambda, I am okay with this approach. Thank you for writing down the considerations and trade-offs, that helped a lot in figuring out on how to proceed.

src/main/java/org/mockito/internal/matchers/VarargMatcher.java
```diff
@@ -9,5 +9,14 @@
...
+ * @deprecated use of this interface is deprecated as the behaviour it promotes has limitations.
+ * It is not recommended for new implementations to implement this method.
+ *
+ * <p>Instead, matchers should implement the {@link org.mockito.ArgumentMatcher#type()} method.
+ * If this method returns the same raw type as a vararg parameter, then Mockito will treat the
+ * matcher as matching the entire vararg array parameter, otherwise it will be treated as matching a single element.
+ * For an example, see {@link org.mockito.ArgumentMatchers#isNull(Class)}.
  */
+@Deprecated
 public interface VarargMatcher extends Serializable {}
```
TimvdLippe, line 22: With this new solution, I think we can remove this interface now and include it in Mockito 5

src/test/java/org/mockito/internal/matchers/AndTest.java
```diff
@@ -0,0 +1,62 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.ArgumentMatcher;
+import org.mockito.Mock;
+import org.mockitoutil.TestBase;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.BDDMockito.given;
+
+public class AndTest extends TestBase {
```
TimvdLippe, line 16: Love these tests!

src/test/java/org/mockitousage/matchers/VarargsTest.java
```diff
@@ -304,24 +300,15 @@ public void shouldNotCaptureVarArgs_1args2captures() {
...
-    /**
-     * As of v2.0.0-beta.118 this test fails. Once the github issues:
-     * <ul>
-     * <li>'#584 ArgumentCaptor can't capture varargs-arrays
-     * <li>#565 ArgumentCaptor should be type aware' are fixed this test must
-     * succeed
-     * </ul>
-     */
     @Test
-    @Ignore("Blocked by github issue: #584 & #565")
```
TimvdLippe, line 316: Does this mean this PR also fixes #565?

src/test/java/org/mockitousage/matchers/VarargsTest.java
```diff
@@ -342,8 +329,281 @@ public void shouldNotMatchVaraArgs() {
...
+
+    @Test
+    public void shouldVerifyExactlyOneVarArg_any() {
+        mock.varargs("one param");
+
+        verify(mock).varargs(any(String.class));
+    }
+
+    @Test
+    @Ignore("Fails due to https://github.com/mockito/mockito/issues/1593")
```
TimvdLippe, line 492: Your PR description claims this is fixed. Can we unignore this test?

TimvdLippe (commented):
I think we are almost there 😄

src/test/java/org/mockitousage/matchers/VarargsTest.java
big-andy-coates: in short, no.

This PR addresses #584, which is what this test tests.
#565 is not vararg related. Not sure why this test as tagged with the issue as it doesn't seem related.

src/test/java/org/mockitousage/matchers/VarargsTest.java
big-andy-coates: Ah, good catch. #1593 isn't resolved until we remove the broken `VarargMatcher` "use the last vararg matcher to match all additional parameters".  I've remove this from the fixed list.

big-andy-coates:
@TimvdLippe I've updated the `since` to reflect the new target release of `5.0.0` and I've marked `ArgumentMatcher` interface as `@FunctionalInterface`.

If possible, I'd prefer we merge this, and I look to remove the `VarargMatcher` interface as a separate PR. Is that OK, or would you rather I remove `VarargMatcher` in this PR?

src/main/java/org/mockito/internal/matchers/VarargMatcher.java
big-andy-coates: As per [comment below](https://github.com/mockito/mockito/pull/2807#issuecomment-1362341232) I'm prefer to remove this in a separate PR, if you don't mind?

big-andy-coates: See https://github.com/mockito/mockito/pull/2807#issuecomment-1362341232

TimvdLippe (approved):
Woohoo, great work!

big-andy-coates:
> @big-andy-coates Yes that's okay to me. It might still be a good idea to include the removal of the interface into Mockito 5.0.0 no? Since it removes an interface, even if it is internal, I think it is cleanest to include in a major release.

Yes, I'll get this done asap to include in the 5.0.0 release.

big-andy-coates:
See https://github.com/mockito/mockito/pull/2835 for pr removing `VarargMatcher`

cpovirk:
This PR does end up introducing an incompatible change to `CapturingMatcher`: It changes the constructor from `CapturingMatcher()` to `CapturingMatcher(Class)`. We don't appear to have very many `CapturingMatcher` subclasses inside Google, so it looks like we'll be able to update the ones that we do have atomically when we pull in this update.

I gather that this change might not land until the 5.0.0 major release, in which case that's obviously the time to make incompatible changes :) You may want to note the change in a `@since` tag (and perhaps release notes, as you might already be planning to do). You could go even further by introducing the new constructor in a release before 5.0.0 so that people at least don't have to update their code atomically. But unless an organization is a much heavier user of `CapturingMatcher` than we are, that's not necessarily worth the trouble.

cpovirk:
Oh, sorry, scratch all that: `CapturingMatcher` is under `internal`, so we are Doing It Wrong :) (I got it mixed up with the changes to the classes in `org.mockito` itself.)

TimvdLippe:
Ah yeah we don't consider `org.mockito.internal` as part of SemVer, so ideally users don't use these internal matchers. Can you port these implementations to use only public API? If not, please file a GitHub issue so we can consider the use case.

cpovirk:
Thanks. I have to imagine they're possible to port. They might actually mostly be implementations of `isA` from before that existed, and I imagine a plain capture will suffice for the rest.

That's not to say that we'll actually do it anytime soon. (I'm sure you of all people are shocked... :)) But I should at least more readily remember how much this is our fault next time.

TimvdLippe:
Haha yeah all good. I am sure you can find some of internal tickets with more explanation on how to tackle that problem. Feel free to message me if you need more information from my side.

cpovirk:
Thanks, I had already come across one :)

I will also report that people _really_ like writing `ArgumentCaptor.forClass(null)`, which now throws NPE. But that looks manageable, too.

dacianf:
Hey @big-andy-coates @TimvdLippe,

I'm trying Mockito 5.2 in Android using Kotlin and I got over a problem with varg. 

Let's say we have the following interface for getting strings:

```
interface TestVargsMock {
    fun getString(res: Int, vararg args: String): String
}
```

When I try to mock it using:

```
whenever(testVargsMock.getString(anyInt(), any()))
            .thenAnswer {
            "It works"
        }
```
works just fine, as expected; it checks if only one argument was used

but if I try `whenever(testVargsMock.getString(anyInt(), any(Array<out String>::class.java)))` it doesn't compile because of a type mismatch error: `Type mismatch: inferred type is Array<out String>! but String was expected`

if I try to mock any of string `whenever(testVargsMock.getString(anyInt(), any(String::class.java)))` mocking is failing with `java.lang.NullPointerException: any(String::class.java) must not be null`

I don't know how should I express `String[].class` in kotlin so the mocking will work. Is there a way to do it or this is a bug that needs to be fixed?

TimvdLippe:
@dacianf For Kotlin related questions, please ask it over at `mockito-kotlin`: https://github.com/mockito/mockito-kotlin/issues/474 Any help appreciated there to fix the Kotlin support!

PR #2834, opened 2022-12-23
Switch the default mockmaker to the inline mockmaker on JDK 17+

reta:

Fixes https://github.com/mockito/mockito/issues/2589

## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/configuration/plugins/DefaultMockitoPlugins.java
+++ b/src/main/java/org/mockito/internal/configuration/plugins/DefaultMockitoPlugins.java
@@ -28,13 +28,14 @@ public class DefaultMockitoPlugins implements MockitoPlugins {
     static final String SUBCLASS_ALIAS = MockMakers.SUBCLASS;
     public static final Set<String> MOCK_MAKER_ALIASES = new HashSet<>();
     static final String MODULE_ALIAS = "member-accessor-module";
+    static final String REFLECTION_ALIAS = "member-accessor-reflection";
 
     static {
         // Keep the mapping: plugin interface name -> plugin implementation class name
         DEFAULT_PLUGINS.put(PluginSwitch.class.getName(), DefaultPluginSwitch.class.getName());
         DEFAULT_PLUGINS.put(
                 MockMaker.class.getName(),
-                "org.mockito.internal.creation.bytebuddy.ByteBuddyMockMaker");
+                "org.mockito.internal.creation.bytebuddy.InlineByteBuddyMockMaker");
         DEFAULT_PLUGINS.put(
                 StackTraceCleanerProvider.class.getName(),
                 "org.mockito.internal.exceptions.stacktrace.DefaultStackTraceCleanerProvider");
@@ -53,9 +54,11 @@ public class DefaultMockitoPlugins implements MockitoPlugins {
                 MockitoLogger.class.getName(), "org.mockito.internal.util.ConsoleMockitoLogger");
         DEFAULT_PLUGINS.put(
                 MemberAccessor.class.getName(),
-                "org.mockito.internal.util.reflection.ReflectionMemberAccessor");
+                "org.mockito.internal.util.reflection.ModuleMemberAccessor");
         DEFAULT_PLUGINS.put(
                 MODULE_ALIAS, "org.mockito.internal.util.reflection.ModuleMemberAccessor");
+        DEFAULT_PLUGINS.put(
+                REFLECTION_ALIAS, "org.mockito.internal.util.reflection.ReflectionMemberAccessor");
         DEFAULT_PLUGINS.put(
                 DoNotMockEnforcer.class.getName(),
                 "org.mockito.internal.configuration.DefaultDoNotMockEnforcer");
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/InlineBytecodeGenerator.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/InlineBytecodeGenerator.java
@@ -205,6 +205,7 @@ public class InlineBytecodeGenerator implements BytecodeGenerator, ClassFileTran
         boolean subclassingRequired =
                 !features.interfaces.isEmpty()
                         || features.serializableMode != SerializableMode.NONE
+                        || features.stripAnnotations
                         || Modifier.isAbstract(features.mockedType.getModifiers());
 
         checkSupportedCombination(subclassingRequired, features);
@@ -416,6 +417,7 @@ public class InlineBytecodeGenerator implements BytecodeGenerator, ClassFileTran
         }
         mocked.clear();
         flatMocked.clear();
+        subclassEngine.clearAllCaches();
         try {
             instrumentation.retransformClasses(types.toArray(new Class<?>[0]));
         } catch (UnmodifiableClassException e) {
--- a/src/main/java/org/mockito/internal/exceptions/stacktrace/DefaultStackTraceCleaner.java
+++ b/src/main/java/org/mockito/internal/exceptions/stacktrace/DefaultStackTraceCleaner.java
@@ -24,11 +24,18 @@ public class DefaultStackTraceCleaner implements StackTraceCleaner {
     private boolean isIn(String className) {
         if (isFromMockitoRunner(className) || isFromMockitoRule(className)) {
             return true;
-        } else if (isMockDispatcher(className) || isFromMockito(className)) {
+        } else if (isMockDispatcher(className)
+                || isFromMockito(className)
+                || isMethodHandle(className)) {
             return false;
         } else {
             return true;
         }
+    }
+
+    /* Some mock makers (like inline) use java.lang.invoke.MethodHandle to dispatch calls */
+    private static boolean isMethodHandle(String className) {
+        return className.startsWith("java.lang.invoke.MethodHandle");
     }
 
     private static boolean isMockDispatcher(String className) {
--- a/src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsSmartNulls.java
+++ b/src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsSmartNulls.java
@@ -8,9 +8,12 @@ import static org.mockito.internal.exceptions.Reporter.smartNullPointerException
 import static org.mockito.internal.util.ObjectMethodsGuru.isToStringMethod;
 
 import java.io.Serializable;
+import java.lang.reflect.Method;
+import java.util.Arrays;
 
 import org.mockito.Mockito;
 import org.mockito.internal.creation.MockSettingsImpl;
+import org.mockito.internal.creation.bytebuddy.MockAccess;
 import org.mockito.internal.debugging.LocationFactory;
 import org.mockito.internal.util.MockUtil;
 import org.mockito.invocation.InvocationOnMock;
@@ -89,9 +92,28 @@ public class ReturnsSmartNulls implements Answer<Object>, Serializable {
             if (isToStringMethod(currentInvocation.getMethod())) {
                 return "SmartNull returned by this unstubbed method call on a mock:\n"
                         + unstubbedInvocation;
+            } else if (isMethodOf(
+                    MockAccess.class, currentInvocation.getMock(), currentInvocation.getMethod())) {
+                /* The MockAccess methods should be called directly */
+                return currentInvocation.callRealMethod();
             }
 
             throw smartNullPointerException(unstubbedInvocation.toString(), location);
+        }
+
+        private static boolean isMethodOf(Class<?> clazz, Object instance, Method method) {
+            if (!clazz.isInstance(instance)) {
+                return false;
+            }
+
+            for (Method m : clazz.getDeclaredMethods()) {
+                if (m.getName().equalsIgnoreCase(method.getName())
+                        && Arrays.equals(m.getParameterTypes(), method.getParameterTypes())) {
+                    return true;
+                }
+            }
+
+            return false;
         }
     }
 }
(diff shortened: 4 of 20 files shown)


TimvdLippe:
Seems like there are a couple of test failures :( Some of these look cryptic with the missing MethodDispatcher file. We had that before and I thought we fixed that? Do we need the fix that we did for `mockito-inline` maybe: https://github.com/mockito/mockito/blob/main/subprojects/inline/src/main/resources/mockito-extensions/org.mockito.plugins.MemberAccessor

reta:
> Seems like there are a couple of test failures :( Some of these look cryptic with the missing MethodDispatcher file. We had that before and I thought we fixed that?

Thanks @TimvdLippe , yeah, I am looking into that but it seems like we have to change not only `MockMaker` default but the `MemberAccessor` as well (as you pointed out) to work with the new defaults.

subprojects/module-test/src/test/java/org/mockito/moduletest/ReplicatingClassLoader.java
```diff
@@ -53,4 +53,9 @@ public Class<?> loadClass(String name) throws ClassNotFoundException {
     protected URL findResource(String moduleName, String name) {
         return Mockito.class.getResource("/" + name);
     }
+
+    @Override
+    public InputStream getResourceAsStream(String name) {
+        return Mockito.class.getResourceAsStream("/" + name);
```
reta, line 59: Fixing `module-test` tests

src/main/java/org/mockito/internal/creation/bytebuddy/InlineBytecodeGenerator.java
```diff
@@ -416,6 +416,7 @@ public synchronized void clearAllCaches() {
         }
         mocked.clear();
         flatMocked.clear();
+        subclassEngine.clearAllCaches();
```
reta, line 419: Fixing `MockitoClearTest` tests

src/test/java/org/mockitousage/configuration/ClassCacheVersusClassReloadingTest.java
```diff
@@ -74,7 +74,8 @@ private static SimplePerRealmReloadingClassLoader.ReloadClassPredicate reloadMoc
         return new SimplePerRealmReloadingClassLoader.ReloadClassPredicate() {
             public boolean acceptReloadOf(String qualifiedName) {
                 return (!qualifiedName.contains("net.bytebuddy")
-                        && qualifiedName.contains("org.mockito"));
+                        && qualifiedName.contains("org.mockito")
+                            && !qualifiedName.contains("org.mockito.internal.creation.bytebuddy.inject"));
```
reta, line 78: Fixing `ClassCacheVersusClassReloadingTest`

src/main/java/org/mockito/internal/configuration/plugins/DefaultMockitoPlugins.java
```diff
@@ -53,7 +53,7 @@ public class DefaultMockitoPlugins implements MockitoPlugins {
                 MockitoLogger.class.getName(), "org.mockito.internal.util.ConsoleMockitoLogger");
         DEFAULT_PLUGINS.put(
                 MemberAccessor.class.getName(),
-                "org.mockito.internal.util.reflection.ReflectionMemberAccessor");
+                "org.mockito.internal.util.reflection.ModuleMemberAccessor");
```
TimvdLippe, line 56: Since we flip the default here, I suppose we have to update the `mockito-subclass` artifact to set the `MemberAccessor` to reflection as well.

reta: > Since we flip the default here, I suppose we have to update the `mockito-subclass` artifact to set the `MemberAccessor` to reflection as well.

The `mockito-subclass`  seems to define member accessor [1] but `mockito-android` and `mockito-proxy` indeed relied on reflection, changed that, thanks!

[1] https://github.com/mockito/mockito/blob/main/subprojects/subclass/src/main/resources/mockito-extensions/org.mockito.plugins.MemberAccessor

src/main/java/org/mockito/internal/creation/bytebuddy/InlineBytecodeGenerator.java
```diff
@@ -205,6 +205,7 @@ public <T> Class<? extends T> mockClass(MockFeatures<T> features) {
         boolean subclassingRequired =
                 !features.interfaces.isEmpty()
                         || features.serializableMode != SerializableMode.NONE
+                        || features.stripAnnotations
```
reta, line 208: Fixing `MocksCreationTest` test cases

src/main/java/org/mockito/internal/util/reflection/InstrumentationMemberAccessor.java
```diff
@@ -188,7 +188,16 @@ public Object newInstance(
...
+                                    // Use handle.asFixedArity() to handle varargs bindings properly
+                                    // (as by default Java will create method handle with
+                                    // asVarargsCollector), fe:
+                                    //
+                                    //   private static class Varargs {
+                                    //     Varargs(String whatever, Observer... observers) {
+                                    //     }
+                                    //   }
+                                    return DISPATCHER.invokeWithArguments(
+                                            handle.asFixedArity(), arguments);
```
reta, line 200: Fixing `ParameterizedConstructorInstantiatorTest` tests (varargs bindings)

TimvdLippe:
@reta Thanks for all the work thus far! This is the last PR for Mockito 5, is there something I can help you out with?

reta:
> @reta Thanks for all the work thus far! This is the last PR for Mockito 5, is there something I can help you out with?

Thanks @TimvdLippe , I am slowly crunching over test cases (it takes time for me to troubleshoot some of them ) so far there are just few left in Mockito core, I hope to fix them in the next couple of days:

```
    DefaultInternalRunnerTest. does_not_fail_second_test_when_first_test_fail
    ReturnsSmartNullsTest. should_return_a_new_mock_that_has_been_defined_with_method_generic_and_provided_in_argument
    ReturnsSmartNullsTest. should_return_a_new_mock_that_has_been_defined_with_method_generic_and_provided_in_var_args
    ReturnsSmartNullsTest. should_return_an_object_that_has_been_defined_with_class_generic
    ParameterizedConstructorInstantiatorTest. should_fail_if_an_argument_instance_type_do_not_match_wanted_type
    PartialMockingWithSpiesTest. shouldStackTraceGetFilteredOnUserExceptions
```

The real blocker it seems like is OSGi, where inline mock maker may not be used as a default: the issue is that `InlineDelegateByteBuddyMockMaker` creates `mockitoboot.jar` and modifies classpath at runtime, not very OSGi friendly way. If you have time to look at it (`osgi-test` module), that would be very helpful. 

Thank you!

TimvdLippe:
@raphw do you have some more context on the previous comment with regards to OSGi?

raphw:
Hmm, the only possible solution I see would be to move `MockMethodDispatcher` to the `java.lang` package. This would then also require that we use unsafe to inject the class instead of the official API. This way, we would no longer be fully Java API compatible.

Alternatively, we could instrument all class loaders to always load the dispatcher class from the boot loader to possibly override any custom behavior. But this is rather intrusive. Maybe we could make this configurable?

What do you think would be the best way forward?

src/main/java/org/mockito/internal/util/reflection/InstrumentationMemberAccessor.java
```diff
@@ -372,17 +381,23 @@ private static void assureArguments(
                 throw new IllegalArgumentException("Cannot access " + target + " on " + owner);
             }
         }
-        if (types.length != values.length) {
+
+        Object[] args = values;
```
reta, line 385: Fixing `ParameterizedConstructorInstantiatorTest`

src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsSmartNulls.java
```diff
@@ -89,9 +92,27 @@ public Object answer(InvocationOnMock currentInvocation) throws Throwable {
             if (isToStringMethod(currentInvocation.getMethod())) {
                 return "SmartNull returned by this unstubbed method call on a mock:\n"
                         + unstubbedInvocation;
+            } else if (isMethodOf(MockAccess.class, currentInvocation.getMock(), currentInvocation.getMethod())) {
```
reta, line 95: Fixing `ReturnsSmartNullsTest` test cases, @TimvdLippe this is a really interesting one, the `ReturnsSmartNulls` essentially creates mock for a mock, this is fine by and large but mocks could implement `MockAccess` (Mockito's internal), and the methods of  `MockAccess` are also stubbed with default answer, failing the tests (no suitable answer exists).

The fix is to call the real methods of the `MockAccess`, the tests are green but I am not entirely sure this is the right place, may be `MockHandlerImpl` could be the one, would appreciate some thoughts here.

Thank you.

TimvdLippe: I think I prefer to have the code here, as our `MockHandlerImpl` is in the hot path and we should keep it as clean and fast as possible. So I would be okay with this.

src/main/java/org/mockito/internal/exceptions/stacktrace/DefaultStackTraceCleaner.java
```diff
@@ -24,13 +24,20 @@ public boolean isIn(StackFrameMetadata e) {
     private boolean isIn(String className) {
         if (isFromMockitoRunner(className) || isFromMockitoRule(className)) {
             return true;
-        } else if (isMockDispatcher(className) || isFromMockito(className)) {
+        } else if (isMockDispatcher(className)
+                || isFromMockito(className)
+                || isMethodHandle(className)) {
```
reta, line 29: Fixing `PartialMockingWithSpiesTest`, @TimvdLippe apologies, but another tricky one: the inline mock maker uses `MethodHandle` to dispatch method calls and this fact leaks into the stacktrace. I have added the filtering here, but we may unintentionally filter out the user's stacktrace as well (fe if test or code under test uses `MethodHandle`), what do you think?

raphw: Wouldn't it be an option to filter all frames after a given frame here? That way, "our" handle frames should be discovered.

reta: > Wouldn't it be an option to filter all frames after a given frame here? That way, "our" handle frames should be discovered.

@raphw I looked at current `StackTraceCleaner`, it operates one frame at a time: there is no context with respect what previous (or next) stack frames are. I think in general it is possible to do, but needs some redesign.

subprojects/programmatic-test/programmatic-test.gradle
```diff
@@ -16,3 +16,7 @@ tasks.javadoc.enabled = false
 
 sourceCompatibility = 11
 targetCompatibility = 11
+
+test {
```
reta, line 20: @TimvdLippe note on this change: I have not found the exact problem yet, but there are 2 test suites which manipulate mock makers. It seems like the order matters, I see the random test failures often. With clean JVM process (`forkEvery = 1`) not a single failure observed.

reta:
> Alternatively, we could instrument all class loaders to always load the dispatcher class from the boot loader to possibly override any custom behavior. But this is rather intrusive. Maybe we could make this configurable?

@raphw  I think this is the best option to try so far and should work with most (if not all) OSGi containters

subprojects/osgi-test/src/test/java/org/mockito/osgitest/OsgiTest.java
```diff
@@ -79,11 +82,19 @@ private static Class<?>[] setUpClasses() throws Exception {
     }
 
     private static Class<?>[] getTestClasses() throws Exception {
-        return new Class<?>[] {
-            loadTestClass("SimpleMockTest"),
-            loadTestClass("MockNonPublicClassFailsTest"),
-            loadTestClass("MockClassInOtherBundleTest")
-        };
+        if (Plugins.getMockMaker() instanceof InlineMockMaker) {
```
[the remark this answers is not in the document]
TimvdLippe: That works for me, thanks!

subprojects/programmatic-test/programmatic-test.gradle
TimvdLippe: Okay, we will monitor CI, but if you don't observe any failures locally, we are probably good to go here. Thanks for figuring that one out!

TimvdLippe (approved):
I have reviewed all of these changes and they LGTM. Massive work @reta, thank you so much!

I am a bit confused as to why some of the tests weren't previously failing in our mockmaker-inline CI branch (the `mock-maker-inline` job in our matrix). However, that's not a problem for this PR and something later to figure out.

@raphw do you mind doing a full check as well, just to make sure everything looks okay from your side? Then we can merge this and publish Mockito 5 🎉

reta:
Thanks @TimvdLippe !

> I am a bit confused as to why some of the tests weren't previously failing in our mockmaker-inline CI branch (the `mock-maker-inline` job in our matrix). However, that's not a problem for this PR and something later to figure out.

I think there is a gap in tests: the tests get mock maker instance (from environment setting) but not the member accessor it supposed to be used with. As such, the tests used reflection, not module accessor, all the time. Does it make sense?

raphw:
This looks very good to me, but I think the OSGi configuration deserves some explicit documrntation.

TimvdLippe:
@reta yes that makes sense! Let's fix that in a separate PR. This looks good to me, thank you so much again!

Issue #2839, opened 2022-12-27
Broken build: `MockMethodDispatcher.raw` not being built


big-andy-coates:
I've noticed that the build can get into a bad state locally, with `MockMethodDispatcher.raw` not being rebuilt if it's missing.

This is because the `copyMockMethodDispatcher` task defined in `inline-mock.gradle` is incorrectly defined. It's currently defined as:

```groovy
task copyMockMethodDispatcher {
    dependsOn compileJava
    outputs.files files("${sourceSets.main.java.outputDir}/org/mockito/internal/creation/bytebuddy/inject")
        .asFileTree
        .matching { include "*.raw" }
        .files
        .collect { "${buildDir}/${it.name}" }

    doLast {
        copy {
            from "${sourceSets.main.java.outputDir}/org/mockito/internal/creation/bytebuddy/inject"
            into "${sourceSets.main.java.outputDir}/org/mockito/internal/creation/bytebuddy/inject"
            include 'MockMethodDispatcher.class'
            rename { String fileName ->
                fileName.replace('.class', '.raw')
            }
        }
    }
}
```

However:

1. It's not defining any inputs!  So the task doesn't know when to re-run.
2. It's defining its output as "$buildDir/MockMethodDispatcher.raw", which is wrong, it should be in the `/org/mockito/internal/creation/bytebuddy/inject` directory...

The net result of this is that Gradle caches the fact the task has run and then even if the `MockMethodDispatcher.raw` output file is deleted, the task does NOT re-run.

It looks like this was done to avoid warnings under Gradle 7.x about tasks sharing output directories.  

A better solution would be to move the renamed file to a new location, outside of the output of `compileJava` and include this in the jar.

PR #2840, opened 2022-12-27
Fix copyMockMethodDispatcher task

big-andy-coates:
fixes: https://github.com/mockito/mockito/issues/2839

The task was incorrectly defining inputs and outputs, resulting in the task being marked 'UP-TO-DATE' and not re-running when the input changed or the output was removed.

The task now copies the 'raw' file into a new output directory and includes that directory in the jar.

This corrects the task inputs and outputs, while also avoiding warnings about shared output directories.


big-andy-coates:
@TimvdLippe I believe you changed this code recently.  When you have a moment, I'd appreciate a review :)

TimvdLippe:
The changes lgtm, but the build fails on it I think. Not sure what is wrong with it though

big-andy-coates:
Ill take a look in a bit

big-andy-coates:
@TimvdLippe, green build now...

TimvdLippe (approved):
Awesome thank you!

PR #2842, opened 2022-12-29
Type aware argument captor

big-andy-coates:
fixes: https://github.com/mockito/mockito/issues/565

Enhance argument captor to be more type aware. With this change it is now possible to create an argument captor that only captures certain subtypes of a parameter type.

For example:

```java
// Given:
int simpleMethod(Collection<?> arg);

// When:
mock.simpleMethod(Set.of());
mock.simpleMethod(List.of());

// Then:
ArgumentCaptor<Collection<?>> captor = ArgumentCaptor.forClass(List.class);
verify(mock).simpleMethod(captor.capture());
assertThat(captor.getAllValues()).containsExactly(List.of());
```

Fix also works with `@Captor` annotations.

## Backwards compatibility

Most code should continue to work as before. The only reason tests might start failing is if someone was creating argument captors for subtypes and then testing for all types, for example:

```java
// Given:
int simpleMethod(Collection<?> arg);

// When:
mock.simpleMethod(Set.of());   // <-- note `Set` passed.

// Then:
ArgumentCaptor<Collection<?>> captor = ArgumentCaptor.forClass(List.class);  // <-- note `List` captor
verify(mock).simpleMethod(captor.capture());
```

...the `verify` call would have previously passed, but with this change will now fail.

IMHO, this is suitably weird that the number of existing users hitting this should be low, and their code can be 'fixed' on current release of Mockito, prior to upgrading.

## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
+++ b/src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
@@ -15,13 +15,13 @@ import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import org.mockito.ArgumentMatcher;
+import org.mockito.internal.util.Primitives;
 
-@SuppressWarnings("unchecked")
 public class CapturingMatcher<T>
         implements ArgumentMatcher<T>, CapturesArguments, VarargMatcher, Serializable {
 
     private final Class<? extends T> clazz;
-    private final List<Object> arguments = new ArrayList<>();
+    private final List<T> arguments = new ArrayList<>();
 
     private final ReadWriteLock lock = new ReentrantReadWriteLock();
     private final Lock readLock = lock.readLock();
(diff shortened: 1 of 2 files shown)


big-andy-coates:
@TimvdLippe, saw https://github.com/mockito/mockito/issues/565 when looking into the varargs stuff and figured it wouldn't be hard to implement.

src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
```diff
@@ -15,13 +15,13 @@
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import org.mockito.ArgumentMatcher;
+import org.mockito.internal.util.Primitives;
 
-@SuppressWarnings("unchecked")
 public class CapturingMatcher<T>
         implements ArgumentMatcher<T>, CapturesArguments, VarargMatcher, Serializable {
```
TimvdLippe, line 21: As a follow-up cleanup, I think we can also remove the `CapturesArguments` interface, since this is the only implementation. Then we can match on `CapturingMatcher` in https://github.com/mockito/mockito/blob/123beb8311ed6303ba71d682cff8434fe36b7f0f/src/main/java/org/mockito/internal/invocation/InvocationMatcher.java#L143

src/main/java/org/mockito/internal/matchers/CapturingMatcher.java
```diff
@@ -33,12 +33,20 @@ public CapturingMatcher(final Class<? extends T> clazz) {
 
     @Override
     public boolean matches(Object argument) {
```
TimvdLippe, line 35: Nit: I think we can change `Object` to `T` here as well? It doesn't change the implementation of the method, but since we are `ArgumentMatcher<T>` here I think that should be safe. However, this is okay for now and we can consider this as part of the follow-up as well.

TimvdLippe (approved):
That's a lot easier than I thought, nice!

PR #2843, opened 2022-12-30
Improve examples for InOrder

ascopes:
Include some context in InOrder examples and add an example that uses a static mock as well.

I spent a while being unable to work out how to integrate InOrder with static mocks before realising the example of how to do this is currently encoded in a specific overload of a specific method within InOrder that I had missed as I was not using it.

## Checklist


ascopes:
Fixed JavaDoc error encountered when building under JDK11

PR #2861, opened 2023-01-13
Catch NoClassDefFoundError in KotlinInlineClassUtil

ZacSweers:
When using R8 in AGP 7.4+, we noticed that this code block failed at runtime with `NoClassDefFoundError`. My theory for why this happens is that something in the build pipeline for Android is optimizing `Class.forName(...)` away into a linked `JvmInline.class` call, but then R8 strips the annotation as unused.

You may think - why would mockito be running under minified conditions??

We actually do this in Slack Android's instrumentation tests to create test conditions more similar to production environments, and have a gradle plugin called [Keeper](https://github.com/slackhq/keeper) that generates proguard rules for classes used in the main app _from_ androidTest sources or libraries.

I'm still investigating how this state occurs, but thought to send by this PR to see if you'd all be open to it as an extra catch.

```
Caused by: java.lang.NoClassDefFoundError: Failed resolution of: Lkotlin/jvm/JvmInline;	
	at org.mockito.internal.util.KotlinInlineClassUtil.<clinit>(KotlinInlineClassUtil.java:20)	
	at org.mockito.internal.util.KotlinInlineClassUtil.unboxUnderlyingValueIfNeeded(Unknown Source)	
	at org.mockito.internal.stubbing.answers.Returns.answer(Returns.java:28)	
	at org.mockito.internal.stubbing.StubbedInvocationMatcher.answer(StubbedInvocationMatcher.java:42)	
	at org.mockito.internal.handler.MockHandlerImpl.handle(MockHandlerImpl.java:103)	
	at com.android.dx.mockito.InvocationHandlerAdapter.invoke(InvocationHandlerAdapter.java:52)
```

## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/util/KotlinInlineClassUtil.java
+++ b/src/main/java/org/mockito/internal/util/KotlinInlineClassUtil.java
@@ -18,8 +18,8 @@ public class KotlinInlineClassUtil {
     static {
         try {
             jvmInlineAnnotation = (Class<Annotation>) Class.forName("kotlin.jvm.JvmInline");
-        } catch (ClassNotFoundException e) {
-            // Do nothing: kotlin is pre 1.5.0
+        } catch (ClassNotFoundException | NoClassDefFoundError e) {
+            // Do nothing: kotlin is pre 1.5.0 or JvmInline is otherwise unavailable
         }
     }
 


ZacSweers:
I've been able to confirm at least part of my theory. When inspecting dalvik bytecode in the final test APK, I see this for the `<clinit>`

```dalvik
# direct methods
.method static constructor <clinit>()V
    .registers 1

    .line 20
    :try_start_0
    const-class v0, Lkotlin/jvm/JvmInline;

    sput-object v0, Lorg/mockito/internal/util/KotlinInlineClassUtil;->jvmInlineAnnotation:Ljava/lang/Class;
    :try_end_4
    .catch Ljava/lang/ClassNotFoundException; {:try_start_0 .. :try_end_4} :catch_4

    :catch_4
    return-void
.end method
```

Where it has indeed optimized away the `Class.forName` to link the `JvmInline` class directly.

With that in mind, I'd understand if you don't want to accept this PR as we do need to have something that automatically sets a keep rule for this use in our gradle plugin.

TimvdLippe:
Thanks for your PR and investigation! Since this is a linker error, I am inclined to classify this as a problem with the R8 configuration. If I recall correctly, you can tell R8 to ignore certain classes, so I suppose you can use that instead to work around this issue?

ZacSweers:
Yup that makes sense 👍

PR #2866, opened 2023-01-14
Annotate `Mockito#{mock,spy}(T... reified)` with `@SafeVarargs`

Stephan202:
This avoids "Unchecked generics array creation for varargs parameter" warnings at the call site.

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -1932,6 +1932,7 @@ public class Mockito extends ArgumentMatchers {
      * @return mock object
      * @since 4.9.0
      */
+    @SafeVarargs
     public static <T> T mock(T... reified) {
         if (reified.length > 0) {
             throw new IllegalArgumentException(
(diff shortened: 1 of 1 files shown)


src/main/java/org/mockito/Mockito.java
```diff
@@ -1932,6 +1932,7 @@ public class Mockito extends ArgumentMatchers {
      * @return mock object
      * @since 4.9.0
      */
+    @SafeVarargs
     public static <T> T mock(T... reified) {
```
Stephan202, line 1936: In a sense there [is already a test](https://github.com/mockito/mockito/blob/a9595f559cf4acdaed92371537ced982e30f508d/src/test/java/org/mockito/MockitoTest.java#L164) for this case, but as Mockito currently doesn't compile with something like `options.compilerArgs << "-Xlint:unchecked" << "-Werror"`, the build currently doesn't fail on it. The changes in this PR are relevant for users that do compile with those flags.

TimvdLippe (approved):
Thanks!

Issue #2855, opened 2023-01-09
Make sure the tests use mock maker with intended member accessor


reta:
It seems like there is a gap in tests: the tests get mock maker instance (from environment setting) but not the member accessor it supposed to be used with. As such, the tests used reflection, not module accessor, all the time (see please https://github.com/mockito/mockito/pull/2834#issuecomment-1374876590).

check that

 - [ ] The mockito message in the stacktrace have useful information, but it didn't help
 - [ ] The problematic code (if that's possible) is copied here;
       Note that some configuration are impossible to mock via Mockito
 - [ ] Provide versions (mockito / jdk / os / any other relevant information)
 - [ ] Provide a [Short, Self Contained, Correct (Compilable), Example](http://sscce.org) of the issue
       (same as any question on stackoverflow.com)
 - [ ] Read the [contributing guide](https://github.com/mockito/mockito/blob/main/.github/CONTRIBUTING.md)

@TimvdLippe I could take it if you don't mind, thank you

TimvdLippe:
Go ahead 👍

PR #2872, opened 2023-01-16
Make sure the tests use mock maker with intended member accessor

reta:
Make sure the tests use mock maker with intended member accessor
Fixes https://github.com/mockito/mockito/issues/2855

## Checklist

### DIFF
--- a/gradle/mockito-core/testing.gradle
+++ b/gradle/mockito-core/testing.gradle
@@ -6,24 +6,43 @@ if (java11 != null) {
     }
 }
 
-task(createTestResources).doLast {
-    def mockMakerFile = new File("$projectDir/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker")
-    if (System.env.MOCK_MAKER != null) {
-        logger.info("Using MockMaker ${System.env.MOCK_MAKER}")
-        mockMakerFile.parentFile.mkdirs()
-        mockMakerFile.createNewFile()
-        mockMakerFile.write(System.env.MOCK_MAKER)
-    } else {
-        logger.info("Using default MockMaker")
+task(createTestResources) {
+    doLast {
+        // Configure MockMaker from environment (if specified), otherwise use default
+        def mockMakerFile = new File("$sourceSets.test.output.resourcesDir/mockito-extensions/org.mockito.plugins.MockMaker")
+        if (System.env.MOCK_MAKER != null && !System.env.MOCK_MAKER.endsWith("default")) {
+            logger.info("Using MockMaker ${System.env.MOCK_MAKER}")
+            mockMakerFile.parentFile.mkdirs()
+            mockMakerFile.createNewFile()
+            mockMakerFile.write(System.env.MOCK_MAKER)
+        } else {
+            logger.info("Using default MockMaker")
+        }
+
+        // Configure MemberAccessor from environment (if specified), otherwise use default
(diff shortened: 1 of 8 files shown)


gradle/mockito-core/testing.gradle
```diff
@@ -6,24 +6,43 @@ if (java11 != null) {
...
+
+    def memberAccessorFile = new File("$sourceSets.test.output.resourcesDir/mockito-extensions/org.mockito.plugins.MemberAccessor")
+    if (memberAccessorFile.exists()) {
+        memberAccessorFile.delete()
+    }
 }
 
-compileTestJava.dependsOn createTestResources
-compileTestJava.finalizedBy removeTestResources
+processTestResources.finalizedBy(createTestResources)
```
reta, line 47: @TimvdLippe the problem turned out to be worth than initially suggested: not only `MemberAccessor` were not picked, none of the `MockMaker`s were exercised except the default one.

What was happening is that the `mockito-extensions` created by `createTestResources` were immediately deleted by `removeTestResources` upon `compileTestJava` finalization. Essentially, all tests were run with default `MockMaker` only.

I fixed that but large number of tests are not compatible with `ProxyMockMaker` because they mock classes, not interfaces. I have added the large number of exclusions like:

```
    @BeforeClass
    public static void check_mock_maker() {
        Assume.assumeThat(Plugins.getMockMaker(), not(instanceOf(ProxyMockMaker.class)));
    }
```

Do you think this is fine? I have ~40 test cases left to troubleshoot

src/main/java/org/mockito/internal/junit/DefaultStubbingLookupListener.java
```diff
@@ -68,17 +69,15 @@ private static List<Invocation> potentialArgMismatches(
...
+                            invocation.getMethod().getName())
                     // If stubbing and invocation are in the same source file we assume they are in
                     // the test code,
                     // and we don't flag it as mismatch:
-                    && !s.getInvocation()
-                            .getLocation()
-                            .getSourceFile()
-                            .equals(invocation.getLocation().getSourceFile())) {
+                    && !Objects.equals(
+                            s.getInvocation().getLocation().getSourceFile(),
```
reta, line 79: Was causing NPE here in case of `ProxyMockMaker` because `s.getInvocation().getLocation().getSourceFile()` is `null`

gradle/mockito-core/testing.gradle
TimvdLippe: Hm, that's a good point. If so many tests from core fail with the proxy mockmaker, I don't think that's worth it. In that case, let's remove it from our build matrix. The project itself should have sufficient coverage then.

subprojects/extTest/src/test/java/org/mockitousage/plugins/switcher/MyPluginSwitch.java
```diff
@@ -14,6 +14,10 @@ public class MyPluginSwitch implements PluginSwitch {
     static List<String> invokedFor = new LinkedList<String>();
 
     public boolean isEnabled(String pluginClassName) {
+        if (!pluginClassName.startsWith("org.mockitousage")) {
```
reta, line 17: Filtering the plugins from the envinronment (like `member-accessor-inline`, `mock-maker-inline`, ...)

TimvdLippe (approved):
Nice! Now we know for sure we have covered all cases. Great work once again :)

PR #2882, opened 2023-01-21
Feat: reified mock overloads

ascopes:
Added reified mock overload methods to match the methods that consume an explicit Class<T> object.

This enables the reified overloads to provide the same functionalities as the older
explicit mock overloads.

A few examples of places where this would be overly useful:

```java
interface AppendableBuilder<T extends Appendable> {
    AppendableBuilder append(Appendable other);
}

AppendableBuilder<StringBuilder> obj = mock(Answers.RETURNS_SELF);
```

```java
List<String> inputArguments = mock("the input list");
List<String> outputArguments = mock("the output list");

...

assertThat(transformer.transform(inputArguments))
    .isSameAs(outputArguments);
    
// ERROR! Expected <the output list> but got <the input list>!
```

The following methods have been added:

- Mockito.<T>mock(String, T...) - Sets the mock name.
- Mockito.<T>mock(Answer, T...) - Sets the default answer.
- Mockito.<T>mock(WithSettings, T...) - Provides custom settings.

I have also added a case where passing null varargs in would result in a NullPointerException previously. Now, it will tell the user that they should not be providing these varargs at all.

These overloads should not conflict with any existing usages of this API in the intended way (that I can think of), as the assumption is that a user probably isn't going to want to be mocking Mockito internals, Strings, or WithSettings objects anyway. If they do need to achieve this, then this is still accessible via the existing Mockito.<T>mock(Class<T>) methods anyway.

## Checklist

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -1924,21 +1924,68 @@ public class Mockito extends ArgumentMatchers {
     public static final Answer<Object> RETURNS_SELF = Answers.RETURNS_SELF;
 
     /**
-     * Creates mock object of requested class or interface.
+     * Creates a mock object of the requested class or interface.
      * <p>
-     * See examples in javadoc for {@link Mockito} class
+     * See examples in javadoc for the {@link Mockito} class.
      *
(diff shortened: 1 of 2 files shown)


TimvdLippe (approved):
Makes sense, thank you!

Issue #2879, opened 2023-01-20
Clean up JDK-8 related code


reta:
Since Mockito `5.0.0` has JDK-11 baseline (🥳 ), we could clean up all JDK-8 code:
 - Remove `Java8LocationImpl`
 - Remove `Java8LocationFactory`
 - Check for any other places JDK-8 is checked for 
 - Check for any pre-JDK-8 code (https://github.com/mockito/mockito/pull/2731)

@TimvdLippe I think this is a good time to clean up, what do you think? (I could take care of that).

Thank you!

TimvdLippe:
Yes I agree. #2731 already started this, but it stalled

PR #2883, opened 2023-01-22
Clean up JDK-8 related code

reta:

Clean up JDK-8 related code.
Fixes https://github.com/mockito/mockito/issues/2879

## Checklist


src/main/java/org/mockito/internal/debugging/LocationFactory.java
```diff
@@ -24,26 +24,6 @@ private interface Factory {
     }
 
     private static Factory createLocationFactory() {
```
TimvdLippe, line 26: Nit: this whole factory can now be removed and we can use the relevant location directly

reta: This is true, thanks @TimvdLippe !

src/main/java/org/mockito/internal/debugging/Java9PlusLocationImpl.java
```diff
@@ -1,304 +0,0 @@
...
-import java.lang.reflect.Method;
-import java.util.Collections;
-import java.util.List;
-import java.util.Set;
-import java.util.function.Function;
-import java.util.function.Predicate;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
-class Java9PlusLocationImpl implements Location, Serializable {
```
TimvdLippe, line 25: The implementation of this class is different than what we still have in `LocationImpl`. Shouldn't we update `LocationImpl` to use this implementation? E.g. delete `LocationImpl` and rename this file to it.

reta: @TimvdLippe this is correct, the implementation is indeed different: `Java9PlusLocationImpl ` uses refection for referencing `StackWalker` APIs, this is not needed anymore because `LocationImpl` uses regular imports instead.

TimvdLippe: Ah of course. Thanks for the clarification!

build.gradle
```diff
@@ -99,7 +99,7 @@ dependencies {
     testUtil sourceSets.test.output
 
     signature 'org.codehaus.mojo.signature:java18:1.0@signature'
-    signature 'net.sf.androidscents.signature:android-api-level-24:7.0_r2@signature'
+    signature 'net.sf.androidscents.signature:android-api-level-26:8.0.0_r2@signature'
```
reta, line 102: @TimvdLippe not really necessary (could be workarounded) but brings in these APIs [1] (seems like not part of v24) please let me know what do you think

```
import java.time.Duration;
import java.time.Period;
```

[1] https://github.com/mockito/mockito/pull/2883/files#diff-98c705b0ceb69f07f9ef4b9e9167ecf84b3617a1e950756298285e56b35df4a8R11-R12

TimvdLippe: Yeah we have talked about this before. Since Mockito 5 now supports 11+, I think this is fine.

subprojects/kotlinReleaseCoroutinesTest/kotlinReleaseCoroutinesTest.gradle
```diff
@@ -17,6 +20,15 @@ repositories {
     maven { url 'https://dl.bintray.com/kotlin/kotlin-eap' }
 }
 
+sourceCompatibility = 11
```
reta, line 23: Fixing this warning:

```
> Task :kotlinTest:compileTestKotlin
'compileTestJava' task (current target is 17) and 'compileTestKotlin' task (current target is 1.8) jvm target compatibility should be set to the same Java version.
By default will become an error since Gradle 8.0+! Read more: https://kotl.in/gradle/jvm/target-validation
Consider using JVM toolchain: https://kotl.in/gradle/jvm/toolchain
```

reta:
@TimvdLippe please let me know if I missed some places, I think we are good to go, thank you!

TimvdLippe (approved):
Awesome thank you!

TWiStErRob:
Note: this was partially reverted in https://github.com/mockito/mockito/pull/2891 to supply Android compatibility.

PR #2889, opened 2023-01-28
Fixes some mistakes and missing details in documentation

ascopes:
- Makes static mock verification and construction initialisers into
  functional interfaces, and adds some missing documentation notes.
  Since Java 8 is the minimum target, this seems reasonable to
  include.
- Fixes a number of grammar/spelling typos.
- Replaces an instance of a JavaDoc comment suggesting to use a
  deprecated method with a reference to the suggested replacement.
- Adds a note about supporting chaining of Spy with InjectMocks into
  the InjectMocks documentation.


TimvdLippe (approved):
Awesome work. Thank you!

Issue #2890, opened 2023-01-30
StackWalker.Option not found on Mockito 5.1.0


sergiandreplace:
We have an Android project using Mockito and we tried to upgrade from 5.0.0 to 5.1.0.

When running the Android tests we always receive an error like this:

```
java.lang.NoClassDefFoundError: Failed resolution of: Ljava/lang/StackWalker$Option;
	at org.mockito.internal.debugging.LocationImpl.stackWalker(LocationImpl.java:138)
	at org.mockito.internal.debugging.LocationImpl.<clinit>(LocationImpl.java:40)
	at org.mockito.internal.debugging.LocationFactory.create(LocationFactory.java:17)
	at org.mockito.internal.debugging.LocationFactory.create(LocationFactory.java:13)
	at org.mockito.internal.matchers.LocalizedMatcher.<init>(LocalizedMatcher.java:19)
	at org.mockito.internal.progress.ArgumentMatcherStorageImpl.reportMatcher(ArgumentMatcherStorageImpl.java:30)
	at org.mockito.ArgumentMatchers.reportMatcher(ArgumentMatchers.java:1036)
	at org.mockito.ArgumentMatchers.any(ArgumentMatchers.java:177)
	...
```

After some investigation we found that:

* StackWalker is only available from Java 9. All our modules are setup to use Java 11.
* We can't see StackWalker class from the IDE
* This only happens on Android instrumental tests, not in "normal" java tests
* The class StackWalker is not available in Android SDK java implementation: https://github.com/AndroidSDKSources/android-sdk-sources-for-api-level-33/tree/master/java/lang (that would explain why is not working only for instrumental tests, as they are executed on an Android device). Checked for latest android versions.

It looks like a very specific Android issue.

For the meantime, we will halt the upgrade, but any help will be great.

Thanks in advance

 - [x] The mockito message in the stacktrace have useful information, but it didn't help
 - [N/A] The problematic code (if that's possible) is copied here;
       Note that some configuration are impossible to mock via Mockito
 - [x] Provide versions (mockito / jdk / os / any other relevant information)
 - [X] Provide a [Short, Self Contained, Correct (Compilable), Example](http://sscce.org) of the issue
       (same as any question on stackoverflow.com)
 - [X] Read the [contributing guide](https://github.com/mockito/mockito/blob/main/.github/CONTRIBUTING.md)

TimvdLippe:
Two questions:
1. Are you using `mockito-android`?
[...]

PR #2891, opened 2023-01-30
StackWalker.Option not found on Mockito 5.1.0

reta:
Fixes https://github.com/mockito/mockito/issues/2890

## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/debugging/Java8LocationImpl.java
+++ b/src/main/java/org/mockito/internal/debugging/Java8LocationImpl.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright (c) 2007 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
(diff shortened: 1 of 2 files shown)


TimvdLippe:
Just for my understanding, this still is compatible with the SecurityManager issues you initially did the work for @reta ? It would be a shame to lose these, but not sure if that's still the case.

reta:
> Just for my understanding, this still is compatible with the SecurityManager issues you initially did the work for @reta ? It would be a shame to lose these, but not sure if that's still the case.

@TimvdLippe no, it should not (but I will double check). The reason is that the `LocationImpl` does not use reflection anymore but imports from JDK. If `StackWalker` won't be available, `LocationImpl` won't be loaded, falling back to JDK-8 implementation.

TWiStErRob:
Note: this partially reverts https://github.com/mockito/mockito/pull/2883 to supply Android compatibility.

PR #2893, opened 2023-01-31
Set up Android Github Action pipeline. Fixes #2892

### LINKED ISSUE #2892: Set up Android Github Action pipeline
Coming from https://github.com/mockito/mockito/issues/2890#issuecomment-1409200309:

> we started an effort to set up some basic Android tests for us: https://github.com/mockito/mockito/tree/main/subprojects/androidTest However, this stalled as we couldn't get it to work on CI. If anybody affected here would like to dedicate some time to prevent these kind of issues from happening, please take a look at the existing test suite and infrastructure and submit a PR to get it passing on CI.

The presence of such Github Action would have helped to prevent regressions like https://github.com/mockito/mockito/issues/2890. 

@TimvdLippe I will try to pull it off (I don't work with Android sadly), it seems like Github Actions have support for it.

reta:

Set up Android Github Action pipeline. 
Fixes [#2892](https://github.com/mockito/mockito/issues/2892)

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,6 +85,41 @@ jobs:
         chmod +x codecov
         ./codecov
 
+  #
+  # Android build job
+  #
+  android:
+    runs-on: macos-latest
+    if: "! contains(toJSON(github.event.commits.*.message), '[skip ci]')"
+
+    # Definition of the build matrix
+    strategy:
+      matrix:
+        java: [ 11 ]
+        entry:
+          - { mock-maker: 'mock-maker-default', member-accessor: 'member-accessor-default' }
+
+    # All build steps
+    steps:
+    - name: 1. Check out code
+      uses: actions/checkout@v3 # https://github.com/actions/checkout
+      with:
+        fetch-depth: '0' # https://github.com/shipkit/shipkit-changelog#fetch-depth-on-ci
+
+    - name: 2. Set up Java ${{ matrix.java }}
+      uses: actions/setup-java@v3
+      with:
+        distribution: 'zulu'
+        java-version: ${{ matrix.java }}
+
+    - name: 3. Run Android tests on Java ${{ matrix.java }} with ${{ matrix.entry.mock-maker }} and ${{ matrix.entry.member-accessor }}
+      uses: reactivecircus/android-emulator-runner@v2
+      with:
+        api-level: 29
+        script: ./gradlew :androidTest:connectedCheck --no-daemon --no-build-cache
+      env:
+        MOCK_MAKER: ${{ matrix.entry.mock-maker }}
+        MEMBER_ACCESSOR: ${{ matrix.entry.member-accessor }}
 
   #
   # Release job, only for pushes to the main development branch
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,6 +96,7 @@ jobs:
     strategy:
       matrix:
         java: [ 11 ]
+        android-api: [26, 27, 28, 29, 30, 31]
 
     # All build steps
     steps:
@@ -110,10 +111,11 @@ jobs:
         distribution: 'zulu'
         java-version: ${{ matrix.java }}
 
-    - name: 3. Run Android tests on Java ${{ matrix.java }}
+    - name: 3. Run Android tests on Java ${{ matrix.java }} with Android API level ${{ matrix.android-api }}
       uses: reactivecircus/android-emulator-runner@v2
       with:
-        api-level: 29
+        arch: x86_64
+        api-level: ${{ matrix.android-api }}
         script: ./gradlew :androidTest:connectedCheck --no-daemon --no-build-cache
 
   #
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,13 +96,11 @@ jobs:
     strategy:
       matrix:
         java: [ 11 ]
-        entry:
-          - { mock-maker: 'mock-maker-default', member-accessor: 'member-accessor-default' }
 
     # All build steps
     steps:
     - name: 1. Check out code
-      uses: actions/checkout@v3 # https://github.com/actions/checkout
+      uses: actions/checkout@v3
       with:
         fetch-depth: '0' # https://github.com/shipkit/shipkit-changelog#fetch-depth-on-ci
 
@@ -112,14 +110,11 @@ jobs:
         distribution: 'zulu'
         java-version: ${{ matrix.java }}
 
-    - name: 3. Run Android tests on Java ${{ matrix.java }} with ${{ matrix.entry.mock-maker }} and ${{ matrix.entry.member-accessor }}
+    - name: 3. Run Android tests on Java ${{ matrix.java }}
       uses: reactivecircus/android-emulator-runner@v2
       with:
         api-level: 29
         script: ./gradlew :androidTest:connectedCheck --no-daemon --no-build-cache
-      env:
-        MOCK_MAKER: ${{ matrix.entry.mock-maker }}
-        MEMBER_ACCESSOR: ${{ matrix.entry.member-accessor }}
 
   #
   # Release job, only for pushes to the main development branch
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,7 +96,7 @@ jobs:
     strategy:
       matrix:
         java: [ 11 ]
-        android-api: [26, 27, 28, 29, 30, 31]
+        android-api: [ 26 ]
 
     # All build steps
     steps:
--- a/subprojects/androidTest/src/androidTest/java/org/mockitousage/androidtest/BasicInstrumentedTests.kt
+++ b/subprojects/androidTest/src/androidTest/java/org/mockitousage/androidtest/BasicInstrumentedTests.kt
@@ -35,7 +35,6 @@ class BasicInstrumentedTests {
     fun mockAndUseBasicClassUsingAnnotatedMock() {
         val basicClass = BasicOpenClassReceiver(mockedViaAnnotationBasicOpenClass)
         basicClass.callDependencyMethod()
-        throw Exception("java")
     }
 
     @Test


.github/workflows/ci.yml
```diff
@@ -85,6 +85,52 @@ jobs:
...
+      with:
+        distribution: 'zulu'
+        java-version: ${{ matrix.java }}
+
+    - name: 3. Setup Android SDK
+      uses: android-actions/setup-android@v2
+
+    - name: 4. Install Android SDK components
+      run: |
+        sdkmanager --install "platforms;android-31"
```
TWiStErRob, line 124: You probably don't even need these two steps, because Android Gradle Plugin will download everything automatically (what is needed) and GitHub Actions hosted runners are set up with Android SDK already.

`3.` is only necessary if you use self-hosted runners.
`4.` is not necessary if you use Gradle.

.github/workflows/ci.yml
```diff
@@ -85,6 +85,52 @@ jobs:
...
+
+    - name: 5. Run Android tests on Java ${{ matrix.java }} with ${{ matrix.entry.mock-maker }} and ${{ matrix.entry.member-accessor }}
+      uses: reactivecircus/android-emulator-runner@v2
+      with:
+        api-level: 29
+        script: ./gradlew :androidTest:connectedCheck
+      env:
+        MOCK_MAKER: ${{ matrix.entry.mock-maker }}
+        MEMBER_ACCESSOR: ${{ matrix.entry.member-accessor }}
```
TWiStErRob, line 133: Could you please add a step (in the same job) for running `gradlew test` too so that unit tests with mocks are also tested. These **should** be just normal JVM tests, but we never know what can break :)

.github/workflows/ci.yml
```diff
@@ -70,7 +70,7 @@ jobs:
         MOCK_MAKER: ${{ matrix.entry.mock-maker }}
         MEMBER_ACCESSOR: ${{ matrix.entry.member-accessor }}
 
-    - name: 7. Upload coverage report
+    - name: 10. Upload coverage report
```
TWiStErRob, line 73: maths :D

.github/workflows/ci.yml
```diff
@@ -85,6 +85,52 @@ jobs:
...
+    # Definition of the build matrix
+    strategy:
+      matrix:
+        java: [ 11 ]
+        entry:
+          - { mock-maker: 'mock-maker-default', member-accessor: 'member-accessor-default' }
+          - { mock-maker: 'mock-maker-inline', member-accessor: 'member-accessor-module' }
+          - { mock-maker: 'mock-maker-subclass', member-accessor: 'member-accessor-module' }
+          - { mock-maker: 'mock-maker-subclass', member-accessor: 'member-accessor-reflection' }
+          - { mock-maker: 'mock-maker-inline', member-accessor: 'member-accessor-reflection' }
```
TWiStErRob, line 104: I have a feeling matrix won't be necessary like this, because mockito-android has its own mockmaker, right?

TWiStErRob (commented):
Should work :)

.github/workflows/ci.yml
TWiStErRob: ah never mind, that would be covered by normal CI, nice 👏

(iff all the conditions in settings.gradle are met in normal CI)

.github/workflows/ci.yml
reta: 👍 thanks! still figuring out how to make it work :)

.github/workflows/ci.yml
reta: 👍

.github/workflows/ci.yml
reta: :D yeah, adding and removing steps, will fix that 😄

.github/workflows/ci.yml
reta: It might be the case indeed, I will check it shortly

.github/workflows/ci.yml
```diff
@@ -85,6 +85,52 @@ jobs:
         chmod +x codecov
         ./codecov
 
+  #
+  # Android build job
```
TWiStErRob, line 89: FYI https://github.com/mockito/mockito/pull/2894
(I don't think it's necessary for this pipeline PR, but was inspired :) )

reta:
@TWiStErRob @TimvdLippe I think we nailed it :)

subprojects/androidTest/src/androidTest/java/org/mockitousage/androidtest/BasicInstrumentedTests.kt
```diff
@@ -35,7 +35,6 @@ class BasicInstrumentedTests {
     fun mockAndUseBasicClassUsingAnnotatedMock() {
         val basicClass = BasicOpenClassReceiver(mockedViaAnnotationBasicOpenClass)
         basicClass.callDependencyMethod()
-        throw Exception("java")
```
reta, line 38: This test seemed to be not valid: the exception was thrown but no validations / versification or what not

reta:
FYI, I tested the Github action on 5.1.0:

```
Starting 6 tests on test(AVD) - 10

org.mockitousage.androidtest.BasicInstrumentedTests > mockAndUseBasicClassWithVerify[test(AVD) - 10] FAILED 
	java.lang.NoClassDefFoundError: Failed resolution of: Ljava/lang/StackWalker$Option;
	at org.mockito.internal.debugging.LocationImpl.stackWalker(LocationImpl.java:138)

org.mockitousage.androidtest.BasicInstrumentedTests > mockAndUseBasicClassUsingLocalMock[test(AVD) - 10] FAILED 
	java.lang.NoClassDefFoundError: org.mockito.internal.debugging.LocationImpl
	at org.mockito.internal.debugging.LocationFactory.create(LocationFactory.java:17)

org.mockitousage.androidtest.BasicInstrumentedTests > mockAndUseBasicInterfaceUsingAnnotatedMock[test(AVD) - 10] FAILED 
	java.lang.NoClassDefFoundError: org.mockito.internal.debugging.LocationImpl
	at org.mockito.internal.debugging.LocationFactory.create(LocationFactory.java:17)

org.mockitousage.androidtest.BasicInstrumentedTests > mockAndUseBasicClassUsingAnnotatedMock[test(AVD) - 10] FAILED 
	java.lang.NoClassDefFoundError: org.mockito.internal.debugging.LocationImpl
	at org.mockito.internal.debugging.LocationFactory.create(LocationFactory.java:17)

org.mockitousage.androidtest.BasicInstrumentedTests > mockAndUseBasicInterfaceAndVerify[test(AVD) - 10] FAILED 
	java.lang.NoClassDefFoundError: org.mockito.internal.debugging.LocationImpl
	at org.mockito.internal.debugging.LocationFactory.create(LocationFactory.java:17)

org.mockitousage.androidtest.BasicInstrumentedTests > mockAndUseBasicInterfaceUsingLocalMock[test(AVD) - 10] FAILED 
	java.lang.NoClassDefFoundError: org.mockito.internal.debugging.LocationImpl
	at org.mockito.internal.debugging.LocationFactory.create(LocationFactory.java:17)

```

.github/workflows/ci.yml
```diff
@@ -85,6 +85,41 @@ jobs:
...
+  android:
+    runs-on: macos-latest
+    if: "! contains(toJSON(github.event.commits.*.message), '[skip ci]')"
+
+    # Definition of the build matrix
+    strategy:
+      matrix:
+        java: [ 11 ]
+        entry:
+          - { mock-maker: 'mock-maker-default', member-accessor: 'member-accessor-default' }
```
TWiStErRob, line 100: Do you want to keep a 1-element matrix?

.github/workflows/ci.yml
```diff
@@ -85,6 +85,41 @@ jobs:
...
+
+    # All build steps
+    steps:
+    - name: 1. Check out code
+      uses: actions/checkout@v3 # https://github.com/actions/checkout
```
TWiStErRob, line 105: Unnecessary comment is unnecessary ;)
All actions are GitHub org/repo@tag (or hash) format. So this mapping is the default resolution.

subprojects/androidTest/src/androidTest/java/org/mockitousage/androidtest/BasicInstrumentedTests.kt
TWiStErRob: Agree, it was probably testing failure, because a CI job is as good as how you can diagnose failures.

.github/workflows/ci.yml
```diff
@@ -85,6 +85,41 @@ jobs:
...
+      uses: actions/setup-java@v3
+      with:
+        distribution: 'zulu'
+        java-version: ${{ matrix.java }}
+
+    - name: 3. Run Android tests on Java ${{ matrix.java }} with ${{ matrix.entry.mock-maker }} and ${{ matrix.entry.member-accessor }}
+      uses: reactivecircus/android-emulator-runner@v2
+      with:
+        api-level: 29
+        script: ./gradlew :androidTest:connectedCheck --no-daemon --no-build-cache
```
TWiStErRob, line 119: Upload resulting report as artifact?

TWiStErRob (commented):
Nice, getting simpler! 👏

.github/workflows/ci.yml
TWiStErRob: Actually might want to introduce more API levels in the future, because they have different JVMs. See my comment on the issue, API 21 fails, while API 29 passes.

reta: Yeah, I think we could remove the mock-maker part

.github/workflows/ci.yml
reta: That's a good question for @TimvdLippe , I don't see test reports uploaded (but only coverage reports).

.github/workflows/ci.yml
reta: I think we settled on level 26 as the minimum: https://github.com/mockito/mockito/blob/main/build.gradle#L102

TWiStErRob: For reference the [official docs](https://developer.android.com/about/versions/oreo/android-8.0#art) only list some new APIs, so even Android 8 doesn't support Java 11. However, as I understand with [desugaring](https://developer.android.com/studio/write/java11-default-support-table) Java 11+ APIs can be used as low as API 1. This is of course not practical, I think an API 21 minimum would be [enough](https://apilevels.com/). That said [official full Java 11 support is only in 33](https://developer.android.com/about/versions/13/features#core-libraries).

Anyway, based on the info, starting with `android-api: [26,27,28,29,30,31,33]` would work.
([32 has no Java changes](https://developer.android.com/sdk/api_diff/32/changes), it was a "feature drop", not given a major version, if it works, it could be included though)

.github/workflows/ci.yml
TWiStErRob: 😲 so how do you know what failed?
https://github.com/mockito/mockito/actions/runs/1865554496/jobs/2625339758

.github/workflows/ci.yml
TWiStErRob: Just to clarify, the matrix should only change the emulator version, the compile/target/minSdk should be hardcoded to highest/lowest values as [Google recommends it (TL;DR: we have ~1 year to use the latest target, otherwise cannot publish)](https://support.google.com/googleplay/android-developer/answer/11926878).

TWiStErRob (approved):
I would include `api-level: ${{ matrix.android-api }}` for [all the versions that pass right now](https://github.com/mockito/mockito/pull/2893#discussion_r1093002196) to preserve compatibility.

.github/workflows/ci.yml
reta: You look into Github action output, like https://github.com/mockito/mockito/pull/2893#issuecomment-1410906236

TWiStErRob: Hmm... that leaves so many questions open:
```
java.lang.NoClassDefFoundError: org.mockito.internal.debugging.LocationImpl
at org.mockito.internal.debugging.LocationFactory.create(LocationFactory.java:17)
```
is a very minimal output, where is it called from? which code path got here? what's the root "caused by"? What got output to stdout/stderr? How long did it take? etc. All of these are recorded in the report. As an example open any `IntgTest` from [this workflow's "Test Results HTML" artifacts](https://github.com/TWiStErRob/net.twisterrob.gradle/actions/runs/4052565235). Anyway up to you if like reading logs :), it would drive me crazy.

reta: :D @TWiStErRob I am totally in for test report publishing (don't get me wrong), I have just included the most interesting part but here is the complete log [1], this is how the test failures are troubleshot at the moment. 

[1] https://github.com/reta/mockito/actions/runs/4057324609/jobs/6982933109

reta:
> I would include `api-level: ${{ matrix.android-api }}` for [all the versions that pass right now](https://github.com/mockito/mockito/pull/2893#discussion_r1093002196) to preserve compatibility.

We may need to exclude some, I believe I have tried `31` first and the emulator part didn't work :(

.github/workflows/ci.yml
```diff
@@ -85,6 +85,37 @@ jobs:
...
+  #
+  android:
+    runs-on: macos-latest
+    if: "! contains(toJSON(github.event.commits.*.message), '[skip ci]')"
+
+    # Definition of the build matrix
+    strategy:
+      matrix:
+        java: [ 11 ]
+        android-api: [26, 27, 28, 29]
```
reta, line 99: @TWiStErRob so I have to remove 30-31 API levels (for now), the emulator GA seems to incorrectly deduct the arch:

```
/bin/sh -c \sdkmanager --install 'system-images;android-31;default;x86' --channel=0 > /dev/null
  Warning: Failed to find package 'system-images;android-31;default;x86'
```

The result is that emulator is not installed and the test run fails. I think we could hack the arch parameters or just get to that later on when it is fixed. I think it looks good now. Thanks!

TWiStErRob: It just doesn't exist: `sdkmanager --list | grep android- | grep default`:
```
  system-images;android-26;default;arm64-v8a                                               | 1            | ARM 64 v8a System Image
  system-images;android-26;default;x86                                                     | 1            | Intel x86 Atom System Image
  system-images;android-26;default;x86_64                                                  | 1            | Intel x86 Atom_64 System Image
  system-images;android-27;default;arm64-v8a                                               | 1            | ARM 64 v8a System Image
  system-images;android-27;default;x86                                                     | 1            | Intel x86 Atom System Image
  system-images;android-27;default;x86_64                                                  | 1            | Intel x86 Atom_64 System Image
  system-images;android-28;default;arm64-v8a                                               | 1            | ARM 64 v8a System Image
  system-images;android-28;default;x86                                                     | 4            | Intel x86 Atom System Image
  system-images;android-28;default;x86_64                                                  | 4            | Intel x86 Atom_64 System Image
  system-images;android-29;default;arm64-v8a                                               | 8            | ARM 64 v8a System Image
  system-images;android-29;default;x86                                                     | 8            | Intel x86 Atom System Image
  system-images;android-29;default;x86_64                                                  | 8            | Intel x86 Atom_64 System Image
  system-images;android-30;default;arm64-v8a                                               | 1            | ARM 64 v8a System Image
  system-images;android-30;default;x86_64                                                  | 10           | Intel x86 Atom_64 System Image
  system-images;android-31;default;arm64-v8a                                               | 4            | ARM 64 v8a System Image
  system-images;android-31;default;x86_64                                                  | 4            | Intel x86 Atom_64 System Image

```
I would be very surprised if the "latest" MacOS, that GitHub just released recently would be x86.

Can you try adding the numbers and forcing the arch?

https://github.com/ReactiveCircus/android-emulator-runner/blob/e1248ae048e31ec2e171adc06be585d06c1f936e/action.yml#L14-L16

```diff
      uses: reactivecircus/android-emulator-runner@v2
      with:
        api-level: ${{ matrix.android-api }}
+       arch: x86_64 # or arm64-v8a (don't know if these Macs are M1/M2 or not)
```

reta: `x86_64` does not work for sure, I tried that before, let me try `arm`

TWiStErRob: What does "does not work for sure" means, because in their examples they also use that:
https://github.com/ReactiveCircus/android-emulator-runner#usage--examples

TWiStErRob: `PANIC: Avd's CPU Architecture 'arm64' is not supported by the QEMU2 emulator on x86_64 host.` sounds good.

reta: It seems like going up to `31` is fine, updated the matrix, after 31 - failures :(

TWiStErRob: I think 33 would work, told you 32 is "special": https://github.com/mockito/mockito/pull/2893#discussion_r1093002196

Notice that in the above table there's no 32 and 33, but there are emulators.

Let's merge it as is, this is a huge improvement over "nothing". We can figure out 32/33 later.

TWiStErRob (approved):

.github/workflows/ci.yml
reta: Oh sorry, I missed that the issue was only specific to 32

TWiStErRob: Let's see why https://issuetracker.google.com/issues/267458959

.github/workflows/ci.yml
TimvdLippe: We don't have test reporting setup. For that, we rely on console output solely and locally reproducing the error. With CI that is sometimes tricky, but that's the setup we got atm. If you have any suggestions, feel free to create an issue and we can discuss further there.

.github/workflows/ci.yml
```diff
@@ -85,6 +85,38 @@ jobs:
...
+  #
+  android:
+    runs-on: macos-latest
+    if: "! contains(toJSON(github.event.commits.*.message), '[skip ci]')"
+
+    # Definition of the build matrix
+    strategy:
+      matrix:
+        java: [ 11 ]
+        android-api: [26, 27, 28, 29, 30, 31]
```
TimvdLippe, line 99: I don't think we want to test so many Android releases, since we don't have a lot of CPU capacity on GitHub actions. Instead, I think only having 26 (which is the oldest version that we still support) is sufficient. If we notice gaps in Android support, we can always extend the matrix, but for now I think just 1 Android version is good enough.

TimvdLippe (changes requested):
Only 1 nit. Do I understand correctly we want to merge #2894 first?

TWiStErRob:
No, let's merge this first. There's no dependency, either one can go first, but this one is ready, mine still needs review.

.github/workflows/ci.yml
TWiStErRob: Good point, very likely if it works on the oldest, it'll work elsewhere too.

TWiStErRob: That said, I would follow this up with a PR for running 33, because that has full on Java 11 support, which can mess things up.

reta: Kept only `26`

TimvdLippe (approved):
Great work, thank you!

TWiStErRob:
Thank you for your patience @reta! Thanks for @TimvdLippe for taking this, it'll definitely help the Android community if this keeps passing on every release.

reta:
> Thank you for your patience @reta! Thanks for @TimvdLippe for taking this, it'll definitely help the Android community if this keeps passing on every release.

Thanks a for tremendous help and guidance @TWiStErRob !

PR #2908, opened 2023-02-08
Fixes #2905 : ThreadLocal classes can be mocked.

### LINKED ISSUE #2905: StackOverflow while mocking a ThreadLocal on Mockito 5.1.1
Hello Folks,

I'm upgrading Mockito in a Java application from 4.9.0 to 5.1.1. We are on JDK 11 (I personally use Temurin 11.0.13).

So, I've a test failing due to a StackOverflow

```
java.lang.StackOverflowError
	at org.mockito.internal.creation.bytebuddy.inject.MockMethodDispatcher.get(MockMethodDispatcher.java:34)
	at java.base/java.lang.ThreadLocal.get(ThreadLocal.java:162)
	at org.mockito.internal.creation.bytebuddy.MockMethodAdvice$SelfCallInfo.checkSelfCall(MockMethodAdvice.java:368)
	at org.mockito.internal.creation.bytebuddy.MockMethodAdvice.isMocked(MockMethodAdvice.java:171)
```
This happens when we mock a class that extends `ThreadLocal`. But I verified that it happens also by mocking `ThreadLocal` as well.

Sounds like a bug, but you are the experts here.

Best regards,
Roberto

robertotru:
Fixes https://github.com/mockito/mockito/issues/2905

## Changes
- `MockMethodAdvice` way of checking if an object was a mock didn't work for `ThreadLocal` classes. The only change here is in method `MockMethodAdvice#isMocked(Object)` where the two conditions are flipped in order to avoid hitting the `get()` method from `ThreadLocal` used to check if the object was calling itself.

_Note_: this was working in Mockito 4.9.0, didn't check what happens in 5.0.0 and 5.1.0.

## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodAdvice.java
+++ b/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodAdvice.java
@@ -168,7 +168,7 @@ public class MockMethodAdvice extends MockMethodDispatcher {
 
     @Override
     public boolean isMocked(Object instance) {
-        return selfCallInfo.checkSelfCall(instance) && isMock(instance);
+        return isMock(instance) && selfCallInfo.checkSelfCall(instance);
     }
 
     @Override
--- a/src/test/java/org/mockitousage/bugs/ThreadLocalTest.java
+++ b/src/test/java/org/mockitousage/bugs/ThreadLocalTest.java
@@ -20,39 +20,47 @@ public class ThreadLocalTest extends TestBase {
 
     @Test
     public void mock_ThreadLocal_does_not_raise_StackOverflowError() {
-        StackOverflowError stackOverflowError = Assertions.catchThrowableOfType(() -> {
-            mock(ThreadLocal.class, RETURNS_MOCKS);
-        }, StackOverflowError.class);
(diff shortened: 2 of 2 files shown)


robertotru:
Hello,

This PR is a possible fix for the issue reported in https://github.com/mockito/mockito/issues/2905
Hope it helps.
Best
R

raphw:
While this will solve the issue for thread local classes, it will introduce it for map objects. Let me have a look if this can be solved differently.

TimvdLippe:
> it will introduce it for map objects

Can we make sure we have a regression test for that as well? I would have merged this PR since all tests are green. Would be good to codify this requirement to ensure we don't regress.

robertotru:
@TimvdLippe and @raphw 

Got it. I can add additional tests, just give me a list of use cases you want me to cover.

Best
R

raphw:
Turns out the recursion check is one level deeper. For the case I meant, the self-call-check is pointless anyways. Turning this the way as suggested should not be an issue.

raphw (approved):

robertotru:
Thanks a lot for the quick processing.

PR #2923, opened 2023-02-22
Feature 2921 generic types

### LINKED ISSUE #2921: TypeBasedCandidateFilter could match field type signature to circumvent type erasure
Hi,
when using `@InjectMocks` in combination with multiple `@Mock` fields that have the same generic type (after type erasure), this fails unless the field names in injectee and test class are equal. 

I stumbled over this when I added a second `Provider<T>` field to both test and injectee class, and I guess most other people doing the same will also fail here as miserably as I did (also because googling for `@InjectMock` doesn't show the relevant [InjectMocks documentation](https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/InjectMocks.html) in the first result page ...)

E.g. this test will fail because I named the test class `@Mock` fields with the suffix `mock`. When deleting that name suffix, the test succeeds.
```java
import static org.junit.jupiter.api.Assertions.assertNotNull;

import javax.inject.Provider;

import org.mockito.InjectMocks;
import org.mockito.Mock;

public class InjectMockTest {
    
    public static class UnderTest {
        Provider<String> stringProvider;
        Provider<Integer> intProvider;
    }

    @Mock
    private Provider<String> stringProviderMock;

    @Mock
    private Provider<Integer> intProviderMock;

    @InjectMocks
    private UnderTest underTest;

    @Test
    public void testInjectMock() {
        assertNotNull(underTest.stringProvider);
        assertNotNull(underTest.intProvider);
    }

}
```
I saw in the debugger that in `TypeBasedCandidateFilter.filterCandidate()` , `candidateFieldToBeInjected.getGenericType().getTypeName()` yields e.g. `"javax.inject.Provider<java.lang.Integer>"`. So if Mockito would maintain the `@Mock` field's `java.lang.reflect.Field` along with the mock object, the generic type names could be compared. 

This would effectively solve the problem of type erasure for this usecase, which I'd guess is quite common ...

jfrantzius:
Fixes #2921 by preserving and using the generic Java type information that was previously lost to type erasure. 

Contains a JUnit test with usage example, demonstrating that fields with generic types do not need to have matching names anymore.

### DIFF
--- a/src/main/java/org/mockito/internal/configuration/injection/filter/TypeBasedCandidateFilter.java
+++ b/src/main/java/org/mockito/internal/configuration/injection/filter/TypeBasedCandidateFilter.java
@@ -25,6 +25,7 @@ public class TypeBasedCandidateFilter implements MockCandidateFilter {
     }
 
     protected boolean isCompatibleTypes(Type typeToMock, Type mockType, Field injectMocksField) {
+        boolean result = false;
         if (typeToMock instanceof ParameterizedType && mockType instanceof ParameterizedType) {
             // ParameterizedType.equals() is documented as:
             // "Instances of classes that implement this interface must implement
@@ -34,7 +35,7 @@ public class TypeBasedCandidateFilter implements MockCandidateFilter {
             // and e.g. Set doesn't equal TreeSet, so roll our own comparison if
             // ParameterizedTypeImpl.equals() returns false
             if (typeToMock.equals(mockType)) {
-                return true;
+                result = true;
             } else {
                 ParameterizedType genericTypeToMock = (ParameterizedType) typeToMock;
                 ParameterizedType genericMockType = (ParameterizedType) mockType;
@@ -42,21 +43,19 @@ public class TypeBasedCandidateFilter implements MockCandidateFilter {
                 Type[] actualTypeArguments2 = genericMockType.getActualTypeArguments();
                 // Recurse on type parameters, so we properly test whether e.g. Wildcard bounds
                 // have a match
-                return recurseOnTypeArguments(
+                result = recurseOnTypeArguments(
                         injectMocksField, actualTypeArguments, actualTypeArguments2);
             }
         } else if (typeToMock instanceof WildcardType) {
             WildcardType wildcardTypeToMock = (WildcardType) typeToMock;
             Type[] upperBounds = wildcardTypeToMock.getUpperBounds();
-            return Arrays.stream(upperBounds)
+            result = Arrays.stream(upperBounds)
                     .anyMatch(t -> isCompatibleTypes(t, mockType, injectMocksField));
         } else if (typeToMock instanceof Class && mockType instanceof Class) {
-            return ((Class) typeToMock).isAssignableFrom((Class) mockType);
+            result = ((Class) typeToMock).isAssignableFrom((Class) mockType);
         } // no need to check for GenericArrayType, as Mockito cannot mock this anyway
 
-        // can only happen if there is a new subclass of Type that we don't know about yet,
-        // must have return statement here to be able to compile
-        return false;
+        return result;
     }
 
     private boolean recurseOnTypeArguments(
--- a/src/main/java/org/mockito/internal/configuration/injection/filter/TypeBasedCandidateFilter.java
+++ b/src/main/java/org/mockito/internal/configuration/injection/filter/TypeBasedCandidateFilter.java
@@ -5,9 +5,12 @@
 package org.mockito.internal.configuration.injection.filter;
 
 import java.lang.reflect.Field;
+import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+
+import org.mockito.internal.util.MockUtil;
 
 public class TypeBasedCandidateFilter implements MockCandidateFilter {
 
@@ -26,7 +29,16 @@ public class TypeBasedCandidateFilter implements MockCandidateFilter {
         List<Object> mockTypeMatches = new ArrayList<>();
         for (Object mock : mocks) {
             if (candidateFieldToBeInjected.getType().isAssignableFrom(mock.getClass())) {
-                mockTypeMatches.add(mock);
+                Type genericTypeToMock = MockUtil.getMockSettings(mock).getGenericTypeToMock();
+                Type genericType = candidateFieldToBeInjected.getGenericType();
+                // be more specific if generic type information is available
+                if (genericTypeToMock!=null || genericType!=null) {
+                    if (genericTypeToMock!=null && genericType!=null && genericTypeToMock.getTypeName().equals(genericType.getTypeName())) {
+                        mockTypeMatches.add(mock);
+                    } // else: filter out mock, as generic types don't match
+                } else {
+                    mockTypeMatches.add(mock);
+                }
             }
         }
 
--- a/src/main/java/org/mockito/internal/configuration/injection/filter/TypeBasedCandidateFilter.java
+++ b/src/main/java/org/mockito/internal/configuration/injection/filter/TypeBasedCandidateFilter.java
@@ -79,22 +79,31 @@ public class TypeBasedCandidateFilter implements MockCandidateFilter {
             Type actualTypeArgument = actualTypeArguments[i];
             Type actualTypeArgument2 = actualTypeArguments2[i];
             if (actualTypeArgument instanceof TypeVariable) {
+                TypeVariable<?> typeVariable = (TypeVariable<?>)actualTypeArgument;
                 // this is a TypeVariable declared by the class under test that turned
                 // up in one of its fields,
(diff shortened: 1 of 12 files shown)


src/main/java/org/mockito/internal/configuration/injection/filter/TypeBasedCandidateFilter.java
```diff
@@ -26,7 +80,16 @@ public OngoingInjector filterCandidate(
...
             if (candidateFieldToBeInjected.getType().isAssignableFrom(mock.getClass())) {
-                mockTypeMatches.add(mock);
+                Type genericMockType = MockUtil.getMockSettings(mock).getGenericTypeToMock();
+                Type genericType = candidateFieldToBeInjected.getGenericType();
+                boolean bothHaveGenericTypeInfo = genericType != null && genericMockType != null;
+                // be more specific if generic type information is available
+                if (!bothHaveGenericTypeInfo || isCompatibleTypes(genericType, genericMockType)) {
+                    mockTypeMatches.add(mock);
+                } else { // filter out mock, as generic types don't match
+                    // System.out.println("types don't match " + candidateFieldToBeInjected + " " +
```
TimvdLippe, line 90: Please remove commented code

src/main/java/org/mockito/internal/configuration/injection/filter/TypeBasedCandidateFilter.java
```diff
@@ -17,6 +25,52 @@ public TypeBasedCandidateFilter(MockCandidateFilter next) {
...
+                } else {
+                    // owner type, raw type or type arguments length don't match
+                    return false;
+                }
+            }
+        } else if (typeToMock instanceof WildcardType) {
+            WildcardType wildcardTypeToMock = (WildcardType) typeToMock;
+            Type[] upperBounds = wildcardTypeToMock.getUpperBounds();
+            return Arrays.stream(upperBounds).anyMatch(t -> isCompatibleTypes(t, mockType));
+        } else if (typeToMock instanceof GenericArrayType && mockType instanceof GenericArrayType) {
```
TimvdLippe, line 64: CodeCov reports this branch is not covered. Can you please expand the coverage here or remove this check?

subprojects/junit-jupiter/src/test/java/org/mockitousage/GenericTypeMockTest.java
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * Verify that InjectMocks will correctly match fields with same generic type but different type parameters,
+     * without using the same field name.
+     */
+    @Test
+    public void testInjectMock() {
+        // this used to fail without any error message hinting at the problem, as soon as a class under test has
+        // a second field of the same generic type, but with different type parameter. The programmer then
+        // had to know that mock field names have to match field names in the class under test.
+        assertNotNull(underTest.stringList);
```
TimvdLippe, line 58: Please define a separate test case for each of these, preferably with an inner class with each Mock, so that it is clear which mock is used in which class

subprojects/junit-jupiter/src/test/java/org/mockitousage/GenericTypeMockTest.java
```diff
@@ -0,0 +1,69 @@
...
+        assertNotNull(underTest.intList);
+        assertNotNull(underTest.anySet);
+        assertNotNull(underTest.intStringCollectionMap);
+
+        assertEquals(stringProviderMock, underTest.stringList);
```
TimvdLippe, line 63: If you are already asserting equals here, then you can remove the `assertNotNull` before

TimvdLippe (changes requested):

src/main/java/org/mockito/internal/configuration/injection/filter/TypeBasedCandidateFilter.java
jfrantzius: The `WildcardType` part is covered:
image
But you are right that the `GenericTypeArrayType` part from line 64 isn't. I'll remove it, as Mockito cannot mock Array types (as I found out in a test)

subprojects/junit-jupiter/src/test/java/org/mockitousage/GenericTypeMockTest.java
jfrantzius: If both mock field and object's field are null, they'd be considered equals, e.g. if both `@Mock` and `@InjectMock` annotations do not work

jfrantzius:
@TimvdLippe please review and resolve where appropriate

src/main/java/org/mockito/internal/configuration/injection/filter/TypeBasedCandidateFilter.java
jfrantzius: Removed it

src/main/java/org/mockito/internal/configuration/injection/filter/TypeBasedCandidateFilter.java
jfrantzius: Removed not covered code

subprojects/junit-jupiter/src/test/java/org/mockitousage/GenericTypeMockTest.java
jfrantzius: Done

jfrantzius:
https://github.com/mockito/mockito/actions/runs/4296719412/jobs/7488794171 
```
Starting 0 tests on test(AVD) - 13
Test run failed to complete. No test results. onError: commandError=true message=Attempt to invoke interface method 'boolean android.app.IActivityManager.startInstrumentation(android.content.ComponentName, java.lang.String, int, android.os.Bundle, android.app.IInstrumentationWatcher, android.app.IUiAutomationConnection, int, java.lang.String)' on a null object reference

* What went wrong:
Execution failed for task ':androidTest:connectedDebugAndroidTest'.
> There were failing tests. See the report at: file:///Users/runner/work/mockito/mockito/subprojects/androidTest/build/reports/androidTests/connected/index.html
```
I don't know what to make of this, maybe just re-run?

jfrantzius:
@TimvdLippe thx for restarting the build. Did you have a chance to take a look at my comments on your review?

Sorry for maybe being pushy, but I have some time on my hands for this at the moment, which might change soon. Also I don't know under what circumstances you are able to devote time to Mockito :)

TimvdLippe:
I am on vacation this week, but should be back next week to review.

jfrantzius:
Alright, have a nice vacation!

FabrizzioRuiz (approved):

src/main/java/org/mockito/internal/configuration/injection/filter/TypeBasedCandidateFilter.java
TimvdLippe: There is still some coverage missing, can you either remove the uncovered branches or add regression tests that cover the logic? https://app.codecov.io/gh/mockito/mockito/pull/2923/blob/src/main/java/org/mockito/internal/configuration/injection/filter/TypeBasedCandidateFilter.java Since this is intricate logic and a hot codepath, I would like to see full coverage here.

TimvdLippe:
@jfrantzius Apart from the missing code coverage, this PR LGTM!

jfrantzius:
@TimvdLippe I added more test coverage

TimvdLippe:
@jfrantzius please look at the code coverage report and see which branches are missing coverage. There are still some missing, for example one where the generic types don't match.

jfrantzius:
I added support for generic injectee classes, e.g.
```java
        public class UnderTestWithTypeParameters<T1, T2> {
            List<T1> t1List;
            Set<T2> t2Set;
        }
```
There is one line that cannot be covered in tests, please see the comment there: https://github.com/jfrantzius/mockito/blob/feature-2921-generic-type/src/main/java/org/mockito/internal/configuration/injection/filter/TypeBasedCandidateFilter.java#L70

jfrantzius:
Could you please approve the workflow @TimvdLippe ?

jfrantzius:
@TimvdLippe I'm off for today, if you trigger the build then the last line missing in coverage should now be covered as well.

jfrantzius:
I guess it's time to setup Spotless in my IDE ...

jfrantzius:
@TimvdLippe build might need another nudge in https://github.com/mockito/mockito/actions/runs/4372977101/jobs/7658608894
```
/bin/sh -c exit $(cat gradle.exit)
Error: The process '/bin/sh' failed with exit code 1
Terminate Emulator
INFO    | Wait for emulator (pid 3622) 20 seconds to shutdown gracefully before kill;you can set environment variable ANDROID_EMULATOR_WAIT_TIME_BEFORE_KILL(in seconds) to change the default value (20 seconds)
```

jfrantzius:
WDYT @TimvdLippe , is it good to merge?

TimvdLippe (approved):
Thanks for all the work, this LGTM!

PR #2942, opened 2023-03-14
#2934 throw exception on multiple matches by type if cannot be reduce…

### LINKED ISSUE #2934: Feature suggestion: TerminalMockCandidateFilter should report error when more than one candidate left
Hi,
when stumbling over #2921 , I realized that Mockito will silently fail to inject `@Mock` objects when there is more than one candidate. Usually, DI containers will throw an exception in that case, otherwise the user first has to understand what went wrong (no injection happened), and then start wondering *why* it failed (e.g. multiple `@Mock` fields with same type, and user will have to know which exactly).

My suggestion would be to have a new error similar to `Reporter.cannotInjectDependency` that `TerminalMockCandidateFilter` then should call if `mocks.size() > 1`

jfrantzius:
Fixes #2934 : throw exception on multiple matches by type if cannot be reduced to single match

### DIFF
--- a/src/main/java/org/mockito/internal/configuration/injection/filter/TypeBasedCandidateFilter.java
+++ b/src/main/java/org/mockito/internal/configuration/injection/filter/TypeBasedCandidateFilter.java
@@ -4,6 +4,8 @@
  */
 package org.mockito.internal.configuration.injection.filter;
 
+import static org.mockito.internal.exceptions.Reporter.moreThanOneMockCandidate;
+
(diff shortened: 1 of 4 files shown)


jfrantzius:
Hi @TimvdLippe , WDYT?

jfrantzius:
@TimvdLippe  Is there a way for me to restart that build, except for a new commit?
image

TimvdLippe (approved):
LGTM, good addition!

PR #2945, opened 2023-03-16
Fixes #2877: removed mockito-inline as subproject and actual subproject

### LINKED ISSUE #2877: Mockito-inline dependency after upgrade 5.0.0
After reading that mockito switched to mockito-inline by default (with 5.0.0) i expected either mockito-core to have a mockito-inline dependency, or mockito-inline has been merged into mockito-core.

Interestingly neither of this is the case (since mockito-inline has a 5.0.0 release and no mockito-inline transitive dependency is found in https://mvnrepository.com/artifact/org.mockito/mockito-core/5.0.0 )

The concrete question is, how should we go about the current explicit dependency on mockito-inline. Should it be removed with 5.0.0 or should it stay?

krzyk:
Fixes #2877 as suggested in https://github.com/mockito/mockito/issues/2877 removed inline from settings.gradle.kts and also the inline subproject itself.
Also removed groovyInlineTest which was testing this subproject.

## Checklist


krzyk:
Sorry I was a bit surprised that ANDROID_HOME needs to be set to run part of android tests.

TimvdLippe:
Seems like this would reduce our test coverage by quite a lot. Can we keep the inline test suite and rather it depending on `mockito-inline`, it depends on `mockito-core` instead which already uses the inline mockmaker by default.

krzyk:
@TimvdLippe Sure, I replaced the dependency in groovyInline tests

krzyk:
@TimvdLippe Yes, I misunderstood your previous comment. Now all the tests are back.

settings.gradle.kts
```diff
@@ -3,11 +3,9 @@ plugins {
 }
 
 include("subclass",
-    "inline",
```
TimvdLippe, line 6: While we need to delete the source files of this artifact, we do need to keep this module so it runs the tests. Please run the code coverage locally to see what it is doing, as it is reporting missing usage.

I recommend renaming this module to `inlineTest`. Then, we need to make sure we don't accidentally publish it, but iirc by appending Test to it, it works. I can double check that later.

krzyk: Oops, forgot about that. I checked the coverage locally and the html report doesnt' show decimals so I assumed 85% was it.
Now after I readded the renamed project it is at 86%.

subprojects/inlineTest/inlineTest.gradle
TimvdLippe, line 1: We need to remove the `java-library.gradle` line from this file, to make sure we don't accidentally start publishing this artifact. I think this should can be the replacement:

```
apply(from = "$rootDir/gradle/dependencies.gradle")

plugins {
    java
}
```

See https://github.com/mockito/mockito/blob/19965ff915bf6c19eeb350f07d8ddfefdb95b1a0/subprojects/memory-test/memory-test.gradle as an example

krzyk: OK, applied

TimvdLippe (approved):
Yup, that looks good to me! Thanks for the PR and the updates 😄

PR #2948, opened 2023-03-25
Fixes 2947: correct visibility check to respect nestmates

raphw:
Since Java 11, subclasses can invoke private constructors if they are nested classes and the access right is registered in the nestmate property of both classes. Without recognizing such rights, Mockito cannot create a synthetic constructor hierarchy that sets fields for spies as the constructor generation is skipped due to a missing super constructor that is considered valid.


TimvdLippe (approved):
Nice!

PR #2949, opened 2023-03-25
Add `ArgumentMatchers#assertArg` method.

### LINKED ISSUE #2285: Assertions on parameters during verification
### Suggestion
I would like to suggest adding an API allowing for assertions during parameter verification without argument captors in simple cases.

### Motivation
In my projects I generally use Mockito and AssertJ for testing. For non trivial method parameters, I often encounter the following ArgumentCaptor pattern:
```java
ClassUnderTest classUnderTest = new ClassUnderTest(otherServiceMock);

classUnderTest.method();

ArgumentCaptor<Param> paramArgumentCaptor = ArgumentCaptor.forClass(Param.class);
verify(otherServiceMock).doStuff(paramArgumentCaptor.capture());
Param paramFromCaptor = paramArgumentCaptor.getValue();
assertThat(paramFromCaptor.getField1()).isEqualTo("value1");
assertThat(paramFromCaptor.getField2()).isEqualTo("value2");
```

This is fine, but a little bit verbose.

An alternative pattern I find myself using lately is using the `argThat` argument matcher to eliminate the captor, resulting in the verify statement looking something like this:
```java
verify(otherServiceMock).doStuff(argThat(param -> {
    assertThat(param.getField1()).isEqualTo("value1");
    assertThat(param.getField2()).isEqualTo("value2");
    return true;
}));
```
In my opinion, at least in simple cases where only assertions on a single parameter are desired, this looks a little cleaner. Just the return statement is a bit pointless.

My suggestion is adding an API that would allow eliminating the return statement in this pattern, for example:

```java
verify(otherServiceMock).doStuff(assertArg(param -> {
    assertThat(param.getField1()).isEqualTo("value1");
    assertThat(param.getField2()).isEqualTo("value2");
}));
```

Or is there already an API that would allow a similar pattern that I've missed?

maciejwalkowiak:
Fixes #2285

Enables convenient alternative to `ArgumentCaptor` to assert arguments during verification:

```java
verify(otherServiceMock).doStuff(assertArg(param -> {
    assertThat(param.getField1()).isEqualTo("value1");
    assertThat(param.getField2()).isEqualTo("value2");
}));
```

Verification succeeds as long as no exception is thrown from the lambda passed to `assertArg`, which means it works with any assertions (JUnit, AssertJ, ...).

Since this is my first PR to Mockito, I may have missed something, or perhaps the documentation should be improved. Feedback is very welcome!

## Checklist

### DIFF
--- a/src/test/java/org/mockitousage/matchers/MatchersTest.java
+++ b/src/test/java/org/mockitousage/matchers/MatchersTest.java
@@ -20,6 +20,7 @@ import static org.mockito.AdditionalMatchers.leq;
 import static org.mockito.AdditionalMatchers.lt;
 import static org.mockito.AdditionalMatchers.not;
 import static org.mockito.AdditionalMatchers.or;
+import static org.mockito.ArgumentMatchers.assertArg;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.ArgumentMatchers.nullable;
 import static org.mockito.Mockito.any;
@@ -52,6 +53,7 @@ import java.util.List;
 import java.util.RandomAccess;
 import java.util.regex.Pattern;
 
+import org.junit.ComparisonFailure;
 import org.junit.Test;
 import org.mockito.ArgumentMatchers;
 import org.mockito.Mockito;
@@ -623,5 +625,24 @@ public class MatchersTest extends TestBase {
         mock.oneArg(Character.valueOf('\u20AC'));
 
         verify(mock, times(2)).oneArg(nullable(Character.class));
+    }
+
+    @Test
+    public void assertArg_matcher() throws Exception {
+        mock.oneArg("hello");
+
+        verify(mock).oneArg(assertArg((String it) -> assertEquals("hello", it)));
+    }
+
+    @Test
+    public void assertArg_matcher_fails_when_assertion_fails() throws Exception {
+        mock.oneArg("hello");
+
+        try {
+            verify(mock).oneArg(assertArg((String it) -> assertEquals("not-hello", it)));
+            fail("Should throw an exception");
+        } catch (ComparisonFailure e) {
+            // do nothing
+        }
     }
 }
--- a/src/test/java/org/mockitousage/matchers/MatchersTest.java
+++ b/src/test/java/org/mockitousage/matchers/MatchersTest.java
@@ -645,4 +645,32 @@ public class MatchersTest extends TestBase {
             // do nothing
         }
     }
+
+    @Test
+    public void can_invoke_method_on_mock_after_assert_arg() throws Exception {
+        mock.oneArg("hello");
+
+        try {
+            verify(mock).oneArg(assertArg((String it) -> assertEquals("not-hello", it)));
+            fail("Should throw an exception");
+        } catch (ComparisonFailure e) {
+            // do nothing
+        }
+
+        mock.oneArg("hello");
+    }
+
+    @Test
+    public void can_verify_on_mock_after_assert_arg() throws Exception {
+        mock.oneArg("hello");
+
+        try {
+            verify(mock).oneArg(assertArg((String it) -> assertEquals("not-hello", it)));
+            fail("Should throw an exception");
+        } catch (ComparisonFailure e) {
+            // do nothing
+        }
+
+        verify(mock).oneArg("hello");
+    }
 }
--- a/src/main/java/org/mockito/ArgumentMatchers.java
+++ b/src/main/java/org/mockito/ArgumentMatchers.java
@@ -14,6 +14,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Consumer;
 import java.util.regex.Pattern;
 
 import org.mockito.internal.matchers.Any;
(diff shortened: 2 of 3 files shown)


TimvdLippe (approved):
Simple and elegant, nice! I wonder if we can even deprecate `ArgumentCaptor` at this point in favor of this, but let's cross that bridge when we get to it.

TimvdLippe (changes requested):
Ah in fact, it is good to call out this new API in our main documentation. To do so, add section 55 to https://github.com/mockito/mockito/blob/0b8685c37be46d43b80638d9e9e5ba9efd35f291/src/main/java/org/mockito/Mockito.java#L1663 and add it to the index at https://github.com/mockito/mockito/blob/0b8685c37be46d43b80638d9e9e5ba9efd35f291/src/main/java/org/mockito/Mockito.java#L114

It's best to include an example usage (such as the one you included in your PR description).

maciejwalkowiak:
@TimvdLippe Javadocs added (feel free to rephrase if you find better wording). I added "(Since 5.3.0)" but I am not sure in which version exactly it would be released.

crohit (approved):

crohit:
Oops I didn't mean to approve! I was just really looking forward for this change and clicked the wrong button!.

maciejwalkowiak:
@TimvdLippe checkstyle issues are fixed now. Please take another look when you have a minute

src/test/java/org/mockitousage/matchers/MatchersTest.java
```diff
@@ -624,4 +626,23 @@ public void nullable_matcher() throws Exception {
...
+
+    @Test
+    public void assertArg_matcher() throws Exception {
+        mock.oneArg("hello");
+
+        verify(mock).oneArg(assertArg((String it) -> assertEquals("hello", it)));
+    }
+
+    @Test
+    public void assertArg_matcher_fails_when_assertion_fails() throws Exception {
```
TimvdLippe, line 638: I just realized that throwing an exception might cause stateful problems in our tracking mechanism. E.g. if we match on arguments and we throw an exception somewhere, I am not sure how Mockito recovers. Given that this PR passes the checks, I think this is fine, but let's codify this explicitly.

Can we add the following test cases?

```java
    @Test
    public void can_invoke_method_on_mock_after_assert_arg() throws Exception {
        mock.oneArg("hello");

        try {
            verify(mock).oneArg(assertArg((String it) -> assertEquals("not-hello", it)));
            fail("Should throw an exception");
        } catch (ComparisonFailure e) {
            // do nothing
        }

        mock.oneArg("hello");
    }

    @Test
    public void can_verify_on_mock_after_assert_arg() throws Exception {
        mock.oneArg("hello");

        try {
            verify(mock).oneArg(assertArg((String it) -> assertEquals("not-hello", it)));
            fail("Should throw an exception");
        } catch (ComparisonFailure e) {
            // do nothing
        }

        verify(mock).oneArg("hello");
    }
```

maciejwalkowiak: @TimvdLippe tests added

TimvdLippe (approved):
Awesome thank you!

sbernard31:
@TimvdLippe,

> Simple and elegant, nice! I wonder if we can even deprecate `ArgumentCaptor` at this point in favor of this, but let's cross that bridge when we get to it.

Even if this new feature is great and more elegant than `ArgumentCaptor` in some cases.
This is less powerful than `ArgumentCaptor` and I don't think this is a good idea to deprecate or remove `ArgumentCaptor`.

We have some use cases where this will be hard to replace : 
- like capturing a arguument and then use it in another `verify`.
- create some utility doing general validation on some argument but returning a captured one for more specific validation.

maciejwalkowiak:
@sbernard31 I agree. An example when `assertArg` is not going to replace argument captors: https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/BannerTests.java#L105

sbernard31:
@TimvdLippe you're right about it seems we can do the same  with `assertArg` combining with `AtomicReference` and it seems I was wrong when I said _"This is less powerful than ArgumentCaptor"_.

But If we compare `ArgumentCaptors` to `assertArg`+`AtomicReference`, I feel ArgumentCaptor is less verbose and more easy to read and this could be a good reason to keep it. But clearly a less strong reason that what I thought at the beginning.

```java 
// assertArg + AtomicReference
AtomicReference<String> reference = new AtomicReference<>();
verify(mock).invocation(assertArg(arg -> reference.set(arg)));
verify(mock).secondInvocation(eq(reference.get()));

// ArgumentCaptor
ArgumentCaptor<String> arg = new  ArgumentCaptor<>();
verify(mock).invocation(arg.capture());
verify(mock).secondInvocation(eq(arg.getValue()));
```

TimvdLippe:
@sbernard31 Right, that's exactly what we should be monitoring in the years to come, which pattern is used more frequently than the other. For now, there is no need to deprecate it, I only saw the similarities between the two API's and hence my comment, but nothing more than that.

edwnmrtnz:
Will this work in kotlin too? Tried it just now and the arg is null.

```
 @Test
    fun `should only fetch active movies`() = runBlocking {
        val repo = Mockito.mock(MoviesRepository::class.java)
        val sut = FetchMoviesUseCase(repo)

        sut.execute(page = 1)

        Mockito.verify(repo).fetch(assertArg { params ->
            Truth.assertThat(params["status"]).isEqualTo("active")
        })
        Unit
    }
```

Above code results to
```
assertArg {
            println(it)
        } must not be null
java.lang.NullPointerException: assertArg {
            println(it)
        } must not be null
```

PR #2961, opened 2023-04-02
Assertion matcher

### LINKED ISSUE #2960: Assertion matcher
`ArgumentMatcher` returns true/false and it's enough for stubs.
However, for mocks false is not enough. When expectation is not meet, it is important why not meet. It means _rich-failure_ is needed, not just `false`

Actually _rich-failure_ can be achieved with `org.assertj.core.matcher.AssertionMatcher`.
It works but it's not handy: every assertion needs a class, when lambda is enough

michaldo:
Fixes #2960

Simplifies using custom matcher with assertions

### DIFF
--- a/ArgumentMatchersTest.java
+++ b/ArgumentMatchersTest.java
@@ -1,0 +1,39 @@
+package org.mockito;
+
+import org.junit.Test;
+import org.mockitoutil.TestBase;
+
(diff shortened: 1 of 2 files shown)


TimvdLippe:
Duplicate of #2949

PR #2965, opened 2023-04-05
Adds  `withoutAnnotations` parameter to `@Mock`

wezleytsai:
Allows setting `MockSettings#withoutAnnotations` using the `@Mock` annotation, similar to other settings that are already available.


src/main/java/org/mockito/Mock.java
```diff
@@ -132,6 +132,13 @@
      */
     String mockMaker() default "";
 
+    /**
+     * Mock will not attempt to preserve all annotation metadata, see {@link MockSettings#withoutAnnotations()}.
+     *
+     * @since 5.3.0
```
wezleytsai, line 138: Not sure if this is the correct next version

TimvdLippe (approved):
Thanks!

PR #2976, opened 2023-04-16
Bugfix 2974 fix npe

jfrantzius:
Fix NPE when root of type hierachy is reached


jfrantzius:
WDYT @TimvdLippe ?

TimvdLippe (approved):
Thanks!

PR #2983, opened 2023-04-21
Fixes 2972: Weakens visibility constraint

raphw:
Breaks spying on encapsulated objects where a super type in a different package declares only a protected constructor.


TimvdLippe (approved):
Awesome, thanks!

PR #2989, opened 2023-04-25
Fix #2915 Forbid spy on mocked interface

### LINKED ISSUE #2915: Verify on Spy Interface incorrect count of actual invocation
Hi,

At first, thanks for this awesome library. 

During a migration from mockito 4.4.0 to mockito 5.1.1 one of my test began to fail. 
It reported that the spied mock has 2 call one a method instead of one. 

I figured that  it has something to do with the new default [mockmaker](https://github.com/mockito/mockito/releases/tag/v5.0.0) since it's working with mockito-subclass.

Still, the message is misleading : it's suggesting that there is an error while the code works as expected.
Either there is a bug in the spy count , either the message should be changed to reflect that verify/spy on interface is impossible. 
(I may also have missed something :p ) 

Short failing example  java 11/ junit 5/ mockito 5.1.1

```java

class MyTest {

    @Test
    void testSpyMockInterfaceMockito5() {
        MInterface spiedInterface = spy(mock(MInterface.class));
        spiedInterface.add(0);
        verify(spiedInterface).add(0);
    }
}

interface MInterface {

    void add(Integer i);
}

/*
## Mockito 5.1.1 
org.mockito.exceptions.verification.TooManyActualInvocations:
mInterface$MockitoMock$b85WZhRq.add(0);
Wanted 1 time:
-> at MyTest.testSpyMockInterfaceMockito5(MyTest.java:21)
But was 2 times:
-> at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
-> at MyTest.testSpyMockInterfaceMockito5(MyTest.java:18)

## mockito 4.4.0 ok
## mockito-subclass 5.1.1 ok
 */
```

bmaggi:
A proposition of implementation for [2915](https://github.com/mockito/mockito/issues/2915) 

I only raise exception for interface since it's working with class.

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -10,6 +10,7 @@ import org.mockito.internal.MockitoCore;
 import org.mockito.internal.creation.MockSettingsImpl;
 import org.mockito.internal.framework.DefaultMockitoFramework;
 import org.mockito.internal.session.DefaultMockitoSessionBuilder;
+import org.mockito.internal.util.MockUtil;
 import org.mockito.internal.verification.VerificationModeFactory;
 import org.mockito.invocation.Invocation;
 import org.mockito.invocation.InvocationFactory;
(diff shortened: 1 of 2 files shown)


bmaggi:
I changed the code forbid all mocks

TimvdLippe (approved):
Thank you!

PR #3004, opened 2023-05-08
Add a Json Matcher

SingingBush:
I find myself needing to create a custom _ArgumentMatcher_ for handling json structures occasionally so thought it may be worth having this available within mockito itself

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -92,6 +92,8 @@ dependencies {
 
     compileOnly libraries.junit4, libraries.hamcrest, libraries.opentest4j
     implementation libraries.objenesis
+    implementation 'jakarta.json:jakarta.json-api:2.1.1'
+
 
     testImplementation libraries.assertj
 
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -92,6 +92,8 @@ dependencies {
 
     compileOnly libraries.junit4, libraries.hamcrest, libraries.opentest4j
     implementation libraries.objenesis
+    implementation 'jakarta.json:jakarta.json-api:2.1.1'
```
TimvdLippe, line 95: `mockito-core` does not and should not have a lot of dependencies. We should not be adding a full JSON parser to it, as the vast majority of users won't need this. Since Mockito is in a lot of places, the lean dependency tree is a must.

Therefore, we can't add this dependency. I understand the use case and the alternative would be to ship our own, which we also don't want.

As such, I don't think we can ship this as part of `mockito-core`. Instead, you can ship this yourself in your own library for those that would like it?

SingingBush: ok no problem. I thought it may be the case as it's an understandable requirement. I have a gist that can be used to copy'n'paste into projects required.

PR #3020, opened 2023-05-25
Edit DoAnswerStyleStubbing.isSet() to return true if there are answers

### LINKED ISSUE #2955: Function name is incorrect.
I saw [DoAnswerStyleStubbing](https://github.com/mockito/mockito/blob/96eb5e0cbff80a1545b1093db1baf4e33a2041e7/src/main/java/org/mockito/internal/stubbing/DoAnswerStyleStubbing.java#L27-L29) class code.
I thought that below is correct.

```java
boolean isSet() {
    return !answers.isEmpty();
}
```
After execute `setAnswers` function with answers, instance will return `isSet is false`.
I think it looks strange.

And [InvocationContainerImpl](https://github.com/mockito/mockito/blob/96eb5e0cbff80a1545b1093db1baf4e33a2041e7/src/main/java/org/mockito/internal/stubbing/InvocationContainerImpl.java#L112-L114) also need to be changed as follow.
```java
public boolean hasAnswersForStubbing() {
    return doAnswerStyleStubbing.isSet();
}
```

ellaella12:
Fixes #2955

### DIFF
--- a/src/main/java/org/mockito/internal/stubbing/DoAnswerStyleStubbing.java
+++ b/src/main/java/org/mockito/internal/stubbing/DoAnswerStyleStubbing.java
@@ -25,7 +25,7 @@ class DoAnswerStyleStubbing implements Serializable {
     }
 
     boolean isSet() {
-        return answers.isEmpty();
+        return answers.size() > 0;
     }
 
     void clear() {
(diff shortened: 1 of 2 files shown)


TimvdLippe (approved):
Thanks!

PR #3028, opened 2023-06-05
[#2898] Build CI on more realistic combinations (and faster)

### LINKED ISSUE #2898: Build CI on more realistic combinations (and faster)
Conversation split from https://github.com/mockito/mockito/pull/2894#discussion_r1096034581

I strongly recommend adopting this approach in this repo: https://jakewharton.com/build-on-latest-java-test-through-lowest-java/
I'm opening this issue to discuss if there's an agreement.

Jake's point in that guide is that you build once, and test on different runtimes, same as your use will download your pre-built jar and run it on different JVMs. 

### JDK incompatibilities in user's runtime

In the current CI setup:
 * the code is compiled on JDK 11 then tests executed
 * in a different matrix, code is compiled on JDK 17 then tests executed
 * then on tags, releases are built on JDK 11 and published (this is what users will download from maven)

This means that we never ever test the real use case, that the JDK 11-built JAR is executed on JDK 17, for example because the Mockito user's CI is running JDK 17.

@reta mentioned
> From experience: the code compiled on older JDKs (fe JDK-8) may fail at runtime (due to slight differences in how JVM interprets bytecode), but it is very rare.

This could be very easily covered by the above approach.

### One of the Gradle plugins don't support the old/new JDK version

With Android Gradle Plugin 8.0 (already in beta, released soon), it will be compiled and packaged with Java 17, forcing any of its users to upgrade their Gradle build setup to use JDK 17 or higher (that is the JVM running Gradle itself). This means that Mockito will either need to add conditionals in settings.gradle.kts similar to the one removed here: https://github.com/mockito/mockito/pull/2894/commits/fe8881e525de437d7fc3ad2e4e0dcd21725420b4 to prevent including the `:androidTest` module on JDK 11 machines/CI matrix. This will practically mean that the real use case of Mockito built on Java 11, running in Java 17 test runtime in an Android project will not be covered.

### Faster CI

Since the build will happen only on one JDK, producing one binary JAR, the build will be more reproducible and better covered. With Jake's approach the individual JDK coverage is done via Gradle tasks and toolchains, meaning contributors can easily run all test on all JDKs without much setup (as opposed to having to install and set `JAVA_HOME` to mimic CI). Additionally the CI time will be also reduced because the extra compilation on each matrix step will not be necessary any more; the freed up time can be used to cover more JDKs!

### Easier adding of new JDKs to test on
Since each JDK coverage is just a task, adding it cannot cause incompatibilities with the hosting Gradle build, because the toolchain will only apply to the specific test task.

---

check that

undermyumbrella1:
Fixes #2898
- builds once 
- uploads build artifacts
- other jobs/other java version will download this build artifact
- then it will run compiled tests to check that they pass

Considerations:
1. originally matrix jobs are used
- but this mean other jobs will have to depend on all jobs in the matrix to complete
- instead of a single job in the matrix to complete = slower
- Therefore, reusable workflows is used instead
- so each run will be a separate gh job

2. Android job depends on build job's build artifacts
- specifically ` Task :androidTest:bundleDebugClassesToRuntimeJar` 
- is this expected behaviour or should android job be able to run successfully, independent of build job?


TimvdLippe:
Thanks for the PR! I can't trigger the CI pipeline on this, so I think the `ci.yml` no longer works as expected. Can you update that file to the expected syntax so that we can trigger a build?

.github/workflows/ci.yml
```diff
@@ -18,72 +18,8 @@ permissions:
...
-        curl -Os https://uploader.codecov.io/latest/linux/codecov.SHA256SUM
-        curl -Os https://uploader.codecov.io/latest/linux/codecov.SHA256SUM.sig
-
-        gpgv codecov.SHA256SUM.sig codecov.SHA256SUM
-        shasum -a 256 -c codecov.SHA256SUM
-
-        chmod +x codecov
-        ./codecov
+  build-and-test:
+   uses: mockito/mockito/.github/workflows/build-and-test.yml@feature/build-once-test-all
```
TWiStErRob, line 22: This might be why it's not runnable: this is referencing a potentially non-existent file because this branch exists only in your fork?

I would recommend using the [relative notation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_iduses):
```suggestion
  build-and-test:
   uses: ./.github/workflows/build-and-test.yml
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,63 @@
...
+
+      - name: 4. Test on Java ${{ inputs.java }} with ${{ inputs.mock-maker }} and ${{ inputs.member-accessor }}
+        run: |
+          ./gradlew test -x testClasses \
+          -x compileTestJava -x compileTestGroovy -x compileTestKotlin -x compileTestBundleJava \
+          -x processTestResources \
+          -x classes -x compileJava -x compileGroovy -x compileKotlin -x compileOtherBundleJava \
+          -x processResources -x processOtherBundleResources \
+          -x preBuild -x preDebugBuild \
+          -x jar
```
TWiStErRob, line 48: This looks risky, if you miss a task (e.g. by project evolution), it might run a mixed build and there might be no visible effect. Also as far as I remember this might fail in Gradle 8.

I think instead of uploading the build directory, it might be a more stable approach to run `publishToAllPublicationsMavenRepository` (see https://github.com/mockito/mockito/blob/main/gradle/java-publication.gradle#L100) and only save that folder with the "production-like" jars. The question is how it would be used, and the answer might be [Gradle dependency substitution](https://docs.gradle.org/current/userguide/resolution_rules.html) done conditionally.

.github/workflows/ci.yml
TWiStErRob, line 22: Why is this indirection necessary? I it looks like you extracted a file to call other files. Is it extracted so that there's no need to list all the *test* jobs in `needs` below?

Note: we can use [`matrix` with `uses`](https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-a-matrix-strategy-with-a-reusable-workflow), so theoretically you can say `needs: [test]` which refers to the whole test matrix.

.github/workflows/test.yml
ellaella12: agree that it is risky, not sure if my approach was right
any ideas on how to only execute test aka compile once test all? as gradlew test still runs build and compile tasks due to dependent tasks

based on the link, the jar files are published to a local folder, hence they cannot be shared between jobs. also compiled classes and jars are needed together so that compiled classes are used to run the test aka compile once test all. is the idea to upload compiled classes and jars in a remote maven repo to be shared between jobs?

.github/workflows/ci.yml
ellaella12: using matrix would mean test jobs will have to wait for ALL build jobs to finish. In reality it just needs to wait for a single build job in the matrix to finish. E.g testing on java 17  with default mock maker and member accessor, only needs to wait for build on java 11 with default mock maker and member accessor. It doesn't need to wait for other combination of mock maker and member accessor to complete.

or is it better to use matrix but it will be slower?

.github/workflows/test.yml
TWiStErRob: Based on my current knowledge there are two ways to go about this:
1. substitute the `main` sourceSet of each project with a pre-built .jar file (based on a `-P` flag), unsure if this is actually feasible. Gradle has a pretty powerful `configuration` and `dependencies` API, so it should be possible.
2. Set up toolchains and create multiple test tasks (see how [Gradle Stutter plugin creates multiple test tasks for each Java * Gradle version](https://github.com/gradle-nexus/publish-plugin/actions/runs/5181270374/jobs/9336551191#step:4:175)).

The second one would create multiple builds (because of the maintainer's requirements to keep the matrix), but the .jar files would be exactly the same. As far as I read it, in the referenced blog post this is not a problem because there's no matrix and the toolchain approach is used.

The first approach would optimize the build pipeline for matrix, but I've never seen an example like this. At my previous company with did a similar `-x` hack for a while, but it fell apart during Gradle/Android Gradle Plugin upgrades. Dependency substitution is a less-hacky way of achieving what `-x` does.

.github/workflows/ci.yml
TWiStErRob: Right now we have:
```mermaid
graph LR;
  build-default-default-->test11-default-default-->release
  build-inline-module-->test11-inline-module-->release
  build-inline-reflection-->test11-inline-reflection-->release
  build-subclass-module-->test11-subclass-module-->release
  build-subclass-reflection-->test11-subclass-reflection-->release
  build-default-default-->test17-default-default-->release
  build-inline-module-->test17-inline-module-->release
  build-inline-reflection-->test17-inline-reflection-->release
  build-subclass-module-->test17-subclass-module-->release
  build-subclass-reflection-->test17-subclass-reflection-->release
```
Based on a quick check there's only one variant of JAR build output. The mockmaker/member parameter is only used to set up test resources, so we should only need to build once, which means that we end up with this:
```mermaid
graph LR;
  build-->test11-default-default-->release
  build-->test11-inline-module-->release
  build-->test11-inline-reflection-->release
  build-->test11-subclass-module-->release
  build-->test11-subclass-reflection-->release
  build-->test17-default-default-->release
  build-->test17-inline-module-->release
  build-->test17-inline-reflection-->release
  build-->test17-subclass-module-->release
  build-->test17-subclass-reflection-->release
```
I think this means that the current structure can be kept and there's no need for workflow calls at all?

TWiStErRob: Note that workflow calls have a really bad developer experience:
image
compared to matrix:
image

.github/workflows/test.yml
ellaella12: the second approach creates different gradle task on the same runner/gh job. From the response in the original issue, it seems different gh job is needed to test for each java version

TWiStErRob: Yeah, that works pretty well with matrix: see [how to dynamically use a task](https://github.com/gradle-nexus/publish-plugin/blob/master/.github/workflows/java-versions.yml#L40).

.github/workflows/ci.yml
```diff
@@ -37,53 +83,50 @@ jobs:
...
+          java-version: ${{ matrix.java }}
+
+      - name: 4. Test on Java ${{ matrix.java }} with ${{ matrix.entry.mock-maker }} and ${{ matrix.entry.member-accessor }}
+        run: |
+          ./gradlew test \
+          -x compileTestJava -x compileTestGroovy -x compileTestKotlin -x compileTestBundleJava \
+          -x classes -x compileJava -x compileGroovy -x compileKotlin -x compileOtherBundleJava \
+          -x processResources -x processOtherBundleResources \
+          -x preBuild -x preDebugBuild \
+          -x jar
```
TWiStErRob, line 111: Tip: when you have just one command you can do this too:
```suggestion
        run: >
          ./gradlew test
          -x compileTestJava -x compileTestGroovy -x compileTestKotlin -x compileTestBundleJava
          -x classes -x compileJava -x compileGroovy -x compileKotlin -x compileOtherBundleJava
          -x processResources -x processOtherBundleResources
          -x preBuild -x preDebugBuild
          -x jar
```
probably worth keeping the `\` version though for consistency with other `run:` blocks.

.github/workflows/ci.yml
TWiStErRob: Can you please resolve this conversation now that you migrated to single file?
(Also the other files are still on the branch, did you want to keep those?)

TWiStErRob: Let's resolve this too.

.github/workflows/test.yml
ellaella12: Sorry, I am still trying to understand this. Starting with build artifacts, instead of `upload-artifacts`, use `publishToAllPublicationsMavenRepository`.

1) this uploads artifacts into a local maven repo. How are these artifacts shared between github jobs? as local repo is local to each runner
2) jar + compiled classes is needed for testing. So is it advisable to upload the entire build folder?

PR #3040, opened 2023-06-16
Add another hint to no argument value captured

ghostbuster91:
Hi, 

I've just hit that and having this case on the list of hints would be extremely helpful.

It is quite common for me to alternate between `thenReturn` and `thenAnswer`. However, if you forgot to change to `thenAnswer` when using `argumentCaptor.getCapture()` your code will still compile but it will fail in runtime. 

Example:
```java
when(service.call(captor.capture())).thenReturn(Response(captor.getValue().toString()))
```
in order to fix it, the code needs to be changed to:
```java
when(service.call(captor.capture())).thenAnswer(i-> Response(captor.getValue().toString()))
```
which I am sure is obvious for a lot of people who use mockito a lot, but it isn't for newcomers and seasonal users. 

## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/exceptions/Reporter.java
+++ b/src/main/java/org/mockito/internal/exceptions/Reporter.java
@@ -711,7 +711,8 @@ public class Reporter {
                 join(
                         "No argument value was captured!",
                         "You might have forgotten to use argument.capture() in verify()...",
-                        "...or you used capture() in stubbing but stubbed method was not called.",
+                        "...or you used capture() in stubbing but stubbed method was not called,",
+                        "...or you used capture.getValue() in the thenReturn() instead of thenAnswer()."
                         "Be aware that it is recommended to use capture() only with verify()",
                         "",
                         "Examples of correct argument capturing:",


kubukoz (approved):

TimvdLippe:
I am not sure if we should promote this use case. Instead, you should return the `i.getArgument()` api if you want to return a value that was passed in. You shouldn't use an `ArgumentCaptor` for that as it is more indirection that the Mockito api already supplies.

TimvdLippe:
@ghostbuster91 Can you let us know if `i.getArgument()` indeed suffices for your use case?

ghostbuster91:
It requires passing an index of the parameter to extract value that we are interested in which in case of 4+ parameters methods is quite cumbersome, as well as manually casting value to a proper type, but yes, it works.

TimvdLippe:
It is quite cumbersome, but that's also the reality with a 4 parameter method if you want to stub everything. Even then, you don't use the `ArgumentCaptor` for all values, so you still would only get 1 parameter from it. Given that, I don't think we want to promote the `ArgumentCaptor` for this particular use case in our documentation, as `i.getArgument()` is a more explicit usage of our AP.

PR #3051, opened 2023-07-04
Gradle 7.6 to 8.2 (conventions to extensions migration, mostly)

TWiStErRob:
**Don't be alarmed by the number of commits, they're 1-4 lines each.**
**When squashing, please remove the commit details, keep only the title!**
_Note: each commit in the PR shows why the change was made, expand the commit messages during review!_

## Summary

A lot of this PR is fixing warning naggings introduced in 8.2, but for 7.x deprecations.
So their preferred migration paths are available in earlier versions too (7.6 / 8.0 / 8.1), and deprecations can usually be fixed way in advance of nagging.
Happy to downgrade to Gradle 7.6/8.0 to fix problems and upgrade Gradle to latest separately, if necessary.

## Remaining problems during `gradlew build` after this PR

<details><summary>Not related to Gradle</summary>

```
> Task :compileJava
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

> Task :errorprone:compileTestJava
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.

> Task :errorprone:test
WARNING: package com.sun.tools.javac.type not in jdk.compiler

> Task :compileTestJava
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use or override a deprecated API that is marked for removal.
Note: Recompile with -Xlint:removal for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
```

</details>

### Gradle vs `configuration :jacocoAgent`

<details><summary>Resolution of the configuration :jacocoAgent was attempted from a context different than the project context.</summary>

```
> Task :osgi-test:test
Resolution of the configuration :jacocoAgent was attempted from a context different than the project context. Have a look at the documentation to understand why this is a problem and how it can be resolved. This behaviour has been deprecated and is scheduled to be removed in Gradle 8.0. See https://docs.gradle.org/7.6/userguide/viewing_debugging_dependencies.html#sub:resolving-unsafe-configuration-resolution-errors for more details.
        at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration.resolveToStateOrLater(DefaultConfiguration.java:607)
[...]

PR #3052, opened 2023-07-04
Rewire Jacoco for Gradle 8/9

TWiStErRob:
Fixes the `:jacocoAgent` warning listed in https://github.com/mockito/mockito/pull/3051, see individual commits to make sense of the "full rewrite". cc @szpak for review because https://github.com/mockito/mockito/pull/1229#discussion_r148213888.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,7 +78,6 @@ jobs:
       with:
         files: build/reports/jacoco/mockitoCoverage/mockitoCoverage.xml
         fail_ci_if_error: true
-        verbose: true
 
   #
   # Android build job
@@ -167,7 +166,6 @@ jobs:
       with:
         files: subprojects/androidTest/build/reports/coverage/androidTest/debug/connected/report.xml
         fail_ci_if_error: true
-        verbose: true
 
   #
   # Release job, only for pushes to the main development branch
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,20 +70,15 @@ jobs:
         MOCK_MAKER: ${{ matrix.entry.mock-maker }}
         MEMBER_ACCESSOR: ${{ matrix.entry.member-accessor }}
 
-    - name: 7. Upload coverage report
-      run: |
-        ./gradlew coverageReport -s --scan && cp build/reports/jacoco/mockitoCoverage/mockitoCoverage.xml jacoco.xml
-        curl https://keybase.io/codecovsecurity/pgp_keys.asc | gpg --no-default-keyring --keyring trustedkeys.gpg --import # One-time step
+    - name: 7. Generate coverage report
+      run: ./gradlew coverageReport --stacktrace --scan
 
-        curl -Os https://uploader.codecov.io/latest/linux/codecov
-        curl -Os https://uploader.codecov.io/latest/linux/codecov.SHA256SUM
-        curl -Os https://uploader.codecov.io/latest/linux/codecov.SHA256SUM.sig
-
-        gpgv codecov.SHA256SUM.sig codecov.SHA256SUM
-        shasum -a 256 -c codecov.SHA256SUM
-
-        chmod +x codecov
-        ./codecov
+    - name: 8. Upload coverage report
+      uses: codecov/codecov-action@v3
+      with:
+        files: build/reports/jacoco/mockitoCoverage/mockitoCoverage.xml
+        fail_ci_if_error: true
+        verbose: true
 
   #
   # Android build job
@@ -166,21 +161,13 @@ jobs:
           ${{ github.workspace }}/emulator.log
           ${{ github.workspace }}/emulator-startup.log
 
+    # :androidTest:connectedCheck (which depends on :androidTest:createDebugAndroidTestCoverageReport) already generated coverage report.
     - name: 5. Upload coverage report
-      run: |
-        # :androidTest:connectedCheck (which depends on :androidTest:createDebugAndroidTestCoverageReport) already generated coverage report.
-
-        curl https://keybase.io/codecovsecurity/pgp_keys.asc | gpg --no-default-keyring --keyring trustedkeys.gpg --import # One-time step
-
-        curl -Os https://uploader.codecov.io/latest/macos/codecov
-        curl -Os https://uploader.codecov.io/latest/macos/codecov.SHA256SUM
-        curl -Os https://uploader.codecov.io/latest/macos/codecov.SHA256SUM.sig
-
-        gpgv codecov.SHA256SUM.sig codecov.SHA256SUM
-        shasum -a 256 -c codecov.SHA256SUM
-
-        chmod +x codecov
-        ./codecov -f subprojects/androidTest/build/reports/coverage/androidTest/debug/connected/report.xml
+      uses: codecov/codecov-action@v3
+      with:
+        files: subprojects/androidTest/build/reports/coverage/androidTest/debug/connected/report.xml
+        fail_ci_if_error: true
+        verbose: true
 
   #
   # Release job, only for pushes to the main development branch
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,7 +180,7 @@ jobs:
         shasum -a 256 -c codecov.SHA256SUM
 
         chmod +x codecov
-        ./codecov -f build/reports/coverage/androidTest/debug/connected/report.xml
+        ./codecov -f subprojects/androidTest/build/reports/coverage/androidTest/debug/connected/report.xml
 
   #
   # Release job, only for pushes to the main development branch
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -166,6 +166,22 @@ jobs:
           ${{ github.workspace }}/emulator.log
           ${{ github.workspace }}/emulator-startup.log
 
+    - name: 5. Upload coverage report
+      run: |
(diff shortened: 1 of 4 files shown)


gradle/root/coverage.gradle
```diff
@@ -1,43 +1,25 @@
-task mockitoCoverage(type: JacocoReport) {
-
-    allprojects { currentProject ->
-        // These tests do not have a `test` task.
-        if (currentProject.name in ['android']) {
-            return
```
TWiStErRob, line 6: Assuming this was meant to exclude the current :androidTest module (because :android is a java module), this is now removed. As an effect of this the coverage might go down, because there are uncovered classes:
image

TimvdLippe: We now do have tests for this module, but it does not report any coverage for these files. Can we update the android CI jobs to collect and upload the coverage as well?

gradle/root/coverage.gradle
```diff
@@ -46,4 +28,19 @@ task mockitoCoverage(type: JacocoReport) {
     }
 }
 
-task coverageReport(dependsOn: ["mockitoCoverage"]) {}
+gradle.taskGraph.whenReady {
+    mockitoCoverage.configure {
```
TimvdLippe, line 32: Can we move this into the task configuration instead? I know this was required back then, but I think they fixed this issue now and we can move it back into the same task.

TWiStErRob: Hmm, I can try. afterEvaluate doesn't work inside register, but maybe whenReady does.

gradle/root/coverage.gradle
TWiStErRob: Ah, so :android doesn't have unit tests, but it has the instrumentation integration tests in :androidTest.

I would double-test (same source code running as unit on JVM and instrumentation) with Robolectric if that's possible for these. The reason being is that coverage is really hard to share between jobs, unless Codecov has some magic for this. Locally when I ran coverageReport, it executed all the tests, if we add :androidTest:testDebugAndroidTest as a dependency like others, then running covrageReport will require an emulator to be running. Will have a look at how Codecov is set up, hopefully it's doable with it.

TimvdLippe: CodeCov already combines all code coverage reports of all test suites, including our subprojects. The last step of the pipeline does this: https://github.com/mockito/mockito/blob/c1eec880a27eee89b68b75a8995a94431ff3fdb3/.github/workflows/ci.yml#L73

I guess we need to add this last step to the android jobs as well, assuming we already gather code coverage?

TWiStErRob: Yeah, that's what I'm doing right now.

This relies on having an .xml file filled up with data. That's why step 7 gathers all subprojects (code in coverage.gradle). So need a similar setup for `:androidTest`. I think I got it working. Pushing as soon as the current CI jobs complete, so I can observe the covarge in the below comment before/after.

gradle/root/coverage.gradle
TWiStErRob: Did you mean this? https://github.com/mockito/mockito/pull/3052/commits/97151b03242ee2ed56618be5fb756058c8189d59 or I misunderstood?

TWiStErRob: Note: I moved it out in https://github.com/mockito/mockito/pull/3052/commits/32894517b02ab0e26a6753acb7de2f6ece69c9b5 because it was running an O(N^2) setup. After each subproject was evaluated, it re-set the trees, but some of the trees were already resolved, and this repeated over and over again.

.github/workflows/ci.yml
```diff
@@ -166,6 +166,22 @@ jobs:
...
+      run: |
+        # :androidTest:connectedCheck (which depends on :androidTest:createDebugAndroidTestCoverageReport) already generated coverage report.
+
+        curl https://keybase.io/codecovsecurity/pgp_keys.asc | gpg --no-default-keyring --keyring trustedkeys.gpg --import # One-time step
+
+        curl -Os https://uploader.codecov.io/latest/macos/codecov
+        curl -Os https://uploader.codecov.io/latest/macos/codecov.SHA256SUM
+        curl -Os https://uploader.codecov.io/latest/macos/codecov.SHA256SUM.sig
+
+        gpgv codecov.SHA256SUM.sig codecov.SHA256SUM
```
TWiStErRob, line 179: it seems to fail here, any idea?

TimvdLippe: In the Java jobs we download the Linux signatures, but here you are downloading the Macos signatures. I suppose locally you are using a Mac, but on CI we are using Linux?

TimvdLippe: Yes we can use that action for both the Java and Android jobs

szpak:
> cc @szpak for review because https://github.com/mockito/mockito/pull/1229#discussion_r148213888.

Wow. It was long time ago. In those times, I was much more up-to-date with the Gradle dewelopment and recommendations for plugin makers :-D.

gradle/root/coverage.gradle
```diff
@@ -1,43 +1,56 @@
...
-
-    allprojects { currentProject ->
-        // These tests do not have a `test` task.
-        if (currentProject.name in ['android']) {
-            return
+allprojects { project ->
+    plugins.withId("java") {
+        project.apply plugin: "jacoco"
+        project.jacoco {
+            toolVersion = '0.8.8'
```
szpak, line 5: Can we switch to [0.8.10](https://github.com/jacoco/jacoco/releases/tag/v0.8.10) already?

szpak: Or maybe better not, to have more clear path in case of regression.

TWiStErRob: Yep, I didn't update so we can compare the outputs.

subprojects/androidTest/androidTest.gradle
```diff
@@ -18,10 +18,37 @@ android {
...
+                plugins.withId("java") {
+                    SourceSetContainer sourceSets = currentProject.sourceSets
+                    // Mimic org.gradle.testing.jacoco.tasks.JacocoReportBase.sourceSets().
+                    // Not possible to set task.javaSources because it's initialized with setDisallowChanges.
+                    //task.javaSources.add({ [ sourceSets.main.allJava.srcDirs ] })
+                    task.classFileCollection.from(
+                        sourceSets.main.output.asFileTree.matching {
+                            exclude("**/internal/util/concurrent/**")
+                        }
+                    )
```
TWiStErRob, line 41: Do we need to exclude this?
```suggestion
                    task.classFileCollection.from(sourceSets.main.output)
```

From an earlier run it looked like it's not used by these tests:
image

TWiStErRob: Why are they excluded in the first place in the normal report?

TimvdLippe: The classes in that folder are from a different library that we bundle. See the license (https://github.com/mockito/mockito/blob/222c14c97bc2ec59ec5e486ddaa6438edded0e5c/src/main/java/org/mockito/internal/util/concurrent/LICENSE) and corresponding readme. Since we repackage this library, we don't include it in our code coverage.

TimvdLippe:
Seems like the Android coverage is now working correctly? We do decrease the coverage, but that's because we miss some stuff in Android.

I think this PR is ready now, is it not? Thanks again for taking this on!

subprojects/androidTest/androidTest.gradle
TWiStErRob: From a previous CI run (screenshot above) it looks like the androidTest is not exercising any lines from it. I wonder if the maintenance of those excludes is worth the minor percentage difference gain for the test coverage. Anyway, it works now.

TWiStErRob:
Yeah, I think this is ready. It's a bit ugly and hacky with the coverage gathering from android modules and exclusions.

And where I managed to generate coverage for a dependency of the androidTest module. The `runtimeOnly` vs `exclude` hack can be simplified if the `androidTest` is split up into `androidInstrumentedTest` and `androidUnitTest` modules.

I like how the `codecov-action` turned out though, way simpler than all the unix magic before.

The coverage change is all expected: https://app.codecov.io/gh/mockito/mockito/pull/3052/indirect-changes?src=pr&el=tree-more&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mockito

It shows that android module is now covered and some of the core mockito classes are also covered more (probably because the coverage files are also aggregated from android module's unit tests). I wouldn't look at it as a decrease, but more as a "more realistic measurement" of the coverage 😅.

If you're ok with the resulted code and tricks, happy to merge this now.

TimvdLippe:
Yeah very happy that we now have a proper view of test coverage of our Android code. It also shows we are missing a bit there.

I was about to merge this, but then I spotted one last nit: the report includes `subprojects/androidTest/src/main/java/org/mockitousage/androidtest/BasicClassesForTesting.kt`. I think we need to exclude `src/main/java` of our `androidTest` project.

TimvdLippe:
Also, with regards to cleaning up the instrumentation and unit tests for Android, I would be happy to merge a PR for that if you fancy picking it up.

TWiStErRob:
Run completed, the file disappeared and other android files are still there.

| Before https://github.com/mockito/mockito/pull/3052/commits/b60e1bb85dbe79dda48dd1196a97c6f966a35083 | After https://github.com/mockito/mockito/pull/3052/commits/b60e1bb85dbe79dda48dd1196a97c6f966a35083 |
| - | - |
| image | image |

TimvdLippe (approved):
Nice work, thank you!

PR #3062, opened 2023-07-14
Gradle build script improvements

TWiStErRob:
A step towards https://github.com/mockito/mockito/issues/2898

This is in preparation for an [other PR](https://github.com/TWiStErRob/mockito/pull/1) introducing toolchains which is a cross-module configuration. I did some generalization and consistency improvements to make it easier to make global changes. Some of the diffs in the PR might look strange, I recommend looking commit by commit to make sense of them.

Many modules already had `java-library.gradle` included. So I created a parallel `java-test.gradle`. With this there's now two named types of modules. Each subproject belongs to one of these types, except:
 * androidTest is an Android module, very special (because no java plugin)
 * bom is published, but not a library

There are some special-setup modules which still extend one of the named modules:
 * rootProject/mockito-core is a java-library (among many other things).
 * groovyInlineTest and groovyTest are Groovy modules too.
 * kotlinReleaseCoroutinesTest and kotlinTest are Kotlin modules too.
 * osgi-test does things.

There were several (test) modules missing Java compatiblility declaration:
 - Now they're all using Java 11 source/target for compilation.

Testing:
 * Generate repo snapshot by changing
   ```
   -repositories { maven { url = "$buildDir/repo" } }
   +repositories { maven { url = "${rootProject.buildDir}/repo" } }
   ```
 * Running `gradlew clean` followed by `gradlew publishAllPublicationsToMavenRepository --no-build-cache`
 * Grab contents of `rootProject/build/repo/`
 * Rename all files that have timestamps replacing timestamp with a constant.
Test matrix:
 * I repeated this on master with JAVA_HOME=JDK11, JAVA_HOME=JDK17.
 * and on the branch with JAVA_HOME=JDK11, JAVA_HOME=JDK17.
 * and on the branch with JAVA_HOME=JDK17 + `-Pmockito.test.java=11`

The changes all looked expected. Diffs were:
 * javadoc timestamps
 * Maven metadata (SNAPSHOT timestamps and artifact order)
 * Java 11 and 17 generate the same decompiled classes, but different class files and constant pools. This is on `master` already.

## Checklist

 - [x] Read the [contributing guide](https://github.com/mockito/mockito/blob/main/.github/CONTRIBUTING.md)
 - [x] PR should be motivated, i.e. what does it fix, why, and if relevant how
 - [x] If possible / relevant include an example in the description, that could help all readers
       including project members to get a better picture of the change
 - [x] Avoid other runtime dependencies
[...]

PR #3064, opened 2023-07-16
Toolchain for Test task

TWiStErRob:
Re-raise of https://github.com/TWiStErRob/mockito/pull/1 in mockito repo.

A step towards https://github.com/mockito/mockito/issues/2898, from here we could create multiple test tasks as a followup, and remove the gradle property and use the task name as dynamic matrix element instead.

Testing: see section in https://github.com/mockito/mockito/pull/3062

Additionally I added `fail(System.getProperty("java.home"))` in a test to see if the JDK was as expected.

I tried to minimize the changes, that's why I created a [separate PR to make it possible to configure all projects at once](https://github.com/mockito/mockito/pull/3062).

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,12 +65,15 @@ jobs:
       run: >
         ./gradlew
         spotlessCheck
+        --stacktrace
+        --scan
 
     - name: 6. Build on Java ${{ matrix.java }} with ${{ matrix.entry.mock-maker }} and ${{ matrix.entry.member-accessor }}
       run: >
         ./gradlew
         build
         idea
+        --stacktrace
         --scan
       env:
         MOCK_MAKER: ${{ matrix.entry.mock-maker }}
@@ -211,6 +214,7 @@ jobs:
           publishToSonatype
           closeAndReleaseStagingRepository
           releaseSummary
+          --stacktrace
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
           NEXUS_TOKEN_USER: ${{secrets.NEXUS_TOKEN_USER}}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,6 +71,7 @@ jobs:
     - name: 6. Build on Java ${{ matrix.java }} with ${{ matrix.entry.mock-maker }} and ${{ matrix.entry.member-accessor }}
       run: >
         ./gradlew
+        -Pmockito.test.java=${{ matrix.java }}
         build
         idea
         --stacktrace
@@ -82,6 +83,7 @@ jobs:
     - name: 7. Generate coverage report
       run: >
         ./gradlew
+        -Pmockito.test.java=${{ matrix.java }}
         coverageReport
         --stacktrace
         --scan
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,16 +62,26 @@ jobs:
 
     - name: 5. Spotless check (single job only). Run './gradlew spotlessApply' locally if this job fails.
       if: matrix.java == 11 && matrix.entry.mock-maker == 'mock-maker-default' # SINGLE-MATRIX-JOB
-      run: ./gradlew spotlessCheck
+      run: >
+        ./gradlew
+        spotlessCheck
 
     - name: 6. Build on Java ${{ matrix.java }} with ${{ matrix.entry.mock-maker }} and ${{ matrix.entry.member-accessor }}
-      run: ./gradlew build idea --scan
+      run: >
+        ./gradlew
+        build
+        idea
+        --scan
       env:
         MOCK_MAKER: ${{ matrix.entry.mock-maker }}
         MEMBER_ACCESSOR: ${{ matrix.entry.member-accessor }}
 
     - name: 7. Generate coverage report
-      run: ./gradlew coverageReport --stacktrace --scan
+      run: >
+        ./gradlew
+        coverageReport
+        --stacktrace
+        --scan
 
     - name: 8. Upload coverage report
       uses: codecov/codecov-action@v3
@@ -195,7 +205,12 @@ jobs:
           java-version: 17
 
       - name: 3. Build and release
-        run: ./gradlew githubRelease publishToSonatype closeAndReleaseStagingRepository releaseSummary
+        run: >
+          ./gradlew
+          githubRelease
+          publishToSonatype
+          closeAndReleaseStagingRepository
+          releaseSummary
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
           NEXUS_TOKEN_USER: ${{secrets.NEXUS_TOKEN_USER}}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,11 +46,11 @@ jobs:
       with:
         fetch-depth: '0' # https://github.com/shipkit/shipkit-changelog#fetch-depth-on-ci
 
-    - name: 2. Set up Java ${{ matrix.java }}
+    - name: 2. Set up Java for running Gradle build
       uses: actions/setup-java@v3
       with:
         distribution: 'zulu'
-        java-version: ${{ matrix.java }}
+        java-version: 17
 
     - name: 3. Validate Gradle wrapper
       if: matrix.java == 11 && matrix.entry.mock-maker == 'mock-maker-default' # SINGLE-MATRIX-JOB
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -46,11 +46,11 @@ jobs:
       with:
         fetch-depth: '0' # https://github.com/shipkit/shipkit-changelog#fetch-depth-on-ci
 
-    - name: 2. Set up Java ${{ matrix.java }}
+    - name: 2. Set up Java for running Gradle build
       uses: actions/setup-java@v3
       with:
         distribution: 'zulu'
-        java-version: ${{ matrix.java }}
+        java-version: 17
```
TWiStErRob, line 53: This will also enable using AGP 8 for testing Android.

gradle.properties
```diff
@@ -16,3 +16,9 @@ org.gradle.jvmargs=-Xmx2048m -Dfile.encoding=UTF-8 \
 android.useAndroidX=true
 # Kotlin code style for this project: "official" or "obsolete":
 kotlin.code.style=official
+
+# Version of JDK to use for running tests.
+# Possible values:
+# - auto -> use Java running Gradle, best for local dev.
+# - A Java major version number, like 11 or 17.
+mockito.test.java=auto
```
TWiStErRob, line 24: Could be an env var, but prop can be documented nicely here and also have a global default.

gradle/test-launcher.gradle
```diff
@@ -0,0 +1,13 @@
+tasks.withType(Test) {
```
TWiStErRob, line 1: Hmm, probably should apply to Android test too.

TimvdLippe:
Do you know why the code coverage has increased with this PR? Looking at the changes, I don't see a particular reason why they would cover more. Or is this the result of running on Java 17 for Android? If so, are we also missing coverage now? I didn't spot anything, but let's double check.

Great change btw. We should definitely start using toolchains and now it is clear what we are using when.

gradle/test-launcher.gradle
TimvdLippe: I never really understood the relation between Java and Android. Why does the Java toolchain matter for Android? I thought they are running their own version.

gradle/test-launcher.gradle
```diff
@@ -0,0 +1,13 @@
...
+    javaLauncher = javaToolchains.launcherFor {
+        languageVersion = providers
+            .gradleProperty("mockito.test.java")
+            .map {
+                if (it == "auto") {
+                    JavaLanguageVersion.of(JavaVersion.current().majorVersion)
```
TimvdLippe, line 7: Shouldn't we now also remove the target and source compatibility definitions, now that we use toolchains?

gradle/test-launcher.gradle
TWiStErRob: Android apps have two types of tests: unit and instrumentation. Unit tests are the same JVM-run tests as any other JAR. Instrumentation is special as in it runs on a device. In the above comment I was referring to these unit tests: https://github.com/mockito/mockito/tree/main/subprojects/androidTest/src/test/java/org/mockitousage/androidtest

Apart from tests Java version matters for Android because they didn't reinvent the wheel (too much), the build still requires .jar/.class inputs, but it spits out .dex files which are essentially over-optimized containers for a transformed bytecode. So it still matters (for AGP) what bytecode target version it's reading (can't polyfill everything, e.g. records), and it matters for Android runtime (on device), because Java cannot be updated on older devices, so we can't call certain methods depending on minSdk.

gradle/test-launcher.gradle
TWiStErRob: The toolchains are only for tests for now. I didn't change the compilation (`compileJava` / `compileTestJava` works as before (always on 11)).

TWiStErRob:
> Do you know why the code coverage has increased with this PR? Looking at the changes, I don't see a particular reason why they would cover more.

@TimvdLippe the coverage changed because master android CI failed, so coverage from instrumentation tests was not included in baseline the PR is compared against. They passed on this PR, so they're covered, that's the increase. If you [re-run this job on master](https://github.com/mockito/mockito/actions/runs/5561152759/jobs/10158631715) so it passes, it should equalize. Please do.

gradle/test-launcher.gradle
TWiStErRob: https://github.com/mockito/mockito/pull/3064/commits/2ed1724177e20d8377c707a38e8b9be77aad4b30

.github/workflows/ci.yml
```diff
@@ -183,19 +198,25 @@ jobs:
...
-      - name: Set up Java 11
+      - name: 2. Set up Java for running Gradle build
         uses: actions/setup-java@v3
         with:
           distribution: 'zulu'
-          java-version: 11
-
-      - name: Build and release
-        run: ./gradlew githubRelease publishToSonatype closeAndReleaseStagingRepository releaseSummary
+          java-version: 17
```
TWiStErRob, line 210: This line means (*).

Before:
 * Gradle Java 11* producing Java 11 bytecode (because targetCompatibility).
 * Test matrix Java 11 producing Java 11 bytecode running on Java 11
 * Test matrix Java 17 producing Java 11 bytecode running on Java 17

After:
 * Gradle Java 17* producing Java 11 bytecode (because targetCompatibility).
 * Test matrix Java 17 producing Java 11 bytecode running on Java 11
 * Test matrix Java 17 producing Java 11 bytecode running on Java 17

The key here is that the built bytecode is always the same, and is tested on different runtimes. Same as user's envs.

I think unless Java 21+ LTS [deprecates target=11](https://inside.java/2022/09/26/quality-heads-up/), we can keep using this pattern. If deprecation happens we can increase minimum, or force a toolchain, but that would remove optimizations/fixes of javac in Java 12-17.

gradle/test-launcher.gradle
TWiStErRob: <sup>1</sup>: `java-version: matrix.java -> 17` [here in `build:` job](https://github.com/mockito/mockito/pull/3064/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR53)
<sup>2</sup>: `java-version: 11 -> 17` [here in `release:` job](https://github.com/mockito/mockito/pull/3064/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR210)
<sup>3</sup>: implicit Gradle Java runtime to explicit toolchain [here](https://github.com/mockito/mockito/pull/3064/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR74)
Java 11 bytecode is forced by targetCompatibility.

Before:
 * build: Test matrix Java 11<sup>1</sup> producing Java 11 bytecode running on Java 11<sup>3</sup>
 * build: Test matrix Java 17<sup>1</sup> producing Java 11 bytecode running on Java 17<sup>3</sup>
 * release: Gradle on Java 11<sup>2</sup> producing Java 11 bytecode.

After:
 * build: Test matrix Java 17<sup>1</sup> producing Java 11 bytecode running on Java 11<sup>3</sup>
 * build: Test matrix Java 17<sup>1</sup> producing Java 11 bytecode running on Java 17<sup>3</sup>
 * release Gradle on Java 17<sup>2</sup> producing Java 11 bytecode.

The key here is that the built bytecode is always the same, and is tested on different runtimes. Same as user's envs.

I think unless Java 21+ LTS [deprecates target=11](https://inside.java/2022/09/26/quality-heads-up/), we can keep using this pattern. If deprecation happens we can increase minimum, or force a toolchain, but that would remove optimizations/fixes of javac in Java 12-17.

TimvdLippe:
@TWiStErRob can you rebase on master? I merged some Dependabot PRs and unfortunately they now clash. That should also fix the coverage issue, as I reran that job

TWiStErRob:
Clean rebase, only the trivial conflict with enterprise bump.

TWiStErRob:
Can Codecov handle force pushes? https://github.com/mockito/mockito/pull/3064#issuecomment-1637202174 is not updated yet.

gradle/test-launcher.gradle
TimvdLippe: Yes this looks fine to me. Thanks for the explanation!

settings.gradle.kts
```diff
@@ -1,5 +1,6 @@
 plugins {
     id("com.gradle.enterprise") version "3.14"
+    id("org.gradle.toolchains.foojay-resolver-convention") version "0.6.0"
```
TimvdLippe, line 3: One last question: I don't see this plugin being used. Do we really need this? I thought Gradle has Java toolchains by default.

TWiStErRob: Yes, in the past (7.x) it was, now not defining it is deprecated: https://github.com/mockito/mockito/pull/3064/commits/25958c56d3270c66b7a4669ad5cbab80b8f02edd

PR #3078, opened 2023-08-04
Fixes #3077 : Report unused stubbing exceptions when test filter is no-op

### LINKED ISSUE #3077: Unused stubbings are not reported when filters are used
Unused stubbings are not reported when a filter is used during the test run. This is documented clearly in the [StrictRunner class](https://github.com/mockito/mockito/blob/f5ad9e9f9a338bcea6180fc474a562acc13f2c84/src/main/java/org/mockito/internal/runners/StrictRunner.java#L47-L52):

```java
// only report when:
// 1. if all tests from given test have ran (filter requested is false)
//   Otherwise we would report unnecessary stubs even if the user runs just single test
// from the class
// 2. tests are successful (we don't want to add an extra failure on top of any existing
// failure, to avoid confusion)
```

Only reporting unused stubbings when all tests are run is reasonable. However, the logic for detecting this case is a little too coarse-grained, since even no-op filters end up disabling the reporting of unused stubbings. This can be troublesome when, for example, JUnit categories are used to distinguish between unit and integration tests.

It'd be nice if we could detect when filters actually prevent one or more tests from being run, and only skip reporting unused stubbings in those cases.

Created based off of https://github.com/gradle/gradle/issues/10694, specifically [this comment](https://github.com/gradle/gradle/issues/10694#issuecomment-1374911274).

This affects a project currently using Mockito version 4.11.0, but I've just verified that it also affects version 5.2.0.

C0urante:
Fixes #3077 

## Summary

Adds granularity to the logic that skips reporting of unused stubbings when tests are filtered. Instead of simply detecting the use of any kind of filter, intercepts uses of that filter and only skips reporting of unused stubbings when the filter actually prevents one or more tests from being run.

This allows filters (such as JUnit category excludes, which themselves can be used to separate unit and integration tests) to be applied during a test run without sacrificing the benefits of strict stubbing.

Testing coverage is added for this newly-handled case (i.e., when a filter does not prevent any tests from being run), as well as existing cases (i.e., when a filter does prevent at least one test from being run).

## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/runners/StrictRunner.java
+++ b/src/main/java/org/mockito/internal/runners/StrictRunner.java
@@ -16,7 +16,7 @@ public class StrictRunner implements InternalRunner {
 
     private final Class<?> testClass;
     private final InternalRunner runner;
-    private boolean filterRequested;
+    private boolean testsSkipped;
 
     /**
      * @param runner - the runner to wrap around
@@ -25,6 +25,7 @@ public class StrictRunner implements InternalRunner {
     public StrictRunner(InternalRunner runner, Class<?> testClass) {
         this.runner = runner;
         this.testClass = testClass;
+        this.testsSkipped = false;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/org/mockito/internal/runners/StrictRunner.java
```diff
@@ -16,7 +16,7 @@ public class StrictRunner implements InternalRunner {
 
     private final Class<?> testClass;
     private final InternalRunner runner;
-    private boolean filterRequested;
+    private volatile boolean testsSkipped;
```
TimvdLippe, line 19: Why does this need to be `volatile`? We try to stay away of that as much as we can, as multi-threading is already a difficult story with Mockito. As I see it, there is no need for this to be shared across instances, so I think this is not necessary?

C0urante: Ah thanks, good catch! I wasn't familiar with the threading model at play here and wanted to be conservative. I've removed the `volatile` modifier.

TimvdLippe (approved):
Thanks!

C0urante:
Thanks @TimvdLippe!

Issue #3087, opened 2023-08-10
Backporting policy


C0urante:
With version 5.0.0, Java 8 support was dropped and Java 11 is required for that version and beyond. If a project is still relying on Java 8 (or 9, or 10), they cannot benefit from any new fixes or features without upgrading their Java version, unless those same changes are backported to 4.x.y and released (or they want to fork and package Mockito themselves).

Given that, it'd be nice to know what the project's official policy on backporting is. Based on recent releases, it doesn't seem like backporting is common (the last release on the 4.x.y line is 4.11.0, which was published in December 2022, a month or so before 5.0.0, which was published in January 2023).

Assuming this practice is intentional but undocumented, a short note added to the README (or possibly CONTRIBUTING?) doc stating that 4.x.y is no longer maintained and no new features, fixes, or security patches will be applied to it could be valuable.

C0urante:
(To be clear: this is not a request for any commits on the 5.x.y line to be backported to 4.x.y, or for any future releases of 4.x.y to take place.)

TimvdLippe:
Yes that understanding is correct. We do not backport changes and have not had any need for that in the past years. Instead, we aim to make breaking changes as minimally impactful as we can, so that upgrading is as easy as we can make it. We do take breaking changes serious for that reason and the Java 8 decision was not taken lightly.

I would be happy to merge a PR that adds such a clarification to our README. I think the first paragraph on the various versions could be expanded upon for that? Do you mind sending us a PR for that?

C0urante:
Sure thing! Opened https://github.com/mockito/mockito/pull/3090

PR #3090, opened 2023-08-11
Fixes 3087 : Add note on backporting to README

C0urante:
Fixes https://github.com/mockito/mockito/issues/3087

Adds a (very brief) note stating that backporting does not take place and older major versions are not supported.


TimvdLippe (approved):
Thank you!

PR #3097, opened 2023-08-17
Default mock of Optional.isEmpty() returns true for RETURN_DEEP_STUBS

### LINKED ISSUE #2865: Default mock of `Optional` is not `empty` when using `RETURN_DEEP_STUBS`
Version affected: Mockito: 4.x & 5.x

If you create a mock with `RETURN_DEEP_STUBS`, then any method that returns a collection will, by default, return an empty collection, i.e. `isEmpty` returns `true`, etc.  This is great, as it means no additional configuration of the mock is required before passing to logic that would attempt to access elements of the collection.

However, the same is not true for `Optional`, (and I'm assuming `OptionalLong` etc): if `RETURN_DEEP_STUBS` is used, methods that return `Optional` aren't treated in any special way and just return a default mock with not stubbing configured. This means `isEmpty` returns `false`, but `get()` returns `null`.   This breaks the contract of `Optional`, causing test failures that aren't actually valid, or requiring manual configuration of the `Optional` mock to do-the-right-thing, cluttering test code.

Here's an example test that demonstrates this:

```java
interface Address {
   Optional<String> getState();
}

@Test
public void shouldDefaultDeepMockOptionalsToEmpty() {
    final Address address = mock(Address.class, RETURNS_DEEP_STUBS);

    assertThat(address.getState()).isEqualTo(Optional.empty());  // <--- fails
}

@Test
public void shouldDefaultMockOptionalsToEmpty() {
    final Address address = mock(Address.class); // <-- no RETURNS_DEEP_STUBS

    assertThat(address.getState()).isEqualTo(Optional.empty());  // <--- passes
}
```

Note the second test, which doesn't use deep mocking, passes.

AndreasTu:
ReturnsDeepStubs now answers with true on Optional.isEmpty() when using RETURN_DEEP_STUBS.
ReturnsDeepStubs now returns real Optionals instead of mocks.

Fixes #2865

## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsDeepStubs.java
+++ b/src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsDeepStubs.java
@@ -54,6 +54,10 @@ public class ReturnsDeepStubs implements Answer<Object>, Serializable {
         MockCreationSettings<?> mockSettings = MockUtil.getMockSettings(invocation.getMock());
 
         Class<?> rawType = returnTypeGenericMetadata.rawType();
+        final var emptyValue = ReturnsEmptyValues.returnCommonEmptyValueFor(rawType);
+        if (emptyValue != null) {
+            return emptyValue;
+        }
         if (!typeMockabilityOf(rawType, mockSettings.getMockMaker()).mockable()) {
             if (invocation.getMethod().getReturnType().equals(rawType)) {
                 return delegate().answer(invocation);
--- a/src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValues.java
+++ b/src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValues.java
@@ -139,14 +139,6 @@ public class ReturnsEmptyValues implements Answer<Object>, Serializable {
             return new TreeMap<>();
         } else if (type == LinkedHashMap.class) {
             return new LinkedHashMap<>();
-        } else if (type == Optional.class) {
-            return Optional.empty();
-        } else if (type == OptionalDouble.class) {
-            return OptionalDouble.empty();
-        } else if (type == OptionalInt.class) {
-            return OptionalInt.empty();
-        } else if (type == OptionalLong.class) {
-            return OptionalLong.empty();
         } else if (type == Stream.class) {
             return Stream.empty();
         } else if (type == DoubleStream.class) {
@@ -160,8 +152,27 @@ public class ReturnsEmptyValues implements Answer<Object>, Serializable {
         } else if (type == Period.class) {
             return Period.ZERO;
         }
-
         // Let's not care about the rest of collections.
+
+        return returnCommonEmptyValueFor(type);
+    }
+
+    /**
+     * Returns empty values for common known types, shared between {@link ReturnsEmptyValues} and {@link ReturnsDeepStubs}.
+     *
+     * @param type the type to check
+     * @return the empty value, or {@code null}
+     */
+    static Object returnCommonEmptyValueFor(Class<?> type) {
+        if (type == Optional.class) {
+            return Optional.empty();
+        } else if (type == OptionalDouble.class) {
+            return OptionalDouble.empty();
+        } else if (type == OptionalInt.class) {
+            return OptionalInt.empty();
+        } else if (type == OptionalLong.class) {
+            return OptionalLong.empty();
+        }
         return null;
     }
 }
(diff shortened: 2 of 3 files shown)


src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsDeepStubs.java
```diff
@@ -48,6 +53,10 @@ public class ReturnsDeepStubs implements Answer<Object>, Serializable {
 
     @Override
     public Object answer(InvocationOnMock invocation) throws Throwable {
```
TimvdLippe, line 55: In this answer, on line 68 we already call `delegate().answer()` where the delegate is `ReturnsEmptyValues`. This class in turn already returns an empty optional: https://github.com/mockito/mockito/blob/076e8ac837c17e81347854feaad69687cd7d2b83/src/main/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValues.java#L142-L149

Is there a different reason why we are not hitting that codepath?

AndreasTu: 1. Because `Optional` returns to be mockable in line 66 `!typeMockabilityOf(rawType, mockSettings.getMockMaker()).mockable()`.
2. The `delegate().answer()` would return a real `Optional`, which was mentioned in the ticket not to be the right thing to return. As the `RETURNS_DEEP_STUBS` also do not return `delegate().answer()` for e.g. `Lists`.

TimvdLippe: In that case I would say the delegation is not working as expected and we should fix that instead. I would like us not to reimplement the logic of returning empty values in another place, which is why we have the delegate. If that is not working properly, I consider that the bug. Can you look into fixing that while we make sure we don't break other use cases?

AndreasTu: @TimvdLippe Yes I can do that.
But just to be sure, that will also return a real Optional in the `RETURNS_DEEP_STUBS` case.
So the observable behavior for the client will be different for e.g. Lists and Optionals, when using  `RETURNS_DEEP_STUBS`.

AndreasTu: @TimvdLippe done

big-andy-coates: Are we sure this is what we want?

This would make the default behaviour consistent, i.e. returning an empty optional, but would change the way users would need to deep mock methods returning optionals (and collections?).  Instead of mocking the calls to the optional's methods, users would need to mock the method returning the optional.

For example:

```java
when(a.getOptional().isPresent()).thenReturn(true);
when(a.getOptional().get()).thenReturn("thing");
```

Would need to become:

```java
when(a.getOptional()).thenReturn(Optional.of("thing"));
```

Or, if they really did still want an optional mock for some reason:

```java
Optional<String> mockOptional = mock(Optional.class);
when(mockOptional.isPresent()).thenReturn(true);
when(mockOptional.get()).thenReturn("thing");
when(a.getOptional()).thenReturn(mockOptional);
```

Personally, I think returning a real empty optional is better as the optional is an immutable object, so returning an immutable instance means the immutability is maintained in testing by default, highlighting unreachable branches etc.

It is a change that would need documenting both in the standard docs and in the release notes to help users upgrade (it's a breaking change).

AndreasTu: @TimvdLippe The proposed change to use `ReturnsEmptyValues` breaks multiple existing (>=12) tests.
Shall I go forward and change these tests to comply to the new behavior?
This seams like an invasive change.

Example:
`org.mockitousage.basicapi.MocksSerializationTest#BUG_ISSUE_399_try_some_mocks_with_current_answers`
`org.mockito.internal.stubbing.defaultanswers.ReturnsGenericDeepStubsTest#will_return_default_value_on_non_mockable_nested_generic`

and more...

TimvdLippe: Hm yes that is too many breakages unfortunately. Still, I would prefer us to not implement this same list twice. Can we come up with a way that uses both?

AndreasTu: I could delegate the check into `ReturnsEmptyValue` where I extract the Optional checks into an own method, then we have the different logic, but near each other to see the differences.

AndreasTu: @TimvdLippe Done.

PR #3123, opened 2023-09-25
Fixes #3005

### LINKED ISSUE #3005: ClassCastException in openMocks when using wildcard types [5.3.1]
When using a combination of `@InjectMocks` and `@Mock` to inject a field that has a wildcard generic type, a ClassCastException occurs:

```
java.lang.ClassCastException: class sun.reflect.generics.reflectiveObjects.WildcardTypeImpl cannot be cast to class java.lang.Class (sun.reflect.generics.reflectiveObjects.WildcardTypeImpl and java.lang.Class are in module java.base of loader 'bootstrap')

	at org.mockito.internal.configuration.injection.filter.TypeBasedCandidateFilter.isCompatibleTypes(TypeBasedCandidateFilter.java:57)
	at org.mockito.internal.configuration.injection.filter.TypeBasedCandidateFilter.lambda$recurseOnTypeArguments$2(TypeBasedCandidateFilter.java:138)
	at java.base/java.util.stream.MatchOps$1MatchSink.accept(MatchOps.java:90)
	at java.base/java.util.stream.Streams$StreamBuilderImpl.tryAdvance(Streams.java:397)
	at java.base/java.util.stream.Streams$ConcatSpliterator.tryAdvance(Streams.java:723)
	at java.base/java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:129)
	at java.base/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:527)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:513)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.MatchOps$MatchOp.evaluateSequential(MatchOps.java:230)
	at java.base/java.util.stream.MatchOps$MatchOp.evaluateSequential(MatchOps.java:196)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.anyMatch(ReferencePipeline.java:632)
	at org.mockito.internal.configuration.injection.filter.TypeBasedCandidateFilter.recurseOnTypeArguments(TypeBasedCandidateFilter.java:136)
	at org.mockito.internal.configuration.injection.filter.TypeBasedCandidateFilter.isCompatibleTypes(TypeBasedCandidateFilter.java:51)
	at org.mockito.internal.configuration.injection.filter.TypeBasedCandidateFilter.filterCandidate(TypeBasedCandidateFilter.java:167)
	at org.mockito.internal.configuration.injection.PropertyAndSetterInjection.injectMockCandidatesOnFields(PropertyAndSetterInjection.java:136)
	at org.mockito.internal.configuration.injection.PropertyAndSetterInjection.injectMockCandidates(PropertyAndSetterInjection.java:113)
	at org.mockito.internal.configuration.injection.PropertyAndSetterInjection.processInjection(PropertyAndSetterInjection.java:81)
	at org.mockito.internal.configuration.injection.MockInjectionStrategy.process(MockInjectionStrategy.java:68)
	at org.mockito.internal.configuration.injection.MockInjectionStrategy.relayProcessToNextStrategy(MockInjectionStrategy.java:91)
	at org.mockito.internal.configuration.injection.MockInjectionStrategy.process(MockInjectionStrategy.java:71)
	at org.mockito.internal.configuration.injection.MockInjectionStrategy.relayProcessToNextStrategy(MockInjectionStrategy.java:91)
	at org.mockito.internal.configuration.injection.MockInjectionStrategy.process(MockInjectionStrategy.java:71)
	at org.mockito.internal.configuration.injection.MockInjection$OngoingMockInjection.apply(MockInjection.java:88)
	at org.mockito.internal.configuration.DefaultInjectionEngine.injectMocksOnFields(DefaultInjectionEngine.java:26)
	at org.mockito.internal.configuration.InjectingAnnotationEngine.injectCloseableMocks(InjectingAnnotationEngine.java:118)
	at org.mockito.internal.configuration.InjectingAnnotationEngine.processInjectMocks(InjectingAnnotationEngine.java:61)
	at org.mockito.internal.configuration.InjectingAnnotationEngine.process(InjectingAnnotationEngine.java:48)
	at org.mockito.MockitoAnnotations.openMocks(MockitoAnnotations.java:81)
	at nl.topicus.mockito.debug.OpenMocksTest.initMocks(OpenMocksTest.java:22)
```

*Mockito version:* 5.3.1
*JDK:* Zulu 20.0.1 and 17.0.7
*OS:* Does not appear to be a factor, reproduced on both Ubuntu 22.04.2 LTS and Windows 10

[Example reproduction code attached](https://github.com/mockito/mockito/files/11420350/mockito-5.3.1-openmocks-classcastexception-wildcardtype.zip)

nineninesevenfour:
Should fix #3005

## Checklist

Fixes several issues with @InjectMocks when generics are involved. Rewrite of matching logic based on resolving the concrete types down the type hierarchy.


nineninesevenfour:
Hmm... not sure, where the separate artefact should exactly hook into.
As far as I can see, only one instance of `AnnotationEngine` is supported. That is the reason why `SpyAnnotationEngine` (for `@Spy`) and `IndependentAnnotationEngine`(for the other annotations) are created on behalf of `InjectingAnnotationEngine`. That means `InjectingAnnotationEngine` cannot just be moved out into a separate component, when you want to keep other annotations in core. So I do not see an easy way to untangle that right now. Additionally the method `AutoCloseable process(Class<?> clazz, Object testInstance);`is too simple to support `@InjectMocks` as it is now. More surrounding information is needed (which mocks exist, where do they come frome, etc.).

The situation for the other extensions loaded through `PluginLoader` seems similar: Please blame me, if I am wrong, but as far as I can see it is always "one or the other" meaning the interfaces do not allow differennt implementations to stack up or exist side by side.

Maybe you have a concrete idea how to achieve what you proposed, then please let me know.

So I propose of the following:
* merge the PR for now to make the issues go away and hold up Mockitos reputation
* we start designing a new SPI mechanism for Mockito 6 that allows augmenting and intercepting the mocking process.
* then with mockito 6 we move out `@InjectMocks` entirely and put it into a separate module.

nineninesevenfour:
@TimvdLippe 
Hello Tim,

I have now moved the changes into a separate submodule "inject-generics". To achieve that I made the following steps:
* Making the interface `MockCandidateFilter` pluggable through the existing plugin mechanism. However, I did not move it into the `org.mockito.plugins` package.
* Introducing a new category of plugins: "internal plugins". The reason is that I think `MockCandidateFilter` is not suitable as a public plugin API. See: [InternalPlugins.java](https://github.com/mockito/mockito/pull/3123/files#diff-3a2ad3229240b0c097e7ba8bb6b13fba8bd409ea6d24adc4f9c02148fd92d76d)
* Move the new classes into the submodule together with a new `TypeWithGenericsCandidateFilter`.
* Reverted other changes in core related to changing `TypeBasedCandidateFilter`.

Is that an approach you can live with?

The name of the new submodule is a matter of taste, if you think of a better name, pls. let me know.
Do you know, if the submodule will be automatically published as "mockito-inject-generics"?

Deprecation of `@InjectMocks` I do not see as a subject of this PR, this should be a separate issue, IMHO. (only one step at a time).

Please let me know, what you think.

TimvdLippe:
As you probably noticed, I haven't had the time to look at this PR. My current precious free time is spent on #3037 which requires our attention at the moment. Hopefully when that discussion settles, we can go back to this PR.

karniemi:
@nineninesevenfour  thanks for working on this! My team is also eagerly waiting for this set of fixes. There seems to be some conflicts already, I wonder if you have time to resolve them before @TimvdLippe hops in to review the changes?

src/main/java/org/mockito/internal/configuration/plugins/DefaultMockitoPlugins.java
```diff
@@ -63,6 +65,8 @@ public class DefaultMockitoPlugins implements MockitoPlugins {
         DEFAULT_PLUGINS.put(
                 DoNotMockEnforcer.class.getName(),
                 "org.mockito.internal.configuration.DefaultDoNotMockEnforcer");
+        DEFAULT_PLUGINS.put(
+                MockCandidateFilter.class.getName(), TypeBasedCandidateFilter.class.getName());
```
TimvdLippe, line 69: While type-safety is nice, we don't want to refer to the concrete implementations here so to avoid loading the classes if they are overridden. Therefore, let's use the string names here instead like the other plugin definitions.

subprojects/inject-generics/inject-generics.gradle
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) 2016 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+apply from: "$rootDir/gradle/dependencies.gradle"
+apply from: "$rootDir/gradle/java-library.gradle"
+apply from: "$rootDir/gradle/root/coverage.gradle"
+
+description = "Filter @InjectMocks according to Java's assignment rules for generics"
```
TimvdLippe, line 9: Should we make this artifact name `inject-mocks` instead? The longterm plan would be to move `@InjectMocks` into this package so that users can rely on that instead. While the current artifact would solve the problem with generics, I ultimately want to move `@InjectMocks` here and move all related machinery into one place.

subprojects/inject-generics/src/test/java/org/mockito/internal/util/reflection/generic/GenericTypeMatchTest.java
```diff
@@ -0,0 +1,421 @@
...
+
+        @ParameterizedTest
+        @CsvSource({
+            "integerList          ,integerList          ,true",
+            "integerList          ,integerCollection    ,true",
+            "integerCollection    ,integerList          ,false",
+            "integerList          ,stringList           ,false",
+            "stringList           ,integerList          ,false",
+
+            "stringList           ,wildcardList         ,true",
```
TimvdLippe, line 70: Nit: add a case for `"integerList           ,wildcardListInteger         ,true",`

subprojects/inject-generics/src/test/java/org/mockito/internal/util/reflection/generic/GenericTypeMatchTest.java
```diff
@@ -0,0 +1,421 @@
...
+        private List<String> stringList;
+        private Collection<?> wildcardCollection;
+
+        private List<Change> changeList;
+
+        @ParameterizedTest
+        @CsvSource({
+            "integerList         ,integerCollectionBox   ,CollectionBox  ,collection  ,true",
+            "integerList         ,stringCollectionBox    ,CollectionBox  ,collection  ,false",
+            "integerList         ,wildcardCollectionBox  ,CollectionBox  ,collection  ,false",
```
TimvdLippe, line 176: In the previous test `stringList` was assignable to `wildcardList`. I would have expected the same here. Why is it false in this case?

subprojects/inject-generics/src/test/java/org/mockitousage/regression/GenericTypeMockTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Tests that verify Mockito can discern mocks by generic types, so if there are multiple mock candidates
+ * with the same generic type but different type parameters available for injection into a given field,
+ * Mockito won't fail to inject (even if mock field name doesn't match under test's field name).
+ */
+@ExtendWith(MockitoExtension.class)
+public class GenericTypeMockTest {
```
TimvdLippe, line 25: Love these!

TimvdLippe (commented):
There's a lot of code here which took me a while to get through. Overall, I like the regression tests and I understand why all of the implementation is required. Therefore, I think this is good to be optional for users to include, as the overhead of these filters can be large, but sometimes is required to get things right.

I left some comments on small improvements that we need to make, but overall this approach LGTM! Thanks again for the work and apologies for the long review time because of the large content of this PR.

Spikhalskiy:
It would be great to see this effort revived. It looks like it addresses multiple regressions that happened in mockito 5.x versions that keep a lot of legacy users, including us, from upgrading.

subprojects/inject-generics/src/test/java/org/mockito/internal/util/reflection/generic/GenericTypeMatchTest.java
Spikhalskiy: From these tests, it looks like "CollectionBox" is matched only with the exact generic type match (and it's coherent across all test cases here)
stringList -> wildcardList is a more solid match because it's a one-to-one collection-to-collection match. While a CollectionBox is not a collection by itself. So I guess in the current implementation it considered a reach to match a wildcarded not-itself-a-collection with a specific collection.

Spikhalskiy: This was corrected in the work continued in #3389

PR #3137, opened 2023-10-09
Allow use of Mockito jar as an agent.

raphw:
Adds premain method such that Mockito can be used as an agent and enable instrumentation, considering upcoming restrictions on dynamic attach.


TimvdLippe (commented):
Changes look good to me. Should we also update our CI to run on 21 or add a regression test in a different way to ensure we remain compatible?

raphw:
Likely a good idea. But probably in a separate PR?

TimvdLippe:
Sure. That works for me. As long as we make sure it does fix the problem on Java 21.

hgschmie:
This is open for two months now. Any ETA when this will hit an actual mockito release?

TimvdLippe:
We are awaiting the resolution of the Maven team on what the official migration path for projects such as Mockito will be: https://github.com/mockito/mockito/issues/3037#issuecomment-1793424678

TimvdLippe:
Superseded by #3437

PR #3144, opened 2023-10-12
Feat: add generic-inferred methods for constructing ArgumentCaptors

ascopes:
Added a new method .captor(U...) on the ArgumentCaptor class that has the same interface as Mockito.mock(U...) with the goal of removing the need for unchecked-warning suppression when users make use of generic types in argument captors.

This allows the following usage without raising compiler warnings:

```java
ArgumentCaptor<List<Thingy>> thingyCaptor = ArgumentCaptor.captor();
```

...rather than the current solution, which is less elegant...

```java
@SuppressWarnings({"unchecked", "rawtypes"})
ArgumentCaptor<List<Thingy>> thingyCaptor = (ArgumentCaptor<List<Thingy>>) ArgumentCaptor.forClass(List.class);
```

A side effect of wanting to write tests for this meant I needed to expose the class held within a captor, so an additional method named `.getCaptorType()` has also been added. I have added this as public as it feels like a useful thing to expose anyway, given the user will know what is held internally. However, if this is unacceptable then I am happy to demote it to package-private or remove it and replace it with something like `assertThat(captor).hasPropertyOrFieldWithValue("clazz", expectedClass);`, or anything else that is suggested.

## Checklist

### DIFF
--- a/src/main/java/org/mockito/ArgumentCaptor.java
+++ b/src/main/java/org/mockito/ArgumentCaptor.java
@@ -135,15 +135,6 @@ public class ArgumentCaptor<T> {
     }
 
     /**
-     * Get the raw class being captured.
-     *
-     * @return the raw class that is being captured by this captor.
-     */
-    public Class<? extends T> getCaptorType() {
-        return clazz;
-    }
-
-    /**
      * Build a new <code>ArgumentCaptor</code>.
      * <p>
      * An <code>ArgumentCaptor</code> will perform type checks (since Mockito 5.0.0).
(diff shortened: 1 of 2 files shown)


src/main/java/org/mockito/ArgumentCaptor.java
```diff
@@ -134,6 +134,15 @@ public List<T> getAllValues() {
         return this.capturingMatcher.getAllValues();
     }
 
+    /**
+     * Get the raw class being captured.
+     *
+     * @return the raw class that is being captured by this captor.
+     */
+    public Class<? extends T> getCaptorType() {
```
TimvdLippe, line 142: Do we need to expose this as public API or do we solely use this for testing? I think it is the latter (which is good), but then let's make this package-protected.

TimvdLippe (commented):
Only 1 nit. The rest LGTM! Nice fix indeed.

src/main/java/org/mockito/ArgumentCaptor.java
ascopes: Sounds good, updated the PR as suggested!

PR #3148, opened 2023-10-18
Disable fail-fast in CI build matrix

valfirst:
This option is needed to be disabled to collect information on failures from all build matrix entries. Otherwise the first failure cancels the executions of all other entries.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,7 @@ jobs:
           - { mock-maker: 'mock-maker-subclass', member-accessor: 'member-accessor-module' }
           - { mock-maker: 'mock-maker-subclass', member-accessor: 'member-accessor-reflection' }
           - { mock-maker: 'mock-maker-inline', member-accessor: 'member-accessor-reflection' }
+      fail-fast: false
 
     # All build steps
     # SINGLE-MATRIX-JOB means that the step does not need to be executed on every job in the matrix


TimvdLippe:
Don't we want this functionality to save some CI resources? I understand that when debugging CI it is benificial to skip fail-fast, but for the most common cases, I think it is better to save some CI resources. WDYT?

valfirst:
@TimvdLippe It's totally up to you. I believe you have better vision here.

While adding Java 21 to CI and updating to JaCoCo 0.8.11 it was challenging for me to capture all failures and to find some patterns (with JaCoCo update errors were swallowed quietly and I was able to make a guess only after disabling fail-fast option). That's why I decided to open this PR. At the same time I see no point to run full CI in case of, for example, Spotless violation - the first failed job is enough to understand the root cause.

Speaking about CI resources savings: all jobs start in parallel, once any job is failed, the rest jobs are cancelled; so some CPU time can be saved here, but still all VMs were allocated here. Savings may be increased if jobs are in batches sequentially and fail fast: run Java 11, if all jobs are passed, then run Java 17, etc. But that's a bit another discussion.

TimvdLippe:
Thanks for the additional context. For CI debugging I agree this flag can be removed. However, most PRs go through the regular flow, so I think the fail-fast default makes sense there. Therefore, let's keep it. Thanks!

PR #3159, opened 2023-10-25
Fixes #3061 : Deprecated mockedTypeIsInconsistentWithDelegatedInstance()

### LINKED ISSUE #3061: Non-existent MockSettings.delegatedInstance() is referenced in error message
Hi,

I noticed that the error message in [Reporter#mockedTypeIsInconsistentWithDelegatedInstanceType](https://github.com/mockito/mockito/blob/main/src/main/java/org/mockito/internal/exceptions/Reporter.java#L938) refers to `withSettings().delegatedInstance()` but that method no longer exists on `MockSettings`.

If there is a some other API to replace it with, let me know, and I'm happy to send a PR with the change.

u4859839:
This PR resolves issue #3061.

Initially I thought it would be possible to replace the delegatedInstance() reference. However, investigating the origins of the mockTypeIsInconsistentWithDelegatedInstance exception, it seems to me that the exception is not needed. It is only thrown in the validateDelegatedInstance method, and this method is never called anywhere in the codebase.  Thus, the exception can be deprecated. 

This will avoid any further confusion with the delegatedInstance() reference. 

This is my first ever pull request, so any feedback is greatly appreciated. 

## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/exceptions/Reporter.java
+++ b/src/main/java/org/mockito/internal/exceptions/Reporter.java
@@ -925,6 +925,12 @@ public class Reporter {
         return details.getCause().getMessage();
     }
 
+    /**
+     * Exception is no longer required. To mock with a delegate of a different type, use delgatesTo().
+     *
+     * @deprecated
+     */
+    @Deprecated
     public static MockitoException mockedTypeIsInconsistentWithDelegatedInstanceType(
             Class<?> mockedType, Object delegatedInstance) {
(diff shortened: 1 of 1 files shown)


TimvdLippe:
If these methods are no longer referenced in the project, then rather than marking them as deprecated, let's remove them altogether. Since these methods are in `org.mockito.internal`, we don't consider its API part of our SemVer guarantee.

PR #3184, opened 2023-11-29
Apply spotless to all java projects

AndreasTu:
This activates spotless checks for all subprojects.

The second commit reformats the files of all subprojects.


TimvdLippe (approved):
Thank you!

Issue #3212, opened 2023-12-22
Documentation on javadoc.io needs attention


michaelMarquard:
Hello 

Since most of your reference documentation is based on javadoc.io I just want to point your attention to the javadoc.io issue https://github.com/maxcellent/javadoc.io/issues/159

It seems that this issue is of minor importance to the javadoc.io team 
but maybe it is important for you... 

As a work around:
Maybe there is a way to generate your documentation without the search field?

Best regards,

Michael

TimvdLippe:
Happy to review a PR that tweaks our JavaDoc configuration. It hasn't had much love lately. The navigation link CSS is also broken, so that needs to be fixed as well (https://github.com/mockito/mockito.github.io/issues/60)

scordio:
Hi @TimvdLippe, I could give a stab at this one as I've been fixing similar issues lately in AssertJ where we have a custom stylesheet (assertj/assertj#3191 and assertj/assertj#3324).

Looking at:

https://github.com/mockito/mockito/blob/3db5df66858ba94aa5776bae97becac62ee45407/.github/workflows/ci.yml#L207-L211

do I understand correctly that Mockito is released with Java 17, i.e., the Javadoc gets generated under Java 17?

Would you consider an upgrade to Java 21, assuming it's compatible with the rest of the build?

This is what we do in AssertJ (still retaining Java 8 compatibility for the actual JAR).

TimvdLippe:
With Mockito 5, we support Java 11+. We currently indeed publish from Java 17 and if upgrading that release job to Java 21 would fix the JavaDoc issues, I am all for it. Would love to review a PR that takes a stab at this!

scordio:
Just for reference, https://www.javadoc.io/doc/org.mockito/mockito-core/5.4.0/index.html appears to be the latest correctly rendered Javadoc.

scordio:
I noticed [`mockito-junit-jupiter`](https://www.javadoc.io/doc/org.mockito/mockito-junit-jupiter/latest/index.html) doesn't have a custom stylesheet. I assume the [stylesheet](https://github.com/mockito/mockito/blob/main/src/javadoc/stylesheet.css) of `mockito-core` should be applied also here, right?

TimvdLippe:
Yes it should indeed be applied there as well

scordio:
I wasn't able to do a 1 to 1 conversion but I tried to replicate the style as much as possible.

Now I only need to figure out why #3226 fails 🙃 

[...]

PR #3226, opened 2024-01-07
Run release job on Java 21

scordio:
Second step for fixing #3212, related to https://github.com/mockito/mockito/issues/3212#issuecomment-1879995932.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -208,7 +208,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'zulu'
-          java-version: 17
+          java-version: 21
 
       - name: 3. Build and release
         run: >


.github/workflows/ci.yml
```diff
@@ -120,7 +120,7 @@ jobs:
       uses: actions/setup-java@v4
       with:
         distribution: 'zulu'
-        java-version: 17
+        java-version: 21
```
TimvdLippe, line 123: This one needs to remain 17 I think

TimvdLippe:
The build appears to fail on `androidTest`. We don't need to run those tests in Java builds afaik, as we run them in de Android builds. Still, I don't think Android supports Java 21, but I do think it can support it via toolchains. Can we set the toolchain to 21 but the compatibility remains on 11 and see if that works?

.github/workflows/ci.yml
```diff
@@ -50,7 +50,7 @@ jobs:
...
```
TimvdLippe, line 53: Since we run `:javadoc` as part of the release job (https://github.com/mockito/mockito/actions/runs/7453524063/job/20279319914#step:4:95) I think we can omit the bump here. I think that will avoid the build issues that you experience, yet still solve our JavaDoc troubles. Thanks again for taking a look!

Issue #1509, opened 2018-10-03
Improve performance of MockitoJUnitRunner


mdeinum:
When working with Mockito and using `MockitoJUnitRunner` combined with `@Mock`/`@Spy` and friends there is quite a performance impact. 

When using `@Mock` each time an `@Test` method is encountered a new instance of the mock is being created. (same applies to `@Spy`). I've rewritten the test to use `Mockito.mock` in an `@BeforeClass` and use `Mockito.reset` in an `@Before` method and performance of the tests went up considerably. 

Resulting from 50% to 80% in some cases. (From 1s to about 245ms for instance). Apparently mock creation is a heavy operation. 

Using some form of caching in the testing infrastructure would/could be beneficial in improving the performance of using `@Mock` and friends.

bric3:
Hi, this is an interesting idea, however I wouldn't make this a default behavior because you never know if production code caches these objects somewhere and perform reference equality. Wjile this is probably not the case in any sane unit test and production code, it is not something we can rely on.

mdeinum:
Not sure why that would be the case in production code (not sure how Mockito would apply there). 

Actually Spring Boot does something similar and reuse mocked beans/instances and resets them before the test. So it appears to work. 

I was thinking something along the lines of caching/reuse it on a per test class basis or provide a plugable strategy of some sorts. I didn't investigate further where the time was actually spend.

PR #3227, opened 2024-01-08
Fixes 1509: Add a lightweight extention quicker but with more limited responsibilities

topizio:
Hi guys, that's the first time I'm trying to contributing to the project, so if my code is in the right direction I might probably need some guidance in order to complete the checklist.

This implementation is meant to be a replacement for tests where automatic mock injections aren't needed but test speed is the priority. This, without allowing mocks to be over-mocked.

You can see hereby the speed on a given test, from:
image

to:
image

The test is defined as:
```java
ExtendWith(MockitoLightweightExtension.class)
class CapabilityFeatureVerifierTest {

	private final FeatureToggleService featureToggleService = mock(FeatureToggleService.class);

	private final CapabilityFeatureVerifier capabilityFeatureVerifier = new CapabilityFeatureVerifier(
			featureToggleService);
}
```

Over-mocking is still detected over mock fields and regular mocks:
 
image

Thanks in advance for your time! 

## Checklist

### DIFF
--- a/subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/MockitoLightweightExtension.java
+++ b/subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/MockitoLightweightExtension.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2024 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockito.junit.jupiter;
+
+import static org.junit.jupiter.api.extension.ExtensionContext.Namespace.create;
+
+import java.lang.reflect.Field;
+import java.util.HashSet;
+import java.util.Set;
+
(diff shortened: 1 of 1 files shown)


TimvdLippe:
Can you please file an issue first where we can discuss your use case. Then we can discuss a potential implementation after.

topizio:
Hi @TimvdLippe, many thanks for your feedback. Isn't https://github.com/mockito/mockito/issues/1509 enough? If not, I'll create another one without any problem and update the PR following this new key

TimvdLippe:
Please file a new issue. That ticket is about the runner, not the JUnit 5 extension.

topizio:
@TimvdLippe , I see your point now, thanks. New issue on its way

topizio:
Replaced by https://github.com/mockito/mockito/pull/3232

PR #3257, opened 2024-02-01
Fixes #3229: Resolve test generic arguments

### LINKED ISSUE #3229: `@Captor` test parameters don't work with primitive type arguments
That's the first time I'm actually contributing to any public project.
I apologize if I'm missing important informations about this issue.

### Description of the error :

I'm using java 21 and the dependency org.mockito:mockito-junit-jupiter version 5.8.0
The issue occurs when using an `ArgumentCaptor` as a test method parameter annotated with `@Captor`, if it is used to capture the argument of a method that takes a primitive type, it throws a `NullPointerException` trying to unbox a `null` value to the primitive type.
This only occurs when using `@Captor` on a parameter. It works fine on a field or as a local `ArgumentCaptor`.
The `@Captor` parameters also work as expected with non-primitive types (Well kinda, I'll go back to this later)

### Reproduction :

Here's the sample code to reproduce the error, followed by the corresponding stacktrace (I had to order the test because the occurence of the error causes all other tests to fail somehow) :

``` java annotate
@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
public class ReproductionTest {
    @Mock
    private Foo foo;

    @Captor
    private ArgumentCaptor<Integer> annotatedFieldCaptor;

    @Test
    @Order(1)
    void this_works_fine() {
        ArgumentCaptor<Integer> localCaptor = ArgumentCaptor.forClass(Integer.class);
        testCaptor(localCaptor);
    }

    @Test
    @Order(2)
    void this_works_too() {
        testCaptor(annotatedFieldCaptor);
    }

    @Test
    @Order(3)
    void this_throws_NullPointerException(@Captor ArgumentCaptor<Integer> annotatedParameterCaptor) {
        testCaptor(annotatedParameterCaptor);
    }

    private void testCaptor(ArgumentCaptor<Integer> captor) {
        doNothing().when(foo).doSomething(captor.capture());
        foo.doSomething(1);
        assertEquals(1, captor.getValue());
    }

    static class Foo {
        void doSomething(int value) {
        }
    }
}
```

Unsurprisingly, the first two tests pass, but the third fails with the following stacktrace :

> java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because the return value of "org.mockito.ArgumentCaptor.capture()" is null
> 
> 	at org.test.ReproductionTest.testCaptor(ReproductionTest.java:44)
> 	at org.test.ReproductionTest.this_throws_NullPointerException(ReproductionTest.java:40)
> 	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
> 	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
> 	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)

### Wait, there's more, and my proposition of solution

I investigated the issue using the debugger and noticed pretty quickly what the problem was. The "capture type" was always set to `Object`, no matter what the generic type inside the `ArgumentCaptor` was (even non primitive). It's not the case for local and field captors, which have the expected "capture type". Knowing that, since the `capture()` method return `defaultValue(clazz)`, it always return `null`. It still works somehow for non primitive method parameters, but `null` cannot be passed as a primitive method parameter, hence the `NullPointerException` about the unboxing. So I investigated a bit more in the `CaptorParameterResolver` class, and I think I know what's wrong. It uses the `CaptorAnnotationProcessor.process` method, which uses
`new GenericMaster().getGenericType(parameter)` to determine the generic type to capture.
However, this leads to the invocation of the method `GenericMaster.getGenericType(Parameter parameter)`.
This method uses `Parameter.getType()` to look for the type parameter, but that actually returns the raw class of the parameter, so the informations about type parameters are lost. In the method `GenericMaster.getGenericType(Field field)` used for fields, it uses `Field.getGenericType()` that actually return the type with type parameters. The equivalent of this method for Parameter is `Parameter.getParameterizedType()`.
I think this is clearly a bug, and the fix seems pretty obvious to me :

In GenericMaster.java, line 30, replace
```return getaClass(parameter.getType());```
by
```return getaClass(parameter.getParameterizedType());```

I tested it with a workaround and it works as intended (as a proof of concept only)
Furthermore, it seems like this method (and all the captor parameter resolving circuit) isn't used anywhere else, so there's very little impact to expect from this fix.

### PR DESCRIPTION
**Pull request by youssef3wi**
This pull request closes #3229 in order to resolve the test arguments. 

## Checklist

### DIFF
--- a/subprojects/junit-jupiter/src/test/java/org/mockitousage/annotation/CaptorAnnotationWithPrimitiveTest.java
+++ b/subprojects/junit-jupiter/src/test/java/org/mockitousage/annotation/CaptorAnnotationWithPrimitiveTest.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2024 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockitousage.annotation;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.doNothing;
+
+@ExtendWith(MockitoExtension.class)
+public class CaptorAnnotationWithPrimitiveTest {
+    @Mock private Foo foo;
+
+    static class Foo {
+        void doSomething(int value) {}
+    }
+
+    @Test
+    public void shouldCaptorPrimitive(@Captor ArgumentCaptor<Integer> captor) {
+        // Given
+        int value = 1;
+
+        // When
+        doNothing().when(foo).doSomething(captor.capture());
+
+        // Then
+        foo.doSomething(value);
+        assertEquals(1, captor.getValue());
+    }
+}
--- a/src/main/java/org/mockito/internal/util/reflection/GenericMaster.java
+++ b/src/main/java/org/mockito/internal/util/reflection/GenericMaster.java
@@ -27,7 +27,7 @@ public class GenericMaster {
      * @param parameter the parameter to inspect
      */
     public Class<?> getGenericType(Parameter parameter) {
-        return getaClass(parameter.getType());
+        return getaClass(parameter.getParameterizedType());
     }
 
     private Class<?> getaClass(Type generic) {
--- a/src/test/java/org/mockito/internal/util/reflection/GenericMasterTest.java
+++ b/src/test/java/org/mockito/internal/util/reflection/GenericMasterTest.java
@@ -7,6 +7,8 @@ package org.mockito.internal.util.reflection;
 import static org.junit.Assert.assertEquals;
 
 import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
 import java.lang.reflect.Type;
 import java.util.*;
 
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: subprojects/junit-jupiter/src/test/java/org/mockitousage/annotation/CaptorAnnotationWithPrimitiveTest.java [hunk 1]
```diff
@@ -0,0 +1,52 @@
...
+    private AutoCloseable closeable;
+
+    static class Foo {
+        void doSomething(int value) {
+        }
+    }
+
+    @BeforeEach
+    void setUp() {
+        closeable = MockitoAnnotations.openMocks(this);
```
TimvdLippe, line 30: Nit: the extension should already do this, doesn't it? To ensure we are purely testing the extension, I think we should remove this.

youssef3wi: You can run/debug the `CaptorAnnotationWithPrimitiveTest` both with and without the changes to see the difference.
Additionally, there is an issue that describes the problem.

youssef3wi: The problem does not lie in the `openMocks` elsewhere; we will encounter issues with `doNothing().when` before the captor problem.

TimvdLippe: If I understand the user report correctly, I think this reproduces the issue as well:

```java
public class CaptorAnnotationWithPrimitiveTest {
    @Mock
    private Foo foo;

    static class Foo {
        void doSomething(int value) {
        }
    }

    @Test
    public void shouldCaptorPrimitive(@Captor ArgumentCaptor<Integer> captor) {
        // Given
        int value = 1;

        // When
        doNothing().when(foo).doSomething(captor.capture());

        // Then
        foo.doSomething(value);
        assertEquals(1, captor.getValue());
    }
}
```

Can you update your unit test and confirm that it fails before and succeeds after? If so, please update the regression test to this instead.

youssef3wi: without_openMocks

youssef3wi: Using the fix
with_fix

TimvdLippe: Thank you!

PR #3319, opened 2024-04-19
Mark reified varargs parameter @Deprecated

bjmi:
see title

### DIFF
--- a/src/main/java/org/mockito/ArgumentCaptor.java
+++ b/src/main/java/org/mockito/ArgumentCaptor.java
@@ -194,7 +194,7 @@ public class ArgumentCaptor<T> {
      */
     @SafeVarargs
     @SuppressWarnings({"varargs", "unchecked"})
-    public static <U> ArgumentCaptor<U> captor(U... reified) {
+    public static <U> ArgumentCaptor<U> captor(@Deprecated U... reified) {
         if (reified == null || reified.length > 0) {
             throw new IllegalArgumentException("Do not provide any arguments to the 'captor' call");
         }
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -1966,7 +1966,8 @@ public class Mockito extends ArgumentMatchers {
      * @since 5.1.0
      */
     @SafeVarargs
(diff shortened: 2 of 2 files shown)


TimvdLippe:
How does this show up in your Idea? I understand we rely on the magic of the compiler to determine the type of the generic argument, but I didn't know you could annotate individual arguments as deprecated. Can you share a screenshot of a usage to show the UX?

bjmi:
Currently IntelliJ IDEA and Eclipse IDE simply ignore the @Deprecated annotation when used on parameters.

But the intention to keep end users from passing a parameter value still applies.
> A program element annotated `@Deprecated` is one that programmers are discouraged from using, typically because it is dangerous, or because a better alternative exists. Compilers warn when a deprecated program element is used or overridden in non-deprecated code. - [Javadoc](https://docs.oracle.com/javase/8/docs/api/java/lang/Deprecated.html)

More information in this [SO answer](https://stackoverflow.com/a/14627381).

What do you think? :)

TimvdLippe:
Since this annotation will not show up in either tooling or the compiler itself, I don't see a lot of use for adding it. When the compiler does emit a warning when a parameter is passed into such a method, I do see the value in adding it. For now, I am inclined to not merge this change. WDYT?

bjmi:
> Since this annotation will not show up in either tooling or the compiler itself, I don't see a lot of use for adding it. When the compiler does emit a warning when a parameter is passed into such a method, I do see the value in adding it. For now, I am inclined to not merge this change. WDYT?

Could you please wait for the outcome of [IDEA-352253](https://youtrack.jetbrains.com/issue/IDEA-352253/Inspection-highlight-use-of-deprecated-parameter)? If it is implemented then the use of avoiding user errors is increased. Otherwise this PR has no real added value.

bjmi:
The proposal in [IDEA-352253](https://youtrack.jetbrains.com/issue/IDEA-352253/Inspection-highlight-use-of-deprecated-parameter) is almost rejected.
This pull request can therefore also be closed.
Thank you for waiting.

TimvdLippe:
Thanks for the update! We can reconsider when IDE's do support this.

PR #3335, opened 2024-05-03
Fixes #3331 : Fix `AdditionalMatchers.and()` and `AdditionalMatchers.or()` not to swap the order of matchers

### LINKED ISSUE #3331: AdditionalMatchers.and() and or() swap matcher order
Consider the following code:

```java
public class MockitoTest {
  @Mock MyService myService;

  @Test
  public void test() {
    Mockito.when(
            myService.doSomething(
                and(any(MyPojo.class),
                    argThat(p -> p.id == 1))))
        .thenReturn(1);

    myService.doSomething("hello");
  }

  interface MyService {
    int doSomething(Object pojo);
  }

  static class MyPojo {
    int id = 1;
  }
}

```

This test throws the following exception:
```
java.lang.ClassCastException: class java.lang.String cannot be cast to class MockitoTest$MyPojo (java.lang.String is in module java.base of loader 'bootstrap'; MockitoTest$MyPojo is in unnamed module of loader 'app')
	at org.mockito.internal.matchers.And.matches(And.java:23)
	at org.mockito.internal.invocation.TypeSafeMatching.apply(TypeSafeMatching.java:35)
	at org.mockito.internal.invocation.MatcherApplicationStrategy.argsMatch(MatcherApplicationStrategy.java:85)
	at org.mockito.internal.invocation.MatcherApplicationStrategy.forEachMatcherAndArgument(MatcherApplicationStrategy.java:71)
	at org.mockito.internal.invocation.InvocationMatcher.argumentsMatch(InvocationMatcher.java:156)
	at org.mockito.internal.invocation.InvocationMatcher.matches(InvocationMatcher.java:82)
	at org.mockito.internal.stubbing.InvocationContainerImpl.findAnswerFor(InvocationContainerImpl.java:92)
	at org.mockito.internal.handler.MockHandlerImpl.handle(MockHandlerImpl.java:91)
	at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.doIntercept(MockMethodInterceptor.java:56)
	at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor$DispatcherDefaultingToRealMethod.interceptAbstract(MockMethodInterceptor.java:161)
	at MockitoTest$MyService$MockitoMock$v3DeXXsQ.doSomething(Unknown Source)
	at MockitoTest.test(MockitoTest.java:23)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
```

Debugger inspection of [`And.matches()`](https://github.com/mockito/mockito/blob/fb9ff6d40442ea98c8c3f8bc9fffbe253bd8df7b/src/main/java/org/mockito/internal/matchers/And.java#L22) shows that `m1` is the lambda and `m2` is the `InstanceOf`, as opposed to what was passed to `and()`. I guess this comes from [`ArgumentMatcherStorageImpl#reportAnd()` and `reportOr()`](https://github.com/mockito/mockito/blob/fb9ff6d40442ea98c8c3f8bc9fffbe253bd8df7b/src/main/java/org/mockito/internal/progress/ArgumentMatcherStorageImpl.java#L42-L59) which pop `m1` before `m2`, swapping the order.

This is with Mockito 5.11.0, Junit 5.10.2 and Java 11.

check that

    _No, it does not provide useful information (here the exception is similar to the one in #2268 but it could be something else like an NPE, depending on the use case)_

dev-jonghoonpark:
related issue: #3331

### Motivation:
- On #3331, we found that AdditionalMatchers.and() and or() swap matcher order

### Modification:
- Changed the assignment order of `popMatcher()` results.

### Result:
- Now, AdditionalMatchers.and() and or() not swap matcher order
- Close #3331

## Checklist

### DIFF
--- a/src/main/java/org/mockito/internal/progress/ArgumentMatcherStorageImpl.java
+++ b/src/main/java/org/mockito/internal/progress/ArgumentMatcherStorageImpl.java
@@ -42,8 +42,8 @@ public class ArgumentMatcherStorageImpl implements ArgumentMatcherStorage {
     public void reportAnd() {
         assertStateFor("And(?)", TWO_SUB_MATCHERS);
 
-        ArgumentMatcher<?> m1 = popMatcher();
         ArgumentMatcher<?> m2 = popMatcher();
+        ArgumentMatcher<?> m1 = popMatcher();
 
         reportMatcher(new And(m1, m2));
     }
@@ -52,8 +52,8 @@ public class ArgumentMatcherStorageImpl implements ArgumentMatcherStorage {
     public void reportOr() {
         assertStateFor("Or(?)", TWO_SUB_MATCHERS);
 
-        ArgumentMatcher<?> m1 = popMatcher();
         ArgumentMatcher<?> m2 = popMatcher();
+        ArgumentMatcher<?> m1 = popMatcher();
 
         reportMatcher(new Or(m1, m2));
     }
--- a/src/test/java/org/mockitousage/matchers/AdditionalMatcherTest.java
+++ b/src/test/java/org/mockitousage/matchers/AdditionalMatcherTest.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2024 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockitousage.matchers;
+
+import org.junit.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.mockito.AdditionalMatchers.and;
+import static org.mockito.AdditionalMatchers.or;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.argThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class AdditionalMatcherTest {
+
+    interface Foo {
+        int doSomeThing(Object param);
+    }
+
+    @Test
+    public void shouldMatchArgumentsSequentiallyWithAnd() {
+        Foo foo = mock(Foo.class);
+
+        when(foo.doSomeThing(and(any(String.class), argThat(String::isEmpty)))).thenReturn(1);
+
+        assertNotEquals(foo.doSomeThing(1), 1);
+    }
+
+    @Test
+    public void shouldMatchArgumentsSequentiallyWithOr() {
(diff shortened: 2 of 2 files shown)


src/test/java/org/mockito/internal/progress/ArgumentMatcherStorageImplTest.java
```diff
@@ -0,0 +1,46 @@
...
+package org.mockito.internal.progress;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.internal.matchers.InstanceOf;
+import org.mockitoutil.TestBase;
+
+import static org.junit.Assert.assertEquals;
+
+public class ArgumentMatcherStorageImplTest extends TestBase {
```
TimvdLippe, line 14: Instead of a test that tests the internals, please create a regression test in `src/test/java/org/mockitousage` where you copy the example in the issue description of #3331 That way, we test against the user code, rather than how Mockito itself is built.

TimvdLippe (changes requested):

src/test/java/org/mockito/internal/progress/ArgumentMatcherStorageImplTest.java
dev-jonghoonpark: @TimvdLippe 

Thank you for your reply.

Could you please check if I understand correctly?

- Move the test under `src/test/java/org/mockitousage`.
- Modify it so that it does not test against the code inside mockito, but against the user's code (black box testing).

dev-jonghoonpark:
@TimvdLippe 

I midified the code based on your feedback, can you check it?

The test code works like this

## and
- If the test is sequential, it will be filtered out by`any(String.class)`, so `doSomeThing()` will not return 1.
- If it's not sequential, the test will fail with an error in `argThat(String::isEmpty)`. (because the input is not a String)

## or
- If the test is sequential, `any(Integer.class)` will pass, so `doSomeThing()` will return 1.
- If it is not sequential, the test will fail with an error in `argThat(String::isEmpty)`. (because the input is not a String)

TimvdLippe (approved):
This looks great, thank you so much!

dev-jonghoonpark:
I'll check why the test is failing and fix it. ; )

TimvdLippe:
You can run `./gradlew spotlessApply` to automatically fix it for you.

PR #3349, opened 2024-05-19
Fixes #3237: Fix NullPointerException in Only.verify

### LINKED ISSUE #3237: Method `Only.verify` throws `NullPointerException`
Method `Only.verify` throws `NullPointerException` with the latest `mockito-core` 5.9.0

Here is a sample code in Kotlin which triggers this issue.
```kotlin
verify(service, only()).doSomething()
```

It triggers the following `NullPointerException` to be thrown.
```
java.lang.NullPointerException: Cannot invoke "org.mockito.invocation.Invocation.getMock()" because "undesired" is null
	at org.mockito.internal.exceptions.Reporter.noMoreInteractionsWanted(Reporter.java:555)
	at org.mockito.internal.verification.Only.verify(Only.java:30)
	at org.mockito.internal.verification.MockAwareVerificationMode.verify(MockAwareVerificationMode.java:30)
	at org.mockito.internal.handler.MockHandlerImpl.handle(MockHandlerImpl.java:75)
	at org.mockito.internal.creation.bytebuddy.MockMethodAdvice.handle(MockMethodAdvice.java:134)
```

A quick analysis shows that it is caused by method `Only.verify` which searches for an optional unverified invocation by execution of method `InvocationsFinder.findFirstUnverified`, but then fails to check for the result for `null` and just calls method `Reporter.noMoreInteractionsWanted` which throws the exception because the invocation was not found.
https://github.com/mockito/mockito/blob/main/src/main/java/org/mockito/internal/verification/Only.java#L29C37-L30
```java
        MatchableInvocation target = data.getTarget();
        List<Invocation> invocations = data.getAllInvocations();
        List<Invocation> chunk = findInvocations(invocations, target);
        if (invocations.size() != 1 && !chunk.isEmpty()) {
            Invocation unverified = findFirstUnverified(invocations);
            throw noMoreInteractionsWanted(unverified, (List) invocations);
        }
```

Note that there is method `NoMoreInteractions.verify` which also makes a call to the same two methods `InvocationsFinder.findFirstUnverified` and `Reporter.noMoreInteractionsWanted` but it does so properly, thus the exception can not be thrown there.
https://github.com/mockito/mockito/blob/main/src/main/java/org/mockito/internal/verification/NoMoreInteractions.java#L25-L28
```java
        Invocation unverified = findFirstUnverified(data.getAllInvocations());
        if (unverified != null) {
            throw noMoreInteractionsWanted(unverified, (List) data.getAllInvocations());
        }
```

So I suggest that the same check for null would be made in method `Only.verify` to avoid the exception.
And I also suggest that static analysis based on annotations would be introduced, then it would be obvious that method `findFirstUnverified` returns a nullable value while method `noMoreInteractionsWanted` does not accept a nullable argument.

baezzys:
## Motivation
In the latest version of mockito-core (5.9.0), the `Only.verify` method throws a `NullPointerException` when it attempts to call `Reporter.noMoreInteractionsWanted` with a null argument. This occurs because the method `InvocationsFinder.findFirstUnverified` can return null, but this is not handled properly in `Only.verify`.

This fixes #3237 

## Changes
- Added a null check for the result of `InvocationsFinder.findFirstUnverified` in `Only.verify`. If the result is null, `Reporter.noMoreInteractionsWanted` is not called.

## Checklist

 - [v] Read the [contributing guide](https://github.com/mockito/mockito/blob/main/.github/CONTRIBUTING.md)
 - [v] PR should be motivated, i.e. what does it fix, why, and if relevant how
 - [v] If possible / relevant include an example in the description, that could help all readers
 - [v] Avoid other runtime dependencies
 - [v] Meaningful commit history ; intention is important please rebase your commit history so that each
 - [v] The pull request follows coding style
 - [v] Mention `Fixes #<issue number>` in the description _if relevant_
 - [v] At least one commit should mention `Fixes #<issue number>` _if relevant_

### DIFF
--- a/src/main/java/org/mockito/internal/verification/Only.java
+++ b/src/main/java/org/mockito/internal/verification/Only.java
@@ -30,7 +30,9 @@ public class Only implements VerificationMode {
         }
         if (invocations.size() != 1 && !chunk.isEmpty()) {
             Invocation unverified = findFirstUnverified(invocations);
-            throw noMoreInteractionsWanted(unverified, (List) invocations);
+            if (unverified != null) {
+                throw noMoreInteractionsWanted(unverified, (List) invocations);
+            }
         }
         if (invocations.size() != 1 || chunk.isEmpty()) {
             throw wantedButNotInvoked(target);
--- a/src/test/java/org/mockito/internal/verification/OnlyTest.java
+++ b/src/test/java/org/mockito/internal/verification/OnlyTest.java
@@ -103,4 +103,34 @@ public class OnlyTest extends TestBase {
         assertTrue(wanted.isVerified());
         assertFalse(different.isVerified());
     }
+
+    @Test
+    public void shouldMarkMultipleInvocationAsVerified() {
+
+        // given
+        InvocationBuilder invocationBuilder = new InvocationBuilder();
+        Invocation wanted = invocationBuilder.mock(mock).simpleMethod().toInvocation();
+        Invocation different = invocationBuilder.mock(mock).differentMethod().toInvocation();
+
+        // when
+        try {
+            only.verify(
+                    new VerificationDataStub(
+                            invocationBuilder.toInvocationMatcher(), wanted, different));
+            fail();
+        } catch (NoInteractionsWanted e) {
+        }
+
+        try {
+            only.verify(
+                    new VerificationDataStub(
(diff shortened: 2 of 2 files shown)


TimvdLippe (approved):
Thanks for the fix and regression test!

PR #3381, opened 2024-06-24
feat: add jpms module-info

### LINKED ISSUE #2604: Consider adding `module-info.class`
Would it be possible to bundle a Java-9-compiled `module-info.class` in with mockito-inline using the multi-version JAR feature (i.e. placing the class in `META-INF/versions/9/module-info.class`).

Although Mockito already distributes as an automatic module, the requirements on bytebuddy and the bytebuddy agent do not get resolved automatically, which results in a somewhat cryptic error the first time you try to use mocks.

I believe that the following would resolve this issue:

```java
module org.mockito {
  requires transitive net.bytebuddy;
  requires transitive net.bytebuddy.agent;

  exports org.mockito;
  exports org.mockito.{other packages here};
}
```

This would reduce the confusion around JPMS for new users, and remove the need to directly depend on implementation detail.

### LINKED ISSUE #2952: Issue with mockito with jdk 19
- [ ] The mockito message in the stacktrace have useful information, but it didn't help:
```
	at ds.tests.OrdArrayTest$ToStringTests.testDisplay(OrdArrayTest.java:622)
Caused by: java.lang.reflect.InvocationTargetException
Caused by: org.mockito.exceptions.base.MockitoInitializationException: 

Could not initialize inline Byte Buddy mock maker.

It appears as if your JDK does not supply a working agent attachment mechanism.
Java               : 19
JVM vendor name    : Oracle Corporation
JVM vendor version : 19.0.2+7
JVM version        : 19.0.2+7
JVM info           : mixed mode
OS name            : Linux
OS version         : 5.4.147-qgki-g65a062a09cb7

Caused by: java.lang.IllegalStateException: Could not self-attach to current VM using external process

[ERROR] Errors: 
[ERROR]   HighArrayTest$ToStringTests.testDisplay:524 » IllegalState Could not initialize plugin: interface org.mockito.plugins.MockMaker (alternate: null)
[ERROR]   OrdArrayTest$ToStringTests.testDisplay:622 » IllegalState Could not initialize plugin: interface org.mockito.plugins.MockMaker (alternate: null)
```

https://github.com/Fernal73/DSAlgos

Worked earlier with Java 11.

xenoterracide:
this is a WIP!

Fixes #2604
Fixes #2952 

## Checklist


TimvdLippe:
For the failures in Javadoc, all `&` and `<` can be replaced with their unicode escape character. E.g. `&amp;` and `&lt;`. For the references to com.sun packages, we can replace them with `@see` I think. Otherwise removing the `@link` and use `<code>` is fine as well.

I don't think this requires updating our Java version, so let's see if we can fix these issues this way. Thanks for the PR and picking this up!

xenoterracide:
Do you guys still want to export `internal` by default? feels unnecessary to me since no one will automagically use jpms unless they have their own module-info, and anyone doing that will probably have been careful about not accessing internal.

Also, do you want to take advantage of github.com/gradlex-org/java-module-testing ? or how do you want to deal with blackbox testing? there probably should be some kind of blackbox test suite that ensures that the appropriate public apis are accessible... although that might be a bigger effort.

xenoterracide:
hmm...

```
org.mockitointegration.NoByteCodeDependenciesTest > pure_mockito_should_not_depend_bytecode_libraries FAILED
    java.lang.AssertionError: 'module-info' has some dependency to ByteBuddy or Objenesis
        at org.mockitointegration.ClassLoadabilityChecker.checkLoadability(ClassLoadabilityChecker.java:41)
        at org.mockitointegration.NoByteCodeDependenciesTest.pure_mockito_should_not_depend_bytecode_libraries(NoByteCodeDependenciesTest.java:50)
```

static meaning optional but is required for mockito to have when it compiles
```
    requires org.opentest4j;
    requires static net.bytebuddy;
    requires static net.bytebuddy.agent;
    requires static junit;
```

xenoterracide:
So, I tried to look at adding a `module-info` to a few of the test suites/projects (might want to look at https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html btw), but all of them seem to depend on the `internal` package. Could be a kick the bucket situation. I made a feature request for test fixtures that support whiteboxing so the suite can be blackbox.

xenoterracide:
@TimvdLippe I think this is ready for a preliminary review. I don't really trust this enough to merge yet. I need some feedback on how to approach testing. Currently I see 2 options

1. Make a lot of internal package classes/methods public to turn the `test` test suite into a blackbox test with its own module-info
2. create a new subprobject or jvm test suite that does this, but that means writing completely new tests, possibly needing to do this multiple times to cover old junit and anything else we're concerned about.

third option but I'm stuck on at this hour is to turn test into a whitebox test. It's currently wanting me to export the test module to junit which I don't seem to be able to do, I created an issue with the plugin author (who probably loves me right now since this is the fourth one ;) ) https://github.com/xenoterracide/mockito/pull/1 

There are surely N other options

TimvdLippe:
@xenoterracide Thanks for the work! We do not want to export the `org.mockito.internal` packages, so our tests should be able to run without that there. I am not sure if I understand why our tests start to fail with the module? As I understand the feature, it requires opt-in to enforce module boundaries, is that not the case?

In case that we are hitting these issues, does that mean that users that rely on `org.mockito.internal` and don't have any specific configuration would start to see their builds failing? In that case, we need to make this a major version bump (maybe we should do that in any case), since that would otherwise lead to unexpected build failures.

With regards to testing, yes ideally we have a separate Gradle project that we use to test these things. For example, I would love it to be possible to test out that no user code can import `org.mockito.internal` on an opt-in basis. That would essentially allow us to delete https://github.com/mockito/mockito/blob/main/subprojects/errorprone/src/main/java/org/mockito/errorprone/bugpatterns/MockitoInternalUsage.java

xenoterracide:
>  I am not sure if I understand why our tests start to fail with the module? As I understand the feature, it requires opt-in to enforce module boundaries, is that not the case?

It does, they fail if you add a module-info to the test suite, which turns it into a module, and all of the tests (I've seen) reference an internal package.

> In case that we are hitting these issues, does that mean that users that rely on org.mockito.internal and don't have any specific configuration would start to see their builds failing?

no, users should not need to rely on it.

> With regards to testing, yes ideally we have a separate Gradle project that we use to test these things.

I can create a module, or perhaps two. based on this it looks like we need to test junit 4 and jupiter, so probably a separate for each.

>  I would love it to be possible to test out that no user code can import org.mockito.internal on an opt-in basis.

I'm certain it can be done. I don't know how to do it ;) . I think this might lead me to make a request in assertj since I can test if something is public/private, maybe they could add is class visible. Honestly seems handy. The current "java 9 module test" project uses the `internal` package and uses the classpath so wouldn't enforce the boundary.

subprojects/junitJupiterExtensionTest/src/test/java/module-info.java
```diff
@@ -0,0 +1,6 @@
+module mockito.junitJupiterExtensionTest.test {
```
TimvdLippe, line 1: Do we also need to specify a `module-info.java` for the main sourceset of this extension? E.g. to denote to users that we require these modules.

TimvdLippe (commented):
> It does, they fail if you add a module-info to the test suite, which turns it into a module, and all of the tests (I've seen) reference an internal package.

Cool, in that case let's not add it and remove `org.mockito.internal` from the file (I see you already did that)

subprojects/junitJupiterExtensionTest/src/test/java/module-info.java
xenoterracide: I'm not certain I understand this question. there is no `main` for this test sub project.

xenoterracide: oof, I think I realized what you're talking about, and that poses a problem since the extension is an external jar and accesses the internal API. I don't know if there's a way to avoid exporting at least a few of the internal packages for that.

Otherwise these look more reasonable to test this works in the real world.

xenoterracide: choices, first I want to get it working before making decisions ;) I'm having trouble making it work with some internals stuff, need to figure out the why's.

I started work on this because spring boot blew up on me just for having mockito on my classpath. There was an easy workaround there, but since it looked like PRs were welcome I decided to take a stab ;).

note: might not have to export all of org.mockito.internal, personally I'd rather ship with a subset of that exposed documenting it'll be "hidden" in the next major version.

xenoterracide: remember a java module is 1 to 1 with a jar, so if an external jar (mockitos or not) depends on any internal package that specific package will have to be exported or perhaps opened. You can export/open to a specific thing though, so perhaps that's enough. Currently though, exporting/opening this would block

```
	at org.mockito/org.mockito.internal.configuration.plugins.PluginLoader$1.invoke(PluginLoader.java:85)
	at jdk.proxy3/jdk.proxy3.$Proxy19.isTypeMockable(Unknown Source)
	at org.mockito/org.mockito.internal.util.MockUtil.typeMockabilityOf(MockUtil.java:78)
	at org.mockito/org.mockito.internal.util.MockCreationValidator.validateType(MockCreationValidator.java:22)
	at org.mockito/org.mockito.internal.creation.MockSettingsImpl.validatedSettings(MockSettingsImpl.java:275)
	at org.mockito/org.mockito.internal.creation.MockSettingsImpl.build(MockSettingsImpl.java:236)
	at org.mockito/org.mockito.internal.MockitoCore.mock(MockitoCore.java:82)
	at org.mockito/org.mockito.Mockito.mock(Mockito.java:2104)
	at org.mockito/org.mockito.internal.configuration.MockAnnotationProcessor.processAnnotationForMock(MockAnnotationProcessor.java:79)
	at org.mockito/org.mockito.internal.configuration.MockAnnotationProcessor.process(MockAnnotationProcessor.java:28)
	at org.mockito/org.mockito.internal.configuration.MockAnnotationProcessor.process(MockAnnotationProcessor.java:25)
	at org.mockito/org.mockito.internal.configuration.IndependentAnnotationEngine.createMockFor(IndependentAnnotationEngine.java:44)
	at org.mockito/org.mockito.internal.configuration.IndependentAnnotationEngine.process(IndependentAnnotationEngine.java:72)
	at org.mockito/org.mockito.internal.configuration.InjectingAnnotationEngine.processIndependentAnnotations(InjectingAnnotationEngine.java:62)
	at org.mockito/org.mockito.internal.configuration.InjectingAnnotationEngine.process(InjectingAnnotationEngine.java:47)
	at org.mockito/org.mockito.MockitoAnnotations.openMocks(MockitoAnnotations.java:81)
	at org.mockito/org.mockito.internal.framework.DefaultMockitoSession.<init>(DefaultMockitoSession.java:43)
	at org.mockito/org.mockito.internal.session.DefaultMockitoSessionBuilder.startMocking(DefaultMockitoSessionBuilder.java:83)
	at org.mockito.junit.jupiter/org.mockito.junit.jupiter.MockitoExtension.beforeEach(MockitoExtension.java:160)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	Suppressed: java.lang.NullPointerException: Cannot invoke "java.util.Set.forEach(java.util.function.Consumer)" because the return value of "org.junit.jupiter.api.extension.ExtensionContext$Store.remove(Object, java.lang.Class)" is null
		at org.mockito.junit.jupiter/org.mockito.junit.jupiter.MockitoExtension.afterEach(MockitoExtension.java:194)
		... 2 more
	at org.mockito/org.mockito.internal.configuration.plugins.DefaultMockitoPlugins.create(DefaultMockitoPlugins.java:105)
	at org.mockito/org.mockito.internal.configuration.plugins.DefaultMockitoPlugins.getDefaultPlugin(DefaultMockitoPlugins.java:79)
	at org.mockito/org.mockito.internal.configuration.plugins.PluginLoader.loadPlugin(PluginLoader.java:75)
	at org.mockito/org.mockito.internal.configuration.plugins.PluginLoader.loadPlugin(PluginLoader.java:49)
	at org.mockito/org.mockito.internal.configuration.plugins.PluginRegistry.<init>(PluginRegistry.java:29)
	at org.mockito/org.mockito.internal.configuration.plugins.Plugins.<clinit>(Plugins.java:26)
	at org.mockito/org.mockito.internal.MockitoCore.<clinit>(MockitoCore.java:71)
	at org.mockito/org.mockito.Mockito.<clinit>(Mockito.java:1683)
	at org.mockito.junit.jupiter/org.mockito.junit.jupiter.MockitoExtension.beforeEach(MockitoExtension.java:156)
	... 2 more
```

xenoterracide: and here I think is a blocking issue. It looks like the code is trying to load `org.junit.jupiter.api.extension.ExtensionContext` reflectively, but junit isn't opening it.

```
module org.junit.jupiter.api {
	requires static transitive org.apiguardian.api;
	requires transitive org.junit.platform.commons;
	requires transitive org.opentest4j;

	exports org.junit.jupiter.api;
	exports org.junit.jupiter.api.condition;
	exports org.junit.jupiter.api.extension;
	exports org.junit.jupiter.api.function;
	exports org.junit.jupiter.api.io;
	exports org.junit.jupiter.api.parallel;

	opens org.junit.jupiter.api.condition to org.junit.platform.commons;
}
```

xenoterracide: line 180 of `InlineDelegateByteBuddyMockMaker` is throwing `java.lang.ClassNotFoundException: org.mockito.internal.creation.bytebuddy.inject.MockMethodDispatcher` which is currently making no sense to me... will keep digging tomorrow. until I've completely burned out or have something that *could work*.

xenoterracide: I'm stumped, I'm certain the problem is that `InlineDelegateByteBuddyMockMaker` isn't being loaded because I checked out what was happening without it. Honestly to get full module-info working I need help someone who understands how that static block in `InlineDelegateByteBuddyMockMaker` works. I tried exporting and opening everything to see if that'd fix it. Easier to play clean up on internals once tests pass.

TimvdLippe:
All right, thanks for all the effort! Let's try to keep things manageable and figure out how we can land these changes. I propose we do the following:

1. Make a separate PR with all of your JavaDoc fixes (don't remove the workaround for `org/mockito/internal` there)
2. We update this PR to only have `src/main/java/module-info.java`

As I understand it, that should work as-is, wouldn't it? If it does, cool let's do that.

If however that still breaks the integration with `InlineDelegateByteBuddyMockMaker`, then @raphw can help out on how to design the loading of it compatible with modules. Do keep in mind that we are generating a file on the fly to load the mockmaker: `org/mockito/internal/creation/bytebuddy/inject/MockMethodDispatcher.class`. This is bundled in our `.jar` on start: https://github.com/mockito/mockito/blob/01da000829488d2ae40875f8ecedef6ac587690a/src/main/java/org/mockito/internal/creation/bytebuddy/InlineDelegateByteBuddyMockMaker.java#L176-L178

xenoterracide:
> As I understand it, that should work as-is, wouldn't it? If it does, cool let's do that.

I don't think so... javadoc PR yes, that'll work fine with or without. Unless you have a clearer idea of testing it, I don't *think* this would work in a modularized runtime at all. Generating that file is probably the problem, but I don't understand exactly how. I wouldn't rely on this being the *only* breakage either. I just know that loading byte buddy is what mine is doing with that extension locally. If there's another path then we need to test that, etc. I would say any general proxy/instrumentation path needs to be done, then the module-info can be reduced down to its minimum. It might have something to do with dynamic classloader registration being deprecated.

I know spring/boot has certain limitations, but I'm not certain if any of that'll apply here, although maybe it'll help https://github.com/spring-projects/spring-framework/issues/32671

raphw:
The static initializer block does two things:

- It attaches to the current VM such that an instance of `Instrumentation` is available.
- It adds a dispatcher to the (unnamed module of the) bootstrap class loader.

I think what is missing is that at the end of the block, it needs to detect if Mockito is loaded as a named module. If so, it needs to add a read edge to the unnamed module of the boot loader. Without that read edge, it cannot interact with the injected dispatcher.

Before, with an implicit module name, it would have that read edge automatically as all unnamed modules are read by default.

xenoterracide:
I rebased on main and flattened the previous commits into a single commit, although that commit message is a mess... I can clean that or anything else up now if needed.

subprojects/junitJupiterExtensionTest/src/test/java/module-info.java
xenoterracide: btw, I noticed that I accidentally re-ordered some imports (fixed) but I could create a checkstyle check for that in another PR-if y'all wanted.

TimvdLippe:
To be able to evaluate this PR, can we revert all changes except `src/main/java/module-info.java` and remove all references to `org.mockito.internal` from it? Then we can figure out how to implement that module read edge that Rafael mentioned and make sure it works as-is. We can later work on junit-jupiter, but for now I would like to take a minimal step first to ensure we get that right, before we roll it out to all other artifacts.

xenoterracide:
Sure, but everything's going to pass because the tests in the main module aren't going to be modularized they'll run on the class path. That's why I left the other things in there.

TimvdLippe:
Right. In that case, I propose we update `subprojects/module-test/` to add a `module-info.java` so that we can test it in that way.

xenoterracide:
you can't not open some internals because the source uses them everywhere across all subprojects. I added the module info there

https://github.com/mockito/mockito/blob/main/subprojects/module-test/src/test/java/org/mockito/moduletest/ModuleAccessTest.java#L11-L13

xenoterracide:
anyways, updated the PR, again I left the bulk of internals commented out so it'd be easy to uncomment for debugging. It'd be better if any/all fix patches were merged prior to merging the actual module-info anyways.

TimvdLippe:
@raphw Seeing the test failures on CI, I think we are going to need the following:

> If so, it needs to add a read edge to the unnamed module of the boot loader.

However, I lack the knowledge on how to do so. Can you help us out with adding that read edge?

raphw:
I will look into it!

raphw:
This should be straight-forward to add as:

```java
InlineBytecodeGenerator.class.getModule().addReads(MockMethodDispatcher.class.getModule());
```

in the last line of the constructor of `InlineBytecodeGenerator`.

TimvdLippe:
@xenoterracide I see you recently rebased this PR, thanks! Can you please apply the fix from Rafael on your PR to hopefully fix some of the CI failures?

xenoterracide:
no, that doesn't fix anything. Honestly there are 2 different issues. 

module-test looks like it's failing because there are 2 copies of bytebuddy module on the classpath.

The one that this was supposed to fix was jupiter usage, and it doesn't help there. I also tried moving it to the static block.

raphw:
What the code does is that it injects a class into the bootstrap class loader. This is needed to make it universally available for Mockito and for other classes that might not be able to even see Mockito. Right now it does however expect that Mockito - when being run as a module - can read this class. If Mockito is an automatic module, it can read all unnamed modules. When it is a named module, it can not read any unnamed modules. The injected class will end up on the unnamed module of the bootstrap class loader. This I think is the problem.

xenoterracide:
So this conversation might be of interest and I'm wondering if there's any relation in the problem
 https://github.com/spring-projects/spring-framework/issues/32671

scordio:
I am trying to move to a module-only build in AssertJ (assertj/assertj#3771) and would be very happy to see the change in this PR happen.

Let me know if there is anything I could do to help, e.g., testing ongoing changes.

TimvdLippe:
I am still eager for us to ship module information for Mockito. That said, the blockers as mentioned above with regards to the the generation of a file is crucial to resolve if we want to move forward. Unfortunately I lack the domain knowledge to give guidance on how we can implement the behavior without breaking module guarantees. If you do, please feel free to help us out here on how we can solve this.

From a maintainers perspective, these are the requirements:
1. All `module-info.java` files must exist for `mockito-core` as well as all artifacts in `mockito-extensions` (which basically only is `mockito-junit-jupiter` as it is the only one with a separate API)
2. We cannot have `org.mockito.internal` in any of these `module-info.java` files
3. We must be able to create the `org/mockito/internal/creation/bytebuddy/inject/MockMethodDispatcher.class` in a module-safe way

scordio:
So far, I managed to get Mockito working on the module path in assertj/assertj#3771 so there is _at least_ one setup working with the current artifacts, therefore it might be doable to replicate the same directly in Mockito.

Once I'm done over there, I'll have a deeper look at this MR and let you know if I have any proposals 🙂

raphw:
So, according to my own ideas, I think this should already do the trick: https://github.com/mockito/mockito/tree/module-prototype

I suggest that we focus on mockito-core for now, and modularize the subprojects by themselves.

I do however get hung up in the javadoc tool. I have no idea why this happens, and I cannot understand what Gradle is trying to run.

xenoterracide:
Thought we fixed the javadoc issue at one. Just don't convince yourself that simply because you have no errors if you don't test against a modularized system. You have to have tests running against the modules; otherwise you may convince yourself that the full module-info works.

raphw:
I created a branch and linked it.

I have a setup to test Byte Buddy with modules, so I wanted to make some general testing with a solution that is as simple as possible.

From that understanding I had hoped to give better advice here.

How did you solve the javadoc issue?

TimvdLippe:
Shall we take Rafaels branch and continue with that? The solution looks promising and I agree we should ship mockito-core module info first before we do anything else.

In terms of testing, I think adding the module info to `module-test` gradle project is sufficient to catch issues. We can also test it by removing an export and verify the tests fail. That way we know for sure they are all set correctly.

raphw:
I would like to void opening and exporting all packages. If this is needed, something is wrong I argue.

About the javadoc? What did we do to resolve this?

TimvdLippe:
Are there any public packages (e.g. not `org.mockito.internal`) that you would want to avoid to export? I would assume that if a package is documented and part of our public API, then we would need to export it as well?

raphw:
I'd argue that internal packages are documented as internal by their package name. If someone upgrade to using Mockito as a module, this would need refactoring anyways.

xenoterracide:
> I would like to void opening and exporting all packages. If this is needed, something is wrong I argue.

I did that because I was trying to debug the issue and ensure it wasn't caused by some lack of export/opens, I've seen some weird things. Like for spring I have to export my.module to spring... because of something they're doing with their own fork of cglib. The idea was to clean it up after it worked.

However, there are packages all over the place consuming internal packages. You'll find out just how much fun this is when you try to do testing without these exports. Which is why I tried to see if I could get a different module working since so I could deal with things that would be required to be publicly available.

> About the javadoc? What did we do to resolve this?

there was a merged PR, please see https://github.com/mockito/mockito/issues/2604 which is mentioned in the description of this PR. It has a reference to that PR.

raphw:
So, for some reason the javadoc task in Gradle does not include all source files, but this is required as the tool cross-references sources to parse the javadoc. Since some sources are skipped, the javadoc fails. This works however, if I run manually:

find /home/rafael/workspace/mockito/mockito-core/src/main/java -type f -name "*.java" | xargs javadoc --module-path '/home/rafael/workspace/mockito/mockito-core/build/classes/java/main:/home/rafael/.gradle/caches/modules-2/files-2.1/net.bytebuddy/byte-buddy/1.15.11/f61886478e0f9ee4c21d09574736f0ff45e0a46c/byte-buddy-1.15.11.jar:/home/rafael/.gradle/caches/modules-2/files-2.1/net.bytebuddy/byte-buddy-agent/1.15.11/a38b16385e867f59a641330f0362ebe742788ed8/byte-buddy-agent-1.15.11.jar:/home/rafael/.gradle/caches/modules-2/files-2.1/junit/junit/4.13.2/8ac9e16d933b6fb43bc7f576336b8f4d7eb5ba12/junit-4.13.2.jar:/home/rafael/.gradle/caches/modules-2/files-2.1/org.hamcrest/hamcrest-core/3.0/254f69f4b0ca22198acfc19fcdd5f96140431c3e/hamcrest-core-3.0.jar:/home/rafael/.gradle/caches/modules-2/files-2.1/org.opentest4j/opentest4j/1.3.0/152ea56b3a72f655d4fd677fc0ef2596c3dd5e6e/opentest4j-1.3.0.jar:/home/rafael/.gradle/caches/modules-2/files-2.1/org.objenesis/objenesis/3.3/1049c09f1de4331e8193e579448d0916d75b7631/objenesis-3.3.jar:/home/rafael/.gradle/caches/modules-2/files-2.1/org.hamcrest/hamcrest/3.0/8fd9b78a8e6a6510a078a9e30e9e86a6035cfaf7/hamcrest-3.0.jar' -classpath '/home/rafael/workspace/mockito/mockito-core/build/generated/resources/inline/java' -d '/home/rafael/workspace/mockito/mockito-core/build/docs/javadoc'

xenoterracide:
@raphw so I updated my branch and added your fix to it, this is from moduletest which obviously has to pass

```
Module net.bytebuddy reads another module named net.bytebuddy
java.lang.module.ResolutionException: Module net.bytebuddy reads another module named net.bytebuddy
	at java.base/java.lang.module.Resolver.resolveFail(Resolver.java:900)
	at java.base/java.lang.module.Resolver.checkExportSuppliers(Resolver.java:718)
	at java.base/java.lang.module.Resolver.finish(Resolver.java:380)
	at java.base/java.lang.module.Configuration.<init>(Configuration.java:139)
	at java.base/java.lang.module.Configuration.resolve(Configuration.java:422)
	at mockito.module.test.test/org.mockito.moduletest.ModuleUtil.layer(ModuleUtil.java:115)
	at mockito.module.test.test/org.mockito.moduletest.ModuleAccessTest.can_access_non_opened_module_with_module_member_accessor(ModuleAccessTest.java:32)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at junit@4.13.2/org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at junit@4.13.2/org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at junit@4.13.2/org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at junit@4.13.2/org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at junit@4.13.2/org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at junit@4.13.2/org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at junit@4.13.2/org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at junit@4.13.2/org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at junit@4.13.2/org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at junit@4.13.2/org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at junit@4.13.2/org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at junit@4.13.2/org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at junit@4.13.2/org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at junit@4.13.2/org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at junit@4.13.2/org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at junit@4.13.2/org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.runTestClass(JUnitTestClassExecutor.java:112)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:58)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:40)
	at org.gradle.api.internal.tasks.testing.junit.AbstractJUnitTestClassProcessor.processTestClass(AbstractJUnitTestClassProcessor.java:54)
	at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.processTestClass(SuiteTestClassProcessor.java:53)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.dispatch.ContextClassLoaderDispatch.dispatch(ContextClassLoaderDispatch.java:33)
	at org.gradle.internal.dispatch.ProxyDispatchAdapter$DispatchingInvocationHandler.invoke(ProxyDispatchAdapter.java:92)
	at jdk.proxy2/jdk.proxy2.$Proxy6.processTestClass(Unknown Source)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker$2.run(TestWorker.java:183)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.executeAndMaintainThreadName(TestWorker.java:132)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.execute(TestWorker.java:103)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.execute(TestWorker.java:63)
	at org.gradle.process.internal.worker.child.ActionExecutionWorker.execute(ActionExecutionWorker.java:56)
	at org.gradle.process.internal.worker.child.SystemApplicationClassLoaderWorker.call(SystemApplicationClassLoaderWorker.java:121)
	at org.gradle.process.internal.worker.child.SystemApplicationClassLoaderWorker.call(SystemApplicationClassLoaderWorker.java:71)
	at gradle.worker/worker.org.gradle.process.internal.worker.GradleWorkerMain.run(GradleWorkerMain.java:69)
	at gradle.worker/worker.org.gradle.process.internal.worker.GradleWorkerMain.main(GradleWorkerMain.java:74)
```

raphw:
This one now works for me: https://github.com/mockito/mockito/pull/3597

I had to stop excluding the internal package from javadoc generation. I wonder if we can filter the internal types from the result, but it does not seem like the tool offers support for this.

xenoterracide:
@raphw fix for javadoc now included, I just commented out the exclude. also, you can simply run `./gradlew test`, and `./gradlew check` before running `./gradlew build`, the formers don't call javadoc

xenoterracide:
I'm going to kill my fork unless anyone thinks any of the work here is useful to keep around

TimvdLippe:
Your work has been super helpful, thanks! I think we were able to incorporate all findings from this PR, so I don't see a need to keep your fork around. Unless you prefer to do so of course.

PR #3391, opened 2024-07-01
Gradle lazy configuration

### LINKED ISSUE #3384: gradle improvements commentary
this isn't a real issue, more of a commentary for educational purposes. if the team wants to make some (debatable?) improvements to the build.

1. Lazy configuration. `tasks.withType(Foo)` would be lazy if changed to `tasks.withType(Foo).configureEach`. This means a task shouldn't be configured unless it needs to be executed.
2. `libraries.depname` could be migrated to a [version catalog](https://docs.gradle.org/current/userguide/platforms.html) and even have the variable name be libraries with the same access keys. I think most people are using `libs.versions.toml` but there is a dsl version
3. switch many subproject testing to [jvm test suites](https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html). I think this will make it easier to understand for contributors what is a released jar vs what's simply a test suite. I also believe it will come with some performance improvements and code reduction due to not having to have another project. note: test is already a jvm test suite.

thachlp:
## Checklist

 - [x] Fixes #3384

### DIFF
--- a/gradle/test-launcher.gradle
+++ b/gradle/test-launcher.gradle
@@ -1,4 +1,4 @@
-tasks.withType(Test) {
+tasks.withType(Test).configureEach {
     javaLauncher = javaToolchains.launcherFor {
         languageVersion = providers
             .gradleProperty("mockito.test.java")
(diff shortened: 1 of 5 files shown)


TimvdLippe (approved):
Thank you!

PR #3230, opened 2024-01-09
Align Javadoc configuration to Java 21 standards

scordio:
> [!IMPORTANT]  
> Please, check it out locally, generate the Javadoc under Java 21, and check the result before merging this PR!

This aligns the Javadoc configuration to the Java 21 standards and also upgrades [`highlight.js`](https://highlightjs.org/) to version 11.9.0, using its cdnjs distribution.

A small preview of the new stylesheet is at https://github.com/mockito/mockito/issues/3212#issuecomment-1880255670.

* Relates to maxcellent/javadoc.io#159
* Depends on #3226
* Fixes #3212
* Fixes mockito/mockito.github.io#60

### Deliverables
- [x] Apply custom stylesheet to `mockito-core`
- [x] Apply custom stylesheet to `mockito-junit-jupiter`

PR #3432, opened 2024-08-25
Allow links to JUnit Jupiter Javadoc, add Gradle version catalog

scordio:
This solves the gap mentioned at https://github.com/mockito/mockito/pull/3230#issuecomment-1890393656.

`javadoc` can create external links to libraries documented with "mismatching" modularity, like for JUnit Jupiter, which offers modular documentation. However, a warning is generated when the mismatch happens, and this is a problem when using the `-Werror` option to treat warnings as errors.

To circumvent this problem without dropping the `-Werror` option, Java 18 added the

`--link-modularity-mismatch=(warn|info)`

command line option to specify how external documentation with wrong modularity should be reported.

See: https://bugs.openjdk.org/browse/JDK-8274639

In addition, this introduces a [`libs.versions.toml`](https://docs.gradle.org/current/userguide/platforms.html#sub:conventional-dependencies-toml) version catalog, suggested at #3384.


scordio:
> All the CI jobs with Java < 18 should skip the Javadoc generation. I'll check how this could be done.

This is resolved with a conditional configuration of the `javadoc` task.

gradle/mockito-junit-jupiter/javadoc.gradle
```diff
@@ -34,6 +34,12 @@ tasks.named("javadoc", Javadoc) {
     options.use = true
     options.windowTitle = "Mockito JUnit Jupiter ${project.version} API"
 
+    // https://bugs.openjdk.org/browse/JDK-8274639
+    if (JavaVersion.current() >= JavaVersion.VERSION_18) {
+        options.addStringOption('-link-modularity-mismatch', 'info')
+        options.links("https://junit.org/junit5/docs/5.11.0/api/")
```
scordio, line 40: Ideally, the JUnit version number shouldn't be hardcoded.

I see it appears here:

https://github.com/mockito/mockito/blob/f6450a01f27d0e4e7e95a52025b000726558774d/gradle/dependencies.gradle#L8

What would be the proper way to expose that value to this file?

TimvdLippe: If I understand correctly, we need to define the `versions` to be exposed on the project as well. For that, we can do the following:

```groovy
ext {
    libraries = [:]
    libraryVersions = [:]
}

libraryVersions.bytebuddy = '1.15.0'
libraryVersions.junitJupiter = '5.11.0'
libraryVersions.errorprone = '2.23.0'
```

And then here this should work:

```groovy
options.links("https://junit.org/junit5/docs/${libraryVersions.junitJupiter}/api/")
```

scordio: Solved with the introduction of [`libs.versions.toml`](https://docs.gradle.org/current/userguide/platforms.html#sub:conventional-dependencies-toml).

TimvdLippe (approved):
Nice fix, thanks for the contribution!

PR #3453, opened 2024-09-28
Move core sources to mockito-core module

### LINKED ISSUE #3444: Move root project to dedicated core folder
Currently the main Mockito source code lives in the root folder. This is not ideal:

1. That means that we have next to each other `src/main/java` and `subprojects`, this is strange and almost confusing for any new contributors. 
2. This doesn't align with what other notable projects (spring, micronaut, micrometer, etc.) do, i.e. they have a root folder and their various components are located in sub-modules of the root folder.
3. The root Gradle build script has to mix configurations for both the root project, `mockito-core` and the subprojects

We could imagine something like this :

```
├── ./mockito-core
└── ./mockito-extensions
│   ├── ./bom
│   └── ./subclass-mockmacker 
└── ./mockito-integration-tests
    ├── ...
    └── ./java-21-tests
```

This will enable a clear separation of the build scripts.

Cons, git will perceive this as a move and could make diffs a tad more difficult, by using a `--follow`. Even in IntelliJ automatically follows the file when looking at the file history (clicking on the colored line expand them with the new name).

Github has improved at [tracking this](https://github.blog/changelog/2022-06-06-view-commit-history-across-file-renames-and-moves/).

I believe the benefit outweighs this issue. 

----
As a side note: This is not the first time that sources are moved in Mockito history, when Mockito moved from SVN to Mercurial, then to git, the source folder was also different (we had `src` then `src/main/java`).

Depends on #3443

bric3:
Moves the core sources to a `mockito-core` module, this has several benefits mentioned in #3444

1. Improves the mental model of the project, in a multi-module project it is weird (if not confusing) to have the main sources in the root project.
2. This aligns well with the practices seen on other notable projects.
3. Improves the build configuration separation of concerns, before all projects and the mockito core were configured in the root `build.gradle`.

With this change the structure will look like

```text
├── ./mockito-bom
├── ./mockito-core
└── ./mockito-extensions
│   ├── ./...
│   └── ./subclass-mockmacker 
└── ./mockito-integration-tests
    ├── ...
    └── ./java-21-tests
```

Previous history is possible using git `--follow`, IJ does that automatically (you can also click the colored line on the left the file history to expand the name).

Fixes #3444

- [x] Check coverage 80% instead of 85.75%
- [x] Check shipkit / maven publication on release, see https://github.com/mockito/mockito/pull/3453#issuecomment-2383789021

## Checklist


bric3:
After a `./gradlew publishToMavenLocal`

```
/Users/brice.dutheil/.m2/repository/org/mockito/
├── mockito-android
│   ├── 5.14.2-SNAPSHOT
│   │   ├── maven-metadata-local.xml
│   │   ├── mockito-android-5.14.2-SNAPSHOT-javadoc.jar
│   │   ├── mockito-android-5.14.2-SNAPSHOT-sources.jar
│   │   ├── mockito-android-5.14.2-SNAPSHOT.jar
│   │   └── mockito-android-5.14.2-SNAPSHOT.pom
│   └── maven-metadata-local.xml
├── mockito-bom
│   ├── 5.14.2-SNAPSHOT
│   │   ├── maven-metadata-local.xml
│   │   └── mockito-bom-5.14.2-SNAPSHOT.pom
│   └── maven-metadata-local.xml
├── mockito-core
│   ├── 5.14.2-SNAPSHOT
│   │   ├── maven-metadata-local.xml
│   │   ├── mockito-core-5.14.2-SNAPSHOT-javadoc.jar
│   │   ├── mockito-core-5.14.2-SNAPSHOT-sources.jar
│   │   ├── mockito-core-5.14.2-SNAPSHOT.jar
│   │   └── mockito-core-5.14.2-SNAPSHOT.pom
│   └── maven-metadata-local.xml
├── mockito-errorprone
│   ├── 5.14.2-SNAPSHOT
│   │   ├── maven-metadata-local.xml
│   │   ├── mockito-errorprone-5.14.2-SNAPSHOT-javadoc.jar
│   │   ├── mockito-errorprone-5.14.2-SNAPSHOT-sources.jar
│   │   ├── mockito-errorprone-5.14.2-SNAPSHOT.jar
│   │   └── mockito-errorprone-5.14.2-SNAPSHOT.pom
│   └── maven-metadata-local.xml
├── mockito-junit-jupiter
│   ├── 5.14.2-SNAPSHOT
│   │   ├── maven-metadata-local.xml
│   │   ├── mockito-junit-jupiter-5.14.2-SNAPSHOT-javadoc.jar
│   │   ├── mockito-junit-jupiter-5.14.2-SNAPSHOT-sources.jar
│   │   ├── mockito-junit-jupiter-5.14.2-SNAPSHOT.jar
│   │   └── mockito-junit-jupiter-5.14.2-SNAPSHOT.pom
│   └── maven-metadata-local.xml
├── mockito-proxy
│   ├── 5.14.2-SNAPSHOT
│   │   ├── maven-metadata-local.xml
│   │   ├── mockito-proxy-5.14.2-SNAPSHOT-javadoc.jar
│   │   ├── mockito-proxy-5.14.2-SNAPSHOT-sources.jar
│   │   ├── mockito-proxy-5.14.2-SNAPSHOT.jar
│   │   └── mockito-proxy-5.14.2-SNAPSHOT.pom
│   └── maven-metadata-local.xml
└── mockito-subclass
    ├── 5.14.2-SNAPSHOT
    │   ├── maven-metadata-local.xml
    │   ├── mockito-subclass-5.14.2-SNAPSHOT-javadoc.jar
    │   ├── mockito-subclass-5.14.2-SNAPSHOT-sources.jar
    │   ├── mockito-subclass-5.14.2-SNAPSHOT.jar
    │   └── mockito-subclass-5.14.2-SNAPSHOT.pom
    └── maven-metadata-local.xml
```

TimvdLippe:
We currently name our folders without the `mockito-` prefix in the extensions folder. In our top-level we do have the prefix. To keep things consistent, I would say we should name the folder `core` since the prefix is implied. Otherwise, I would expect the `android` folder to be named `mockito-android`. WDYT?

bric3:
I pondered on that, I thought it was better to have the top level projects starting with `mockito-`. E.g. if we use `core`, then it would be placed after the `config` folder but before the `doc` and `gradle` folders, which I thought not wrong but odd.

raphw:
I agree it should be consistent. A newcomer might otherwise always wonder. Either way works for me.

TimvdLippe:
To make reviewing this easier, let's decide on a folder structure for the extensions and land that in a separate PR.

I personally don't mind our sources folders order to be mixed with other configuration, but I do also appreciate the grouping. Given the option, I do agree with the proposal that all projects are prefixed with `mockito-`. Do we also want that for `mockito-extensions`, given that is not a published artifact?

While these are mostly philosophical questions, I think we should list the options, make a decision and move on. Ultimately I don't think there is a perfect option, so let's be pragmatic as well. But I do want us to make a decision before we move, as we already have moved the folder structure for extensions and now do it again. Let's try to keep our git history as clean as we can, avoiding multiple moves.

TimvdLippe:
@bric3 Can you please update the wording of the commit messages to follow the structure we use for the other commits?

```
Switch Gradle Enterprise plugin to Develocity
```

```
Move core sources to `mockito-core`

Previously, the source code for the `mockito-core` artifact lived in
`/src/main/java` and the configuration was part of the `rootProject`.
With this change, we align the `mockito-core` as a separate artifact
similar to the recent move to split up the extension projects into
`mockito-extensions`.

Fixes #3444
```

build.gradle
```diff
@@ -21,31 +21,14 @@ buildscript {
...
+apply from: "$rootDir/gradle/root/ide.gradle"
+apply from: "$rootDir/gradle/license.gradle"
 
-apply plugin: 'base'
-base.archivesName = 'mockito-core'
-
-apply from: 'gradle/shipkit.gradle'
-
-apply from: 'gradle/root/ide.gradle'
-apply from: 'gradle/root/gradle-fix.gradle'
```
TimvdLippe, line 38: I don't see this file in either the `rootProject` or in `mockito-core`. Does that mean it is unused now and should we delete it? It seems like it should be part of the `rootProject`, but looking at the contents of the file, I am not sure if it is still relevant on recent versions of Gradle.

TimvdLippe (approved):

build.gradle
bric3: Yeah I didn't add the fix since it is not anymore necessary.
My plan is to drop this file in subsequent work on build scripts. I can do it now though.

bric3:
> Can you please update the wording of the commit messages.

Ah yes, I'm using conventional commits now and didn't realize this didn't follow the convention there.

PR #3437, opened 2024-09-03
Allow for installing a Java agent within the Mockito jar, without exposing Byte Buddy's attach mechanism.

raphw:
A Java agent can also be attached directly to Mockito, what is also more clean as it will not affect other libraries that use Byte Buddy. Also, this does not require exposing the underlying Byte Buddy version and contains this logic within Mockito.

Currently, the build does not seem to retain the two additional manifest entries. Might it be that this is overridden by the bundle plugin for OSGi?

PR #3454, opened 2024-09-30
fix: gradle mockitoAgent configuration should not be transitive

bric3:
Fix the documentation for the gradle configuration. Before, the documentation I suggested was working for single jar agents, but `mockito-core` has some transitive dependencies, so they need to be removed ; they will be looked up on the classpath where mockito is already defined.

```kotlin
val mockitoAgent = configurations.create("mockitoAgent")
dependencies {
    testImplementation(libs.mockito) // all mockito and transitive dependencies are there
    mockitoAgent(libs.mockito) { isTransitive = false } // only mockito-core
}
tasks {
    test {
        jvmArgs("-javaagent:${mockitoAgent.asPath}")
    }
}
```

See #3037 
Follow up to #3437

## Checklist

### DIFF
--- a/src/main/java/org/mockito/Mockito.java
+++ b/src/main/java/org/mockito/Mockito.java
@@ -176,20 +176,31 @@ import java.util.function.Function;
  * to function without an explicit setup to enable instrumentation, and the JVM will always display a warning.
  *
  * <p>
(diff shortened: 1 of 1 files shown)


TimvdLippe (approved):
Looks good!

PR #3459, opened 2024-10-03
Avoid attach warning if Byte Buddy is configured for command-line attach.

raphw:
Avoids priniting a warning if the Byte Buddy agent is initialized, but not Mockito.


bric3 (commented):
Good catch! 
Code looks good to me. I believe it is possible to write tests for this.

src/main/java/org/mockito/internal/PremainAttachAccess.java
```diff
@@ -0,0 +1,66 @@
...
+                // agent as a fallback if the Mockito agent is not attached.
+                instrumentation =
+                        doGetInstrumentation(
+                                Installer.class, "getInstrumentation", systemClassLoader);
+            } catch (IllegalStateException ignored) {
+            }
+        }
+        if (instrumentation == null) {
+            if (ClassFileVersion.ofThisVm().isAtLeast(ClassFileVersion.JAVA_V21)) {
+                System.out.println(
```
bric3, line 36: **thought:** I totally forgot after looking back at the code base maybe we can use `Plugins.getMockitoLogger().log(...)` here instead of the `System.out.println`.

src/main/java/org/mockito/internal/PremainAttachAccess.java
```diff
@@ -0,0 +1,66 @@
...
+                        "Mockito is currently self-attaching to enable the inline-mock-maker. This "
+                                + "will no longer work in future releases of the JDK. Please add Mockito as an agent to your "
+                                + "build what is described in Mockito's documentation: "
+                                + "https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#0.3");
+            }
+            // Attempt to dynamically attach, as a last resort.
+            instrumentation = ByteBuddyAgent.install();
+        }
+        if (!instrumentation.isRetransformClassesSupported()) {
+            throw new IllegalStateException(
```
bric3, line 46: **question:** OK with ISE, but maybe we should use `MockitoException`.

bric3 (commented):
I'm undecided on the approach to write the test there. Either use our internal `ClassLoaders` test util which is a bit old and may not cover everything. Or just write a simple suite with different agent config.

src/main/java/org/mockito/internal/PremainAttachAccess.java
raphw: I wanted to avoid the stack trace filtering here as it might be a pure mockito stack?

raphw:
As for tests: we could start a VM with arguments and one without, indeed. Then capture the error output and see if it is empty in the one case, and contains the error message in the other?

src/main/java/org/mockito/internal/PremainAttachAccess.java
bric3: Current pr code uses st out, that said I like your idea on the default method.

src/main/java/org/mockito/internal/PremainAttachAccess.java
bric3: Indeed I didn't thought about filtering. Let's keep it that way.

bric3:
@raphw yeah I got a pr that was merged in yours where I created different gradle tasks with different vm args. Feel free to tweak the code if needed.

raphw:
Indeed, I missed that. Thanks for that!

Using the Mockito logger now.

raphw:
Something seems to not work when mergning the changes to the build. Could you have a look @bric3?

bric3:
@raphw I ixed the tast tasks. However I must revert my suggested change on the logger. Calling `Plugins.getMockitoLogger()` in the `PremainAttachAccess` cannot work currently due to a circular dependency on `Plugins.registry`.

```mermaid
flowchart TD
    A["Mockito.mock(...)"] --> B
    B["MockitoCore.mock(...)"] -->|Class init| C
    C["MockitCore.&lt;clinit&gt;"] --> D
    D["Plugins.&lt;clinit&gt;"] -->|Loads mockmaker| E
    E["InlineByteBuddyMockMaker&lt;linit&gt;"] --> F
    F["PremainAttachAccess.getInstrumentation()"] -->|if java 21 and no javaagent| G
    G["Plugins.getMockitoLogger()"] -->|Plugins class still initializing, Plugins.registry is null| D
```

The last commit will issue the warning on stderr rather than mockito logger. Sorry the fuss.

<details><summary>Stack relevant snippets</summary>

```
Could not initialize plugin: interface org.mockito.plugins.MockMaker (alternate: null)
	at org.mockito.internal.configuration.plugins.PluginLoader$1.invoke(PluginLoader.java:85)
	at jdk.proxy3/jdk.proxy3.$Proxy13.isTypeMockable(Unknown Source)
	at org.mockito.internal.util.MockUtil.typeMockabilityOf(MockUtil.java:78)
	at org.mockito.internal.util.MockCreationValidator.validateType(MockCreationValidator.java:22)
	at org.mockito.internal.creation.MockSettingsImpl.validatedSettings(MockSettingsImpl.java:275)
	at org.mockito.internal.creation.MockSettingsImpl.build(MockSettingsImpl.java:236)
	at org.mockito.internal.MockitoCore.mock(MockitoCore.java:82)
	at org.mockito.Mockito.mock(Mockito.java:2162)
	at org.mockito.Mockito.mock(Mockito.java:2135)

```

```
	at org.mockito.internal.configuration.plugins.DefaultMockitoPlugins.create(DefaultMockitoPlugins.java:105)
	at org.mockito.internal.configuration.plugins.DefaultMockitoPlugins.getDefaultPlugin(DefaultMockitoPlugins.java:79)
	at org.mockito.internal.configuration.plugins.PluginLoader.loadPlugin(PluginLoader.java:75)
	at org.mockito.internal.configuration.plugins.PluginLoader.loadPlugin(PluginLoader.java:49)
	at org.mockito.internal.configuration.plugins.PluginRegistry.<init>(PluginRegistry.java:29)
	at org.mockito.internal.configuration.plugins.Plugins.<clinit>(Plugins.java:26)
	at org.mockito.internal.MockitoCore.<clinit>(MockitoCore.java:71)
	at org.mockito.Mockito.<clinit>(Mockito.java:1741)
```

```
	at org.mockito.internal.creation.bytebuddy.InlineDelegateByteBuddyMockMaker.<init>(InlineDelegateByteBuddyMockMaker.java:254)
	at org.mockito.internal.creation.bytebuddy.InlineByteBuddyMockMaker.<init>(InlineByteBuddyMockMaker.java:23)
	at java.base/jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:62)
	... 50 more
Caused by: java.lang.NullPointerException: Cannot invoke "org.mockito.internal.configuration.plugins.PluginRegistry.getMockitoLogger()" because "org.mockito.internal.configuration.plugins.Plugins.registry" is null
	at org.mockito.internal.configuration.plugins.Plugins.getMockitoLogger(Plugins.java:83)
	at org.mockito.internal.PremainAttachAccess.getInstrumentation(PremainAttachAccess.java:42)
	at org.mockito.internal.creation.bytebuddy.InlineDelegateByteBuddyMockMaker.<clinit>(InlineDelegateByteBuddyMockMaker.java:138)
```


</details>

bric3 (approved):

src/main/java/org/mockito/internal/PremainAttachAccess.java
bric3: Reverted, see https://github.com/mockito/mockito/pull/3459#issuecomment-2396220498

Issue #3446, opened 2024-09-17
Use convention plugins


bric3:
At this time the build script is based on applying other gradle scripts. This approach is old by 10 year in the early days of gradle. THe preferred approach nowadays is to use convention plugins.

For example in Gradle 8.10, a `gradle init` with the app template, will create a `buildSrc` folder with convention plugins, e.g that configure the java conventions, etc.

We should do the same.

This task can be done incrementally.

PR #3464, opened 2024-10-07
Move to Gradle conventions plugins

bric3:
This PR port most of the existing sparse Groovy DSL scripts as [_convention plugins_](https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html#sec:convention_plugins), this is done to switch from decade old accumulation of scripting from the early days of Gradle to best practice of today. Fortunately our script were well split. 

For the full motivation see #3446

At this stage, this PR port is almost a 1 to 1, i.e. it doesn't aim at introducing or changing the build behavior. It is not complete, the coverage, shipkit, spotless, licenses still remain to be converted.

**tl;dr:** what this PR does is changing `apply from: ....gradle` to a convention plugin `plugins { id("mockito....-conventions") }`. THese plugin are added in the standard `buildSrc`

* Create the `buildSrc` project for the conventions plugins
  * Declares the moved plugin in the version catalog file

* Java lib config
  * `java-library.gradle` --> `mockito.library-conventions.gradle.kts`, which applies other convention plugins below
  * `java-library.gradle` --> `mockito.java-library-conventions.gradle.kts`, also sets up Java related configurations found in the root `build.gradle`, and applies `mockito.java-conventions`
  * `osgi.gradle` --> `mockito.osgi-conventions.gradle.kts`, this only applies the plugin as config differ between mockito-core and mockito-junit-jupiter.
  * `java-publication.gradle` --> `mockito.publication-conventions.gradle.kts`
  * `checkstyle.gradle` --> `mockito.checkstyle-conventions.gradle.kts`
  * `errorprone.gradle` merged in `mockito.java-library-conventions.gradle.kts`

  *  animalSniffer config in `mockito-core/build.gradle` and `mockito-junit-jupiter/build.gradle` --> extracted to `mockito.java-backward-compatibility-checks-conventions.gradle.kts`

* Tests
  * `java-test.gradle` --> `mockito.test-conventions.gradle.kts` (which applies `mockito.test-jfr-profiler-conventions`, `mockito.java-conventions`)
  * `jfr.gradle` --> `mockito.test-jfr-profiler-conventions.gradle.kts`
  * `test-launcher.gradle` --> `mockito.test-launcher-conventions.gradle.kts`
  * `retry-test.gradle` --> `mockito.test-retry-conventions.gradle.kts`

* Javadoc
  * `mockito-core/javadoc.gradle` and `mockito-junit-jupiter/javadoc.gradle` --> `mockito.javadoc-conventions.gradle.kts`
  * `mockito-core/src/main/javadoc` -> `config/javadoc/resources`, because it's shared with junit jupiter extension.

Now Mockito Libraries (core and extensions) declare this

```
plugins {
    id("mockito.library-conventions")
    id("mockito.java-backward-compatibility-checks-conventions") // if needed
    id("mockito.javadoc-conventions") // if needed
}
```

And integration tests have this

```
plugins {
    id("java")
    id("mockito.test-conventions")
}
```

Android test project also applies our convention.

```
plugins {
    id("com.android.application")
    id("kotlin-android")
    id("mockito.test-conventions")
}
```

-----

**Two approaches for the review**

1. Take a look at all the [file changes](https://github.com/mockito/mockito/pull/3464/files).
2. Review commit by commit, starting with this one : https://github.com/mockito/mockito/pull/3464/commits/94132ac72f553406b17d81aea46482042da820a3, navigate back and forth with previous / next.

-----

## Checklist


build.gradle
```diff
@@ -20,45 +19,16 @@ buildscript {
...
 
 subprojects { proj ->
-    repositories {
-        mavenCentral()
-        google()
-    }
     plugins.withId('java') {
-        proj.apply from: "$rootDir/gradle/errorprone.gradle"
-        proj.apply from: "$rootDir/gradle/jfr.gradle"
         proj.apply from: "$rootDir/gradle/spotless.gradle"
```
TimvdLippe, line 32: Shouldn't we apply this in `buildSrc/src/main/kotlin/mockito.java-conventions.gradle.kts` instead? That's the convention we should apply to all Java projects, so I think it should apply `spotless` as well. That way, we git rid of the `subprojects` call as well (which is discouraged because of performance implications)

buildSrc/src/main/kotlin/mockito.java-library-conventions.gradle.kts
```diff
@@ -0,0 +1,40 @@
...
+    id("net.ltgt.errorprone")
+}
+
+afterEvaluate {
+    if (!base.archivesName.get().startsWith("mockito-")) {
+        // TODO throw GradleException("Project module should be prefixed with `mockito-` : ${project.projectDir})")
+        logger.warn("Project module should be prefixed with `mockito-` : ${project.projectDir})")
+        base.archivesName = "mockito-" + project.name
+    }
+}
```
TimvdLippe, line 16: I don't think the `afterEvaluate` is required here and we should use the `project.name` instead to have the correct logic. Then, we can also immediately throw and enforce this invariant.

```suggestion
if (!project.getName().startsWith("mockito-")) {
    throw new GradleException("Project module should be prefixed with `mockito-` : ${project.projectDir})")
}
```
```

File: buildSrc/src/main/kotlin/mockito.publication-conventions.gradle.kts [hunk 3]
```diff
@@ -0,0 +1,164 @@
+import gradle.kotlin.dsl.accessors._2a0859b2b1624a8c1f98b2dbe4a3b98d.javadoc
```
TimvdLippe, line 1: I am not that familiar with Kotlin Gradle scripts yet, but to me this looks like an internal import that we should not be using. Is this intentional?

buildSrc/src/main/kotlin/mockito.publication-conventions.gradle.kts
```diff
@@ -0,0 +1,164 @@
...
+group = "org.mockito"
+
+// generic publication conventions
+publishing {
+    publications {
+        withType(MavenPublication::class).configureEach {
+            pom {
+                artifactId = project.base.archivesName.get()
+                name = artifactId
+                afterEvaluate {
```
TimvdLippe, line 21: Same concern with `afterEvaluate` here. Can we instead define a `mockitoPublicationExtension` to configure the description?

mockito-core/build.gradle
```diff
@@ -1,20 +1,15 @@
...
+    id("mockito.java-backward-compatibility-checks-conventions")
+    id("mockito.javadoc-conventions")
 }
 
 description = 'Mockito mock objects library core API and implementation'
 base.archivesName = 'mockito-core'
 
-
-apply from: "$rootDir/gradle/java-library.gradle"
 apply from: "$rootDir/gradle/coverage.gradle"
```
TimvdLippe, line 10: Do we want to tackle all `apply from` applications in this PR? I don't mind doing them in a separate PR as well, but curious to what your preferred ordering of work is.

TimvdLippe (changes requested):

mockito-core/build.gradle
bric3: Indeed, I prefer to proceed with the rest in another PR. I stopped where the changes makes sense.

Especially for coverage, where I noticed we may be able to [simply the aggregation](https://github.com/orgs/mockito/projects/1/views/1?pane=issue&itemId=80276438).

mockito-extensions/mockito-android/build.gradle
```diff
@@ -1,4 +1,7 @@
-apply from: "$rootDir/gradle/java-library.gradle"
+plugins {
+    id("mockito.java-library-conventions")
+    id("mockito.publication-conventions")
```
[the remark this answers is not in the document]
bric3: It's applied to `mockito.library-conventions`, but not `mockito.java-library-conventions`. The reason is that `mockito.library-conventions` has osgi, which is applied to other jvm projects. 

Also, `checkstyle` was not applied to this project, but I can add it.

I could create a plugin `mockito.android-library-conventions` without the osgi setup in order to have symmetry with the `mockito.library-conventions` plugin.

build.gradle
bric3: Yes that was my plan, but I ran into some issue with spotless, so I'd rather do it in another [task](https://github.com/orgs/mockito/projects/1/views/1?pane=issue&itemId=83037086).

buildSrc/src/main/kotlin/mockito.java-library-conventions.gradle.kts
bric3: Good catch! I edited your suggestion to use the kotlin property access `.name` instead of `.getName()`

buildSrc/src/main/kotlin/mockito.publication-conventions.gradle.kts
bric3: Thank you it shouldn't be here !

buildSrc/src/main/kotlin/mockito.publication-conventions.gradle.kts
bric3: This was in the original script, let's see if it's necessary now that it is in a convention plugin.

bric3: Done

TimvdLippe (approved):
Thanks for all the work and fixing the commit messages as well, much appreciated!

TimvdLippe:
Ah, I merged some plugin updates and didn't realize that would cause merge conflicts again :(

PR #3476, opened 2024-10-16
Fixes #3469: Prevent NPEs for Disabled Mocks

### LINKED ISSUE #3469: Disabled Mock Triggers NPE when Checking isMock
check that
 - [X] The mockito message in the stacktrace have useful information, but it didn't help
 - [X] The problematic code (if that's possible) is copied here;
 - [X] Provide versions (mockito / jdk / os / any other relevant information)
 - [X] Provide a [Short, Self Contained, Correct (Compilable), Example](http://sscce.org) of the issue

There is the ability to trigger a NullPointerException by calling `isMock` on a disabled mock. This behavior should probably be changed to either return true, because a DisabledMock is still a mock or `isMock` should throw a `DisabledMockException` because you are interacting with a disabledMock.
 
Problematic Code is: `DisabledMockHandler#getMockSettings` returning `null` here breaks the logic for determining if an object a mock.
```
@Override
    public MockCreationSettings getMockSettings() {
        return null;
    }
```
NPE Source is in the following loop of `MockUtil#getMockHandlerOrNull`. 
```
for (MockMaker mockMaker : mockMakers.values()) {
            MockHandler<?> handler = mockMaker.getHandler(mock);
            if (handler != null) {
                assert getMockMaker(handler.getMockSettings().getMockMaker()) == mockMaker;
                return handler;
            }
        }
```

Stack Trace Snippet:
```
java.lang.NullPointerException: Cannot invoke "org.mockito.mock.MockCreationSettings.getMockMaker()" because the return value of "org.mockito.invocation.MockHandler.getMockSettings()" is null
	at org.mockito.internal.util.MockUtil.getMockHandlerOrNull(MockUtil.java:160)
	at org.mockito.internal.util.MockUtil.isMock(MockUtil.java:147)
	at org.mockito.internal.util.DefaultMockingDetails.isMock(DefaultMockingDetails.java:32)
	at org.mockito.internal.framework.DefaultMockitoFrameworkTest.behavior_after_clear_inline_mocks_is_mock(DefaultMockitoFrameworkTest.java:226)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
```

The following unit test reproduces the issue. This unit test can be added to `DefaultMockitoFrameworkTest` and will fail for an NPE. This test should probably be modeled to either trigger a `DisabledMockException` or `return true` because a DisabledMock is still a mock.
```

    @Test
    public void behavior_after_clear_inline_mocks_is_mock() {
        // clearing mocks only works with inline mocking
        assumeTrue(Plugins.getMockMaker() instanceof InlineMockMaker);
        PersonWithName obj = mock(PersonWithName.class);
        when(obj.getMyName()).thenReturn("Bob");
        assertEquals("Bob", obj.getMyName());
        assertTrue(mockingDetails(obj).isMock());
        framework.clearInlineMocks();

        try {
            mockingDetails(obj).isMock();
        } catch (DisabledMockException e) {
            return;
        }
        Assert.fail("Should have thrown DisabledMockException");
    }
 ```

Kramermp:
This change makes a couple of changes surronding the the clearMock and clearMock apis.
1. All disabled mocks gets the INSTANCE of DisabledMockCreationSettings. This prevents any NPEs from attempting to access the MockCreationSettings, but will throw a MockDisabledException if attempting to actually read the settings. This will block calls such as: isMock(), etc.
2. This change syncs the logic between the clearMocks() and clearMock() logic to make the customer experience the same whether you are attempting to clear all mocks or just and individual mock.

These changes are being made mainly to introduce some more consistent behavior for mocks that have been cleared. Currently, the user experience is less than desirable to have NPEs be tripped depending on which clear method was called on the mock. Additionally, I think this disabling approach brings the design more in-line with some of the documentation about the behavior of a cleared mock. 

## Checklist

 - [X ] Avoid other runtime dependencies

### DIFF
--- a/mockito-core/src/main/java/org/mockito/internal/creation/bytebuddy/InlineDelegateByteBuddyMockMaker.java
+++ b/mockito-core/src/main/java/org/mockito/internal/creation/bytebuddy/InlineDelegateByteBuddyMockMaker.java
@@ -532,7 +532,11 @@ class InlineDelegateByteBuddyMockMaker
                 entry.remove(mock);
             }
         } else {
-            mocks.remove(mock);
+            mocks.put(
+                    mock,
+                    new MockMethodInterceptor(
+                            DisabledMockHandler.HANDLER,
+                            DisabledMockHandler.HANDLER.getMockSettings()));
         }
     }
 
@@ -541,8 +545,7 @@ class InlineDelegateByteBuddyMockMaker
         mockedStatics.getBackingMap().clear();
 
         for (Entry<Object, MockMethodInterceptor> entry : mocks) {
-            MockCreationSettings settings = entry.getValue().getMockHandler().getMockSettings();
-            entry.setValue(new MockMethodInterceptor(DisabledMockHandler.HANDLER, settings));
+            clearMock(entry.getKey());
         }
     }
 
--- a/mockito-core/src/main/java/org/mockito/internal/framework/DisabledMockHandler.java
+++ b/mockito-core/src/main/java/org/mockito/internal/framework/DisabledMockHandler.java
@@ -9,7 +9,19 @@ import org.mockito.exceptions.misusing.DisabledMockException;
 import org.mockito.invocation.Invocation;
 import org.mockito.invocation.InvocationContainer;
 import org.mockito.invocation.MockHandler;
+import org.mockito.listeners.InvocationListener;
+import org.mockito.listeners.StubbingLookupListener;
+import org.mockito.listeners.VerificationStartedListener;
 import org.mockito.mock.MockCreationSettings;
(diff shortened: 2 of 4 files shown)


TimvdLippe (approved):
Thanks a lot for the improvements!

PR #3494, opened 2024-11-04
Bump errorprone from 2.23.0 to 2.35.1 v2

### LINKED ISSUE #3493: Update ErrorProne to the latest version
Over the past year (?) we have received Dependabot updates for ErrorProne. However, they fail the pipeline as the newest versions of ErrorProne catch existing issues in the codebase. An example PR from Dependabot: https://github.com/mockito/mockito/pull/3488

We need to update our version, as well as fix the pre-existing issues it catches. This is a good first issue to pick up, as it only requires compilation to succeed on the latest ErrorProne version.

kbs1027:
- Replaced `assertThat(set).isEqualTo(set)` with `assertThat(set.equals(set)).isTrue()` in order to explicitly verify reflexivity and avoid triggering ErrorProne's SelfAssertion warning. 

- Fixes #3493 

## Checklist

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,6 +1,6 @@
 [versions]
 bytebuddy = "1.15.8"
-errorprone = "2.23.0"
+errorprone = "2.35.1"
 junit4 = "4.13.2"
 junit-jupiter = "5.11.3"
 junit-platform = "1.11.3"
--- a/mockito-core/src/test/java/org/mockito/internal/util/collections/HashCodeAndEqualsSafeSetTest.java
+++ b/mockito-core/src/test/java/org/mockito/internal/util/collections/HashCodeAndEqualsSafeSetTest.java
@@ -133,7 +133,7 @@ public class HashCodeAndEqualsSafeSetTest {
     @Test
     public void isEqualToItself() {
         HashCodeAndEqualsSafeSet set = HashCodeAndEqualsSafeSet.of(mock1);
-        assertThat(set).isEqualTo(set);
+        assertThat(set.equals(set)).isTrue();
     }
 
     @Test
--- a/mockito-core/src/test/java/org/mockitousage/basicapi/ReplacingObjectMethodsTest.java
+++ b/mockito-core/src/test/java/org/mockitousage/basicapi/ReplacingObjectMethodsTest.java
@@ -33,7 +33,7 @@ public class ReplacingObjectMethodsTest extends TestBase {
         Object mock = Mockito.mock(ObjectMethodsOverridden.class);
         Object otherMock = Mockito.mock(ObjectMethodsOverridden.class);
 
-        assertThat(mock).isEqualTo(mock);
+        assertThat(mock.equals(mock)).isTrue();
         assertThat(mock).isNotEqualTo(otherMock);
         assertThat(mock.hashCode()).isNotEqualTo(otherMock.hashCode());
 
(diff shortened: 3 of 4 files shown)


kbs1027:
@TimvdLippe 
It seems that the CI is failing on Java 11 because the com/google/errorprone/bugpatterns/BugChecker file in errorprone version 2.35.1 has been compiled with Java 17. Do you have any suggestions on how to handle this situation?

kbs1027:
@TimvdLippe 
It seems necessary to split the dependency configuration based on whether the Java version is 17 or above.
what do you think about this?

TimvdLippe:
Seems like this release changed behaviour: https://github.com/google/error-prone/releases/tag/v2.33.0

Can we update our Gradle toolchain to set the correct language versions maybe?

kbs1027:
@TimvdLippe  Starting from version 2.33.0, Error Prone is compiled with Java 17. If we try to compile it with a version lower than Java 17, it will fail in the CI process. Even if we set the toolchain to use Java 17, it still won’t compile if the CI is running Java 11. To handle this, I’m thinking of dynamically checking the conditions and configuring the Error Prone version accordingly:
```
if (JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_17)) {
    implementation libs.errorprone.test.api;
} else {
    implementation "com.google.errorprone:error_prone_test_helpers:2.23.0";
}
```
What do you think about this approach?

kbs1027:
> Can we update our Gradle toolchain to set the correct language versions maybe?

So, what you're saying is that if CI is running on Java 11, I should exclude it from compilation and set up the toolchain to enable usage starting from Java 17?

TimvdLippe:
Yes, on Java 11, we should not configure ErrorProne at all. Therefore, we still use a single version of ErrorProne and make it work on all versions. To do so, add the if-statement as you posted around the configuration of the `JavaCompile` tasks.

bric3:
It seems the CI is running Java 17 already.

https://github.com/mockito/mockito/blob/7a58ebe1e511f3f04fedb808324a88fc666204d1/.github/workflows/ci.yml#L49-L54

Error prone tests however run with the configured tool chain, which include a JDK 11.

If the ABI of `com.google.errorprone:error_prone_test_helpers` (libs.errorprone.test.api) is compatible between the two versions we could declare the "api" as compile only and inject the right dependency in the test taks

Something like that 


```diff 
+val errorProneJava11: Configuration by configurations.creating
+val errorProneJava17: Configuration by configurations.creating
+
 dependencies {
     implementation(libs.errorprone)
 
     testImplementation("junit:junit:4.13.2")
-    testImplementation(libs.errorprone.test.api)
+    testCompileOnly(libs.errorprone.test.api)
+    
+    errorProneJava11(libs.errorprone.test.helpers11)
+    errorProneJava17(libs.errorprone.test.helpers17)
 }
 
 tasks {
    test {
+        when (providers.gradleProperty("mockito.test.java").get()) {
+            "auto" -> { check current jvm }
+            "11" -> classpath += errorProneJava17
+            else -> classpath += errorProneJava17
+        }
+
    }
 }
```


<details><summary>(Or maybe use a dependency substitution)</summary>

```diff 
+configurations {
+    testRuntimeOnly {
+        val errorProneVersion = when (providers.gradleProperty("mockito.test.java").get()) {
+            "auto" -> { check() current jvm }
+            "11" -> libs.versions.errorprone11.get()
+            else -> libs.versions.errorprone17.get()
+        }
+        resolutionStrategy.eachDependency {
+            if (requested.group == "com.google.errorprone") {
+                useVersion(errorProneVersion)
+            }
+        }
+    }
+}
```

</details>

-----

Thinking loudly: That said, when I see the the "auto" value I think we are too relaxed here, and instead we may want to enforce the JDK by default (maybe have multiple test tasks, or test suites each using the right JDK).

PR #3509, opened 2024-11-15
feat(Optional<Callable<?>>): add new function

luizanurnberg:
This PR refactors the MockMethodAdvice class to replace nullable return values with Optional. This change enhances the readability and safety of the code by explicitly indicating that some methods may not always return a value. It reduces the risk of NullPointerException by requiring the caller to handle cases where no value is present, aligning with modern Java practices for handling optional values.


TimvdLippe:
@luizanurnberg Can you please remove all commits that are unrelated to this PR and add regression tests that showcase the need for these changes? We typically do not accept structural changes, unless they fix existing bugs in our codebase (to avoid churn and potentially break users in unexpected ways).

TimvdLippe:
Closing per the above comment. We would love to see your contributions, but please ensure you have clean unique commits in your PR so can we review and merge them.

PR #3510, opened 2024-11-15
Feature/thread lock

luizanurnberg:
This PR refactors the ThreadVerifiesContinuouslyInteractingMockTest class to improve thread safety by adding a ReentrantLock around critical sections. The goal is to ensure that interactions with the mock object are safe and consistent when accessed from multiple threads, preventing race conditions and improving the reliability of the test in a multithreaded environment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,7 +89,7 @@ jobs:
         --scan
 
     - name: 8. Upload coverage report
-      uses: codecov/codecov-action@v5
+      uses: codecov/codecov-action@v4
       env:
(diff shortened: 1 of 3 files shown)


TimvdLippe:
@luizanurnberg Can you please remove all commits that are unrelated to this PR and add regression tests that showcase the need for these changes? We typically do not accept structural changes, unless they fix existing bugs in our codebase (to avoid churn and potentially break users in unexpected ways).

TimvdLippe:
Closing per the above comment. We would love to see your contributions, but please ensure you have clean unique commits in your PR so can we review and merge them.

Issue #3528, opened 2024-12-06
Mockito 5.14 + Java 21 + Maven surefire agent config : Cannot mock final class


lwiddershoven:
Reproduction: https://github.com/lwiddershoven/mockito-issue

When maven (or IntelliJ) run my test it fails:
```
[ERROR] IssueTest.issue -- Time elapsed: 0.159 s <<< ERROR!
org.mockito.exceptions.base.MockitoException: 

Cannot mock/spy class java.lang.reflect.Method
Mockito cannot mock/spy because :
 - final class
        at IssueTest.issue(IssueTest.java:12)
```

I use:
- Mockito 5.14.2
- JVM OpenJDK Runtime Environment Temurin-21.0.5+11 (build 21.0.5+11-LTS)
- Maven: Apache Maven 3.9.9 (8e8579a9e76f7d015ee5ec7bfcdc97d26018693

I have applied the maven configuration from: https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#mockito-instrumentation

Mockito works fine when mocking normal classes, but specifically not final classes.

The pom configuration has been picked up as the Mockito and javaagent warnings do not show up in the maven output anymore.

Relevant part of the pom:
```xml
 <plugins>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-dependency-plugin</artifactId>
                <executions>
                    <execution>
                        <goals>
                            <goal>properties</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-surefire-plugin</artifactId>
                <version>${surefire-plugin.version}</version>
                <configuration>
                    <!-- Java 21+ does not allow dynamic loading/  some mockito functionality - this is how we enable that explicitly-->
                    <argLine>@{argLine} -javaagent:${org.mockito:mockito-core:jar}</argLine>
                    <systemPropertyVariables>
                        <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
                        <maven.home>${maven.home}</maven.home>
                    </systemPropertyVariables>
                </configuration>
            </plugin>
        </plugins>
```

I'm not sure what more I can do to test or fix this. Any ideas?

bric3:
The dependency in your reproducer is 

```
<dependency>
    <groupId>io.quarkus</groupId>
[...]

Issue #3518, opened 2024-11-21
Mockito cannot mock this class: class org.apache.lucene.index.IndexReader


parnmatt:
We are in the process of bumping to Lucene 10.0.0, and JDK 21.
We are using the latest Mockito version available 5.14.2.

There have been some internal changes in this major version of Lucene; however, now our mocked tests are failing trying to mock the `IndexReader`. We get the lovely exception to reach out below:

```
org.mockito.exceptions.base.MockitoException: 
Mockito cannot mock this class: class org.apache.lucene.index.IndexReader.

If you're not sure why you're getting this error, please open an issue on GitHub.


Java               : 21
JVM vendor name    : Eclipse Adoptium
JVM vendor version : 21.0.4+7-LTS
JVM name           : OpenJDK 64-Bit Server VM
JVM version        : 21.0.4+7-LTS
JVM info           : mixed mode
OS name            : Linux
OS version         : 5.19.0-35-generic


You are seeing this disclaimer because Mockito is configured to create inlined mocks.
You can learn about inline mocks and their limitations under item #39 of the Mockito class javadoc.

Underlying exception : org.mockito.exceptions.base.MockitoException: Unsupported settings with this type 'org.apache.lucene.index.IndexReader'
[...]
Caused by: org.mockito.exceptions.base.MockitoException: Unsupported settings with this type 'org.apache.lucene.index.IndexReader'
	at net.bytebuddy.TypeCache.findOrInsert(TypeCache.java:168)
	at net.bytebuddy.TypeCache$WithInlineExpunction.findOrInsert(TypeCache.java:399)
	at net.bytebuddy.TypeCache.findOrInsert(TypeCache.java:190)
	at net.bytebuddy.TypeCache$WithInlineExpunction.findOrInsert(TypeCache.java:410)
	... 11 more
```
---

The bare minimum is enough to trigger this:
```java
IndexReader indexReader = mock(IndexReader.class);
```

In a different test we did have:
```java
private final IndexSearcher indexSearcher = mock(IndexSearcher.class, Mockito.RETURNS_DEEP_STUBS);
// ...
IndexReader indexReader = new IndexReaderStub(new SamplingFields(fieldTermsMap));
when(indexSearcher.getIndexReader()).thenReturn(indexReader);
```
[...]

PR #3531, opened 2024-12-07
Improves a bit the error reporting

bric3:
Improves a bit the error report
- shows mock mack in more cases
- friendly report of `abstract sealed class` being non mockable

Should help diagnosing in cases like
* #3518
* #3528

## Checklist

### DIFF
--- a/mockito-core/build.gradle.kts
+++ b/mockito-core/build.gradle.kts
@@ -1,14 +1,11 @@
 import com.android.build.gradle.internal.tasks.factory.dependsOn
-import org.jetbrains.gradle.ext.settings
-import org.jetbrains.gradle.ext.taskTriggers
+import com.diffplug.gradle.spotless.SpotlessExtension
 
 plugins {
     id("mockito.library-conventions")
     id("mockito.java-backward-compatibility-checks-conventions")
     id("mockito.javadoc-conventions")
     id("java-test-fixtures")
-
-    alias(libs.plugins.ideaExt)
 }
 
 description = "Mockito mock objects library core API and implementation"
@@ -207,43 +204,50 @@ fun mockitoExtensionConfigFile(project: Project, mockitoExtension: String) =
 //<editor-fold defaultstate="collapsed" desc="Generate version class file">
 // Manifest cannot be used because Android strips the manifest from the jar.
 
-val generatedVersionClassDir: Provider<Directory> = layout.buildDirectory.dir("generated/sources/version/java")
+val generateVersionClass by tasks.registering(GenerateVersionClass::class)
 
 sourceSets.main {
     java {
-        srcDir(generatedVersionClassDir)
+        srcDir(generateVersionClass)
     }
 }
 
-tasks {
-    val generateVersionClass by registering {
-        val versionClassContent = """
-        package org.mockito.internal.util;
-        class MockitoVersion {
-            public static final String VERSION = "${project.version}";
-        }
-        """.trimIndent()
+tasks.withType<Checkstyle> {
+    exclude("org/mockito/internal/util/MockitoVersion.java")
+}
 
-        val versionClass = file(generatedVersionClassDir.map { it.file("org/mockito/internal/util/MockitoVersion.java") })
+configure<SpotlessExtension> {
+    java {
+        targetExclude(relativePath(generateVersionClass.get().generatedVersionClassDir) + "/**")
+    }
+}
 
-        doLast {
-            versionClass.run {
+abstract class GenerateVersionClass : DefaultTask() {
+    @OutputDirectory
+    val generatedVersionClassDir: DirectoryProperty = project.objects.directoryProperty()
+        .convention(project.layout.buildDirectory.dir("generated/sources/version/java"))
+
+    @Input
+    val version: Property<String> = project.objects.property(String::class.java)
+        .convention(project.version.toString())
+
+    @TaskAction
+    fun generate() {
+        generatedVersionClassDir.file("org/mockito/internal/util/MockitoVersion.java").get()
+            .asFile
+            .run {
                 parentFile.mkdirs()
                 createNewFile()
-                writeText(versionClassContent)
+                writeText(
+                    // language=Java
+                    """
+                    package org.mockito.internal.util;
+                    class MockitoVersion {
+                        public static final String VERSION = "${version.get()}";
+                    }
+                    """.trimIndent()
+                )
             }
-        }
-    }
-    compileJava.dependsOn(generateVersionClass)
-
-    rootProject.idea {
-        project {
-            settings {
-                taskTriggers {
-                    afterSync(generateVersionClass)
-                }
-            }
-        }
     }
 }
 //</editor-fold>
--- a/mockito-core/build.gradle.kts
+++ b/mockito-core/build.gradle.kts
@@ -1,10 +1,14 @@
 import com.android.build.gradle.internal.tasks.factory.dependsOn
+import org.jetbrains.gradle.ext.settings
+import org.jetbrains.gradle.ext.taskTriggers
 
 plugins {
     id("mockito.library-conventions")
     id("mockito.java-backward-compatibility-checks-conventions")
     id("mockito.javadoc-conventions")
     id("java-test-fixtures")
+
+    alias(libs.plugins.ideaExt)
 }
 
 description = "Mockito mock objects library core API and implementation"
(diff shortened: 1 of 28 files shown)


mockito-core/src/main/java/org/mockito/internal/creation/bytebuddy/InlineDelegateByteBuddyMockMaker.java
```diff
@@ -554,7 +558,10 @@ public TypeMockability isTypeMockable(final Class<?> type) {
         return new TypeMockability() {
             @Override
             public boolean mockable() {
-                return INSTRUMENTATION.isModifiableClass(type) && !EXCLUDES.contains(type);
+                return INSTRUMENTATION.isModifiableClass(type)
+                        && !EXCLUDES.contains(type)
+                        && !(Modifier.isAbstract(type.getModifiers())
```
TimvdLippe, line 563: Should we instead update `INSTRUMENTATION` to handle this? If that does not make sense, then I would vouch for creating a method that returns a boolean and we use it here. That makes it more readable and we can use it down below in the reason as well.

bric3: That's not possible as the instrumentation comes from the JDK.

But I was thinking about moving some type related checks in our `TypeSupport`. Currently, it only has a check for `isSealed`.

TimvdLippe: WDYT about introducing a `TypeSupport.INSTANCE.ismockable()` where we implement the logic? That way, the separation of concerns is on the class boundary.

mockito-core/build.gradle.kts
```diff
@@ -200,3 +204,47 @@ fun mockitoExtensionConfigFile(project: Project, mockitoExtension: String) =
...
+            versionClass.run {
+                parentFile.mkdirs()
+                createNewFile()
+                writeText(versionClassContent)
+            }
+        }
+    }
+    classes.dependsOn(generateVersionClass)
+
+    rootProject.idea {
```
TimvdLippe, line 239: Rather than manually configuring the dependency here, I think we can use the output of the Task declared here as inputs of the `srcDir`: https://discuss.gradle.org/t/right-way-to-generate-sources-in-gradle-7/41141/3

E.g. I think the folowing code is equivalent to what you wrote:

```kotlin

val generatedVersionClassDir: Provider<Directory> = layout.buildDirectory.dir("generated/sources/version/java")

tasks {
    val generateVersionClassTask by registering {
        val versionClassContent = """
        package org.mockito.internal.util;
        class MockitoVersion {
            public static final String VERSION = "${project.version}";
        }
        """.trimIndent()

        val versionClass = file(generatedVersionClassDir.map { it.file("org/mockito/internal/util/MockitoVersion.java") })

        doLast {
            versionClass.run {
                parentFile.mkdirs()
                createNewFile()
                writeText(versionClassContent)
            }
        }
    }

    sourceSets.main {
        java {
            srcDir(generateVersionClassTask)
        }
    }
}
```

This also avoids us adding the IntelliJ extension and needing to use `afterSync`, which afaik is a performance penalty.

Also do note that this only works if the task has a configured `@OutputDirectory`. Therefore, we most likely need to create a custom Gradle task in `buildSrc/src/main/kotlin` and then configure the task here. Something like this:

```kotlin
tasks {
    // I don't know what the correct Kotlin way is to declare this task, as I only used Groovy Gradle
    val generateVersionClassTask by registering (GenerateMockitoVersionClass) {`
        mockitoVersion = project.version
        generatedVersionClassDir = layout.buildDirectory.dir("generated/sources/version/java")
    }

    sourceSets.main {
        java {
            srcDir(generateVersionClassTask)
        }
    }
}
```

Here `GenerateMockitoVersionClass` has an `@OutputDirectory` of `generatedVersionClassDir` and an `@Input` for `mockitoVersion`

TimvdLippe (changes requested):

mockito-core/build.gradle.kts
bric3: Interesting I'll take a look

mockito-core/src/main/java/org/mockito/internal/creation/bytebuddy/InlineDelegateByteBuddyMockMaker.java
bric3: I'm not sure about moving that logic here at this time, because

1. The mockability depends on the mock maker
2. If we revisit this code, I think we may want to revisit `TypeMockability` as well, e.g. instead of returning string, I wonder if we could use a list of enum values instead.

mockito-core/build.gradle.kts
bric3: Done

bric3:
Putting this back to draft due to CI matrix not happy with the changes.

bric3:
Superseded by #3535, because GHA workflow is not triggered anymore on this PR.

Sathya-ac9:
Hi,

I am facing an issue while trying to mock or spy on ZoomSDK in an Android project using Mockito. The following error occurs:
`org.mockito.exceptions.base.MockitoException:
Cannot mock/spy class us.zoom.sdk.ZoomSDK
Mockito cannot mock/spy because :
final class
at org.mockito.internal.runners.DefaultInternalRunner$1$1.evaluate(DefaultInternalRunner.java:50)
at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
at org.mockito.internal.runners.DefaultInternalRunner$1.run(DefaultInternalRunner.java:100)
at org.mockito.internal.runners.DefaultInternalRunner.run(DefaultInternalRunner.java:107)
at org.mockito.internal.runners.StrictRunner.run(StrictRunner.java:41)
at org.mockito.junit.MockitoJUnitRunner.run(MockitoJUnitRunner.java:163)
at org.junit.runners.Suite.runChild(Suite.java:128)
at org.junit.runners.Suite.runChild(Suite.java:27)
at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
at org.junit.runner.JUnitCore.run(JUnitCore.java:115)
at androidx.test.internal.runner.TestExecutor.execute(TestExecutor.java:68)
at androidx.test.internal.runner.TestExecutor.execute(TestExecutor.java:59)
at androidx.test.runner.AndroidJUnitRunner.onStart(AndroidJUnitRunner.java:463)
at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:2415)`

The error indicates that ZoomSDK is a final class, and Mockito cannot mock or spy on final classes by default.

**Seeking Help/Suggestions:**

- Is there a workaround for mocking or spying on final classes in Mockito?
- Should I consider using a different approach or library for testing with ZoomSDK?
- Any guidance on how to resolve this issue would be greatly appreciated.

Thanks in advance!

Best Regards,
Sathya

bric3:
@Sathya-ac9 Please open a new issue.

Issue #3474, opened 2024-10-15
Revise root build.gradle


bric3:
Root `build.gradle` needs to be updated to latest best practices. This is a follow-up of #3446

* Some plugins are declared the [old legacy way](https://docs.gradle.org/current/userguide/plugins.html#sec:old_plugin_application)

   ```groovy
   buildscript {
       repositories {
           ...
       }
       dependencies {
           classpath "..."
       }
   }

   apply plugin: '...' // actually happening in an applied script.
   ```

   This should now use the `plugins` declaration, and can eventually be improved to use convention plugins rather than scripts.

* Scripts are still in the `./gradle` folder which is not standard.
* Nexus publication and shipkit can be applied together as a convention plugin.
* Build script declaration don't seem to be picked up by dependabot or renavabot.

**Todo**

* [x] #3532
* [x] #3533

PR #3533, opened 2024-12-09
Makes shipkit / nexusPublish a convention plugin for root

bric3:
Sibling of #3532 

Following on the trend to follow best practice, the `shipkit.gradle` script is now a convention plugin for the root project.

One more step to help on https://github.com/mockito/mockito/issues/3474

## Checklist


buildSrc/src/main/kotlin/mockito.root.releasing-conventions.gradle.kts
```diff
@@ -0,0 +1,73 @@
+import gradle.kotlin.dsl.accessors._edf0c2271024fd9311959755ab251c7b.ext
```
TimvdLippe, line 1: Nit: is this import intentional? It seems unused.

gradle/libs.versions.toml
```diff
@@ -11,6 +11,9 @@ gradleplugin-animal-sniffer = "1.7.2"
 gradleplugin-aQute-bnd = "7.0.0"
 gradleplugin-errorprone = "4.1.0"
 gradleplugin-spotless = "6.25.0"
+gradleplugin-nexusPublish = "2.0.0-rc-1"
+gradleplugin-shipkit-changelog = "2.0.1"
+gradleplugin-shipkit-autoVersion = "2.1.0"
```
TimvdLippe, line 16: Nit: since there is only 1 plugin that uses this version, I think it is duplicate to put it here as well. Therefore, if there is only one artifact with a particular version reference, I would prefer to use `version` instead of `version.ref`

buildSrc/src/main/kotlin/mockito.root.releasing-conventions.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+    id("org.shipkit.shipkit-github-release")
+
+    id("io.github.gradle-nexus.publish-plugin")
+}
+
+val isSnapshot = version.toString().endsWith("SNAPSHOT")
+tasks {
+    generateChangelog {
+        githubToken = providers.environmentVariable("GITHUB_TOKEN").get()
+        previousRevision = project.ext["shipkit-auto-version.previous-tag"].toString()
```
TimvdLippe, line 15: In a follow-up PR, let's refactor this so that we also use an environment variable for this.

buildSrc/src/main/kotlin/mockito.root.releasing-conventions.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+        repository = "mockito/mockito"
+    }
+
+    githubRelease {
+        enabled = !isSnapshot
+        dependsOn(generateChangelog)
+        githubToken = providers.environmentVariable("GITHUB_TOKEN").get()
+        newTagRevision = providers.environmentVariable("GITHUB_SHA").get()
+        repository = generateChangelog.get().repository
```
TimvdLippe, line 24: WDYT about defining constants values instead of referencing separate tasks here. That removes the need for eager configuration

TimvdLippe (changes requested):
Some nits, the rest LGTM

buildSrc/src/main/kotlin/mockito.root.releasing-conventions.gradle.kts
bric3: Indeed that came to mind. I avoided this here to make a 1:1 translation. But this would be better in my opinion.

PR #3551, opened 2024-12-16
Avoid warning when dynamic attach is enabled

raphw:
Checks the command line arguments for the `EnableDynamicAgentLoading` argument and does not warn about dynamic attach, if the option is set.


mockito-core/src/main/java/org/mockito/internal/PremainAttachAccess.java
```diff
@@ -34,16 +40,41 @@ public static Instrumentation getInstrumentation() {
...
+                                            .getClass()
+                                            .getMethod("getInputArguments")
+                                            .invoke(runtimeMXBean);
+                    dynamicAgentLoading =
+                            arguments.stream()
+                                    .anyMatch(
+                                            argument ->
+                                                    argument.contains(
+                                                            "-XX:+EnableDynamicAgentLoading"));
+                } catch (Exception ignored) {
```
TimvdLippe, line 62: Can we update the tests we have for this feature to ensure this as well? If I remember correctly we had those tests in the `.gradle` files somewhere

mockito-core/src/main/java/org/mockito/internal/PremainAttachAccess.java
```diff
@@ -34,16 +40,41 @@ public static Instrumentation getInstrumentation() {
...
-                // it is later calling this method to access the instrumentation.
-                System.err.println(
-                        "Mockito is currently self-attaching to enable the inline-mock-maker. This "
-                                + "will no longer work in future releases of the JDK. Please add Mockito as an agent to your "
-                                + "build as described in Mockito's documentation: "
-                                + "https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#0.3");
+                boolean dynamicAgentLoading;
+                try {
+                    Object runtimeMXBean =
+                            Class.forName("java.lang.management.ManagementFactory")
```
TimvdLippe, line 46: Why do we need to use reflection here?

raphw: Android does not know these classes and would fail if Mockito had a hard link to them. Now the invocation will just fail.

mockito-core/src/main/java/org/mockito/internal/PremainAttachAccess.java
raphw: Fair point, I tried to add a test!

bric3 (approved):

TimvdLippe (approved):
Nice!

schabe77:
This doesn't work for me. That's because 
https://github.com/mockito/mockito/blob/e04dbbe3c83badb0d7d6d9cd700f456e227a9bb3/mockito-core/src/main/java/org/mockito/internal/PremainAttachAccess.java#L55

throws a 
`java.lang.IllegalAccessException: class org.mockito.internal.PremainAttachAccess cannot access class sun.management.RuntimeImpl (in module java.management) because module java.management does not export sun.management to unnamed module @78b66d36`

But it works, when getInputArguments is requested without Reflection:
```
        RuntimeMXBean runtimeMXBean = (RuntimeMXBean)
                Class.forName("java.lang.management.ManagementFactory")
                        .getMethod("getRuntimeMXBean")
                        .invoke(null);
       dynamicAgentLoading =
                runtimeMXBean.getInputArguments().stream()
                        .anyMatch(
                                argument ->
                                        argument.contains(
                                                "-XX:+EnableDynamicAgentLoading"));
```

raphw:
Indeed, this is an annoying implication of the module system if a method is resolved against an overriding class, not the interface. I will need to change that to get the interface class and to invoke the method on it.

raphw:
Suggested fix: https://github.com/mockito/mockito/blob/e04dbbe3c83badb0d7d6d9cd700f456e227a9bb3/mockito-core/src/main/java/org/mockito/internal/PremainAttachAccess.java#L50-L55

PR #3552, opened 2024-12-16
Improve error message and give suggestion on how to resolve problem when mocking a sealed interface or abstract class.

raphw:
Improve the error message of sealed types to suggest mocking a non-abstract subclass which allows to use a mock by type substitution.

### DIFF
--- a/mockito-core/src/main/java/org/mockito/internal/creation/bytebuddy/InlineDelegateByteBuddyMockMaker.java
+++ b/mockito-core/src/main/java/org/mockito/internal/creation/bytebuddy/InlineDelegateByteBuddyMockMaker.java
@@ -114,14 +114,12 @@ class InlineDelegateByteBuddyMockMaker
         Throwable initializationError = null;
 
         // ByteBuddy internally may attempt to fork a subprocess. In Java 11 and Java 19, the
-        // Java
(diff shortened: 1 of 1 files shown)


TimvdLippe (approved):
Looks good to me. Seems like we have duplicate code between the inline mockmakers. Shall we deduplicate that?

raphw:
Can you point to where? If it's the same checks, certainly, to make them behave consistent.

TimvdLippe:
Ah it was added here: https://github.com/mockito/mockito/pull/3534 Was it accidentally reverted by @bric3 later? Not sure if I am following what happened to these changes.

bric3:
This is certainly weird. The diff though looks the same.

#3534 was committed as 7c277acac on main.

Note I tried to handle that in #3533, but the CI is failing in an obscure way so it's nothing changed on that front.

bric3:
**This PR #3552**



**#3534**

raphw:
I was confused myself, but thought I just never created a PR.

TimvdLippe:
Given that https://github.com/mockito/mockito/commit/7c277acac44986450090f6540a32305be342e324 exists on `main`, I don't think we need this PR anymore. Let's revisit when it turns out we are missing something.

PR #3562, opened 2025-01-03
Add extendsFrom testImplementation to Gradle's mockitoAgent configura…

Chessray:
…tion. Without this, I encountered a build error.

## Checklist


TimvdLippe:
@bric3 I don't think this is correct, especially as it works for other users. Therefore, I assume this is a user problem, but do let me know what you think.

bric3:
Agreed, this patch looks incorrect. The mockito agent configuration should only have mockito-core because it's purpose is to be passed as a single (jar) path to the `-javaagent` flag therefore adding anything in this configuration is likely incorrect.

PR #3566, opened 2025-01-05
Refine reflection when calling management factory

raphw:
By invoking the management factory return object's class, a module boundary can be breached as the method should be invoked on the interface type which is exported.


StefanMuellerCH:
Thank you @raphw for the fix. Is it possible to make a mockito fix-version with this fix included? That would be great, thanks alot for your work!

PR #3585, opened 2025-02-10
Fixes #3584 Improve Gradle Kotlin documentation to add agent instead of overwriting the jvmArgs

### LINKED ISSUE #3584: Documentation "how to set up mockito-core as a Java agent" for Gradle Kotlin should be improved
The current version of documentation recommends to set the jvmArgs as the jvmArgs
```
tasks {
    test {
        jvmArgs("-javaagent:${mockitoAgent.asPath}")
     }
}
``` 

This is working when only Mockito agent is in use. As soon as another agent is expected (example for code coverage), this will fail as this overrides any previously registered agent. It will also remove all previously set jvmArgs.

_**To reproduce:**_
Both from IntelliJ
* Run tests requiring Mockito agent coinfigured as documented, it will succeed
* Re-run the same test with coverage, test will fail

_**Fix**_
The fix is trivial, instead of setting the jvmArgs, we need to add the entry instead.
```
tasks {
    test {
        jvmArgs.add("-javaagent:${mockitoAgent.asPath}")
     }
}
```
Tested with Grade 8.12 and Mockito 5.15.2

See https://github.com/mockito/mockito/blob/83ee69b8b927e166f7a8acf6e789a26a58f63567/mockito-core/src/main/java/org/mockito/Mockito.java#L194

tdltdl:
As per #3584, current documentation recommend to set the jvmArg in the gradle.build.kts extract. This overwrites the value of jvmArgs. This, for example, prevents to run the coverage from IntelliJ. 
With this improved version, the agent is added to the jvmArgs, leaving the existing ones untouched.

### DIFF
--- a/mockito-core/src/main/java/org/mockito/Mockito.java
+++ b/mockito-core/src/main/java/org/mockito/Mockito.java
@@ -191,7 +191,7 @@ import java.util.function.Function;
  * }
  * tasks {
  *     test {
- *         jvmArgs("-javaagent:${mockitoAgent.asPath}")
+ *         jvmArgs.add("-javaagent:${mockitoAgent.asPath}")
  *     }
  * }
  * </code></pre>


bric3 (approved):

bric3:
Thanks @tdltdl that's a good point.

bric3:
@tdltdl Note that [`jvmArgs(...)`](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.testing.Test.html#org.gradle.api.tasks.testing.Test:jvmArgs(java.lang.Object[]))[\[2\]](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.testing.Test.html#org.gradle.api.tasks.testing.Test:jvmArgs(java.lang.Iterable)) is adding the parameters not replacing it. In Kotlin DSL to replace the args one would use [`setJvmArgs(...)`](https://docs.gradle.org/current/javadoc/org/gradle/process/JavaForkOptions.html#setJvmArgs(java.util.List))[\[2\]](https://docs.gradle.org/current/javadoc/org/gradle/process/JavaForkOptions.html#setJvmArgs(java.util.List))

However maybe it's simpler to understand the idea of this configuration by explicitly mentioning the `add` method.

tdltdl:
> @tdltdl Note that [`jvmArgs(...)`](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.testing.Test.html#org.gradle.api.tasks.testing.Test:jvmArgs(java.lang.Object%5B%5D))[[2]](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.testing.Test.html#org.gradle.api.tasks.testing.Test:jvmArgs(java.lang.Iterable)) is adding the parameters not replacing it. In Kotlin DSL to replace the args one would use [`setJvmArgs(...)`](https://docs.gradle.org/current/javadoc/org/gradle/process/JavaForkOptions.html#setJvmArgs(java.util.List))[[2]](https://docs.gradle.org/current/javadoc/org/gradle/process/JavaForkOptions.html#setJvmArgs(java.util.List))
> 
> However, maybe it's simpler for everyone to understand the idea of this configuration (adding a the `-javaagent` flag) by explicitly mentioning the `add` method there.

Without explicit add, the code coverage was not working anymore on IntelliJ. Maybe the issue is at Gradle? Also the groovy version is jvmArgs += "-javaagent:${configurations.mockitoAgent.asPath}" (note the plus sign sign)

Issue #2826, opened 2022-12-18
Inconsistent behaviour of @InjectMocks annotation


el-hopaness-romtic:
Consider classes A and B
```
public class A {
    String call() {
        return null;
    }
}
```

```
public class B extends A {
}
```

And class C, contructor of which uses inctances of both A and B
```
public class C {
    private final A a;
    private final B b;

    C(A a, B b) {
        this.a = a;
        this.b = b;
    }

    public void callFields() {
        System.out.println("A " + a.call());
        System.out.println("B " + b.call());
    }
}
```

In that case following test does not behave consistently
```
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.extension.ExtendWith;
import org.mockito.InjectMocks;
import org.mockito.Mock;
import org.mockito.junit.jupiter.MockitoExtension;

import static org.mockito.Mockito.when;

@ExtendWith(MockitoExtension.class)
class MockitoTest {
    @Mock
    A a;
    @Mock
    B b;
    @InjectMocks
    C c;

    @Test
    void test() {
        when(a.call()).thenReturn("A");
        when(b.call()).thenReturn("B");
        c.callFields();
    }
}

```
![image](https://user-images.githubusercontent.com/60661917/208298558-691d0052-e710-444b-8f88-1a8e3de483be.png)
![image](https://user-images.githubusercontent.com/60661917/208298575-347af0e6-d7c8-4d3c-9b34-c4b8c444d9b1.png)

org.junit.jupiter:junit-jupiter:5.9.1
org.mockito:mockito-junit-jupiter:4.8.1
openjdk version "17.0.3" 2022-04-19 LTS

HermitDede:
This issue caused me at least 1-2 hours to figure out. There was no clear explanation or warning for the root cause. It was behaving differently on each context. I came here to post this issue, but looks like this has been ongoing for 1 year 😞

PR #3595, opened 2025-02-22
Fixes #2807: enhance constructor injection

CQ-Developer:
## changelist
- store mock object's type info (in form of java.lang.reflect.Field)
- pick the most suitable candidate by `ConstructorArgumentVoter`
Assume `Aoo` is mock object, `Boo` is Constructor parameter.
1. if Aoo and Boo have same type, Aoo will get 3 points
2. if Aoo and Boo have same generic type, Aoo will get 3 points.
3. ...
the `ConstructorArgumentVoter` will pick the candidate who got the most votes as injection object.

### DIFF
--- a/mockito-core/src/main/java/org/mockito/internal/configuration/DefaultInjectionEngine.java
+++ b/mockito-core/src/main/java/org/mockito/internal/configuration/DefaultInjectionEngine.java
@@ -5,6 +5,7 @@
 package org.mockito.internal.configuration;
 
 import java.lang.reflect.Field;
+import java.util.Map;
 import java.util.Set;
 
 import org.mockito.internal.configuration.injection.MockInjection;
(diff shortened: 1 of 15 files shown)


CQ-Developer:
@ChristianSchwarz @TimvdLippe please help to review this PR, I  test it in my code, and it's working fine, thanks a lot.

TimvdLippe:
Your PR does not mention a corresponding issue where we discuss the merits. Please open an issue first where we can discuss the use case.

Note that injection is typically a topic we take longer to decide what to do, given that Mockito is not an injection framework.

pascalgn:
The (new) logic in ConstructorInjection looks very similar to TypeBasedCandidateFilter, should it be consolidated?

el-hopaness-romtic:
Hello @CQ-Developer, does it fix https://github.com/mockito/mockito/issues/2826 by any chance? Also, if it helps you, i have noticed if i use setter or field injection in my case, it works fine, it's only constructor injection that is faulty

CQ-Developer:
> Hello @CQ-Developer, does it fix #2826 by any chance? Also, if it helps you, i have noticed if i use setter or field injection in my case, it works fine, it's only constructor injection that is faulty

yes, field injection is fine. this pr will fix the constructor injection. i believe this will fix #2826 .

CQ-Developer:
I found that you may remove `@InjectMocks` in the future according to #1518.
So...

PR #3597, opened 2025-02-23
Add support for including module-info in Mockito.

### LINKED ISSUE #2604: Consider adding `module-info.class`

### LINKED ISSUE #2952: Issue with mockito with jdk 19

raphw:
Adds `module-info.java` to describe module descriptor.

Fixes #2604
Fixes #2952
Closes #3381


raphw:
I understand why this is failing now.

1. We add the `MockMethodDispatcher` class to the `org.mockito.internal.creation.bytebuddy.inject` package.
2. We compile the module-info.java where the compiler adds all packages in scope to the module's package list.
3. Even though we remove the file later and rename it to raw, the package is still in the package list.
4. The class loader never attempts to check another class loader (the boot loader) for this package. It is never found.

Could we "post-process" the module-info.class after the javac application? We could use ASM to apply a module visitor on it. All we would need to do would be to remove the package in question from the module info file.

raphw:
This would be the code we need to run in the post-processing:

        ClassReader reader = new ClassReader("module-info.class");
        ClassWriter writer = new ClassWriter(reader, 0);
        reader.accept(new ClassVisitor(Opcodes.ASM9, writer) {
            @Override
            public ModuleVisitor visitModule(String name, int access, String version) {
                return new ModuleVisitor(Opcodes.ASM9, super.visitModule(name, access, version)) {
                    @Override
                    public void visitPackage(String packaze) {
                        if (!packaze.equals("org/mockito/internal/creation/bytebuddy/inject")) {
                            super.visitPackage(packaze);
                        }
                    }
                };
            }
        }, 0);

raphw:
@TimvdLippe Any idea how to effectivate this code? https://github.com/mockito/mockito/pull/3597/files

The Class writer returns a byte array that would need to replace the original file.

mockito-core/build.gradle.kts
```diff
@@ -73,6 +78,24 @@ tasks {
         into(generatedInlineResource.map { it.dir("org/mockito/internal/creation/bytebuddy/inject") })
 
         rename("(.+)\\.class", "$1.raw")
+
+        /*TODO: How to do this?
```
TimvdLippe, line 82: If I understand correctly, you are curious how to hook this up with Gradle? If that's the case, then you should wrap this in a `doLast` block similar to what we do in `createTestResources`. In that case, the code doesn't run during configuration, but it runs in the execution phase of Gradle.

raphw: Ok, I gave this several attempts, but I have still not become friends with Gradle. How do I get hold if the `module-info.class` file in `doLast`? All I want, is to get hold of that file, and transform it.

raphw: I managed, but there is another obstacle. There must not be a jar entry for a folder in *inject*, as this triggers the class loader to reserve the package name and avoid the boot loader. I thought I adjusted this, by moving up the raw class file one level, but now I am stuck again as it fails me on some obscure jar bundling error. If any Gradle user knows how to resolve this, I'd be thankful for the hint. Otherwise, I will try again next week.

raphw:
@TimvdLippe or @bric3 - I currently try to move the raw class file into a different folder (and therewith try to avoid that the class loader registers the package that this folder represents, so it can be loaded from the boot loader).

But this does not seem to work. I adjusted the copy task:


        from(sourceSets.main.flatMap { it.java.classesDirectory }
            .map { it.file("org/mockito/internal/creation/bytebuddy/inject/MockMethodDispatcher.class") })
        into(generatedInlineResource.map { it.dir("org/mockito/internal/creation/bytebuddy") })

        rename(".*", "inject_MockMethodDispatcher.raw")

Why does this not work? For me, the *jar* task fails now with:

> Task :mockito-core:jar FAILED
error  : Classes found in the wrong directory: {org/mockito/internal/creation/bytebuddy/MockMethodDispatcher.class=org.mockito.internal.creation.bytebuddy.inject.MockMethodDispatcher}

TimvdLippe:
You probably fixed the Jar task, as in the Gradle build scan (https://scans.gradle.com/s/x55h4q2hpkgeq/timeline?type=org.gradle.api.tasks.bundling.Jar#76bq4yvmlsst2) it is successful. Looking at the build failure, it seems to be an OSGi issue:

```
> Task :mockito-extensions:mockito-junit-jupiter:verifyOSGi FAILED
      ⇒ Bundle: org.mockito.junit-jupiter cannot be resolved
              ⇒ Import-Package: org.mockito.internal.configuration; version=[5.15.0,6.0.0); status=INTERNAL cannot be resolved
                      ⇒ because Import-Package requirement for: org.mockito.internal.creation.bytebuddy.inject could not be provided by any available bundle or dependency
```

We probably need to update our OSGi instead to ignore this error? A possible solution I found after googling this error: https://github.com/bndtools/bnd/issues/2227#issuecomment-491310275

```
tasks.jar {
    withConvention(BundleTaskConvention::class) {
        bnd(mapOf(
                "-fixupmessages" to "^Classes found in the wrong directory: \\\\{META-INF/versions/9/module-info\\\\.class=module-info}$"
        ))
    }
}
```

raphw:
Indeed. I had to clean out my local build to get there.

I think I figured it out. Needed to fix some assumptions, but runs locally now. I think this does the trick altogether.

TimvdLippe:
Nice work! Glad to see we were able to figure it out together.

I do want us to see if we can find an alternative fix for our Javadoc though. I really do want us to omit our internal classes in our Javadoc as it is too easy to read over. With the fixes we made in Gradle, maybe we can add the line back and it "just" works?

raphw:
I found a workaround where I define the sources explicily. It's a Gradle limitation as the missing packages still need to be included in the source task.

TimvdLippe (approved):
Happy to see this all work and our Javadoc as well 🎉 I will publish a new minor release after this gets in

sormuras:
Nice 🧩 work!

Issue #3605, opened 2025-03-07
Inline mocking configuration snippet for Gradle should allow task relocatability


jprinet:
The Mockito documentation snippet to configure inline mocking references the java agent with an absolute path in the JVM arguments ([there](https://github.com/mockito/mockito/blob/c81be5deedfd966bd17c9619769a34c74d9779e6/mockito-core/src/main/java/org/mockito/Mockito.java#L194)).

This actually breaks relocatability as described [in the Gradle documentation](https://docs.gradle.org/current/userguide/build_cache_concepts.html#relocatability) in the Gradle documentation.

[The recommended approach](https://docs.gradle.org/current/userguide/caching_java_projects.html#dealing_with_file_paths) is to use a `CommandLineArgumentProvider`

PR #3606, opened 2025-03-07
Adjust inline mocking snippet to allow task relocatability

jprinet:
## Issue
Inline mocking configuration snippet for Gradle should allow task relocatability

## Fix
Use a `CommandLineArgumentProvider`

Fixes https://github.com/mockito/mockito/issues/3605

## Checklist

### DIFF
--- a/mockito-core/src/main/java/org/mockito/Mockito.java
+++ b/mockito-core/src/main/java/org/mockito/Mockito.java
@@ -181,7 +181,7 @@ import java.util.function.Function;
  * <p>
  * To explicitly attach Mockito during test execution, the library's jar file needs to be specified as <code>-javaagent</code>
  * as an argument to the executing JVM. To enable this in Gradle, the following example adds Mockito to all test
- * tasks using <strong>Kotlin DSL</strong>:
+ * tasks using <strong>Kotlin DSL</strong>. Using a <code>CommandLineArgumentProvider</code> ensures task relocatability as recommended <a href="https://docs.gradle.org/current/userguide/caching_java_projects.html#dealing_with_file_paths">there</a>:
  *
  * <pre class="code"><code class="kotlin">
  * val mockitoAgent = configurations.create("mockitoAgent")
@@ -191,8 +191,18 @@ import java.util.function.Function;
  * }
  * tasks {
  *     test {
- *         jvmArgs.add("-javaagent:${mockitoAgent.asPath}")
+ *         jvmArgumentProviders.add(
+ *             objects.newInstance&lt;JavaAgentArgumentProvider&gt;().apply {
+ *                 classpath.from(mockitoAgent)
+ *             }
+ *         )
  *     }
+ * }
+ * abstract class JavaAgentArgumentProvider : CommandLineArgumentProvider {
+ *     {@literal @}get:Classpath
+ *     abstract val classpath: ConfigurableFileCollection
+ *
+ *     override fun asArguments() = listOf("-javaagent:${classpath.singleFile.absolutePath}")
  * }
  * </code></pre>
  *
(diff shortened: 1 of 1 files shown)


jprinet:
Thanks @marcphilipp for noticing 🙏

mockito-core/src/main/java/org/mockito/Mockito.java
```diff
@@ -191,9 +191,19 @@
...
  *     test {
- *         jvmArgs.add("-javaagent:${mockitoAgent.asPath}")
+ *         jvmArgumentProviders.add(
+ *             objects.newInstance&lt;JavaAgentArgumentProvider&gt;().apply {
+ *                 classpath.from(mockitoAgent)
+ *             }
+ *         )
  *     }
  * }
+ * abstract class JavaAgentArgumentProvider : CommandLineArgumentProvider {
```
TimvdLippe, line 201: Can this be part of Gradle? Mockito isn't the only Java agent out there. We prefer not to ship too much custom build code (if we can) and also not force on our users.

jprinet: Many Java arguments could be referenced with an absolute path and the approach is to provide a generic mechanism to handle that.
There is no plan to have an equivalent of the `JavaAgentArgumentProvider` part of Gradle as far as I know

TimvdLippe: I wonder if instead of elaborate usage instructions (this quite a few lines just to get started with Mockito), we should promote something like https://github.com/mockito/mockito/discussions/3603 instead

TimvdLippe: @bric3 With this PR, all Gradle users would need to add 20 lines of build code just to get started with Mockito. Compared to 1 line that it was before, I don't think that is a reasonable increase. Instead, I would like us to streamline the installation with a Gradle plugin:

```groovy
plugins {
  id 'org.mockito.agent-configuration'
}
```

Or even native Gradle support along the lines of:

```groovy
val mockitoAgent = configurations.create("mockitoAgent")
dependencies {
    testImplementation(libs.mockito)
    mockitoAgent(libs.mockito) { isTransitive = false }
}
javaAgents {
    libs.mockito
}
```

jprinet: I understand that you want to make the configuration snippet minimal, a plugin would definitely work and is the recommended approach for this (note that the plugin linked as an example is not using a `CommandLineArgumentProvider` either).
Another option would be a comment on the snippet mentioning that the setup is not optimal for performances and link to the relevant documentation

TimvdLippe: As per https://github.com/mockito/mockito/discussions/3603#discussioncomment-12479344 I don't think we should add the extra build setup code as suggested in this PR. Jerome responded in that discussion thread that we can also point to the Gradle docs reference. I am okay with that, but I also know that folks rarely read such documentation and instead copy-and-paste the code as-is.

Therefore, if it turns out that a simple `-javaagent` argument breaks caching for Gradle, I think it is in Gradle's best interest to provide a fluent API to make such pitfalls smaller. For now, we can ship with the link as added in this PR and if there is a build performance problem with Mockito, I will redirect folks to Gradle for further guidance.

TLDR: Please keep the link to the Gradle documentation, but remove all other build logic changes.

mockito-core/src/main/java/org/mockito/Mockito.java
```diff
@@ -203,6 +203,7 @@
  *     mockitoAgent
  * }
  * dependencies {
+ *     testImplementation(libs.mockito)
```
jprinet, line 206: This dependency was apparently missing

mockito-core/src/main/java/org/mockito/Mockito.java
jprinet: I updated the PR accordingly 👍

TimvdLippe:
@jprinet If you would like to keep track of how often this build code is copied, here's a GitHub search link: https://github.com/search?q=%22val+mockitoAgent+%3D+configurations.create%28%5C%22mockitoAgent%5C%22%29%22&type=code

mockito-core/src/main/java/org/mockito/Mockito.java
bric3: Sorry for the late reply, but I totally align with @TimvdLippe there. Supporting passing dependencies to `-javaagent` JVM flag is definitely in the interest any build tool. I don't know about Gradle plans, but maybe your example @jprinet could land in the Gradle cookbook (cc @oleg-nenashev).

PR #3610, opened 2025-03-13
Remove Arrays.asList from critical stubbing path in GenericMetadataSu…

adrianroos:
…pport

On Android, the implementation of java.util.Arrays.ArrayList.toArray has a non-trivial implementation that ends up calling into android.os.Process.myUid().

Consequently, attempting to stub static methods in android.os.Process with doReturn will end up calling into myUid() before the stubbing completes, triggering an infinite recursion.

To work around that, replace the addAll(asList(...)) call with Collections.addAll. As an added benefit this avoids an array copy.

This upstreams https://r.android.com/3500336.

## Checklist


adrianroos:
Thank you for the review!

I considered adding a test for this, but reproducing the issue requires both static mocking and running on the Android runtime; the Android platform tests use [ExtendedMockito](https://moltmann.github.io/dexmaker/com/android/dx/mockito/inline/extended/ExtendedMockito.html) to achieve that, but the latest version of that doesn't work with Mockito 5.

Issue #3690, opened 2025-07-20
Mockito fails to mock method with generic array return type due to raw type extraction of generic parameter S[]


g000ner:
Hi team!
I have abstract class `DaService` which has 2 generic parameters and implemented method `public S doSmth(T o, Class<S> clazz)` :

```
public abstract class DaService<T, S> {
    public S doSmth(T o, Class<S> clazz) {
        try {
            return clazz.newInstance();
        } catch (InstantiationException | IllegalAccessException e) {
            throw new RuntimeException(e);
        }
    }
}
```

Class `HttpDaService<T, S>` extends `DaService` :

```
public abstract class HttpDaService<T, S> extends DaService<T, S[]> {
}
```

And finally `HttpDaServiceImpl` which extends  `HttpDaService` :

```
public class HttpDaServiceImpl extends HttpDaService<Object, Object>{
}
```

When I try to mock object of `HttpDaServiceImpl` and use `when().thenReturn()` on method `doSmth()` I have an `MockitoException`:

```
@ExtendWith(MockitoExtension.class)
public class HttpDaServiceImplTest {
    @Mock
    HttpDaServiceImpl daService;

    @Test
    void when_doSmth_thenReturnObjectsArray() {
        when(daService.doSmth(any(), any())).thenReturn(new Object[4]);
        Object[] objects = daService.doSmth(new Object(), Object[].class);
        assertEquals(4, objects.length);
    }
}
```

```
org.mockito.exceptions.base.MockitoException: Raw extraction not supported for : 'S[]'

	at org.example.HttpDaServiceImplTest.when_doSmth_thenReturnObjectsArray(HttpDaServiceImplTest.java:19)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

It seems the problem with generic parameter `S[]` in class `HttpDaService`. I found that the call to `resolveGenericReturnType()` in `org.mockito.internal.stubbing.answers.InvocationInfo` for method `doSmth(T o, Class<S> clazz)` returns a `TypeVariableImpl` instead of the expected `GenericArrayTypeImpl` :

```
public boolean isVoid() {
        final MockCreationSettings mockSettings =
                MockUtil.getMockHandler(invocation.getMock()).getMockSettings();
        Class<?> returnType =
[...]

PR #3699, opened 2025-08-05
Test: reproduce issue with mocking generic array return type (S[])

michelleSOS-L:
This PR adds a failing test to reproduce issue #3690.

The issue occurs when attempting to mock a method that returns a generic array type like `S[]`. Due to how generics are resolved, Mockito throws a `MockitoException: Raw extraction not supported for : 'S[]'`.

### Test Case
The test adds a class hierarchy:
- `DaService<T, S>` with method `S doSmth(T, Class<S>)`
- `HttpDaService<T, S>` extends `DaService<T, S[]>`
- `HttpDaServiceImpl` extends `HttpDaService<Object, Object>`

The test attempts:
```java
when(daService.doSmth(any(), any())).thenReturn(new Object[4]);

### DIFF
--- a/mockito-core/src/test/java/org/mockito/bugs/GenericArrayMockingTest.java
+++ b/mockito-core/src/test/java/org/mockito/bugs/GenericArrayMockingTest.java
@@ -1,0 +1,40 @@
+package org.mockito.bugs;
+
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+import static org.mockito.Mockito.*;
+import static org.junit.Assert.*;
+
+public class GenericArrayMockingTest {
+
+    public abstract static class DaService<T, S> {
+        public S doSmth(T o, Class<S> clazz) {
+            try {
+                return clazz.newInstance();
+            } catch (InstantiationException | IllegalAccessException e) {
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
+    public abstract static class HttpDaService<T, S> extends DaService<T, S[]> {}
+
+    public static class HttpDaServiceImpl extends HttpDaService<Object, Object> {}
+
+    @Mock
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by TimvdLippe** [comment]
The test looks good. To merge this, can you also attempt to fix the issue?

**Comment by michelleSOS-L** [comment]
> The test looks good. To merge this, can you also attempt to fix the issue?

isVoid() no longer resolves generic return types (which failed for GenericArrayType like S[]). It now uses Method#getReturnType() to determine whether the method is void. This is sufficient for the check and avoids the crash path.”

**Comment by TimvdLippe** [comment]
```
> Task :mockito-extensions:mockito-errorprone:checkstyleTest
Error: eckstyle] [ERROR] /home/runner/work/mockito/mockito/mockito-core/src/main/java/org/mockito/internal/stubbing/answers/InvocationInfo.java:14:8: Unused import - org.mockito.internal.util.MockUtil. [UnusedImports]

Error: eckstyle] [ERROR] /home/runner/work/mockito/mockito/mockito-core/src/main/java/org/mockito/internal/stubbing/answers/InvocationInfo.java:16:8: Unused import - org.mockito.internal.util.reflection.GenericMetadataSupport. [UnusedImports]
Error: eckstyle] [ERROR] /home/runner/work/mockito/mockito/mockito-core/src/main/java/org/mockito/internal/stubbing/answers/InvocationInfo.java:18:8: Unused import - org.mockito.mock.MockCreationSettings. [UnusedImports]
```

Please run `./gradlew build` locally to make sure all checks pass.

File: GenericArrayMockingTest.java [hunk 1]
```diff
@@ -0,0 +1,40 @@
+package org.mockito.bugs;
```
TimvdLippe, line 1: I think this file is here by mistake? The other test is already sufficient

PR #3701, opened 2025-08-06
Raw type used on MockCreationListener changed to generic type

### LINKED ISSUE #3700: MockCreationListener uses raw type for generic class MockCreationSettings
When creating an implementation of MockCreationListener, the following warning is shown upon compile:

```
warning: [rawtypes] found raw type: MockCreationSettings
                                  final MockCreationSettings settings) {
                                        ^
  missing type arguments for generic class MockCreationSettings<T>
  where T is a type-variable:
    T extends Object declared in interface MockCreationSettings
1 warning
```

On the `MockCreationListener` interface, the MockCreationSettings does not specify type, nor use a generic type.
The issue can be replicated by adding the following class to any project and extend any JUnit5 test using `@ExtendWith` with the class.
```
public class MyMockExtension implements BeforeAllCallback, AfterEachCallback, AfterAllCallback {

     private static final List<Object> ALL_MOCKS = new ArrayList<>();
     private static final MockitoListener MOCK_CREATION_LISTENER = new MockCreationListenerImpl();

    @Override
    public void afterAll(final ExtensionContext context) {
        Mockito.framework().removeListener(MOCK_CREATION_LISTENER);
    }

    @Override
    public void beforeAll(final ExtensionContext context) {
        Mockito.framework().addListener(MOCK_CREATION_LISTENER);
    }

    private static class MockCreationListenerImpl implements MockCreationListener {

        @Override
        public void onMockCreated(final Object mock,
                                  final MockCreationSettings settings) {
            ALL_MOCKS.add(mock);
        }

        @Override
        public void onStaticMockCreated(final Class<?> mock,
                                        final MockCreationSettings settings) {
            ALL_STATIC_MOCKS.add(mock);
        }

    }
}
```

**Issue found in version 5.7.0 but still exists in version 5.18.0**

Arsarneq:
Fixes #3700, so external projects implementing `MockCreationListener` does no longer get a raw types warning they do not have control over, by using generic type T.
I did see more classes using raw type when using `MockCreationSettings` class, but this will at least remove the warning from projects using the `MockCreationListener`.

When using the XXX implementation should now use generic type, instead of raw type:

```
public class NoOpTestListener<T> implements MockitoTestListener<T> {

    @Override
    public void testFinished(TestFinishedEvent event) {}

    @Override
    public void onMockCreated(Object mock, MockCreationSettings<T> settings) {}
}
```
Instead of 
```
public class NoOpTestListener implements MockitoTestListener {

    @Override
    public void testFinished(TestFinishedEvent event) {}

    @Override
    public void onMockCreated(Object mock, MockCreationSettings settings) {}
}
```

## Checklist

### DIFF
--- a/mockito-core/src/main/java/org/mockito/internal/junit/MismatchReportingTestListener.java
+++ b/mockito-core/src/main/java/org/mockito/internal/junit/MismatchReportingTestListener.java
@@ -14,7 +14,7 @@ import org.mockito.plugins.MockitoLogger;
 /**
  * Reports stubbing argument mismatches to the supplied logger
  */
-public class MismatchReportingTestListener implements MockitoTestListener {
+public class MismatchReportingTestListener<T> implements MockitoTestListener<T> {
 
     private final MockitoLogger logger;
     private List<Object> mocks = new LinkedList<>();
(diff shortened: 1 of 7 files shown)


TimvdLippe:
While I do agree we should do this, unfortunately this is a breaking change. Therefore, we would be able to merge this once we get to Mockito 6, although we don't currently have plans to publish a new major version.

Issue #3563, opened 2025-01-03
Reproducible Build: need to inject JDK distribution details to rebuild


hboutemy:
rebuilding Mockito releases to check thatbuild can really be reproduced bit for bit: https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/mockito/mockito-core/README.md

for release 5.15.2, the only issue I get is in 3 .jar (not all):
```
├── META-INF/MANIFEST.MF
│ @@ -1,13 +1,13 @@
│  Manifest-Version: 1.0
│  Bundle-ManifestVersion: 2
│  Bundle-Name: mockito-errorprone
│  Bundle-SymbolicName: mockito-errorprone
│  Bundle-Version: 5.15.2
│ -Created-By: 21.0.5 (Azul Systems, Inc.)
│ +Created-By: 21.0.5 (Eclipse Adoptium)
│  Import-Package: com.google.errorprone;version="[2.23,3)",com.google.er
│   rorprone.bugpatterns;version="[2.23,3)",com.google.errorprone.fixes;v
```

which means the build is in theory reproducible, but in practice for these 3 jars I need to use EXACTLY `21.0.5 (Azul Systems, Inc.)`

questions:
- can I inject a build property to override the value automatically detected from effective JDK details?
- do you have another strategy for rebuilders?

hboutemy:
notice: I tried injecting `-Pmanifest.createdBy='21.0.5 (Azul Systems, Inc.)'` as done for Junit https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/junit/junit5/README.md
but Mockito build did not take the value...

TimvdLippe:
The relevant implementation in JUnit: https://github.com/junit-team/junit5/commit/2a4326b572b03e189ed93e4a6b109eb32af4429a#diff-b6c0ec6facb84213a6f79b874524c6093403e575490d547a04ac8e6102d33cba

```
"Created-By" to (buildParameters.manifest.createdBy.orNull
    ?: "${System.getProperty("java.version")} (${System.getProperty("java.vendor")} ${System.getProperty("java.vm.version")})"),
```

[...]

PR #3704, opened 2025-08-11
build: Allow overriding 'Created-By' for reproducible builds

JuyeopJang:
### Summary

This pull request resolves issue #3563 by allowing the `Created-By` entry in the JAR manifest to be overridden via a Gradle property. This is a crucial step towards enabling fully reproducible builds for Mockito.

### Key Changes

- Modified `mockito.java-library-conventions.gradle.kts` to check for a `-Pmanifest.createdBy` Gradle property.
- If the property is present, its value is used for the `Created-By` manifest attribute.
- If not, the build defaults to the existing behavior, using the current JVM's version and vendor.

### Manual Verification

As this change affects the build process itself, I have manually verified the two primary scenarios.

The verification focuses on the `mockito-core` module as a representative sample. Since the change was made in the `mockito.java-library-conventions` plugin using `tasks.withType<Jar>()`, this logic is guaranteed by Gradle to be applied to all modules that produce a JAR file. Therefore, a successful test on one core module is sufficient to confirm the correctness of the implementation for the entire project.

An automated test was not added at this stage, but I'm happy to add one if the maintainers deem it necessary.

---

**Scenario 1: Default Behavior (No Property)**

Verifies that the manifest falls back to the default JVM information when no property is provided.

- **Command:**
  ```bash
  ./gradlew clean :mockito-core:jar && unzip -p mockito-core/build/libs/mockito-core-*.jar META-INF/MANIFEST.MF

- **Output:**
  ```text
  Manifest-Version: 1.0
  Bundle-License: MIT
  Bundle-ManifestVersion: 2
  Bundle-Name: Mockito Mock Library for Java. Core bundle requires Byte 
   Buddy and Objenesis.
  Bundle-SymbolicName: org.mockito.mockito-core
  Bundle-Version: 5.18.1.SNAPSHOT
  Can-Retransform-Classes: true
  Created-By: 21.0.7 (Microsoft)
  ...

**Scenario 2: Overridden Behavior (Property Injected)**

Verifies that the manifest uses the value from the Gradle property when provided.

- **Command:**
  ```bash
  ./gradlew clean :mockito-core:jar -Pmanifest.createdBy="21.0.5 (Azul Systems, Inc.)" && unzip -p mockito-core/build/libs/mockito-core-*.jar META-INF/MANIFEST.MF

- **Output:**
  ```text
  Manifest-Version: 1.0
  Bundle-License: MIT
  Bundle-ManifestVersion: 2
  Bundle-Name: Mockito Mock Library for Java. Core bundle requires Byte 
   Buddy and Objenesis.
  Bundle-SymbolicName: org.mockito.mockito-core
  Bundle-Version: 5.18.1.SNAPSHOT
  Can-Retransform-Classes: true
  Created-By: 21.0.5 (Azul Systems, Inc.)
  ...

### DIFF
--- a/buildSrc/src/main/kotlin/mockito.java-library-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/mockito.java-library-conventions.gradle.kts
@@ -33,4 +33,14 @@ tasks {
             unix("rw-r--r--") // 0644
         }
     }
+
+    withType<Jar>().configureEach {
+        manifest {
+            val createdByProperty = project.findProperty("manifest.createdBy") as? String
+            attributes(
+                "Created-By" to (createdByProperty
+                    ?: "${System.getProperty("java.version")} (${System.getProperty("java.vendor")})")
+            )
+        }
+    }
 }


TimvdLippe:
Thank you!

injae-kim:
Nice work @JuyeopJang !! 👍

PR #3708, opened 2025-08-13
feat: Add support for JDK21 Sequenced Collections.

### LINKED ISSUE #3659: Add JDK21 sequenced collections for ReturnsEmptyValues
JDK 21 introduced [Sequenced Collections (JEP 431)](https://openjdk.org/jeps/431) which added new interfaces to the `java.util` Collections API. Using these interfaces as return types leads to Mockito returning `null` when the method is unstubbed since they are obviously missing in `org.mockito.internal.stubbing.defaultanswers.ReturnsEmptyValues`.

Is it possible to add this new API or impossible because of Java 11 baseline?

BeomSeogKim:
This PR resolves issue #3659 by adding support for JDK 21's Sequenced Collections.

Description

JEP 431 introduced new SequencedCollection, SequencedSet, and SequencedMap interfaces in JDK 21. When a mock method with one of these return types was unstubbed, Mockito would previously return null instead of an empty collection.

Implementation Details

To address this while maintaining Java 11 compatibility, ReturnsEmptyValues has been updated. The key changes are:

A new private method, returnValueForSequencedCollection, was introduced to handle the reflection-based checks in isolation.

```java
private Object returnValueForSequencedCollection(Class<?> type) {
    String typeName = type.getName();
    if ("java.util.SequencedCollection".equals(typeName)) {
        return new ArrayList<>();
    } else if ("java.util.SequencedSet".equals(typeName)) {
        return new LinkedHashSet<>();
    } else if ("java.util.SequencedMap".equals(typeName)) {
        return new LinkedHashMap<>();
    }
    // Return null to allow other checks to proceed.
    return null;
}
```

This method is called at the beginning of returnValueFor to prioritize these special version-specific checks.

```java
Object returnValueFor(Class<?> type) {
    // 1. Handle special Sequenced Collections first.
    Object sequencedCollection = returnValueForSequencedCollection(type);
    if (sequencedCollection != null) {
        return sequencedCollection;
    }

    // ... rest of the code
```
Fixes #3659

## Checklist

### DIFF
--- a/mockito-core/src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesTest.java
+++ b/mockito-core/src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesTest.java
@@ -176,6 +176,33 @@ public class ReturnsEmptyValuesTest extends TestBase {
         assertEquals("seconds of empty " + fqcn, 0L, seconds);
     }
 
+    @Test
+    public void should_return_empty_sequenced_collection_on_java21() throws Exception {
+        Class<?> sequencedCollectionClass = getClassOrSkipTest("java.util.SequencedCollection");
+        Object result = values.returnValueFor(sequencedCollectionClass);
+        assertNotNull("SequencedCollection should return non-null value", result);
+        assertTrue("Should return empty collection", ((Collection<?>) result).isEmpty());
+        assertTrue("Should return ArrayList instance", result instanceof ArrayList);
+    }
+
+    @Test
+    public void should_return_empty_sequenced_set_on_java21() throws Exception {
+        Class<?> sequencedSetClass = getClassOrSkipTest("java.util.SequencedSet");
+        Object result = values.returnValueFor(sequencedSetClass);
+        assertNotNull("SequencedSet should return non-null value", result);
+        assertTrue("Should return empty set", ((Set<?>) result).isEmpty());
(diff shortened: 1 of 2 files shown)


mockito-core/src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesTest.java
```diff
@@ -176,6 +176,33 @@ public void should_return_empty_duration() throws Exception {
         assertEquals("seconds of empty " + fqcn, 0L, seconds);
     }
 
+    @Test
+    public void should_return_empty_sequenced_collection_on_java21() throws Exception {
+        Class<?> sequencedCollectionClass = getClassOrSkipTest("java.util.SequencedCollection");
```
bric3, line 181: @mockito/developers I wonder if we should split this test for each JDK?

bric3 (approved):

mockito-core/src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesTest.java
TimvdLippe: Yes, we can use `Assume.assumeThat` for this: https://github.com/search?q=repo%3Amockito%2Fmockito+assumeThat&type=code However, let's do that in a follow-up PR to keep things manageable.

@BeomSeogKim do you mind picking that up next?

TimvdLippe (approved):
Thanks!

PR #3727, opened 2025-09-18
Return completed futures for unstubbed Future/CompletionStage in ReturnsEmptyValues

### LINKED ISSUE #3724: Future<void> methods on mocks throw "type 'Null' is not a subtype of type   'Future<void>'"
When using mockito: ^5.0.0 with null safety, mocking a method that returns a
  Future<void> can lead to a type 'Null' is not a subtype of type 'Future<void>'
  error at runtime. This happens even when the method call is stubbed with
  thenAnswer((_) async => {}) or thenReturn(Future.value()). The only workaround
  is to create a fake implementation of the class instead of using a mock.

  Steps to reproduce:

   1. Create a class with a method that returns Future<void>.
   2. Create a mock of that class using class MockMyClass extends Mock implements
      MyClass {}.
   3. In a test, stub the method call on the mock to return a Future<void>.
   4. Call the method on the mock.

  Expected behavior:

  The method call should not throw an error.

  Actual behavior:

  A type 'Null' is not a subtype of type 'Future<void>' error is thrown.

Woongi9:
Fixes #3724 

Updated `ReturnsEmptyValues#returnValueFor` so that for return types
`Future`, `CompletableFuture`, and `CompletionStage`, Mockito now returns
`CompletableFuture.completedFuture(null)` instead of `null`.

### Tests
- Added `ReturnsEmptyValuesFutureTest` to verify the new default behavior:
  - `Future<String>` returns a completed future with `null` result.
  - `CompletableFuture<Void>` returns a completed future that can be joined without exceptions.
  - `CompletionStage<Integer>` returns a completed stage that can be safely chained.

## Checklist

 - [ ] The pull request follows coding style (run `./gradlew spotlessApply` for auto-formatting)
 - [ ] At least one commit should end with `Fixes #<issue number>` _if relevant_


Woongi9:
@TimvdLippe 
Reviews and feedback are welcome — I’ll be happy to fix anything needed.

mockito-core/src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesFutureTest.java
```diff
@@ -0,0 +1,58 @@
...
+    interface ReturnCompletableFuture {
+        CompletableFuture<Void> v();
+    }
+
+    interface ReturnCompletionStage {
+        CompletionStage<Integer> s();
+    }
+
+    @Test
+    public void returnsCompletedFuture_forFuture() throws Exception {
```
TimvdLippe, line 33: These tests look good to me. Rather than a separate test suite, can you move them into `mockito-core/src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesTest.java` instead?

TimvdLippe (changes requested):
1 nit, the rest LGTM!

mockito-core/src/test/java/org/mockito/internal/stubbing/defaultanswers/ReturnsEmptyValuesFutureTest.java
Woongi9: Done, I’ve moved the tests into ReturnsEmptyValuesTest. Thanks for the review!

TimvdLippe (approved):
Thanks!

honnix:
Looks like this is a breaking change, so maybe worth noting in the release notes (although the minor version bump might already indicate that).

PR #3729, opened 2025-09-18
Introducing the Ability to Mock Construction of Generic Types (#2401)

### LINKED ISSUE #2401: Introducing the Ability to Mock Construction of Generic Types
First of all, thank you for building an awesome testing library!

I have recently started playing with mocked object construction. It's fantastic that this functionality is now available through plain ol' Mockito. It makes for some pretty elegant tests, especially when combined with `@Mock` using the `MockitoExtension`.

One limitation I've encountered is around mocking the constructions of parameterized types.

Here's what I'd like to accomplish:
```
class DefaultKafkaConsumerTest {

    @Mock
    private MockedConstruction<KafkaConsumer<String, String>> kafkaConsumerMockedConstruction;

    @Test
    void testMyThing() {
        new DefaultKafkaConsumer<String, String>();
        final KafkaConsumer<String, String> mockConsumer = kafkaConsumerMockedConstruction.constructed().get(0);
        // ... assert
    }
```

Unfortunately, this doesn't seem to work, as indicated by this message:
```
Mockito cannot infer a static mock from a raw type for kafkaConsumerMockedConstruction

Instead of @Mock MockedConstruction you need to specify a parameterized type
For example, if you would like to mock Sample.class, specify

@Mock MockedConstruction<Sample>

as the type parameter. If the type is itself parameterized, it should be specified as raw type.
```

As suggested, I've updated my tests as follows:
```
class DefaultKafkaConsumerTest {

    @Mock
    private MockedConstruction<KafkaConsumer> kafkaConsumerMockedConstruction;

    @Test
    void testMyThing() {
        new DefaultKafkaConsumer<String, String>();

        final KafkaConsumer<String, String> mockConsumer = kafkaConsumerMockedConstruction.constructed().get(0);
        // ... assert
    }
```

I'm less fond of suppressing warnings everywhere. I've extracted the unchecked operation into a helper method, but I think that just covers up the problem.

I'm curious to understand the technical reason why the MockedConstruction generic parameter has to be a raw type. It's interesting that I can create generic mocks and argument captors, but I can't do this through mocked construction.

Thanks for your time!

giulong:
MockedStatic and MockedConstruction threw a MockitoException when the provided type was itself parameterized, suggesting to use a raw type instead. For instance, `MockedConstruction<MyGenericType<String>>` was not allowed.

This commit allows to mock generic types without throwing an exception and without forcing users to deal with `raw use of parameterized class` warnings.

Fixes #2401

## Checklist


mockito-core/src/main/java/org/mockito/internal/configuration/MockAnnotationProcessor.java
```diff
@@ -88,6 +88,9 @@ static Class<?> inferParameterizedType(Type type, String name, String sort) {
                 if (arguments[0] instanceof Class<?>) {
                     return (Class<?>) arguments[0];
                 }
+                if (arguments[0] instanceof ParameterizedType) {
```
giulong, line 91: I tried to respect the current flow by just adding the new condition, but since both `MockedStatic` and `MockedConstruction` take only one generic parameter, this block can be simplified to something like:
```java
if (type instanceof ParameterizedType) {
    ParameterizedType parameterizedType = (ParameterizedType) type;
    Type[] arguments = parameterizedType.getActualTypeArguments();
    return (Class<?>)
            (arguments[0] instanceof Class<?>
                    ? arguments[0]
                    : ((ParameterizedType) arguments[0]).getRawType());
}
```

Should I apply this change?

TimvdLippe (changes requested):
1 small nit. The rest LGTM!

mockito-core/src/main/java/org/mockito/internal/configuration/MockAnnotationProcessor.java
```diff
@@ -84,11 +84,20 @@ static Class<?> inferParameterizedType(Type type, String name, String sort) {
         if (type instanceof ParameterizedType) {
             ParameterizedType parameterizedType = (ParameterizedType) type;
             Type[] arguments = parameterizedType.getActualTypeArguments();
-            if (arguments.length == 1) {
-                if (arguments[0] instanceof Class<?>) {
-                    return (Class<?>) arguments[0];
-                }
+            if (arguments.length != 1) {
+                throw new IllegalArgumentException(
+                        "Incorrect number of type arguments for "
```
giulong, line 89: Throwing an `IllegalArgumentException` as in [InstrumentationMemberAccessor](https://github.com/mockito/mockito/blob/main/mockito-core/src/main/java/org/mockito/internal/util/reflection/InstrumentationMemberAccessor.java#L405-L411) and other places.

Issue #1803, opened 2019-10-10
Verification is logged as invocation if verbose logging is enabled


liptga:
Hi,

First of all thanks for your great work. During debugging my test I noticed a possible bug.

Consider the following test using Mockito 3.1.0 version (along with JUnit 5.4.2 and Assertj 3.11.1):

```
public class InteractionCountBugTest {

	@Test
	public void verboseLoggingShouldNotLogVerifications() {
		@SuppressWarnings("unchecked")
		List<String> list = mock(List.class, withSettings().verboseLogging());
		list.size(); // <- this invocation is printed (OK)
		verify(list).size(); // <- this invocation is printed (not OK)
	}
}
```

The test fails, since 4 invocations were logged instead of 2.

What do you think?

SeriyBg:
@mockitoguy if you don't mind, I will try to look into it. I will very much appreciate if you could point me in the right direction if you have time :)

liptga:
I took a look into the code. For me it seems to be quite simple. You just grab the org.mockito.internal.progress.ThreadSafeMockingProgress.mockingProgress() method, and get the current MockingProgress instance for the thread. This MockingProgress should have a new method like this:

`boolean isOngoingVerification()`

That method could be implemented in MockingProgress:

```
public boolean isOngoingVerification() {
  return verificationMode != null;
}
```

Then in VerboseMockInvocationLogger simply nothing should be logged, if verification is being done.

What do you think @mockitoguy ?

By the way @SeriyBg  if you implement this, you could probably build in my original test from the first version of the comment at https://github.com/mockito/mockito/issues/1803#issue-505166487

SeriyBg:
@liptga if you have working code you're welcome to create a PR. Just let me know whether I should work on this issue or not :)

liptga:
Please go on with the fix. I just took a short look, since you wanted to have a direction ;)

SeriyBg:
Thanx for the help!

SeriyBg:
@liptga, unfortunately, this fix doesn't work, I assume that this is because of ThreadLocal. Will look for other possible fixes

liptga:
@SeriyBg Could I have a look on your fix somewhere? Maybe as a branch inside your fork.

SeriyBg:
[...]

PR #3737, opened 2025-09-25
fix(verbose-logging): avoid duplicate verbose log from verification callbacks (#1803)

bhawnapannu2701:
Problem
When `verboseLogging()` is enabled, the same invocation is printed again during `verify(...)`, producing duplicate verbose lines.

Root cause
The invocation listener runs both at call time and again from the verification flow, so the identical invocation is logged twice.

Fix
In `VerboseMockInvocationLogger`, detect verification call stacks and, within a short window, suppress only the identical duplicate of the same invocation signature. Genuine consecutive method calls remain logged.

Tests
Added `org.mockito.logging.VerboseLoggingTest` which captures output and asserts exactly one verbose header is printed when verification follows a real call.

Risks
Low; the change only affects verbose logging output, no API or functional behavior.

How to verify
`gradlew :mockito-core:test --tests org.mockito.logging.VerboseLoggingTest`

### DIFF
--- a/mockito-core/src/main/java/org/mockito/internal/debugging/VerboseMockInvocationLogger.java
+++ b/mockito-core/src/main/java/org/mockito/internal/debugging/VerboseMockInvocationLogger.java
@@ -22,6 +22,18 @@ public class VerboseMockInvocationLogger implements InvocationListener {
 
     private int mockInvocationsCounter = 0;
 
+    /**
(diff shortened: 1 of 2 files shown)


bhawnapannu2701:
Opened issue #3739  as requested to discuss the design for avoiding duplicate verbose logs

PR #3753, opened 2025-10-21
Support primitives in GenericArrayReturnType.

### LINKED ISSUE #3752: ClassNotFoundException when stubbing array of primitive type on Android
Suppose:
```
public class Foo {
  public byte[] getData() { ... }
}
```

```
public class MyTest {
  @Test
  public void myTest() {
    Foo mockFoo = mock(Foo.class);
    when(mockFoo.getData()).thenReturn(null);
  }
}
```

When running as an instrumentation test on an Android API 30 emulator with latest Mockito, this results in:

```
java.lang.IllegalStateException: This was not supposed to happen.
	at org.mockito.internal.util.reflection.GenericMetadataSupport$GenericArrayReturnType.rawType(GenericMetadataSupport.java:565)
	at org.mockito.internal.stubbing.answers.InvocationInfo.isVoid(InvocationInfo.java:98)
	at org.mockito.internal.stubbing.answers.Returns.validateFor(Returns.java:34)
	at org.mockito.internal.stubbing.InvocationContainerImpl.addAnswer(InvocationContainerImpl.java:65)
	at org.mockito.internal.stubbing.InvocationContainerImpl.addAnswer(InvocationContainerImpl.java:56)
	at org.mockito.internal.stubbing.OngoingStubbingImpl.thenAnswer(OngoingStubbingImpl.java:32)
	at org.mockito.internal.stubbing.BaseStubbing.thenReturn(BaseStubbing.java:34)
	at com.example.MyTest.myTest.(MyTest.java:50)
	... 38 trimmed
Caused by: java.lang.ClassNotFoundException: [Lbyte;
	at java.lang.Class.classForName(Native Method)
	at java.lang.Class.forName(Class.java:454)
	at org.mockito.internal.util.reflection.GenericMetadataSupport$GenericArrayReturnType.rawType(GenericMetadataSupport.java:556)
	... 46 more
Caused by: java.lang.ClassNotFoundException: byte
	at java.lang.Class.classForName(Native Method)
	at java.lang.BootClassLoader.findClass(ClassLoader.java:1358)
	at java.lang.BootClassLoader.loadClass(ClassLoader.java:1418)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:312)
	... 49 more
Caused by: java.lang.NoClassDefFoundError: Class not found using the boot class loader; no stack trace available
```

After some investigation, it looks like `Method.getGenericReturnType()` is implemented differently on Android and this results in a different code path that wasn't designed to handle primitives. I'll open a PR for this.

jselbo:
Fixes #3752

In my testing, for a method returning `byte[]`, `Method#getGenericReturnType()` on a normal JVM returns a `Class`, the same as `byte[].class`.

On Android, the implementation is completely different than OpenJDK's. Code ref:
https://cs.android.com/android/platform/superproject/main/+/main:libcore/ojluni/src/main/java/java/lang/reflect/Method.java;l=175-177;bpv=0;bpt=1

It uses Android-specific reflection APIs under the hood, and this returns a `GenericArrayType` with component type `byte.class`.

This is my attempt to handle this scenario, by adding primitive support to `GenericArrayReturnType`.

## Checklist

 - [ X] Read the [contributing guide](https://github.com/mockito/mockito/blob/main/.github/CONTRIBUTING.md)
 - [ X] PR should be motivated, i.e. what does it fix, why, and if relevant how
 - [ X] If possible / relevant include an example in the description, that could help all readers
 - [ X] Avoid other runtime dependencies
 - [ X] Meaningful commit history ; intention is important please rebase your commit history so that each
 - [ X] The pull request follows coding style (run `./gradlew spotlessApply` for auto-formatting)
 - [ X] Mention `Fixes #<issue number>` in the description _if relevant_
 - [ X] At least one commit should end with `Fixes #<issue number>` _if relevant_

### DIFF
--- a/mockito-integration-tests/android-tests/src/androidTest/java/org/mockitousage/androidtest/BasicInstrumentedTests.kt
+++ b/mockito-integration-tests/android-tests/src/androidTest/java/org/mockitousage/androidtest/BasicInstrumentedTests.kt
@@ -2,13 +2,16 @@ package org.mockitousage.androidtest
 
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import org.junit.After
+import org.junit.Assert.assertArrayEquals
 import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
+import org.mockito.ArgumentMatchers.anyList
 import org.mockito.Mock
 import org.mockito.MockitoAnnotations
 import org.mockito.Mockito.mock
 import org.mockito.Mockito.verify
+import org.mockito.Mockito.`when`
 
 @RunWith(AndroidJUnit4::class)
 class BasicInstrumentedTests {
@@ -73,5 +76,17 @@ class BasicInstrumentedTests {
         val receiver = BasicInterfaceReceiver(basicInterface)
         receiver.callInterfaceMethod()
         verify(basicInterface).interfaceMethod()
+    }
+
+//Regression test for issue #3752
+
+    @Test
+    fun mockMethodWithArrayOfPrimitiveReturnType() {
+        val mock = mock(HasArrayOfPrimitivesReturnType::class.java)
+        `when`(mock.getData(anyList())).thenReturn(byteArrayOf(1, 2, 3))
+
+        val actual = mock.getData(emptyList())
+
+        assertArrayEquals(byteArrayOf(1, 2, 3), actual)
     }
 }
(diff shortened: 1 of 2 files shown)


jselbo:
Sorry, I did not think of a good way to add tests. The new Primitives logic is trivial and maybe doesn't need a test. The `GenericArrayReturnType` class is a private implementation detail. I suppose I could make it package-private and construct it directly in the test to verify its logic?

jselbo:
Thanks, added an integration test. I confirmed the test fails without the changes in `GenericMetadataSupport` and passes with them.

TimvdLippe (approved):
Thanks for the fix!

PR #3771, opened 2025-12-14
Add tests for Sets utility class

thswlsqls:
## Summary

Add comprehensive test coverage for `Sets` utility class which provides factory methods for creating mock-safe hash sets and regular sets. This improves code coverage and ensures correct behavior of set creation methods.

## Changes

- Added `SetsTest` class with 17 test methods covering all three static methods:
  - `newMockSafeHashSet(Iterable<Object>)` - 4 tests
  - `newMockSafeHashSet(Object...)` - 3 tests
  - `newSet(T...)` - 10 tests
- Test scenarios include:
  - Mock objects handling
  - Regular objects handling
  - Empty collections
  - Null handling (with proper AssertJ type casting)
  - Duplicate elements
  - Order preservation (LinkedHashSet behavior)
  - Error cases (null array parameter)
  - Set operations (add, remove)

**Note**: AssertJ's `contains()` and `containsExactly()` methods require explicit type casting when checking for null values (e.g., `contains((Object) null)`) to avoid `NullPointerException` caused by varargs array creation.

## Testing

All tests pass:
```bash
./gradlew :mockito-core:test --tests "org.mockito.internal.util.collections.SetsTest"

### DIFF
--- a/mockito-core/src/test/java/org/mockito/internal/util/collections/SetsTest.java
+++ b/mockito-core/src/test/java/org/mockito/internal/util/collections/SetsTest.java
@@ -14,6 +14,10 @@ import java.util.Set;
 
 import org.junit.Test;
 
+/**
+ * Tests for {@link Sets} utility class which provides factory methods
+ * for creating mock-safe hash sets and regular sets.
+ */
 public class SetsTest {
 
     interface TestInterface {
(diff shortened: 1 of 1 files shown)


TimvdLippe (changes requested):
Thanks for these! They make sense to me. Only nit is about the comments.

PR #3785, opened 2026-01-19
Avoid mocking of internal static utilities

### LINKED ISSUE #3778: Static mocking of UUID.class corrupted under JDK 25
```java
import static org.mockito.Mockito.mockStatic;

import java.util.UUID;

class DummyTest {

	static final UUID dummyUUID = new UUID(1, 1);

	@Test
	void testUUIDMock() {
		System.out.println("Before mocking: " + dummyUUID);
		try (final var mockUUID = mockStatic(UUID.class)) {
			System.out.println("After mocking: " + dummyUUID);
			assertEquals("00000000-0000-0001-0000-000000000001", dummyUUID.toString());
		}
	}
}
```
fails and prints:
```
Before mocking: 00000000-0000-0001-0000-000000000001
After mocking: ----
org.opentest4j.AssertionFailedError: 
Expected :00000000-0000-0001-0000-000000000001
Actual   :----
```

Interestingly, editing the sample to
```java
	@Test
	void testUUIDMock() {
		System.out.println("Before mocking: " + dummyUUID);
		try (final var mockUUID = mockStatic(UUID.class)) {
			System.out.println("After mocking: " + dummyUUID);
		}
		System.out.println("After try: " + dummyUUID);
	}
```
leads to
```
Before mocking: 00000000-0000-0001-0000-000000000001
After mocking: ----
After try: 00000000-0000-0001-0000-000000000001
```

BTW, the `�` seem to be `null` chars.

openJDK 25.0.1+8-LTS
mockito-core:5.21.0
byte-buddy:1.18.3

With jdk21 this test is green.

raphw:
Avoids mocking private static methods, as well as package-private static methods of java.* classes, as those are primarily internal utilities that will cause non-intuitive changes of methods. This are not invoked for non-static mocks either. Fixes #3778.

Minor risk: people might have relied on stubbing those internal methods.

### DIFF
--- a/mockito-core/src/main/java/org/mockito/internal/creation/bytebuddy/InlineBytecodeGenerator.java
+++ b/mockito-core/src/main/java/org/mockito/internal/creation/bytebuddy/InlineBytecodeGenerator.java
@@ -133,7 +133,18 @@ public class InlineBytecodeGenerator implements BytecodeGenerator, ClassFileTran
                                         .bind(MockMethodAdvice.Identifier.class, identifier)
                                         .to(MockMethodAdvice.class))
                         .method(
-                                isStatic().and(not(BytecodeGenerator.isGroovyMethod(true))),
+                                isStatic()
+                                        .and(
+                                                not(
+                                                        isPrivate()
+                                                                .or(
+                                                                        isDeclaredBy(
+                                                                                        nameStartsWith(
+                                                                                                "java."))
+                                                                                .<MethodDescription>
+                                                                                        and(
+                                                                                                isPackagePrivate()))))
+                                        .and(not(BytecodeGenerator.isGroovyMethod(true))),
                                 Advice.withCustomMapping()
                                         .bind(MockMethodAdvice.Identifier.class, identifier)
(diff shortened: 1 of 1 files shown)


bric3 (approved):
Totally agree
Do you have a story behind this change. Is it related to JEP 500?

raphw:
It only fails with Java 25 because they changed the internal implementation to use private static utilities which are now mocked. Before, the code was inlined in the method, so it worked for instances. That's why I think that is a better default.

TimvdLippe:
Can we reuse the `@DoNotMock` infrastructure? https://github.com/mockito/mockito/blob/main/mockito-core/src/main/java/org/mockito/DoNotMock.java has the annotation, but I believe we have some other custom lists as well. We block mocking `List` if I remember correctly.

raphw:
In this case, it is about methods of classes, not the class in general. Also, it those are not referencable by the API. I think this is the right approach here.

alwibrm:
Hi, will this be merged soon? This behaviour in Mockito is currently preventing us from upgrading to Java 25.

PR #3789, opened 2026-02-26
Replace mockito-android mock maker with dexmaker-mockito-inline fork

### LINKED ISSUE #3788: Proposal: integrate dexmaker-mockito-inline for modern Android support
Unfortunately, the `mockito-android` mockmaker is not very useful in the modern Android ecosystem. It relies entirely on the subclass mockmaker, so it cannot mock final classes. The Android ecosystem has become increasingly Kotlin centric - both Google's first-party libraries and app code. In Kotlin all classes are final by default unless explicitly marked with `open` keyword. This means `mockito-android` is not able to mock much at all.

For a long time our Android teams have been using https://github.com/linkedin/dexmaker with `dexmaker-mockito-inline` to support Mockito in Android tests. It has a mockmaker that does special bytecode generation for Dalvik.

As a point of comparison, MockK (I would say Mockito's biggest competitor in the Kotlin world) seems to have copied/forked dexmaker-mockito-inline in their Android "proxy" implementation:
https://github.com/mockk/mockk/blob/master/modules/mockk-agent-android/README.txt

A downside of dexmaker-mockito-inline for a while was that it only supported Android P (API 28) and up, but according to Android Studio's version distribution dialog this accounts for > 95% of distribution. It doesn't seem like much of a limitation at this point.

I'd like to replace mockito-android's mock maker implementation with one built on top of dexmaker. Any concerns about this?

dexmaker has an Apache 2.0 license so I think it would be okay to include its source. Maybe a better option would be to explicitly take a dependency or discuss transferring ownership?

jselbo:
Fork dexmaker-mockito-inline (https://github.com/linkedin/dexmaker/tree/main/dexmaker-mockito-inline) to replace the subclass-based mock maker with a JVMTI-based inline mock maker that supports mocking final classes and methods on Android.

This allows Mockito to be useful in a modern Kotlin Android ecosystem where classes are `final` by default unless explicitly marked with the `open` keyword. The subclass mock maker cannot mock final classes, and the inline mock maker does not work on Android.

Users currently work around this limitation by:
- avoiding mocking unowned Kotlin code and adding `open` to owned Kotlin classes they want to mock
- or running tests on JVM with a simulated Android environment via Robolectric
- or using MockK (note: they similarly forked dexmaker-mockito-inline in 2018: https://github.com/mockk/mockk/tree/master/modules/mockk-agent-android)

Risks/Considerations:
- Backwards compatibility
  - I think changing mockito-android distribution from `jar` to `aar` will be transparent for Gradle users, but not sure about other build systems.
  - Now requires min API 28 (Android P) - the instrumentation test build will fail if the test's manifest specifies a lower min SDK. According to the Android Studio version distribution dialog, Android P+ covers 95% of Android users. We may have Mockito users running tests on pre-API 28 emulators, but we can encourage them to stay on an older version of mockito-android or upgrade the tests to run on a newer API.
- Forking 
  - Instead of forking, we could probably simply depend on the dexmaker-mockito-inline artifact. The downside of this seems to be lack of control and limiting future improvements.
- Licensing
  - `external/`:
    - `jdk/jvmti.h` is from Oracle OpenJDK under GPL-2.0
    - `slicer/` is from AOSP under Apache 2.0
  - `agent.cc` and all new Java sources except `InlineAndroidMockMaker.java` are copied from dexmaker-mockito-inline licensed under Apache 2.0. Confusingly, the headers of these sources state "Copyright (C) 2017 The Android Open Source Project" even though they seem to originate from the dexmaker repo itself. I opened https://github.com/linkedin/dexmaker/issues/208 asking about this. I retained the headers as-is.
- Dispatcher
  -  dexmaker-mockito-inline includes a pre-built dispatcher.jar with dex bytecode to allow instrumenting methods with JVMTI. I followed the pattern of MockK and build the dispatcher from source instead of having a prebuilt.

Technical changes:
- Convert mockito-android from Java library (JAR) to Android library (AAR) to allow packaging native code
- Add native NDK/CMake code (JVMTI agent + AOSP slicer library)
- Fork dexmaker Java sources under org.mockito.android.internal.creation.inline
- Add InlineAndroidMockMaker entry-point delegating to InlineDexmakerMockMaker
- Build dispatcher.jar via new mockito-android-dispatcher submodule
- Require min SDK 28 (Android P) via manifest for JVMTI support
- Conditionally include mockito-android in the Gradle project based on Android SDK availability
- Add integration tests for final class mocking

Note for reviewers: the only "new" code is InlineAndroidMockMaker.java. Everything else is a copy/fork.

Fixes #3788

## Checklist

This adds a new compile dependency for mockito-android on `com.linkedin.dexmaker:dexmaker` which allows runtime generation of Android dex byte code, similar to Byte-Buddy on the JVM.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,7 +107,7 @@ jobs:
     strategy:
       matrix:
         # Minimum supported and maximum available.
-        android-api: [ 26, 33 ]
+        android-api: [ 28, 33 ]
 
     # All build steps
     steps:


raphw:
I do not know enough on the background to decide if forking or linking is the better solution, but I wonder: Instead of forking, we could probably simply depend on the dexmaker-mockito-inline artifact. The downside of this seems to be lack of control and limiting future improvements.

If the limitation is in the future and if no API is exposed, could we go for linking until we need to potentially fork?

jselbo:
Thinking about it more, I think linking for now makes sense for simplicity. I will open a new PR.

PR #3790, opened 2026-02-26
Fix StackOverflowError with AbstractList after using mockSingleton

jselbo:
(I noticed this issue when integrating with Mockito-Kotlin and running its test suite)

`WeakHashMap.get()` calls `hashCode()` on its keys, which for instrumented classes triggers the byte buddy advice, calling `isMocked() -> getSingletonMockInterceptor() -> WeakHashMap.get()` recursively.

This adds a thread-local recursion guard around the `WeakHashMap.get()` call.

Alternatively, we could use a different map implementation like IdentityHashMap which doesn't call `hashCode()` but I didn't want to diverge from `mockStatic` which also uses a `WeakHashMap` (though I'm not convinced the weak semantics are important)

### DIFF
--- a/mockito-core/src/main/java/org/mockito/internal/creation/bytebuddy/InlineDelegateByteBuddyMockMaker.java
+++ b/mockito-core/src/main/java/org/mockito/internal/creation/bytebuddy/InlineDelegateByteBuddyMockMaker.java
@@ -942,7 +942,7 @@ class InlineDelegateByteBuddyMockMaker
                         join(
                                 "The singleton instance "
                                         + instance.getClass().getName()
-                                        + " is already registered as a mock",
+                                        + " is already registered as a mock in the current thread",
                                 "",
                                 "To create a new mock, the existing mock registration must be deregistered"));
             }


raphw (changes requested):
This is a good guardrail. Could I asked you to add the same mechanism in mockedStatics? It might call a hashCode that delegates to a static utility, what would cause the same issue, so we should handle this consistently.

jselbo:
Actually I think it's not susceptible to the same issue because for `isMockedStatic` we pass a `Class` type to `WeakHashMap.containsKey()` which can't be instrumented.

raphw:
Indeed. Would need a custom hashCode method in a future JVM that calles a static method outside of `Class`, what sounds unlikely. I am however wondering why hashCode is called at all. I would expect identity maps since hashcodes and equality can be broken on user objects. Let me have a quick look.

raphw:
Ah, I think we should avoid the weak hash map here: https://github.com/mockito/mockito/blob/main/mockito-core/src/main/java/org/mockito/internal/creation/bytebuddy/InlineDelegateByteBuddyMockMaker.java#L698

If the hashCode/equals method is broken this will yield weird results, and that even more if two equal singleton objects exist. I would go for an `IndentityHashMap<MockitoWeakReference<Object>>` and use identity comparison on the (pseudo) `MockitoWeakReference`. For bucketing, use system hash code.

raphw:
With this, there should never be calls on the object itself and the guardrail is no longer needed.

jselbo:
I am wondering if the weak reference semantics are still important for this impl. `IdentityHashMap<MockWeakReference<Object>, ..>`  is not exactly the same as `WeakHashMap`, and I think we'd need extra care/complexity to actually achieve weak ref behavior in the same way.

I think for "singletons" the weak ref behavior doesn't really matter. For Java enums at least, they should never be GCed.

jselbo:
I'm wary of building a new map implementation from scratch to satisfy both weakness and identity. I found some discussion on https://stackoverflow.com/questions/22910375/combo-of-identityhashmap-and-weakhashmap which points to Guava or https://github.com/plume-lib/hashmap-util/blob/main/src/main/java/org/plumelib/util/WeakIdentityHashMap.java (unclear how battle-tested this is). I guess neither are really viable options,.

It seems to me a recursion guard (like the original commit) is simpler to maintain even though it's theoretically incomplete. What do you think @raphw ?

raphw:
I think we need to support identity and weak references, we cannot choose as lacking either will lead to subtle bugs, and with Mockito's reach users will run into those. In our case, I would like to avoid the dependency, as Guava can quickly lead to conflicts. Wrapping the mocks seems like a realistic goal as we do not expose the map as general purpose. I believe that is the solution we should aim for.

jselbo:
It seems that `IdentityHashMap<MockWeakReference<Object>, ..>` is not a functional solution because the mock can be GC'ed but the keys in the map still exist, and it's unclear how we'd do lookups. I think the only way to get both behaviors is a new map impl.

raphw (approved):
Last nit pick: add a note to the exception that states that the instance is already mocked that this is for the current thread, to make it obvious.

PR #3792, opened 2026-03-04
Replace mockito-android mock maker implementation with dexmaker-mockito-inline

jselbo:
See https://github.com/mockito/mockito/pull/3789 for a previous discussion on this

For now it is probably better to just depend on dexmaker-mockito-inline . If we need more flexibility or the dexmaker project is not responsive, we can fork it in the future.

Note this means `mockito-android` is now published as an Android library (.aar instead of .jar). This also removes the graceful fallback behavior if this mock maker was included on a non-Android JVM - now invoking a Mockito API will fail with a runtime exception instead of falling back to the subclass mockmaker.

I also updated the Android sections in the Javadoc appropriately 

## Checklist

This adds a new compile dependency for mockito-android on `com.linkedin.dexmaker:dexmaker-mockito-inline` which allows runtime generation of Android dex byte code, similar to Byte-Buddy on the JVM.


raphw:
This looks good to me! I think a best start and we can always move to a fork if there is a need that justifies this step. Also easier to maintain and upgrade for a non-expert!

davwheat:
Would you be able to highlight the breaking change this results in more clearly within the changelog?

We still support SDK26 in our app, so run tests on that version, and I've just spent the last 6 hours banging my head against the wall trying to work out how I have broken all my instrumented tests :p

jselbo:
@davwheat Sorry about the bad experience. Good call - I will update the release notes to more clearly state this breaking change!

PR #3798, opened 2026-03-22
Fixes #2398 : Implement suppress static initialization leveraging java agent

### LINKED ISSUE #2398: How can I avoid 拖
> Hey,
>
> First thanks for reporting, in order to help us to classify issue can you make sure the following check boxes are checked ?
>
> If this is about mockito usage, the better way is to reach out to
>
>  - stackoverflow : https://stackoverflow.com/questions/tagged/mockito
>  - the mailing-list  : https://groups.google.com/forum/#!forum/mockito / mockito@googlegroups.com
>    (Note mailing-list is moderated to avoid spam)
>
> _This block can be removed_
> _Something wrong in the template fix it here `.github/ISSUE_TEMPLATE.md`

check that

 - [ ] Provide versions (mockito / jdk / os / any other relevant information)

mirkoalicastro:
### Suppress static initializers for specified classes

Fixes: #2398

**Summary**

Adding `@SuppressStaticInitializationFor` annotation.  It prevents `<clinit>` (static initializer blocks) from executing for specified classes. This is useful when a class under test has heavy static initialization (database connections, file I/O, native calls) that fails or is undesirable in test environments.

Usage example:
```
@SuppressStaticInitializationFor({"com.example.HeavyService", "com.example.LegacyClient"})
class MyTest {
    // HeavyService and LegacyClient will have their static initializers suppressed.
    // All their static fields will have JVM zero-value defaults (null, 0, false).
}
```

**Why this is now possible**

This feature was historically only available through _PowerMock_, which used a custom classloader-per-test-class architecture to reload classes with modified bytecode. Mockito couldn't offer this because its mock maker after class loading... too late to intercept `<clinit>`.

However, two things have changed:

1. Mockito 5 made the inline mock maker the default. The inline mock maker uses java.lang.instrument and registers a ClassFileTransformer, which is the JVM's official hook for modifying class bytecode. When Mockito is loaded as a -javaagent (premain), this transformer sees every class during initial loading (when classBeingRedefined == null). This is the exact interception point needed to strip `<clinit>` before the JVM executes it.
2. The premain infrastructure already exists. The ClassFileTransformer (InlineBytecodeGenerator.transform()) is already wired into the instrumentation pipeline. Suppressing `<clinit>` is a natural extension: when a class name is in the suppression set and it's being loaded for the first time, the transformer uses ASM to replace the `<clinit>` method body with a single RETURN instruction, leaving all static fields at their JVM zero-value defaults.

PowerMock's classloader approach provided per-test-class isolation (each test got a fresh copy of the class). The instrumentation approach is JVM-global: once a class is loaded with its `<clinit>` suppressed, it stays that way. This is a deliberate trade-off: it avoids the complexity and compatibility issues of custom classloaders, at the cost of requiring JVM forking (forkEvery = 1 for gradle / reuseForks=false for maven) when multiple test classes need different suppression behavior.

In addition to the unit test coverage, I've also installed mockito in my local maven repo and created a project that uses the feature to suppress a static initialization. Then, I configured reuseforks=false, to test that the changes can be reverted, and both the tests passed. Code:

Class to suppress:
```
public class HeavyService {
    public static final String CONNECTION;
    static {
        CONNECTION = initConnection();
    }

    private static String initConnection() {
        throw new RuntimeException("Cannot connect to database in tests!");
    }

    public String serve() {
        return CONNECTION;
    }
}
```

Test number 1:
```
@SuppressStaticInitializationFor({"org.example.HeavyService"})
class HeavyServiceTest {

    @Test
    void static_initializer_is_suppressed() {
        // Without suppression, this would throw "Cannot connect to database in tests!"
        // With suppression, CONNECTION is null (JVM zero-value default)
        assertNull(HeavyService.CONNECTION);
    }

    @Test
    void can_create_instance_without_explosion() {
        HeavyService service = new HeavyService();
        assertNull(service.serve());
    }
}
```

Test number 2:
```
class HeavyService2Test {

    @Test
    void static_initializer_fails() {
        // Without suppression, this will throw "Cannot connect to database in tests!"
        assertThrows(Error.class, () -> {
            String connection = HeavyService.CONNECTION;
        });
    }

    @Test
    void cannot_create_instance() {
        assertThrows(Error.class, HeavyService::new);
    }
}
```

Maven config:
```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>org.example</groupId>
    <artifactId>mvnproject</artifactId>
    <version>1.0-SNAPSHOT</version>

    <properties>
        <maven.compiler.source>11</maven.compiler.source>
        <maven.compiler.target>11</maven.compiler.target>
        <mockito.version>5.23.1-SNAPSHOT</mockito.version>
        <junit.version>5.13.4</junit.version>
        <argLine></argLine>
    </properties>

    <dependencies>
        <dependency>
            <groupId>org.mockito</groupId>
            <artifactId>mockito-core</artifactId>
            <version>${mockito.version}</version>
            <scope>test</scope>
        </dependency>
        <dependency>
            <groupId>org.mockito</groupId>
            <artifactId>mockito-junit-jupiter</artifactId>
            <version>${mockito.version}</version>
            <scope>test</scope>
        </dependency>
        <dependency>
            <artifactId>junit-jupiter</artifactId>
            <version>${junit.version}</version>
            <scope>test</scope>
        </dependency>
    </dependencies>

    <build>
        <plugins>
            <plugin>
                <artifactId>maven-dependency-plugin</artifactId>
                <executions>
                    <execution>
                        <goals>
                            <goal>properties</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
            <plugin>
                <version>3.5.2</version>
                <configuration>
                    <argLine>@{argLine} -javaagent:${org.mockito:mockito-core:jar}</argLine>
                    <reuseForks>false</reuseForks>
                </configuration>
            </plugin>
        </plugins>
    </build>
</project>
```

## Checklist

### DIFF
--- a/mockito-core/src/main/java/org/mockito/internal/creation/bytebuddy/InlineDelegateByteBuddyMockMaker.java
+++ b/mockito-core/src/main/java/org/mockito/internal/creation/bytebuddy/InlineDelegateByteBuddyMockMaker.java
@@ -578,8 +578,20 @@ class InlineDelegateByteBuddyMockMaker
                             "",
                             "Add -javaagent:mockito-core.jar to your JVM arguments."));
         }
+        validateNotAlreadyLoaded(INSTRUMENTATION, classNames);
+    }
+
+    /**
+     * Validates that none of the given class names are already loaded by the JVM.
+     *
+     * @param instrumentation the instrumentation instance to query for loaded classes
+     * @param classNames the class names to check
+     * @throws MockitoException if any of the classes are already loaded
+     */
+    static void validateNotAlreadyLoaded(
+            Instrumentation instrumentation, Collection<String> classNames) {
         for (String name : classNames) {
-            for (Class<?> loaded : INSTRUMENTATION.getAllLoadedClasses()) {
+            for (Class<?> loaded : instrumentation.getAllLoadedClasses()) {
                 if (loaded.getName().equals(name)) {
                     throw new MockitoException(
                             "Cannot suppress static initialization for '"
@@ -588,7 +600,6 @@ class InlineDelegateByteBuddyMockMaker
                 }
             }
         }
-        bytecodeGenerator.addSuppressedClasses(classNames);
     }
 
     @Override
--- a/mockito-core/src/test/java/org/mockito/internal/creation/bytebuddy/SuppressClinitTest.java
+++ b/mockito-core/src/test/java/org/mockito/internal/creation/bytebuddy/SuppressClinitTest.java
@@ -5,10 +5,13 @@
 package org.mockito.internal.creation.bytebuddy;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.hamcrest.CoreMatchers.instanceOf;
 
+import java.lang.instrument.Instrumentation;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Set;
 
@@ -22,6 +25,8 @@ import net.bytebuddy.jar.asm.MethodVisitor;
 import net.bytebuddy.utility.OpenedClassReader;
 import org.junit.Assume;
 import org.junit.Test;
+import org.mockito.exceptions.base.MockitoException;
+import org.mockito.internal.PremainAttachAccess;
 import org.mockito.internal.configuration.plugins.Plugins;
 import org.mockito.plugins.MockMaker;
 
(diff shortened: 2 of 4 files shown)


mirkoalicastro:
Different approach has been taken: https://github.com/mockito/mockito/pull/3801 (merged).

PR #3801, opened 2026-03-30
Fixes #2398 : Implement global suppress static initialization leveraging java agent

### LINKED ISSUE #2398: How can I avoid 拖

mirkoalicastro:
### Suppress static initializers for specified classes

Fixes: #2398

**Summary**

Adding system property `mockito.suppress.clinit` to permanently prevent `<clinit>` (static initializer blocks) from executing for specified classes. This is useful when a class under test has heavy static initialization (database connections, file I/O, native calls) that fails or is undesirable in test environments. The tests are consistent and **do not depend on class loading order**.

Usage example:
```
<plugin>
  <version>3.5.2</version>
  <configuration>
    <argLine>@{argLine} -javaagent:${org.mockito:mockito-core:jar} -Dmockito.suppress.clinit=org.example.HeavyService,org.example.HeavyService2</argLine>
    </configuration>
</plugin>
```

**Why this is now possible**

This feature was historically only available through _PowerMock_, which used a custom classloader-per-test-class architecture to reload classes with modified bytecode. Mockito couldn't offer this because its mock maker after class loading... too late to intercept `<clinit>`.

However, two things have changed:

1. Mockito 5 made the inline mock maker the default. The inline mock maker uses java.lang.instrument and registers a ClassFileTransformer, which is the JVM's official hook for modifying class bytecode. When Mockito is loaded as a -javaagent (premain), this transformer sees every class during initial loading (when classBeingRedefined == null). This is the exact interception point needed to strip `<clinit>` before the JVM executes it.
2. The premain infrastructure already exists. The ClassFileTransformer (InlineBytecodeGenerator.transform()) is already wired into the instrumentation pipeline. Suppressing `<clinit>` is a natural extension: when a class name is in the suppression set and it's being loaded for the first time, the transformer uses ASM to replace the `<clinit>` method body with a single RETURN instruction, leaving all static fields at their JVM zero-value defaults.

PowerMock's classloader approach provided per-test-class isolation (each test got a fresh copy of the class). The instrumentation approach is JVM-global: once a class is loaded with its `<clinit>` suppressed, it stays that way. Same approach is being used here for the feature.

In addition to the unit test coverage, I've also installed mockito in my local maven repo and created a project that uses the feature to suppress a static initialization.

## Checklist

### DIFF
--- a/mockito-core/src/test/java/org/mockito/internal/creation/bytebuddy/ClinitSuppressionTransformerTest.java
+++ b/mockito-core/src/test/java/org/mockito/internal/creation/bytebuddy/ClinitSuppressionTransformerTest.java
@@ -10,8 +10,14 @@ import net.bytebuddy.jar.asm.MethodVisitor;
 import net.bytebuddy.utility.OpenedClassReader;
 import org.junit.Test;
 
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.ObjectStreamClass;
+import java.io.Serializable;
 import java.lang.reflect.Field;
 import java.util.Collections;
 import java.util.HashSet;
@@ -22,7 +28,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 public class ClinitSuppressionTransformerTest {
 
     // Test target class with a static initializer
-    static class WithClinit {
+    static class WithClinit implements Serializable {
         static String value = "initialized";
         static int number = 42;
 
@@ -94,16 +100,7 @@ public class ClinitSuppressionTransformerTest {
         // Load the transformed class with a custom classloader
         String className = WithClinit.class.getName();
         ClassLoader customLoader =
-                new ClassLoader(getClass().getClassLoader()) {
-                    @Override
-                    protected Class<?> loadClass(String name, boolean resolve)
-                            throws ClassNotFoundException {
-                        if (name.equals(className)) {
-                            return defineClass(name, transformed, 0, transformed.length);
-                        }
-                        return super.loadClass(name, resolve);
-                    }
-                };
+                new ClassOverridingLoader(getClass().getClassLoader(), className, transformed);
 
         Class<?> loaded = customLoader.loadClass(className);
         Field valueField = loaded.getDeclaredField("value");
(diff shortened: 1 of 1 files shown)


raphw:
This looks good mostly, but two things:
1. ClassFileTransformer should return null when a class is not transformed. 
2. I wonder if the argument should go to the agent arg, not a property. This would better bind the agent and argument together.

Not entirely sure about two.

mirkoalicastro:
> This looks good mostly, but two things:
> 
> 1. ClassFileTransformer should return null when a class is not transformed.

Good catch. Fixed. It now returns null instead of the original byte array when no `<clinit>` is found.

> 2. I wonder if the argument should go to the agent arg, not a property. This would better bind the agent and argument together.
> 
> Not entirely sure about two.

I actually thought of this. The agent arg approach binds the agent and argument together very nicely, but I thought it could limit future extensibility. If there will ever be the need of additional premain-time parameters, system properties scale naturally:
`-Dmockito.suppress.clinit=org.example.HeavyService,org.example.HeavyService2 -Dmockito.other=value`
Instead, with agent args, since we can only pass one string, it would require introducing a delimiter format, like:
`mockito.suppress.clinit=org.example.HeavyService,org.example.HeavyService2;mockito.other=value`
and parsing the string by ourselves.. splitting by `;`, then extracting `key=value`, etc. It could get messy.

What do you think?

raphw:
Not quite sure myself.

On second thought the system property might however be a better option as you suggest it.

People might want to enable dynamic attach for test VMs to ease using Mockito. Here a system property is cleaner, and you are right about future extension.

Let's keep it this way.

raphw:
One small comment improvement, otherwise this is good to merge.

mirkoalicastro:
> Let's keep it this way.

Perfect.

> One small comment improvement, otherwise this is good to merge.

Sure, which improvement you have in mind?

raphw:
The pending comment on explaining why we do not delete the clinit. See above.

mirkoalicastro:
> The pending comment on explaining why we do not delete the clinit. See above.

I can't see pending comments (maybe not submitted?), but I get the idea of deleting the clinit entirely, instead of empty clinit with only return 😅 . It makes sense and simplifies the code. I've pushed the change

raphw:
Sorry, I did not make myself clear here.

I asked you to add a comment in the code that the clinit is stubbed, not deleted, to retain the implicit serialization UID. This to avoid that someone considers the optimization in the future, to simply remove clinit, what would break serialization.

Could you please bring back the previous version and add that comment?

mirkoalicastro:
> Could you please bring back the previous version and add that comment?

Ah, I see. Reverted and added the comment.

mirkoalicastro:
@raphw I wanted to add a unit test for it, but I couldn't reproduce it. Let me know if the comment that I've added is enough for this 🙏

raphw:
Deserialization should fail for a class with an implicit version UID if the serialized version comes from a class with initializer. You'd need to add a serialized version of a class with initializer without the agent active.

mirkoalicastro:
@raphw Got it. I've added a test that serialiye an instance of the original class with clinit and deserializes it using the transformed bytecode. Also, I've checked that the test would fail if the clinit is suppressed by complete removal.

raphw:
Looks good to me. Any comments from @mockito/developers?

ddingel:
Thank you for contributing this! I already have a use case.

As in my use case I have a lot of classes with static initializers (third party integration I would like to mock away) do you think adding a prefix / packaging version is sensible?
Like:

```
            "-Dmockito.suppress.clinit.packages=some.thirdparty."
```

PR #3815, opened 2026-04-22
Fixes #3814 : implement package-level clinit suppression

### LINKED ISSUE #3814: Package-level clinit suppression
Thank you for contributing this! I already have a use case.

As in my use case I have a lot of classes with static initializers (third party integration I would like to mock away) do you think adding a prefix / packaging version is sensible?
Like:

```
            "-Dmockito.suppress.clinit.packages=some.thirdparty."
```

_Originally posted by @ddingel in https://github.com/mockito/mockito/issues/3801#issuecomment-4207738949_
            
I see this as an enhancement of the feature - as this requirement is real and could simplify test configuration for interaction with legacy systems

mirkoalicastro:
Fixes #3814 .

With this PR, the `mockito.suppress.clinit` property can accept both an exact class name (e.g., `com.example.MyClass`) or a package wildcard ending in `.*` (e.g., `com.example.*`) that matches the package and all sub-packages.

In addition to unit tests, I've also installed locally and created a simple project using both the features.

## Checklist

### DIFF
--- a/mockito-core/src/main/java/org/mockito/internal/PremainAttach.java
+++ b/mockito-core/src/main/java/org/mockito/internal/PremainAttach.java
@@ -7,9 +7,9 @@ package org.mockito.internal;
 import org.mockito.internal.creation.bytebuddy.ClinitSuppressionTransformer;
 
 import java.lang.instrument.Instrumentation;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.function.Predicate;
 
 /**
  * Allows users to specify Mockito as a Java agent where the {@link Instrumentation}
@@ -28,10 +28,10 @@ public class PremainAttach {
         if (PremainAttach.instrumentation != null) {
             return;
         }
-        Set<String> classes =
-                parseSuppressClinitProperty(System.getProperty(SUPPRESS_CLINIT_PROPERTY));
-        if (!classes.isEmpty()) {
-            instrumentation.addTransformer(new ClinitSuppressionTransformer(classes), false);
+        Predicate<String> shouldSuppress =
+                buildClinitPredicate(System.getProperty(SUPPRESS_CLINIT_PROPERTY));
+        if (shouldSuppress != null) {
+            instrumentation.addTransformer(new ClinitSuppressionTransformer(shouldSuppress), false);
         }
         PremainAttach.instrumentation = instrumentation;
     }
@@ -41,23 +41,47 @@ public class PremainAttach {
     }
 
     /**
-     * Parses a comma-separated list of class names into a set, trimming whitespace from each
-     * entry and ignoring empty entries.
+     * Parses the {@code mockito.suppress.clinit} property value into a predicate over
+     * fully-qualified class names. The property is a comma-separated list; each entry is
+     * either an exact class name ({@code com.example.MyClass}) or a package wildcard ending
+     * in {@code .*} ({@code com.example.*}) that matches the package and all sub-packages.
+     * Whitespace is trimmed; empty entries are ignored.
      *
      * @param property the property value, may be {@code null}
-     * @return an unmodifiable set of class names
+     * @return the predicate, or {@code null} if no entries are configured
      */
-    static Set<String> parseSuppressClinitProperty(String property) {
+    static Predicate<String> buildClinitPredicate(String property) {
         if (property == null || property.trim().isEmpty()) {
-            return Collections.emptySet();
+            return null;
         }
-        Set<String> result = new HashSet<>();
+        Set<String> classes = new HashSet<>();
+        Set<String> packages = new HashSet<>();
         for (String name : property.split(",")) {
             String trimmed = name.trim();
-            if (!trimmed.isEmpty()) {
-                result.add(trimmed);
+            if (trimmed.isEmpty()) {
+                continue;
+            }
+            if (trimmed.endsWith(".*")) {
+                // Drop only the trailing '*' and keep the preceding '.' so that
+                // "com.example." correctly rejects "com.examples.Foo".
+                packages.add(trimmed.substring(0, trimmed.length() - 1));
+            } else {
+                classes.add(trimmed);
             }
         }
-        return Collections.unmodifiableSet(result);
+        if (classes.isEmpty() && packages.isEmpty()) {
+            return null;
+        }
+        return name -> {
+            if (classes.contains(name)) {
+                return true;
+            }
+            for (String prefix : packages) {
+                if (name.startsWith(prefix)) {
+                    return true;
+                }
+            }
+            return false;
+        };
     }
 }
--- a/mockito-core/src/main/java/org/mockito/internal/PremainAttach.java
+++ b/mockito-core/src/main/java/org/mockito/internal/PremainAttach.java
@@ -31,6 +31,10 @@ public class PremainAttach {
         Predicate<String> shouldSuppress =
                 buildClinitPredicate(System.getProperty(SUPPRESS_CLINIT_PROPERTY));
         if (shouldSuppress != null) {
+            System.err.println(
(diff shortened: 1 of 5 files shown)


mirkoalicastro:
Hi @raphw . Any chance you could look at this? 🙏

raphw:
I have been thinking, but I am not sure if this is a good idea. New classes might do useful things in their initializers, like settings fields, and this could lead to unexpected breakage. I think this should list class names for this reason, not packages. As for the principle of least surpise. @mockito/developers What do you think?

jselbo:
I dunno, it seems harmless to me. This property is already kind of a power-user scenario so they probably know what they are doing.

bric3:
> As for the principle of least surpise. @mockito/developers What do you think?

> I dunno, it seems harmless to me. This property is already kind of a power-user scenario so they probably know what they are doing.

I think #3801 is already somewhat already in that area. I think having a wildcard makes sense, and barely change that feature. Also, like Joshua mentioned, it's kind of a niche feature.

However, I do not think the whole feature is not harmless either ; while the mechnism is explicit in the build descriptor via a system property, it's not guaranteed to be visible enough. So I do think Mockito should log something, there's no `println` in the original PR. That feature is sufficiently hidden that it should print something at least for late contributors to such a code base.

Given it happens in the premain, I'm not sure we should use `MockitoLogger` (or other logger frameworks, this could cause too early class initialization), so when adding the log only rely on `println`.

mirkoalicastro:
Thanks everyone for taking the time to review.

@raphw I think that those are fair concerns. However, I personally think that the risk is very low since users need to explicitly set the property with a wildcard pattern, so it's unlikely someone would accidentally incur into it. Also, in legacy codebases, listing all the individual classes that need suppression can become fragile as new classes are added, in addition to being verbose.

In general, I find it a great suggestion about adding a log line when a class initializer is suppressed to improve visibility.

raphw:
For my part, we can merge this than. There is no best solution, and I expect that people mostly mock packages they own. Libraries avoid extensive logic in static initializers for a reason. If there is no other comment, I will merge this within this week.

mockito-core/src/main/java/org/mockito/internal/PremainAttach.java
```diff
@@ -41,23 +41,47 @@ public static Instrumentation getInstrumentation() {
...
             String trimmed = name.trim();
-            if (!trimmed.isEmpty()) {
-                result.add(trimmed);
+            if (trimmed.isEmpty()) {
+                continue;
             }
+            if (trimmed.endsWith(".*")) {
+                // Drop only the trailing '*' and keep the preceding '.' so that
+                // "com.example." correctly rejects "com.examples.Foo".
+                packages.add(trimmed.substring(0, trimmed.length() - 1));
```
bric3, line 71: **suggestion:** Maybe, update the `Mockito.java` documentation section. Eventually noting that globs in the middle are not supported.

mockito-core/src/test/java/org/mockito/internal/PremainAttachTest.java
```diff
@@ -6,52 +6,92 @@
...
+                        "com.example.ClassA, com.package.*, com.example.ClassB");
 
-        assertThat(result).isEmpty();
+        assertThat(matcher).isNotNull();
+        assertThat(matcher.test("com.example.ClassA")).isTrue();
+        assertThat(matcher.test("com.example.ClassB")).isTrue();
+        assertThat(matcher.test("com.package.Anything")).isTrue();
+        assertThat(matcher.test("com.package.nested.Thing")).isTrue();
+        assertThat(matcher.test("com.example.ClassC")).isFalse();
+        assertThat(matcher.test("com.packages.Other")).isFalse();
```
bric3: **suggestion:** Maybe a add a test, that covers today's unsupported syntax, like `com.*.bar.*`.

mockito-core/src/test/java/org/mockito/internal/PremainAttachTest.java
```diff
@@ -6,52 +6,92 @@
...
+        assertThat(matcher.test("org.example.Foo")).isFalse();
     }
 
     @Test
-    public void parse_suppress_clinit_property_returns_empty_when_blank() {
-        Set<String> result = PremainAttach.parseSuppressClinitProperty("   ");
+    public void parse_suppress_clinit_property_mixes_classes_and_packages() {
+        Predicate<String> matcher =
+                PremainAttach.buildClinitPredicate(
+                        "com.example.ClassA, com.package.*, com.example.ClassB");
```
bric3: **nitpick:** `com.package` would be an invalid `package` in Java.

mockito-core/src/main/java/org/mockito/internal/PremainAttach.java
```diff
@@ -28,10 +28,10 @@ public static void premain(String arg, Instrumentation instrumentation) {
...
             return;
         }
-        Set<String> classes =
-                parseSuppressClinitProperty(System.getProperty(SUPPRESS_CLINIT_PROPERTY));
-        if (!classes.isEmpty()) {
-            instrumentation.addTransformer(new ClinitSuppressionTransformer(classes), false);
+        Predicate<String> shouldSuppress =
+                buildClinitPredicate(System.getProperty(SUPPRESS_CLINIT_PROPERTY));
+        if (shouldSuppress != null) {
+            instrumentation.addTransformer(new ClinitSuppressionTransformer(shouldSuppress), false);
```
bric3, line 38: **suggestion:** I suggest to add the println here, to avoid spamming the console. Just emitting the the value of the system property should be enough in my opinion, something like: `Suppressing class initializer for '${System.getProperty(SUPPRESS_CLINIT_PROPERTY)}' (see mockito.suppress.clinit system property)`

So the user sees the configuration point, it's value, and what it does.

bric3 (approved):
I pre-approve, but please add the println statement.

mockito-core/src/main/java/org/mockito/internal/PremainAttach.java
mirkoalicastro: good call-out, done

mockito-core/src/test/java/org/mockito/internal/PremainAttachTest.java
mirkoalicastro: added

mockito-core/src/test/java/org/mockito/internal/PremainAttachTest.java
mirkoalicastro: changed

mockito-core/src/main/java/org/mockito/internal/PremainAttach.java
mirkoalicastro: I tried this out in another project using Maven Surefire and got this warning:

```[WARNING] Corrupted channel by directly writing to native stream in forked JVM 1```

And the message didn't show up. 

After checking, Surefire uses stdout as its communication protocol with the forked JVM: https://maven.apache.org/surefire/maven-surefire-plugin/faq.html

`System.err.println` could work as a workaround (stderr isn't used by the protocol), but not sure that is appropriate. I'll do more research on this later, but let me know what you think if you have other ideas.

bric3: > After checking, Surefire uses stdout as its communication protocol with the forked JVM: https://maven.apache.org/surefire/maven-surefire-plugin/faq.html

🤦 

OK let's use stderr

mirkoalicastro: Fair enough. Added!

bric3: thx

bric3 (approved):

PR #3821, opened 2026-05-16
Add `captorForClass` static alias of `ArgumentCaptor.forClass`

### LINKED ISSUE #3820: Add plain static alias of captorForClass for static imports
Static imports with plain name, regardless of coding conventions, may help.

Check the following code snippet for a book's coding block whose all calls are statically imported for the brevity.

```java
        final var inOrder = inOrder(subscriber);
        inOrder.verify(subscriber, times(1)).onSubscribe(notNull());
        final var elementCaptor = forClass(Byte.class); // <<<<<<<<<<<<<<<<<<
        inOrder.verify(subscriber, times(HelloWorld.BYTES)).onNext(elementCaptor.capture());
        inOrder.verify(subscriber, times(1)).onComplete();
        inOrder.verifyNoMoreInteractions();
        final var expected = hello_world_byte_array();
        final var elements = elementCaptor.getAllValues();
        assertEquals(expected.length, elements.size());
        for (int i = 0; i < elements.size(); i++) {
            assertEquals(expected[i], elements.get(i));
        } // @formatter:on
```

It will be helpful when the `ArgumentCaptor` has `captorForClass`.

This is not a redundant signature.

Even the JUnit's [`Arguments`](https://docs.junit.org/6.2.0/api/org.junit.jupiter.params/org/junit/jupiter/params/provider/Arguments.html) class added `arguments(...)` method analogues to the `of(...)` method.

A verify thin wrapper will be ok.

```java
public static <U, S extends U> ArgumentCaptor<U> captorForClass(Class<S> clazz) {
    return forClass(clazz);
}
```

onacit:
Introduces ArgumentCaptor.captorForClass(Class) as a thin delegate to ArgumentCaptor.forClass(Class). The new name reads well when statically imported, where the bare forClass(...) reference is ambiguous and does not convey that an ArgumentCaptor is being built.

This follows the same precedent set by Arguments.arguments(...) in JUnit Jupiter, which was added as a more expressive static-import friendly alias of Arguments.of(...).

Example - before, with everything statically imported the construction site is ambiguous:

    import static org.mockito.ArgumentCaptor.forClass;

    final var elementCaptor = forClass(Byte.class); // forClass of what?
    inOrder.verify(subscriber, times(HelloWorld.BYTES)).onNext(elementCaptor.capture());

After:

    import static org.mockito.ArgumentCaptor.captorForClass;

    final var elementCaptor = captorForClass(Byte.class);
    inOrder.verify(subscriber, times(HelloWorld.BYTES)).onNext(elementCaptor.capture());

forClass is preserved unchanged, so existing call sites are not affected. A unit test mirroring the existing forClass coverage is added in ArgumentCaptorTest.

Fixes #3820

Checklist

 - [x] Read the contributing guide
 - [x] The pull request follows coding style (run ./gradlew spotlessApply for auto-formatting)
 - [x] Mention Fixes #<issue number> in the description if relevant
 - [x] At least one commit should end with Fixes #<issue number> if relevant

### DIFF
--- a/mockito-core/src/main/java/org/mockito/ArgumentCaptor.java
+++ b/mockito-core/src/main/java/org/mockito/ArgumentCaptor.java
@@ -149,6 +149,34 @@ public class ArgumentCaptor<T> {
     }
 
     /**
+     * Static alias of {@link #forClass(Class)}.
+     * <p>
+     * Intended for use with static imports where the plain {@code forClass} name is
+     * ambiguous or insufficiently expressive. The body delegates to
+     * {@link #forClass(Class)} and behaves identically.
+     * <p>
+     * Example usage:
+     *
+     * <pre class="code"><code class="java">
+     *   import static org.mockito.ArgumentCaptor.captorForClass;
+     *
+     *   ArgumentCaptor&lt;Person&gt; argument = captorForClass(Person.class);
+     *   verify(mock).doSomething(argument.capture());
+     *   assertEquals("John", argument.getValue().getName());
+     * </code></pre>
+     *
+     * @param clazz Type matching the parameter to be captured.
+     * @param <S> Type of clazz
+     * @param <U> Type of object captured by the newly built ArgumentCaptor
+     * @return A new ArgumentCaptor
+     * @see #forClass(Class)
+     * @since 5.24.0
+     */
+    public static <U, S extends U> ArgumentCaptor<U> captorForClass(Class<S> clazz) {
+        return forClass(clazz);
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


jselbo:
Seems reasonable to me. In the Kotlin repo we have `val myCaptor = argumentCaptor<Foo>` "static" function as a point of comparison. This is more readable when statically imported.

Though I wonder if it would make more sense to expose this on `Mockito` class? I think it's common for users to `import static org.mockito.Mockito.*` so they would discover this more easily.

PR #3823, opened 2026-05-20
Improve Gradle agent docs for configuration cache

### LINKED ISSUE #3822: Gradle documentation for setting up instrumentation with Gradle configuration cache
Hi everyone!

I noticed that the current documentation for setting up the `mockitoAgent` with Gradle will not work when you enable the configuration cache (e.g. by adding `org.gradle.configuration-cache=true` to your `gradle.properties`).

I get the following error message:
```
Calculating task graph as no cached configuration is available for tasks: clean spotlessApply build
> Task :test FAILED

1 problem was found storing the configuration cache.
- Task `:test` of type `org.gradle.api.tasks.testing.Test`: invocation of 'configurations' references a Gradle script object from a Groovy closure at execution time, which is unsupported with the configuration cache.
  See https://docs.gradle.org/9.5.1/userguide/configuration_cache_requirements.html#config_cache:requirements:gradle_model_types
```

However, this approach seems to work:

```
configurations {
    mockitoAgent
}

test {
    useJUnitPlatform()
    jvmArgumentProviders.add(new CommandLineArgumentProvider() {
        @InputFiles
        @PathSensitive(PathSensitivity.RELATIVE)
        FileCollection mockitoAgent = configurations.mockitoAgent

        @Override
        Iterable<String> asArguments() {
            ["-javaagent:${mockitoAgent.asPath}", "-XX:+EnableDynamicAgentLoading"]
        }
    })
}
```

Since I'm not the greatest expert in Gradle, I wanted to share my solution. Maybe this approach can be improved? Maybe you want to use it to improve the [Javadoc documentation](https://github.com/mockito/mockito/blob/414a99970fe2448f203075e35d1eb40d6d0492b2/mockito-core/src/main/java/org/mockito/Mockito.java)?

I use Gradle 9.5.1 + Mockito 5.23.0

Thanks!

Christopher

codingkiddo:
## Summary

This updates the Groovy Gradle example for explicitly attaching Mockito as a Java agent to use `CommandLineArgumentProvider`.

The previous example accessed `configurations.mockitoAgent` directly from the test task configuration, which can fail when Gradle configuration cache is enabled. The revised example follows the provider-based pattern suggested in #3822 and keeps the documented setup compatible with configuration-cache usage.

The surrounding wording is also adjusted so it no longer says that `CommandLineArgumentProvider` is omitted from the examples.

Fixes #3822

## Checklist

### DIFF
--- a/mockito-core/src/main/java/org/mockito/Mockito.java
+++ b/mockito-core/src/main/java/org/mockito/Mockito.java
@@ -204,7 +204,7 @@ import java.util.function.Function;
  * <p>
  * To explicitly attach Mockito during test execution, the library's jar file needs to be specified as <code>-javaagent</code>
  * as an argument to the executing JVM. To enable this in Gradle, the following example adds Mockito to all test
- * tasks using <strong>Kotlin DSL</strong>. Although omitted for simplicity, using a <code>CommandLineArgumentProvider</code> is recommended by Gradle to ensure task relocatability (<a href="https://docs.gradle.org/current/userguide/caching_java_projects.html#dealing_with_file_paths">documentation</a>):
+ * tasks using <strong>Kotlin DSL</strong>. Using a <code>CommandLineArgumentProvider</code> is recommended by Gradle to ensure task relocatability (<a href="https://docs.gradle.org/current/userguide/caching_java_projects.html#dealing_with_file_paths">documentation</a>):
  *
  * <pre class="code"><code class="kotlin">
  * val mockitoAgent = configurations.create("mockitoAgent")
@@ -233,7 +233,16 @@ import java.util.function.Function;
  * }
  * tasks {
  *     test {
- *         jvmArgs += "-javaagent:${configurations.mockitoAgent.asPath}"
+ *         jvmArgumentProviders.add(new CommandLineArgumentProvider() {
+ *             @InputFiles
+ *             @PathSensitive(PathSensitivity.RELATIVE)
+ *             FileCollection mockitoAgent = configurations.mockitoAgent
+ *
+ *             @Override
+ *             Iterable<String> asArguments() {
+ *                 ["-javaagent:${mockitoAgent.asPath}"]
+ *             }
+ *         })
  *     }
  * }
  * </code></pre>


mockito-core/src/main/java/org/mockito/Mockito.java
```diff
@@ -204,7 +204,7 @@
  * <p>
  * To explicitly attach Mockito during test execution, the library's jar file needs to be specified as <code>-javaagent</code>
  * as an argument to the executing JVM. To enable this in Gradle, the following example adds Mockito to all test
- * tasks using <strong>Kotlin DSL</strong>. Although omitted for simplicity, using a <code>CommandLineArgumentProvider</code> is recommended by Gradle to ensure task relocatability (<a href="https://docs.gradle.org/current/userguide/caching_java_projects.html#dealing_with_file_paths">documentation</a>):
+ * tasks using <strong>Kotlin DSL</strong>. Using a <code>CommandLineArgumentProvider</code> is recommended by Gradle to ensure task relocatability (<a href="https://docs.gradle.org/current/userguide/caching_java_projects.html#dealing_with_file_paths">documentation</a>):
```
ccudennec-otto, line 207: I totally missed that sentenced because I was too focused on the black code section in the docs. 😆 (see generated documentation [here](https://javadoc.io/doc/org.mockito/mockito-core/latest/org.mockito/org/mockito/Mockito.html#0.3))
I think your PR makes this much clearer 👍 

Not sure if actually both code examples should be included in the docs?

raphw:
Makes sense to make this the default recommendation to copy-paste.

ccudennec-otto:
Thanks for the PR + adding the example to the docs! 🌻

codingkiddo:
Thanks for checking this!

I kept both examples because the existing documentation already included Kotlin DSL and Groovy DSL sections, and this PR only updates the Groovy example to use the configuration-cache-friendly `CommandLineArgumentProvider` pattern.

That said, I’m happy to adjust the structure if maintainers prefer a simpler docs section — for example, keeping only one example, or updating both Kotlin and Groovy examples to use the same provider-based approach.

PR #3827, opened 2026-05-26
Fix NotAMockException when using mockSingleton with MockitoJUnit

### LINKED ISSUE #3826: NotAMockException when using mockSingleton with MockitoSession or MockitoJUnit
Example:

```
    @Rule public MockitoRule rule = MockitoJUnit.rule();

    @Test
    public void myTest() {
        try (MockedSingleton<MyEnum> ignored = Mockito.mockSingleton(MyEnum.A)) {}
    }
```
Results in:

```
Argument passed to Mockito.mockingDetails() should be a mock, but is an instance of class com.example.MyEnum!
org.mockito.exceptions.misusing.NotAMockException: Argument passed to Mockito.mockingDetails() should be a mock, but is an instance of class com.example.MyEnum!
...
```

jselbo:
Fixes #3826

The cause of this is that `mockSingleton` implementation overloaded the existing `MockingProgress#mockingStarted` callback. When using `MockitoJUnit` rule, `UniversalTestListener` uses this callback to track created mocks. If the scoped singleton mock is closed, then the `testFinished` downstream logic would look up mocking details on the singleton which is no longer a mock leading to this exception.

It was probably a mistake to reuse the `mockingStarted` callback originally - let's mimic the `mockStatic` path which uses its own dedicated callback for reporting created static mocks.

Note that the unused-stubbing detector stuff (used with "strict stubs" strictness) currently does not support static mocks, and it does not support singleton mocks either. That seems okay; a future PR can add support if desired.

_____________________

## Checklist

### DIFF
--- a/mockito-core/src/main/java/org/mockito/internal/MockitoCore.java
+++ b/mockito-core/src/main/java/org/mockito/internal/MockitoCore.java
@@ -102,7 +102,7 @@ public class MockitoCore {
         checkDoNotMockAnnotation(creationSettings);
         MockMaker.SingletonMockControl<T> control = createSingletonMock(instance, creationSettings);
         control.enable();
-        mockingProgress().mockingStarted(instance, creationSettings);
+        mockingProgress().singletonMockingStarted(instance, creationSettings);
         return new MockedSingletonImpl<>(control);
     }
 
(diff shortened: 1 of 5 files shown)


raphw:
Good catch, thanks!

PR #3829, opened 2026-05-29
Fix clearInlineMocks not de-registering singleton mocks

mirkoalicastro:
## Description

**Problem:** singleton mocks were leaking since clearInlineMocks() didn't clear their mocking state.

clearInlineMocks() and clearInlineMock(...) left singleton mocks registered, so re-mocking the same instance threw "already registered as a mock" exception. 

This PR fixes it:
- clearInlineMocks() clears the mockedSingletons registry
- ~clearInlineMock(...) removes the instance from that registry~: discussed and agreed to not cover.

Without the changes, both the newly-introduce unit tests would fail.

## Checklist

### DIFF
--- a/mockito-core/src/main/java/org/mockito/internal/creation/bytebuddy/InlineDelegateByteBuddyMockMaker.java
+++ b/mockito-core/src/main/java/org/mockito/internal/creation/bytebuddy/InlineDelegateByteBuddyMockMaker.java
@@ -581,10 +581,6 @@ class InlineDelegateByteBuddyMockMaker
             for (Map<Class<?>, ?> entry : mockedStatics.getBackingMap().target.values()) {
                 entry.remove(mock);
             }
-        } else if (isRegisteredSingleton(mock)) {
-            for (Map<Object, ?> entry : mockedSingletons.getBackingMap().target.values()) {
-                entry.remove(mock);
-            }
         } else {
             mocks.put(
                     mock,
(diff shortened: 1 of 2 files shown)


mockito-core/src/main/java/org/mockito/internal/creation/bytebuddy/InlineDelegateByteBuddyMockMaker.java
```diff
@@ -581,6 +581,10 @@ public void clearMock(Object mock) {
             for (Map<Class<?>, ?> entry : mockedStatics.getBackingMap().target.values()) {
                 entry.remove(mock);
             }
+        } else if (isRegisteredSingleton(mock)) {
+            for (Map<Object, ?> entry : mockedSingletons.getBackingMap().target.values()) {
+                entry.remove(mock);
```
jselbo: since `clearMock(mock)` is called for every normal object mock and `isRegisteredSingleton` checks the singleton map on every thread it's been instantiated on, I'm worried about the cost of this new block. It also doesn't seem critical - clearing the map from `clearAllMocks()` seems sufficient to achieve the goal of this PR.

The normal way of "clearing" a scoped mock is by closing its handle (the `MockedSingleton`) - I think calling `clearMock(MyEnum.A)` is not a likely usage.

What do you think about removing this block?

jselbo:
Also I am curious, are you using the new `mockSingleton` API? Or did AI find this bug?

mockito-core/src/main/java/org/mockito/internal/creation/bytebuddy/InlineDelegateByteBuddyMockMaker.java
mirkoalicastro: Initially, I added it for symmetry. But I tend to agree with you. I pushed the changes. Thanks!

mirkoalicastro:
> Also I am curious, are you using the new `mockSingleton` API? Or did AI find this bug?

Actually, I was curious about your last commit (https://github.com/mockito/mockito/commit/af081a35a9c85f10158d69b8109a57f8af04a477), so I started looking into it and found the discrepancy in the clearMock. Then AI highlighted the other discrepancy 😅

jselbo (approved):

PR #3832, opened 2026-06-12
Replace constant hashcode

mirkoalicastro:
A few classes (InterceptedInvocation, SerializableMethod, Equals) returned a constant number (1) as their hashcode implementation, collapsing every instance into one bucket and degrading HashMap/HashSet lookups to linear complexity.

This PR fixes it (and removes corresponding `TODO` from the code) so that each hashcode derives from the same fields taken into consideration for their equals. 

## Checklist

### DIFF
--- a/mockito-core/src/main/java/org/mockito/internal/invocation/InterceptedInvocation.java
+++ b/mockito-core/src/main/java/org/mockito/internal/invocation/InterceptedInvocation.java
@@ -168,9 +168,10 @@ public class InterceptedInvocation implements Invocation, VerificationAwareInvoc
 
     @Override
     public int hashCode() {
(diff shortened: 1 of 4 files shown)


raphw:
If I remember correctly, the problem is hashing class identity. This will break with serialization. Serializing on strings should be fine, though, that's a thing we should certainly improve.

mirkoalicastro:
> If I remember correctly, the problem is hashing class identity. This will break with serialization. Serializing on strings should be fine, though, that's a thing we should certainly improve.

You're right, and I learned something here 😄 

I've changed it to hash on the class names instead of the Class objects to achieve a portable hashCode implementation. The equals contract still holds: equal methods resolve to the same classes and therefore the same names.

Thanks for the catch!

Let me know your thoughts.

PR #3835, opened 2026-06-18
Fixes #2740 : Add mockMaker option to @Spy annotation

### LINKED ISSUE #2740: Add option to specify mockmaker for `@Spy`
To expand on #2626, we should also support this option for `@Spy`, which will require updating our spy annotation injection engine.

mirkoalicastro:
Spy had no way to select a MockMaker per field, so the annotation-based spies always used the globally configured maker. Mock supports mockMaker since 4.8.0, so this PR closes that gap by implementing the same for Spy.

In addition to that, removing TODOs coming from the duplication of the spy-creation logic, extracted into SpyAnnotationUtil.

Example
```
@Spy(mockMaker = MockMakers.SUBCLASS) Foo spyOnFoo = new Foo("arg");
```

Fixes #2740 

## Checklist

### DIFF
--- a/mockito-core/src/main/java/org/mockito/Spy.java
+++ b/mockito-core/src/main/java/org/mockito/Spy.java
@@ -12,6 +12,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
 import org.mockito.junit.MockitoJUnitRunner;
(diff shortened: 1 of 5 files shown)


raphw:
Good catch, and looks good to me.

PR #3840, opened 2026-07-06
Migrate extensions-tests to JUnit Jupiter

mirkoalicastro:
The extensions tests pulled hamcrest-core transitively via JUnit 4. There was a TODO to exclude it (linked to #154) that was blocked on that dependency.
This PR excludes it migrating the module's JUnit 4 unit tests to JUnit Jupiter, dropping its dependencies, and replacing the stale TODO with an active exclude of the hamcrest dependency.

### DIFF
--- a/mockito-core/src/testFixtures/java/org/mockitoutil/TestBase5.java
+++ b/mockito-core/src/testFixtures/java/org/mockitoutil/TestBase5.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright (c) 2026 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
+package org.mockitoutil;
+
+import static org.mockito.Mockito.mock;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.PrintStream;
+
+import org.assertj.core.api.Condition;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.mockito.MockitoAnnotations;
+import org.mockito.StateMaster;
+import org.mockito.internal.MockitoCore;
+import org.mockito.internal.configuration.ConfigurationAccess;
+import org.mockito.internal.debugging.LocationFactory;
+import org.mockito.internal.invocation.InterceptedInvocation;
+import org.mockito.internal.invocation.InvocationBuilder;
+import org.mockito.internal.invocation.InvocationMatcher;
+import org.mockito.internal.invocation.SerializableMethod;
+import org.mockito.internal.invocation.mockref.MockStrongReference;
+import org.mockito.invocation.Invocation;
+
+/**
+ * JUnit 5 (Jupiter) counterpart of {@link TestBase}.
+ */
+public class TestBase5 {
+
+    /**
+     * Condition to be used with AssertJ
+     */
+    public static Condition<Throwable> hasMessageContaining(final String substring) {
+        return new Condition<Throwable>() {
+            @Override
+            public boolean matches(Throwable e) {
+                return e.getMessage().contains(substring);
+            }
+        };
+    }
+
+    @AfterEach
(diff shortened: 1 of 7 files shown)


jselbo:
Sorry, I don't really see the value in this. I don't see the problem with these tests depending on Hamcrest via JUnit4. Maybe I am missing the context from ten years ago, but I'd prefer just removing the TODO.

If we want to migrate all integration tests to Jupiter JUnit that's another conversation

mirkoalicastro:
> Sorry, I don't really see the value in this. I don't see the problem with these tests depending on Hamcrest via JUnit4. Maybe I am missing the context from ten years ago, but I'd prefer just removing the TODO.
> 
> If we want to migrate all integration tests to Jupiter JUnit that's another conversation

I get your point about the TODO. I agree that we could simply drop it, but I see it as good hygiene to get rid of JUnit 4 (maintenance mode), and hamcrest also goes away as a side effect. I can retitle the PR as a Jupiter migration instead of removal of hamcrest via the Jupiter migration. The changes stay the same, but it better reflects the PR.

mockito-integration-tests/extensions-tests/src/test/java/org/mockitousage/plugins/stacktrace/PluginStackTraceFilteringTest.java
```diff
@@ -4,30 +4,38 @@
...
+import org.mockito.internal.configuration.ConfigurationAccess;
+import org.mockito.junit.jupiter.MockitoExtension;
 import org.mockitousage.IMethods;
-import org.mockitoutil.TestBase;
-
-import static org.junit.Assert.fail;
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.Mockito.verify;
 
-public class PluginStackTraceFilteringTest extends TestBase {
```
jselbo, line 19: Not sure about this - why not make a JUnit5 compatible `TestBase` instead of duplicating the logic inline in this test?

jselbo:
I guess that makes sense. I still don't think the `exclude()` in the gradle file is necessary (does JUnit5 even include hamcrest?) so if we drop that and address my comment in `PluginStackTraceFilteringTest` I'm good with merging this.

mirkoalicastro:
> I guess that makes sense. I still don't think the `exclude()` in the gradle file is necessary (does JUnit5 even include hamcrest?) so if we drop that and address my comment in `PluginStackTraceFilteringTest` I'm good with merging this.

Makes sense to me. I doublecheck and indeed JUnit5 doesn't bring it, so I simply dropped it.

Which comment in `PluginStackTraceFilteringTest` you meant? I can't see it

jselbo (changes requested):

mockito-integration-tests/extensions-tests/src/test/java/org/mockitousage/plugins/stacktrace/PluginStackTraceFilteringTest.java
mirkoalicastro: Fair point! I've added a JUnit5 counterpart of TestBase.

mockito-core/src/testFixtures/java/org/mockitoutil/TestBase5.java
```diff
@@ -0,0 +1,129 @@
...
+
+    public static void makeStackTracesClean() {
+        ConfigurationAccess.getConfig().overrideCleansStackTrace(true);
+    }
+
+    public void resetState() {
+        new StateMaster().reset();
+    }
+
+    public static Invocation getLastInvocation() {
```
jselbo: Let's remove all the stuff that isn't used yet.

mirkoalicastro: Done!

jselbo (approved):

PR #3844, opened 2026-07-12
Print object fields via reflection when `toString()` is not implemented

asolntsev:
### Motivation
In my project, it often happens that value object doesn't have its own `toString()` method. And comes from a JAR that I cannot easily change. 

For such objects, it's highly inconvenient to debug failing tests showing just `MyClass@12345` in failure message.

@TimvdLippe  @raphw 

### The problem
In verification failure messages, Mockito uses `toString()` method to print out the argument object. For objects not having their own `toString()`, the useless description like "MyClass@12345" was generated (the default `Object.toString()` implementation).

### The solution
Now, ValuePrinter reflects over an object's declared fields when its class doesn't override `toString()`, thus producing readable description of the object. Too long field values are truncated.

Example of error message:

```java
mock.run(
    Credentials[username=john, password=secret]
);
-> at org.mockitousage.matchers.ReflectivePrintingTest.refEq_printsObjectFieldsUsingReflection(CustomToStringTest.java:53)
Actual invocations have different arguments:
mock.run(
    Credentials[username=usr, password=pwd]
);
-> at org.mockitousage.matchers.ReflectivePrintingTest.refEq_printsObjectFieldsUsingReflection(CustomToStringTest.java:51)
```

Where
```java
/**
 * Class without own "toString()" method
 */
 class Credentials {
     private String username;
     private String password;
}
```

## Checklist

 - [ ] At least one commit should end with `Fixes #<issue number>` _if relevant_

### DIFF
--- a/mockito-core/src/main/java/org/mockito/internal/matchers/text/ValuePrinter.java
+++ b/mockito-core/src/main/java/org/mockito/internal/matchers/text/ValuePrinter.java
@@ -4,9 +4,17 @@
  */
 package org.mockito.internal.matchers.text;
 
+import static java.util.Comparator.comparing;
+
 import java.lang.reflect.Array;
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.stream.Collectors;
+import org.mockito.internal.configuration.plugins.Plugins;
+import org.mockito.plugins.MemberAccessor;
 
 /**
  * Prints a Java object value in a way humans can read it neatly.
@@ -146,11 +154,78 @@ public class ValuePrinter {
         return sb.toString();
     }
 
+    private static final int MAX_FIELD_VALUE_LENGTH = 100;
+
+    /**
+     * Return string representation of given object.
+     * If 'value' has its own {@code toString()} method, it's called.
+     * If 'value' does not declare its own {@code toString()} then {@code Object.toString()} is useless for a
+     * human-readable diff, so fields are printed reflectively instead.
+     *
+     * @param value any java object
+     * @return String representation of this object (non-nullable)
+     */
     private static String descriptionOf(Object value) {
         try {
+            if (!declaresOwnToString(value.getClass())) {
+                return reflectiveToString(value);
+            }
             return String.valueOf(value);
         } catch (RuntimeException e) {
             return value.getClass().getName() + "@" + Integer.toHexString(value.hashCode());
         }
+    }
+
+    private static boolean declaresOwnToString(Class<?> type) {
+        try {
+            return type.getMethod("toString").getDeclaringClass() != Object.class;
+        } catch (NoSuchMethodException neverHappensBecauseObjectDeclaresToString) {
+            return false;
+        }
+    }
+
+    private static String reflectiveToString(Object value) {
+        MemberAccessor accessor = Plugins.getMemberAccessor();
+        StringBuilder result = new StringBuilder(value.getClass().getSimpleName()).append('[');
+        boolean first = true;
+        for (Class<?> type = value.getClass();
+                type != null && type != Object.class;
+                type = type.getSuperclass()) {
+            for (Field field : fields(type)) {
+                if (!first) {
+                    result.append(", ");
+                }
+                first = false;
+                result.append(field.getName())
+                        .append('=')
+                        .append(fieldValue(accessor, field, value));
+            }
+        }
+        return result.append(']').toString();
+    }
+
+    /**
+     * Sorts fields by name - this guarantees the same output every time, thus avoiding flaky tests.
+     */
+    private static Iterable<Field> fields(Class<?> type) {
+        return Arrays.stream(type.getDeclaredFields())
+                .filter(field -> !Modifier.isStatic(field.getModifiers()))
+                .filter(field -> !field.isSynthetic())
+                .sorted(comparing(Field::getName))
+                .collect(Collectors.toList());
+    }
+
+    private static String fieldValue(MemberAccessor accessor, Field field, Object target) {
+        try {
+            return truncate(String.valueOf(accessor.get(field, target)));
+        } catch (RuntimeException | IllegalAccessException fieldNotReadable) {
+            return "<unavailable>";
+        }
+    }
+
+    private static String truncate(String text) {
+        return text.length() > MAX_FIELD_VALUE_LENGTH
+                ? text.substring(0, MAX_FIELD_VALUE_LENGTH) + "..."
+                : text;
     }
 }
--- a/mockito-core/src/test/java/org/mockitousage/matchers/ReflectivePrintingTest.java
+++ b/mockito-core/src/test/java/org/mockitousage/matchers/ReflectivePrintingTest.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright (c) 2026 Mockito contributors
+ * This program is made available under the terms of the MIT License.
+ */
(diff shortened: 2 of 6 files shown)


mockito-core/src/main/java/org/mockito/internal/matchers/text/ValuePrinter.java
```diff
@@ -146,11 +154,78 @@ private static String printChar(char value) {
...
+
+    private static boolean declaresOwnToString(Class<?> type) {
+        try {
+            return type.getMethod("toString").getDeclaringClass() != Object.class;
+        } catch (NoSuchMethodException neverHappensBecauseObjectDeclaresToString) {
+            return false;
+        }
+    }
+
+    private static String reflectiveToString(Object value) {
```
jselbo, line 187: This is probably not nearly as robust as Apache commons `ReflectionToStringBuilder` but checking the source I don't think it would be trivial to pare down for this repo:
https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/builder/ReflectionToStringBuilder.java

mockito-core/src/main/java/org/mockito/internal/matchers/text/ValuePrinter.java
```diff
@@ -146,11 +154,78 @@ private static String printChar(char value) {
...
+        return result.append(']').toString();
+    }
+
+    /**
+     * Sorts fields by name - this guarantees the same output every time, thus avoiding flaky tests.
+     */
+    private static Iterable<Field> fields(Class<?> type) {
+        return Arrays.stream(type.getDeclaredFields())
+                .filter(field -> !Modifier.isStatic(field.getModifiers()))
+                .filter(field -> field.getName().indexOf('$') < 0)
```
jselbo: use `Field.isSynthetic` ?

mockito-core/src/main/java/org/mockito/internal/matchers/text/ValuePrinter.java
```diff
@@ -146,11 +154,78 @@ private static String printChar(char value) {
...
+                .sorted(comparing(Field::getName))
```
jselbo, line 214: why would the order of `type.getDeclaredFields()` be non-deterministic? Not sure why sorting is necessary

mockito-core/src/main/java/org/mockito/internal/matchers/text/ValuePrinter.java
```diff
@@ -146,11 +154,78 @@ private static String printChar(char value) {
...
+
+    private static String fieldValue(MemberAccessor accessor, Field field, Object target) {
+        try {
+            return truncate(String.valueOf(accessor.get(field, target)));
+        } catch (RuntimeException | IllegalAccessException fieldNotReadable) {
+            return "<unavailable>";
+        }
+    }
+
+    private static String truncate(String text) {
```
jselbo, line 226: what if the different value gets truncated? This wouldn't be very useful then. Ideally this would have some behavior to report only the differing fields, or ensure the difference isn't truncated

jselbo:
This seems reasonable at first glance, and low risk if it's only scoped to `refEq`. Still, I left some comments

mockito-core/src/main/java/org/mockito/internal/matchers/text/ValuePrinter.java
asolntsev: I think we don't need all the features of `ReflectionToStringBuilder`. 
I would say, it's not "robust", it just allows configuration: "skip null values" etc. We don't need it.

mockito-core/src/main/java/org/mockito/internal/matchers/text/ValuePrinter.java
asolntsev: Good point, fixed.

mockito-core/src/main/java/org/mockito/internal/matchers/text/ValuePrinter.java
asolntsev: Java reflection doesn't guarantee the order of fields/methods.

From `Class.getDeclaredFields` javadoc:

> The elements in the returned array are not sorted and are not in any particular order.

mockito-core/src/main/java/org/mockito/internal/matchers/text/ValuePrinter.java
asolntsev: While it's theoretically possible, it hopefully doesn't happen too often. 
Anyway, it will be better than now. :) 

Yes, I totally agree that ideally, Mockito should report the differences in the fields. For too long string, probably report the exact position of diff in the string. Maybe in another PR?..

asolntsev:
@jselbo It's scoped not only to `refEq`, but also to `eq` and `same`. BUT it's scoped for objects that don't have own `toString` method. So anyway, it's much better than previously :)

mockito-core/src/main/java/org/mockito/internal/matchers/text/ValuePrinter.java
jselbo: Looks like you need to push a new commit with the change

.gitignore
```diff
@@ -32,3 +32,5 @@ local.properties
 # Used to check reproducibility of Mockito jars
 checksums/
 /.tool-versions
+
+/.claude/settings.local.json
```
jselbo: please remove

mockito-core/src/test/java/org/mockitousage/matchers/ReflectivePrintingTest.java
```diff
@@ -0,0 +1,88 @@
...
+    private final LoginService mock = mock();
+
+    @Test
+    public void printsObjectFieldsUsingReflection() {
+        mock.run(actual);
+
+        assertThatThrownBy(() -> verify(mock).run(expected))
+                .isInstanceOf(AssertionError.class)
+                .hasMessageContaining("Argument(s) are different! Wanted:")
+                .hasMessageContaining("Credentials[password=secret, username=john]")
```
jselbo, line 48: I wonder if there is a privacy/security concern here. Suppose test data has real sensitive information (like reading some secret keys -- probably bad practice but I'm sure it happens). Now Mockito will print the private fields which would otherwise not be exposed if the class doesn't implement `toString()`. This could be a bad surprise for someone.

mockito-core/src/main/java/org/mockito/internal/matchers/text/ValuePrinter.java
asolntsev: Ooops :) 
Pushed.

.gitignore
asolntsev: No problems, I've removed the file.

P.S. Though, it's usually a common practice to ignore this file.

mockito-core/src/test/java/org/mockitousage/matchers/ReflectivePrintingTest.java
asolntsev: Valid concern.
But I would say it should not be a problem in tests.
1. Mockito is mostly used in unit-tests (and sometimes integration tests). Usually people use test data, not real sensitive data in tests.
2. And if they use some real passwords in tests, they should be ready for the risk that these passwords will appear in logs, test reports or CI notifications.
3. Usually Jenkins, GitHub Actions etc. do mask the secure parameters like password.

jselbo:
Ok I'm comfortable merging this. Thanks for this improvement!

jselbo (approved):

PR #3846, opened 2026-07-15
Update decommissioned s01.oss.sonatype.org snapshot URL to Central Portal

kobihikri:
Hi, and thank you for Mockito.

Two references still point at the decommissioned Sonatype OSSRH host `s01.oss.sonatype.org`, which Sonatype has retired in favour of the Central Portal:

- `README.md` (line 76): the "How to release" snapshot download URL.
- `buildSrc/src/main/kotlin/mockito.root.releasing-conventions.gradle.kts` (line 40): the "SNAPSHOTS released to:" log message.

Your build's actual publishing config has already migrated — the same file (line 66) defines `snapshotRepositoryUrl = uri("https://central.sonatype.com/repository/maven-snapshots/")`. So these two strings just lag the real endpoint. This PR updates both `s01.oss.sonatype.org/content/repositories/snapshots` references to `https://central.sonatype.com/repository/maven-snapshots`, matching your own `snapshotRepositoryUrl`.

For transparency: I used AI assistance to spot and draft this; I verified all three locations against the current source myself.

### DIFF
--- a/buildSrc/src/main/kotlin/mockito.root.releasing-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/mockito.root.releasing-conventions.gradle.kts
@@ -37,7 +37,7 @@ tasks {
                 logger.lifecycle(
                     """
                     RELEASE SUMMARY
-                      SNAPSHOTS released to: https://s01.oss.sonatype.org/content/repositories/snapshots/org/mockito/mockito-core
+                      SNAPSHOTS released to: https://central.sonatype.com/repository/maven-snapshots/org/mockito/mockito-core
                       Release to Maven Central: SKIPPED FOR SNAPSHOTS
                       Github releases: SKIPPED FOR SNAPSHOTS
                     """.trimIndent()
--- a/buildSrc/src/main/kotlin/mockito.root.releasing-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/mockito.root.releasing-conventions.gradle.kts
@@ -37,7 +37,7 @@ tasks {
                 logger.lifecycle(
                     """
                     RELEASE SUMMARY
-                      SNAPSHOTS released to: https://central.sonatype.com/repository/maven-snapshots/org/mockito/mockito-core
+                      SNAPSHOTS released to: https://central.sonatype.com/repository/maven-snapshots/org/mockito/mockito-core/
                       Release to Maven Central: SKIPPED FOR SNAPSHOTS
                       Github releases: SKIPPED FOR SNAPSHOTS
                     """.trimIndent()
--- a/buildSrc/src/main/kotlin/mockito.root.releasing-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/mockito.root.releasing-conventions.gradle.kts
@@ -37,7 +37,7 @@ tasks {
                 logger.lifecycle(
                     """
                     RELEASE SUMMARY
-                      SNAPSHOTS released to: https://central.sonatype.com/repository/maven-snapshots/org/mockito/mockito-core/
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/kotlin/mockito.root.releasing-conventions.gradle.kts
```diff
@@ -37,7 +37,7 @@ tasks {
                 logger.lifecycle(
                     """
                     RELEASE SUMMARY
-                      SNAPSHOTS released to: https://s01.oss.sonatype.org/content/repositories/snapshots/org/mockito/mockito-core
+                      SNAPSHOTS released to: https://central.sonatype.com/repository/maven-snapshots/org/mockito/mockito-core
```
jselbo: The new links don't work - they go to a 404 page.

I recently fixed these links in the mockito-kotlin repo. I believe it should be:

https://central.sonatype.com/artifact/org.mockito/mockito-core

jselbo (changes requested):

kobihikri:
Thanks @jselbo, and sorry for the slow turnaround — you were right, the links did 404. Fixed in 5c09d00.

The cause turned out to be a missing trailing slash rather than the wrong path. I checked all four candidates before changing anything:

| URL | HTTP |
|---|---|
| `…/repository/maven-snapshots/org/mockito/mockito-core` (what I filed) | 404 |
| `…/repository/maven-snapshots/org/mockito/mockito-core/` | 200 |
| `https://central.sonatype.com/artifact/org.mockito/mockito-core` (your suggestion) | 200 |
| `https://s01.oss.sonatype.org/…/mockito-core` (the decommissioned one) | 404 |

I went with the trailing slash rather than your suggestion for one reason, and I'd rather say it than silently differ: the `artifact/…` page is the **release** artifact view, while both lines I touched describe where **snapshots** are published ("released as \`-SNAPSHOT\` version to Sonatype snapshot repo at …" and "SNAPSHOTS released to: …"). The trailing-slash URL serves the actual snapshot directory listing for `org/mockito/mockito-core`, and it matches the form this file already uses for `snapshotRepositoryUrl` on line 66.

That said, you know this project's docs far better than I do — if you'd prefer the browsable `central.sonatype.com/artifact/org.mockito/mockito-core` page in one or both spots, say the word and I'll switch it right away. Happy either way.

Disclosure: I used AI assistance (Claude Opus 5) to help identify this and draft the change. I verified every URL myself with `curl` and take responsibility for it.

jselbo:
Ok if I open https://central.sonatype.com/repository/maven-snapshots/org/mockito/mockito-core/ in a browser I still see an error though it's no longer a 404:
> This maven2 hosted repository is not directly browseable at this URL.

kobihikri:
Yeah, you're right — I only checked the status code and never actually opened it in a browser. Sorry about that.

Switched both lines to your URL in c75da06.

My reasoning for the trailing slash was that these two lines are about where snapshots go, and the `artifact/…` page is the release view. But that distinction doesn't count for much if the page just tells you it isn't browseable. Yours is the one someone following the link can actually use.

Same disclosure as before: I used AI assistance (Claude Opus 5) to help find this and draft the change, and I checked the URLs myself.

jselbo:
I'd prefer to interact with a human, not Claude.

I see now the context of the log message and readme are specifically about snapshots, but based on a related page
https://central.sonatype.com/service/rest/repository/browse/maven-snapshots 

I see
> Due to an ongoing issue with the underlying SNAPSHOT hosting service, browsing access for SNAPSHOT releases is unavailable until further notice. You can continue to publish and consume SNAPSHOT releases as usual, but you cannot browse them through the UI.

I am not even sure this project still publishes snapshots to Maven. I'm going to close this since this PR doesn't seem very thoughtful about Mockito specifically.

Issue #757, opened 2016-11-15
"mock-maker-inline" prevents collection of code coverage for spys


Godin:
Hi,

Being one of [JaCoCo (Java Code Coverage Library)](http://www.jacoco.org/jacoco/index.html) developers, I was curious about interoperability of Mockito 2.2.x and JaCoCo on-the-fly instrumentation (using agent), so did some tests and came across following behavior:

with `mock-maker-inline`:
![mock-maker-inline](https://cloud.githubusercontent.com/assets/138671/20305994/94290260-ab38-11e6-96ed-5d274d06cdda.png)
without:
![normal](https://cloud.githubusercontent.com/assets/138671/20305987/8c27b174-ab38-11e6-9230-6fccaee64541.png)

Here is small example that reproduces this behavior - [example.zip](https://github.com/mockito/mockito/files/592003/example.zip) , `mvn clean verify` to run tests and generate JaCoCo report.

* Mockito 2.2.16
* JaCoCo 0.7.7
* JDK 8u102

As a workaround possible to use JaCoCo offline instrumentation (`mvn clean verify -Djacoco-offline` on the given example), but we don't recommend this mode and it is not available nor in EclEmma (plugin for Eclipse based on JaCoCo), nor when using JaCoCo in IntellIJ IDEA.

Also should be noted that the same happens for IntelliJ IDEA's own built-in coverage engine:

![intellij-mock-maker-inline](https://cloud.githubusercontent.com/assets/138671/20306440/8d6f6eb2-ab3a-11e6-9d07-e9a0108f8794.png)
vs
![intellij-normal](https://cloud.githubusercontent.com/assets/138671/20306446/93948e44-ab3a-11e6-9b73-ee4d78d3d55f.png)

Unfortunately, I didn't had time (yet) to dig deeper into code of Mockito to understand why this happens. Best guess based on investigation of similar issue with PowerMock: Mockito's agent bypasses other agents by reading class files from disk.

TimvdLippe:
@Godin Your initial description is about code coverage in general, but this issue only affects mocks created with `@Spy` or `Mockito.spy` right?

TimvdLippe:
On a different note: my expectation for the code coverage would be that `getName` does not have any coverage, since that method is mocked. So that means that nor the default mock maker nor the inline mock maker generate the correct coverage imo

Godin:
@TimvdLippe 

[...]

PR #3847, opened 2026-07-28
Preserve method parameters when clearing inline mocks

### LINKED ISSUE #3763: The class's MethodParameters are removed after clearAllCaches is invoked
Hi team. 

The test case (3rd assertion) below fails because the class's `MethodParameters` are removed after `clearAllCaches` is invoked.

```java
public class MockitoTest {

    @Test
    void testClearAllCaches_Interface() throws NoSuchMethodException {
        // 1st assertion
        assertThat(TestInterface.class.getMethod("get", String.class).getParameters()[0].getName())
                .as("Before mock the interface")
                .isEqualTo("s");

        mock(TestInterface.class);
        // 2nd assertion
        assertThat(TestInterface.class.getMethod("get", String.class).getParameters()[0].getName())
                .as("After mock the interface")
                .isEqualTo("s");

        Mockito.clearAllCaches();
        // 3rd assertion
        assertThat(TestInterface.class.getMethod("get", String.class).getParameters()[0].getName())
                .as("After Mockito clear all caches")
                .isEqualTo("s");
    }

    public interface TestInterface {

        void get(String s);
    }
}
```

It appears that during `clearAllCaches`, the `TestInterface` class is retransformed by a empty `ClassFileTransformer`, which causes this issue.

> It fails on OpenJDK 17, 21 and 24,  even though https://bugs.openjdk.org/browse/JDK-8240908 was marked as fixed.

```java
// InlineBytecodeGenerator.java
@Override
public byte[] transform(
        ClassLoader loader,
        String className,
        Class<?> classBeingRedefined,
        ProtectionDomain protectionDomain,
        byte[] classfileBuffer) {
    if (classBeingRedefined == null
            || !mocked.contains(classBeingRedefined)
                        && !flatMocked.contains(classBeingRedefined)
            || EXCLUDES.contains(classBeingRedefined)) {
        // Get here during clearAllCaches.
        return null;
    } else {
        try {
            return byteBuddy
                    .redefine(
                            classBeingRedefined,
                            //        new ClassFileLocator.Compound(
                            ClassFileLocator.Simple.of(
                                    classBeingRedefined.getName(), classfileBuffer)
                            //            ,ClassFileLocator.ForClassLoader.ofSystemLoader(
                            //        )
                                )
                    // Note: The VM erases parameter meta data from the provided class fil
                    // (bug). We just add this information manually.
                    .visit(new ParameterWritingVisitorWrapper(classBeingRedefined))
                    .visit(mockTransformer)
                    .make()
                    .getBytes();
        } catch (Throwable throwable) {
            lastException = throwable;
            return null;
        }
    }
}
```

Could you please help confirm?

DragonFSKY:
## Summary

Preserve explicit MethodParameters metadata when Mockito.clearAllCaches() removes inline instrumentation.

During cache clearing, tracked types were removed before retransformation, so the transformer could not restore parameter metadata on affected JVMs. Keep a short-lived snapshot of the types being cleared, run the existing parameter visitor for them, and skip mock advice so instrumentation is still removed.

## Verification

- The regression fails with arg0 on unchanged main under Temurin 17.0.19 and passes with this change.
- ./gradlew :mockito-core:check on Temurin 17.0.20: 2,282 passed, 75 conditionally skipped.

Fixes #3763


jselbo:
Thanks, this seems reasonable. I hope @raphw  is able to review because he made a similar fix back in the day: a928c122f7eb431664150f32fef8325f14a97ac8

raphw:
I remember that we removed it again because it broke code coverage and other tools: https://github.com/mockito/mockito/issues/757

I think the only real fix for this is to update the JVM.

raphw:
I (Claude) just tried on Java 19 and the bug is indeed fixed on the JDK as we reported it. So ideally you run on a JDK 21 or higher and you should be good.

raphw:
Actually, it seems like the OpenJDK bug is not working. I replayed this patch with a modern version of Jacoco and it seems to work now. Probably we did not try to register forretransformation and that did not play well.

There is however a bit of stray in the PR.

private volatile Set<Class<?>> clearing = Collections.emptySet();

This will be a problem as transformation can happen in parallel. It is likely rare but if any, this should be a thread local.

raphw:
Looks good with the thread local changes.

raphw (approved):

PR #3850, opened 2026-08-01
Support useConstructor with mockConstruction

### LINKED ISSUE #3811: Support useConstructor with mockConstruction
This question was sort-of asked in #2769 already, but it got closed without much consideration. I wonder if, a few years after, it's time to reconsider making `useConstructor` work with `mockConstruction`.

Take an example like this one:
```scala
val mockSettingsFactory: Function[MockedConstruction.Context, MockSettings] = (ctx: MockedConstruction.Context) => {
      withSettings
        .useConstructor(
          ctx.arguments().get(0),
          ctx.arguments().get(1),
          true
        )
    }

val mockedMasterClient = mockConstruction(classOf[MasterClient], mockSettingsFactory)
```
`MasterClient` should have its selected constructor called but the call is simply captured. This leaves us with an uninitialized instance with `final` fields that can't be changed.

I understand that this uncovers code smells and whatnot, but I'm working on an already-large PR for an open-source project, and I simply have to draw a line as to what I can refactor. I think that this feature would be useful for many working with legacy code.

arimu1:
## Checklist

## Motivation

Fixes #3811

Construction mocks intentionally skip the real constructor body (bytecode short-circuits after calling a super constructor with default args). That leaves fields — including `final` fields — uninitialized, so `CALLS_REAL_METHODS` / partial-mock usage on constructed instances fails with NPEs or wrong state.

`MockSettings.useConstructor(...)` already initializes regular mocks/spies via the selected constructor. Users expected the same setting to work with `mockConstruction`, including forwarding intercepted constructor arguments from `MockedConstruction.Context`.

## How

When a construction mock's settings request `useConstructor`:

1. While construction interception is suspended (`isSuspended`), create a temporary real instance with `ConstructorInstantiator` using the configured constructor args.
2. Copy that instance's field values onto the intercepted mock with the existing `LenientCopyTool` (handles the class hierarchy, including finals where the member accessor allows it).
3. Register the intercepted object as the mock as before.

Default construction-mock behavior is unchanged when `useConstructor` is not set.

## Example

```java
try (MockedConstruction<Foo> mocked =
        mockConstruction(
                Foo.class,
                context ->
                        withSettings()
                                .useConstructor(
                                        context.arguments().get(0),
                                        context.arguments().get(1),
                                        true)
                                .defaultAnswer(CALLS_REAL_METHODS))) {
    Foo foo = new Foo("a", 2, false);
    // constructor-selected fields (including finals) are initialized on the mock
}
```

## Testing

- Added tests in `ConstructionMockTest` covering fixed args, context-forwarded args, multi-arg initialization (issue-style), and the default “fields stay null without useConstructor” behavior.
- `./gradlew :mockito-integration-tests:inline-mocks-tests:test --tests org.mockitoinline.ConstructionMockTest` — 16/16 green
- `./gradlew :mockito-core:test --tests org.mockitousage.constructor.*` — 28/28 green
- `./gradlew spotlessApply`


raphw:
It is a bit of a strange use case, to be honest, but I agree that we should prefer consistency. Since the PR is sound, I think we can include this.

PR #3859, opened 2026-09-02
Capture arguments once per invocation when verifying over time

### LINKED ISSUE #345: Verification using After and ArgumentCaptor returns unexpected size of captured values list

ARV0007:
# Capture arguments once per invocation when verifying over time

Fixes #345

## What's wrong

Capturing arguments is a side effect of *evaluating* a verification mode:

```java
public static void markVerified(Invocation invocation, MatchableInvocation wanted) {
    invocation.markVerified();
    wanted.captureArgumentsFrom(invocation);
}
```

`VerificationOverTimeImpl` polls the delegate repeatedly until its duration elapses, so every poll
captures the arguments of every matching invocation again. With `after()` the loop also doesn't sleep
on the success path — `sleep()` only happens inside `handleVerifyException` — so it spins for the full
duration.

The result is that an `ArgumentCaptor` collects one copy per *poll* instead of one per *invocation*.
A standalone reproduction of the loop, with three invocations and a 200 ms window, collects:

| verification mode | captured | expected |
|---|---:|---:|
| `after(200).times(3)`   | 13,845,153 | 3 |
| `after(200).atLeast(3)` |  9,130,371 | 3 |
| `after(200).atMost(3)`  | 3 | 3 |

Those counts are from one run on an Apple silicon MacBook Air; the exact figure scales with CPU speed
and with the configured duration, so it varies between machines. The shape doesn't: it grows with
poll count rather than with invocation count, which is what makes it exhaust the heap on longer
durations, as reported on the issue since 2016.

`atMost` escapes it only incidentally: it filters already-verified invocations for its own counting
logic, so from the second poll onward it captures nothing.

## Why the fix isn't in the checkers

The obvious change — filter already-verified invocations in `NumberOfInvocationsChecker` and
`AtLeastXNumberOfInvocationsChecker`, as `AtMost` does — is deliberately **not** what this PR does.

That filtering existed and was removed in 5b892f5 (#819), because it broke a later verification's
ability to capture invocations an earlier one had already marked verified.
`ArgumentCaptorDontCapturePreviouslyVerifiedTest` guards that behaviour, and it should stay guarded.

The duplication isn't a property of the checkers. It's a property of the polling loop, which is the
only thing that evaluates a verification mode more than once for a single `verify()` call. So the fix
stays there.

## The change

`VerificationOverTimeImpl.verify()` now polls against a wrapper that captures the arguments of any
given invocation at most once, comparing invocations by identity:

```java
public void captureArgumentsFrom(Invocation invocation) {
    if (alreadyCaptured.add(invocation)) {
        delegate.captureArgumentsFrom(invocation);
    }
}
```

The wrapper is constructed inside `verify()`, so it lives exactly as long as one verification.

- A later verification builds a new wrapper with an empty set, so it still captures every matching
  invocation, including already-verified ones. #819's behaviour is untouched — that path doesn't go
  through `VerificationOverTimeImpl` at all.
- Invocations arriving during the window are still captured, exactly once each, so `after()` keeps
  reporting everything it saw.
- The loop structure, the early-failure path and `canRecoverFromFailure` are unchanged, so
  `should_fail_early_when_at_most_is_used` and `should_fail_early_when_never_is_used` still hold.
- `timeout()` returns on first success and never polls twice, so it is unaffected.
- Every `getTarget()` call site in `mockito-core` is interface-typed, so substituting the wrapper is
  safe.

## Tests

- Re-enables `shouldReturnListOfArgumentsWithSameSizeAsGivenInTimesVerification` and
  `shouldReturnListOfArgumentsWithSameSizeAsGivenInAtLeastVerification`, disabled with
  `@Ignore("TODO review after #936")`.
- Adds `should_capture_previously_verified_invocations_in_a_later_verification`, pinning the #819
  contract through the `after()` path specifically.

## Not addressed here

`after()` still busy-spins for its full duration on the success path, because `sleep()` sits only in
the failure branch. Captures are bounded now so the memory problem is gone, but the CPU burn isn't.
Happy to open that as a separate issue — it seemed better kept out of this change.

## Checklist

- [x] Read the contributing guide
- [x] PR is motivated
- [x] Example included in the description
- [x] Avoids other runtime dependencies
- [x] Meaningful commit history
- [x] Follows coding style (`./gradlew spotlessApply`)
- [x] Mentions `Fixes #345`
- [x] A commit ends with `Fixes #345`


raphw:
The diagnosis is right, and so is the reasoning for keeping the fix out of the checkers. I reproduced it locally: with the production change reverted, after(200).times(3) collects ~27k values instead of 3, and all three of your new/re-enabled tests fail without it and pass with it. The approach looks good though.

A few things to fix:

1. It breaks org.mockito.verification.TimeoutTest: should_fail_because_verification_fails, should_pass_even_if_first_verification_fails and should_try_to_verify_correct_number_of_times all fail. The delegate now gets the wrapper rather than the caller's VerificationData, so doThrow(error).when(mode).verify(data) no longer matches and nothing is thrown. Switching those stubbings to verify(any()) fixes it; the tests are about the polling mechanics, not the instance identity.
2. Drop the data == null ? null : … guard. It's only there because VerificationOverTimeImplTest calls impl.verify(null). Please give that test a mocked VerificationData instead. I'd rather not have a null branch in production code to accommodate a test.
3. The "timeout() is unaffected" argument needs one more sentence. It holds for the usual path because the checkers throw before markVerified, but NumberOfInvocationsChecker.checkNumberOfInvocationsNonGreedy captures inside its loop before it can throw, which is reachable via inOrder + timeout().calls(n). Wrapping unconditionally is the right call — the justification just shouldn't rest on that path being impossible.

With those three changes the full mockito-core suite works. Could you fix those?

